PR #2, opened 2021-01-05
Bugfix/rm 7392 screenreader compatible null values

MichaelKetting:


Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValue.cs
```diff
@@ -654,6 +669,8 @@ protected bool IsSelectStatementEnabled
 
     void IBocReferenceValue.PopulateDropDownList (DropDownList dropDownList)
     {
+      // PopulateDropDownList should be moved to the renderer, the BocReferenceValue should only provide a list of items, see also BocEnumValue.
+
```
a-ctor, line 673: Wäre das nicht besser als Task?

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocEnumValueUserControl.ascx.cs
```diff
@@ -106,8 +106,10 @@ override protected void OnLoad (EventArgs e)
 
     GenderField.LoadUnboundValue ((Gender?)null, IsPostBack);
 
+    UnboundMarriageStatusField.Property = (IBusinessObjectEnumerationProperty) CurrentObject.BusinessObjectClass.GetPropertyDefinition("MarriageStatus");
     //UnboundMarriageStatusField.LoadUnboundValue (person.MarriageStatus, IsPostBack);
     UnboundReadOnlyMarriageStatusField.LoadUnboundValue (person.MarriageStatus, IsPostBack);
+    DisabledUnboundMarriageStatusField.Property = (IBusinessObjectEnumerationProperty) CurrentObject.BusinessObjectClass.GetPropertyDefinition("MarriageStatus");
```
a-ctor, line 112: Mir kommen die changes unrelated vor?

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValue.cs
MichaelKetting: In theory. Thing is, I don't think this will ever get scheduled unless I happen to stumble over it during a refactoring that touches this part of the code. It's a) more work than it seems so there won't be time to just-do-it and b) completely workable at this point. So, making a task for it right now will just add a couple of random bytes to the backlog and when it comes around, you won't know that there's a task for it. and adding the task and a todo comment just felt like too much hassle at this point.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocEnumValueUserControl.ascx.cs
MichaelKetting: That's why the commit doesn't have a task-number to it. The change itself is needed at this time to fix a problem on the testpage impacting this issue. If it where production code, this would be a bug and a bugfix-branch. Since it's just test-code, it's fine.

PR #5, opened 2021-01-12
RM-7535: Don't trigger autopostback when closing the dropdown

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7535


Stefan-Ilic (approved):

MichaelKetting (changes requested):
I just realized, there's a specification bug: If you compare the behavior with a plain HTML select, the change event is triggered when closing the drop down by clicking on the down-arrow. There is no reason for the behavior of the autocomplete to be different. So, we can just close this bug as invalid (fixversion=won't fix) with an explanation  that this behavior is identical to BocReferenceValue. Sorry!

iam-mholle:
Issue closed as `Won't fix`.

PR #8, opened 2021-02-01
RM-7574 Flaky Test: WebTests fail with WxeSequenceCounter mismatch

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7574

### DIFF
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Controls/BocListAsGridUserControl.ascx
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Controls/BocListAsGridUserControl.ascx
@@ -49,7 +49,7 @@
         ShowSortingOrder="True"
         
         Width="100%"
-        Height="10em"
+        Height="35em"
         runat="server">
         <OptionsMenuItems>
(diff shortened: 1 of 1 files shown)


raffaelfoidl:
The error mentioned in the issue description could be reproduced relativley reliably at the TC agent (with one exceptional test run, at least once out of 1.000 executions).

With the enlarged container in this fix, I executed the test 12.300 times in total (3x 100, 2x 1.000, 1x 10.000) and they were all green. Granted, the 10.000-run was on a Friday afternoon (so probably no peak workload for the build server), but I hope that does not influence the expressiveness of the stability tests negatively.

MichaelKetting (changes requested):
I'm worried this fixes the symptom, not the cause of the issue. we need to first make sure that we do everything we can in the test framework to prevent this from happening. E.g. what if we scroll the page? what if we wait? What if we add a retry?

Alternatively, create a follow-up task to improve the webtesting framework to cover this case and just fix the test page now so we can get stable builds. if so, please create the ticket and then send this PR to re-approve.

raffaelfoidl:
For posterity: I created the follow-up task https://re-motion.atlassian.net/browse/RM-7589.

PR #13, opened 2021-02-04
RM-7576 Fix issue where remaining browser processes are not closed correctly

a-ctor:
https://re-motion.atlassian.net/browse/RM-7576

While this might not fix the problem of the hanging edge process, it fixes the cleanup of the browser processes. Previously, a thrown exception would cause processes to not get shut down at all. This problem is fixed by shutting down all processes at once using a shared timeout.

### DIFF
--- a/Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
+++ b/Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
@@ -15,9 +15,12 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 using JetBrains.Annotations;
 using log4net;
 using Remotion.Utilities;
@@ -118,32 +121,75 @@ namespace Remotion.Web.Development.WebTesting.Utilities
     {
       ArgumentUtility.CheckNotNull ("process", process);
 
+      GracefulProcessShutdown (new[] { process }, timeout);
+    }
+
+    /// <summary>
+    /// Tries to shutdown a list of processes gracefully and if they are not closed after <paramref name="timeout"/> it ensures that they are closed.
+    /// </summary>
+    /// <param name="processes">The processes that should be shut down.</param>
+    /// <param name="timeout">The wait timeout after a shutdown has been requested.</param>
+    /// <remarks>
+    /// <p>
+    /// This method will block until the processes have exited. Shutdown procedure is as follows:
+    /// </p>
+    /// <list type="number">
+    /// <item>
+    /// <description>Check if the processes have already exited.</description>
+    /// </item>
+    /// <item>
+    /// <description>Close the main windows of the processes and wait for the processes to exit or the <paramref name="timeout"/> to run out.</description>
+    /// </item>
+    /// <item>
+    /// <description>If the processes are still not closed, kill them and wait for the processes to exit or the <paramref name="timeout"/> is reached.</description>
+    /// </item>
+    /// </list>
+    /// </remarks>
+    public static void GracefulProcessShutdown (IReadOnlyList<Process> processes, TimeSpan timeout)
+    {
+      ArgumentUtility.CheckNotNullOrItemsNull (nameof (processes), processes);
+
       var timeoutInMilliseconds = (int) timeout.TotalMilliseconds;
       if (timeoutInMilliseconds < 0)
         throw new ArgumentOutOfRangeException ("timeout", "Timeout can not be smaller that zero.");
 
+      IReadOnlyList<Process> remainingProcesses = processes.ToList();
+
+      // Clear out any processes that already exited
+      remainingProcesses = WaitForProcessesExits (remainingProcesses, TimeSpan.Zero);
+
+      // Try to gracefully close the process
+      var anyMainWindowClosed = remainingProcesses.Any (CloseMainWindow);
+      if (anyMainWindowClosed)
+        remainingProcesses = WaitForProcessesExits (remainingProcesses, timeout);
+
+      // Force closing the process and wait for process exits
+      foreach (var remainingProcess in remainingProcesses)
+        KillProcess (remainingProcess);
+
+      remainingProcesses = WaitForProcessesExits (remainingProcesses, timeout);
+      if (remainingProcesses.Any())
+        throw CreateProcessesDidNotExitInTimeException (remainingProcesses, timeout);
+    }
+
+    private static bool CloseMainWindow (Process process)
+    {
       try
       {
-        if (process.HasExited)
-          return;
+        return process.CloseMainWindow();
       }
-      catch (Win32Exception)
+      catch (InvalidOperationException)
       {
-        // HasExited can throw Win32Exceptions in certain cases (for example when the process is running with admin rights).
-        // We ignore the exception, because it is still possible to kill the process and the following methods do not throw the same exception.
-        // Tested with process running with admin rights.
-        // If there is an unexpected exception which causes any of the other methods to fail, we have to decide how we want to handle that.
+        // Thrown if the process has already exited, or no process is associated with the Process object.
+        // https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.closemainwindow#remarks
+        // This exception can be swallowed safely, as we are sure that a process is associated and the process has
+        // already exited, therefore the method can return true at this point.
+        return true;
       }
-      
-      // Try to gracefully close the process
-      if (process.CloseMainWindow())
-      {
-        // No exception is thrown when calling .WaitForExit(..) on an already exited process.
-        if (process.WaitForExit (timeoutInMilliseconds))
-          return;
-      }
+    }
 
-      // Force closing the process
+    private static void KillProcess (Process process)
+    {
       try
       {
         process.Kill();
@@ -160,14 +206,44 @@ namespace Remotion.Web.Development.WebTesting.Utilities
         // https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.kill#System_Diagnostics_Process_Kill
         // This exception can be swallowed safely, as we are sure that a process is associated and the process has
         // already exited, therefore the method can return at this point.
-        return;
+      }
+    }
+
+    /// <summary>
+    /// Waits for the specified <paramref name="processes"/> to exit or until the <paramref name="sharedTimeout"/> is reached.
+    /// </summary>
+    /// <param name="processes">The processes that should be waited on to exit.</param>
+    /// <param name="sharedTimeout">The wait timeout after which waiting will be canceled.</param>
+    /// <returns>A subset of the specified <paramref name="processes"/> that did not exit withing the <paramref name="sharedTimeout"/>.</returns>
+    [MustUseReturnValue]
+    private static IReadOnlyList<Process> WaitForProcessesExits (IEnumerable<Process> processes, TimeSpan sharedTimeout)
+    {
+      var remainingProcesses = new List<Process>();
+
+      // Go through all processes and call .WaitForExit on each using a shared timeout
+      // If the timeout runs out still go through all remaining process to make sure to clear any exited processes
+      var stopwatch = Stopwatch.StartNew();
+      foreach (var process in processes)
+      {
+        var timeToTimeoutEnd = sharedTimeout - stopwatch.Elapsed;
+
+        // Use 0 (process.WaitForExit returns instantly) if we exceeded our timeout, otherwise use the remaining timeout
+        var remainingTimeoutInMilliseconds = Math.Max ((int) timeToTimeoutEnd.TotalMilliseconds, 0);
+        if (!process.WaitForExit (remainingTimeoutInMilliseconds))
+          remainingProcesses.Add (process);
       }
 
-      // No exception is thrown when calling .WaitForExit(..) on an already exited process.
-      if (process.WaitForExit (timeoutInMilliseconds))
-        return;
+      return remainingProcesses;
+    }
 
-      throw new InvalidOperationException (string.Format ("Process '{0}' (id: '{1}') did not exit in the expected amount of time.", process.ProcessName, process.Id));
+    private static Exception CreateProcessesDidNotExitInTimeException (IEnumerable<Process> processes, TimeSpan timeout)
+    {
+      var stringBuilder = new StringBuilder();
+      stringBuilder.AppendFormat ("The following processes did not exit within the timeout of {0:0.##} seconds:", timeout.TotalSeconds).AppendLine();
+      foreach (var process in processes)
+        stringBuilder.AppendFormat (" - Process '{0}' (id: {1})", process.ProcessName, process.Id).AppendLine();
+
+      return new InvalidOperationException (stringBuilder.ToString());
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -160,14 +249,27 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
-        return;
+    private static void WaitForProcessesExits (List<Process> processes, TimeSpan timeout)
+    {
+      // Go through all processes and call .WaitForExit on each using a shared timeout
+      // If the timeout runs out still go through all remaining process to make sure to clear any exited processes
+      var timeoutStart = DateTime.UtcNow;
+      for (var i = 0; i < processes.Count; i++)
+      {
+        var elapsedTimeSinceTimeoutStart = DateTime.UtcNow - timeoutStart;
+        var timeToTimeoutEnd = timeout - elapsedTimeSinceTimeoutStart;
```
iam-mholle, line 263: Is there a benefit in using the following `DateTime` calc logic over just using a `Stopwatch`? I think it would make the logic a litte more concise and understandable.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -160,14 +249,27 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
         // This exception can be swallowed safely, as we are sure that a process is associated and the process has
         // already exited, therefore the method can return at this point.
-        return;
       }
+    }
 
-      // No exception is thrown when calling .WaitForExit(..) on an already exited process.
-      if (process.WaitForExit (timeoutInMilliseconds))
-        return;
+    private static void WaitForProcessesExits (List<Process> processes, TimeSpan timeout)
```
iam-mholle, line 255: I think timeout should be named `sharedTimeout`, `totalTimeout` or something similar to indicate that the `timeout` does not apply to every process anew. May not be relevant here as this is a private method 🤷🏼‍♂️

iam-mholle, line 255: I think the name of the method should indicate that the collection is modified and exited processes are removed, this aspect is not deducible from the signature, or it could just return a new collection with non-exited processes instead of mutating.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -134,16 +136,103 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+
+      // Force closing the process and wait for process exits
+      remainingProcesses.ForEach (KillProcess);
+      WaitForProcessesExits (remainingProcesses, timeout);
+
+      if (remainingProcesses.Count == 0)
+        return;
+
+      // At this point we could not close all the process and therefore thrown an exception
+      static Exception CreateProcessDidNotExitInTimeException (Process process)
```
iam-mholle, line 206: Local methods should be placed at the end of the containing method.

iam-mholle (changes requested):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -160,14 +248,34 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
-      if (process.WaitForExit (timeoutInMilliseconds))
-        return;
+    /// <summary>
+    /// Waits for the specified <paramref name="processes"/> to exit or until the <paramref name="sharedTimeout"/> is reached.
+    /// </summary>
+    /// <param name="processes">The processes that should be waited on to exit.</param>
+    /// <param name="sharedTimeout">The wait timeout after which waiting will be canceled.</param>
+    /// <returns>A subset of the specified <paramref name="processes"/> that did not exit withing the <paramref name="sharedTimeout"/>.</returns>
+    [MustUseReturnValue]
+    private static List<Process> WaitForProcessesExits (IEnumerable<Process> processes, TimeSpan sharedTimeout)
```
iam-mholle, line 261: Return type should be `IReadOnlyCollection<>` here imo. No use of returning a mutable collection, an immutable one makes more sense here.

iam-mholle (changes requested):

iam-mholle (approved):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -134,16 +136,103 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+    /// </item>
+    /// <item>
+    /// <description>Close the main windows of the processes and wait for the processes to exit or the <paramref name="timeout"/> to run out.</description>
+    /// </item>
+    /// <item>
+    /// <description>If the processes are still not closed, kill them and wait for the processes to exit or the <paramref name="timeout"/> to run out.</description>
+    /// </item>
+    /// </list>
+    /// </remarks>
+    public static void GracefulProcessShutdown (IReadOnlyList<Process> processes, TimeSpan timeout)
```
MichaelKetting, line 179: Isn't the single-process variant of this method just an overload? Can be do this:
`public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan timeout)
{
  GracefulProcessShutdown (new[]{process}, timeout);
}`

And, is the single-process overload actually needed in the first place?

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -134,16 +136,103 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+      if (anyMainWindowClosed)
+        remainingProcesses = WaitForProcessesExits (remainingProcesses, timeout);
+
+      // Force closing the process and wait for process exits
+      foreach (var remainingProcess in remainingProcesses)
+        KillProcess (remainingProcess);
+
+      remainingProcesses = WaitForProcessesExits (remainingProcesses, timeout);
+
+      if (remainingProcesses.Count == 0)
```
MichaelKetting, line 203: Would this be cleaner if we inverted to the if and just had the exception in the conditonal and otherwise just a regular method exit? We could then also say remainingProcesses.Any() in the IF.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -134,16 +136,103 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+        return;
+
+      // At this point we could not close all the process and therefore thrown an exception
+      throw remainingProcesses.Count == 1
+          ? CreateProcessDidNotExitInTimeException (remainingProcesses[0])
+          : new AggregateException (remainingProcesses.Select (CreateProcessDidNotExitInTimeException));
+    }
+
+    private static Exception CreateProcessDidNotExitInTimeException (Process process)
```
MichaelKetting, line 212: move two methods down to keep the methods in their invocation order.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -160,14 +249,34 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+      var remainingProcesses = new List<Process>();
+
+      // Go through all processes and call .WaitForExit on each using a shared timeout
+      // If the timeout runs out still go through all remaining process to make sure to clear any exited processes
+      var stopwatch = Stopwatch.StartNew();
+      foreach (var process in processes)
+      {
+        var timeToTimeoutEnd = sharedTimeout - stopwatch.Elapsed;
+
+        // Use 0 (.WaitForExit returns instantly) if we exceeded our timeout, otherwise use the remaining timeout
```
MichaelKetting, line 273: "Process.WairForExit(...)"

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
a-ctor: The single process overload is no longer needed, but I wasn't sure if it needs to be kept. If it is not needed I would remove it.

iam-mholle: Keep in mind that there could be some dependencies on this, as this is a public class.

MichaelKetting: Well, we do have the DependDB and it would be easy to adapt a broken invocation. But I don't mind keeping the overload.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -118,32 +120,81 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
     {
       ArgumentUtility.CheckNotNull ("process", process);
 
+      GracefulProcessShutdown(new []{process}, timeout);
```
MichaelKetting, line 123: whitsepace: 
`new[] { process }`

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -118,32 +120,81 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+    /// </p>
+    /// <list type="number">
+    /// <item>
+    /// <description>Check if the processes have already exited.</description>
+    /// </item>
+    /// <item>
+    /// <description>Close the main windows of the processes and wait for the processes to exit or the <paramref name="timeout"/> to run out.</description>
+    /// </item>
+    /// <item>
+    /// <description>If the processes are still not closed, kill them and wait for the processes to exit or the <paramref name="timeout"/> to run out.</description>
```
MichaelKetting, line 143: ...or until the timeout is reached

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -160,14 +211,40 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
-      // No exception is thrown when calling .WaitForExit(..) on an already exited process.
-      if (process.WaitForExit (timeoutInMilliseconds))
-        return;
+      return remainingProcesses;
+    }
 
-      throw new InvalidOperationException (string.Format ("Process '{0}' (id: '{1}') did not exit in the expected amount of time.", process.ProcessName, process.Id));
+    private static Exception CreateProcessDidNotExitInTimeException (Process process)
+    {
+      return new InvalidOperationException (
```
MichaelKetting, line 246: 1. pass the timeout and use it in the message
2. add all processes, that way we can forgoe the aggregateexception
The following processes did not exit within the timeout of {0} seconds:
- Process 'Name' (id: ID)
- ...

ID is an INT, don't need quotes when writing {id: 123)

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -160,14 +206,44 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
-      if (process.WaitForExit (timeoutInMilliseconds))
-        return;
+      return remainingProcesses;
+    }
+
+    private static Exception CreateProcessesDidNotExitInTimeException (IEnumerable<Process> processes, TimeSpan timeout)
+    {
+      var stringBuilder = new StringBuilder();
+      stringBuilder.AppendLine ($"The following processes did not exit within the timeout of {timeout.TotalSeconds:0.##} seconds:");
+      foreach (var process in processes)
```
MichaelKetting, line 243: Do you know
`stringBuilder = processes.Aggregate(sb, (p, sb) => sb.AppendLine (..., p.ID))`
Personally, I like the short style, but it depends on how much you like to read linq :D

Also, please use AppendFormat(...).AppendLine(). It's better for memory management.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
a-ctor: I prefer the "normal" foreach over the `.Aggregate` call in this situation because less work is needed to understand what it does. Aggregate using the same (mutable) variable is effectively a `.ForEach` instead of an `.Aggregate` and makes the code more complicated for not much benefit. Sadly, `.ForEach` is only available on `List<T>` as I would prefer it over a normal foreach iteration.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -239,9 +239,9 @@ private static IReadOnlyList<Process> WaitForProcessesExits (IEnumerable<Process
     private static Exception CreateProcessesDidNotExitInTimeException (IEnumerable<Process> processes, TimeSpan timeout)
     {
       var stringBuilder = new StringBuilder();
-      stringBuilder.AppendLine ($"The following processes did not exit within the timeout of {timeout.TotalSeconds:0.##} seconds:");
+      stringBuilder.AppendFormat ("The following processes did not exit within the timeout of {0:0.##} seconds:", timeout.TotalSeconds);
       foreach (var process in processes)
-        stringBuilder.AppendLine ($" - Process '{process.ProcessName}' (id: {process.Id})");
+        stringBuilder.AppendFormat (" - Process '{0}' (id: {1})", process.ProcessName, process.Id);
```
MichaelKetting, line 244: The `.AppendLine()` is missing at the end. That's the downside of using AppEndFormat, but downside that's acceptable.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -239,9 +239,9 @@ private static IReadOnlyList<Process> WaitForProcessesExits (IEnumerable<Process
     private static Exception CreateProcessesDidNotExitInTimeException (IEnumerable<Process> processes, TimeSpan timeout)
     {
       var stringBuilder = new StringBuilder();
-      stringBuilder.AppendLine ($"The following processes did not exit within the timeout of {timeout.TotalSeconds:0.##} seconds:");
+      stringBuilder.AppendFormat ("The following processes did not exit within the timeout of {0:0.##} seconds:", timeout.TotalSeconds);
```
MichaelKetting, line 242: `.AppendLine()` at the end.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
MichaelKetting: Perhaps. I can work with both and freely admit to having a personal preference to the Aggregate Syntax *). But there is no hard and fast rule :)

*) It goes with the SB acting as a collecting parameter and the Aggregate-pattern, once you know and understand it, just feels more streamlined that way.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
a-ctor: re-motion rewrite to F# confirmed? 😜

MichaelKetting: We don't talk about Fight Club.

MichaelKetting (approved):

PR #16, opened 2021-02-08
RM-7583: Update object binding web test to allow in session only persistence

a-ctor:
https://re-motion.atlassian.net/browse/RM-7583

### DIFF
--- a/Remotion/ObjectBinding/Sample/InMemoryReflectionBusinessObjectStorageProvider.cs
+++ b/Remotion/ObjectBinding/Sample/InMemoryReflectionBusinessObjectStorageProvider.cs
@@ -1,0 +1,76 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Remotion.Collections;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Sample
+{
+  public class InMemoryReflectionBusinessObjectStorageProvider : IReflectionBusinessObjectStorageProvider
+  {
+    private readonly AutoInitDictionary<Type, Dictionary<Guid, MemoryStream>> _reflectionBusinessObjectData =
+        new AutoInitDictionary<Type, Dictionary<Guid, MemoryStream>> (() => new Dictionary<Guid, MemoryStream>());
+
+    public InMemoryReflectionBusinessObjectStorageProvider ()
+    {
+    }
+
+    /// <inheritdoc />
+    public IReadOnlyCollection<Guid> GetObjectIDsForType (Type type)
+    {
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      return _reflectionBusinessObjectData
+          .SelectMany (e => e.Value.Keys)
+          .ToArray();
+    }
+
+    /// <inheritdoc />
+    public Stream GetReadObjectStream (Type type, Guid id)
+    {
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      var oldMemoryStream = _reflectionBusinessObjectData[type].GetValueOrDefault (id);
+      if (oldMemoryStream == null)
+        return null;
+
+      // Create a new MemoryStream because the old one might have been disposed
+      var newMemoryStream = new MemoryStream (oldMemoryStream.ToArray(), false);
+      _reflectionBusinessObjectData[type][id] = newMemoryStream;
+
+      oldMemoryStream.Dispose();
+
+      return newMemoryStream;
+    }
+
+    /// <inheritdoc />
+    public Stream GetWriteObjectStream (Type type, Guid id)
+    {
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      _reflectionBusinessObjectData[type].GetValueOrDefault (id)?.Dispose();
+
+      var newMemoryStream = new MemoryStream();
+      _reflectionBusinessObjectData[type][id] = newMemoryStream;
+
+      return newMemoryStream;
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Sample/ReadFallbackReflectionBusinessObjectStorageProviderDecorator.cs
+++ b/Remotion/ObjectBinding/Sample/ReadFallbackReflectionBusinessObjectStorageProviderDecorator.cs
@@ -1,0 +1,69 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Sample
+{
+  public class ReadFallbackReflectionBusinessObjectStorageProviderDecorator : IReflectionBusinessObjectStorageProvider
+  {
+    private readonly IReflectionBusinessObjectStorageProvider _readFallbackBusinessObjectStorageProvider;
+    private readonly IReflectionBusinessObjectStorageProvider _innerBusinessObjectStorageProvider;
+
+    public ReadFallbackReflectionBusinessObjectStorageProviderDecorator (
+        IReflectionBusinessObjectStorageProvider innerBusinessObjectStorageProvider,
+        IReflectionBusinessObjectStorageProvider readFallbackBusinessObjectStorageProvider)
+    {
+      ArgumentUtility.CheckNotNull ("innerBusinessObjectStorageProvider", innerBusinessObjectStorageProvider);
+      ArgumentUtility.CheckNotNull ("readFallbackBusinessObjectStorageProvider", readFallbackBusinessObjectStorageProvider);
+
+      _innerBusinessObjectStorageProvider = innerBusinessObjectStorageProvider;
+      _readFallbackBusinessObjectStorageProvider = readFallbackBusinessObjectStorageProvider;
+    }
+
+    /// <inheritdoc />
+    public IReadOnlyCollection<Guid> GetObjectIDsForType (Type type)
+    {
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      return _readFallbackBusinessObjectStorageProvider.GetObjectIDsForType (type)
+          .Concat (_innerBusinessObjectStorageProvider.GetObjectIDsForType (type))
+          .Distinct()
+          .ToArray();
+    }
+
+    /// <inheritdoc />
+    public Stream GetReadObjectStream (Type type, Guid id)
+    {
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      return _innerBusinessObjectStorageProvider.GetReadObjectStream (type, id)
+             ?? _readFallbackBusinessObjectStorageProvider.GetReadObjectStream (type, id);
+    }
+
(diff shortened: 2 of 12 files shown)


iam-mholle (approved):
Looking good

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+
+      var reflectionBusinessObjectStorageProvider = GetReflectionBusinessObjectStorageProviderForCurrentSession();
+      return reflectionBusinessObjectStorageProvider.GetWriteObjectStream (type, id);
+    }
+
+    private IReflectionBusinessObjectStorageProvider GetReflectionBusinessObjectStorageProviderForCurrentSession ()
+    {
+      var httpContext = _httpContextProvider.GetCurrentHttpContext();
+      var reflectionBusinessObjectStorageProvider = httpContext.Session[c_sessionStateEntryName] as IReflectionBusinessObjectStorageProvider;
+      if (reflectionBusinessObjectStorageProvider != null)
```
MichaelKetting, line 74: invert if, reduces code to one return and also removes the not-complexity from the condition.

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.IO;
+using Remotion.Utilities;
+using Remotion.Web.Infrastructure;
+
+namespace Remotion.ObjectBinding.Sample
+{
+  public class SessionStateReflectionBusinessObjectStorageProvider : IReflectionBusinessObjectStorageProvider
+  {
+    private const string c_sessionStateEntryName = "SessionStateReflectionBusinessObjectStorageProviderEntry";
```
MichaelKetting, line 27: could also make this static and use typeof(session...).Fullname

MichaelKetting (changes requested):
Still need to review the inmememorygroup, inmemoryobject, and inmemorystream. I'd like to find a way to make the code simpler/more compact while preserving threadsafety. 

Maybe, it's enough to just make a locking-decorator for the Businessobjectstorageprovider. when it's inside the session, we can just do a lock over the methods as a whole. that wya, we can just pass memory streams along and don't neeed all the overhead. 

if that's a good idea, let's try that one and drop the current types after the final review (in case the idea doesn't work out after all).

a-ctor:
> 
> 
> Still need to review the inmememorygroup, inmemoryobject, and inmemorystream. I'd like to find a way to make the code simpler/more compact while preserving threadsafety.
> 
> Maybe, it's enough to just make a locking-decorator for the Businessobjectstorageprovider. when it's inside the session, we can just do a lock over the methods as a whole. that wya, we can just pass memory streams along and don't neeed all the overhead.
> 
> if that's a good idea, let's try that one and drop the current types after the final review (in case the idea doesn't work out after all).

@MichaelKetting 

Since the stream is consumed at the outermost points in `XmlReflectionBusinessObjectStorageProvider` it is not as simple as creating a decorator for `IReflectionBusinessObjectStorageProvider`. Without creating another indirection for the XML serialization the locking must be done in the `XmlReflectionBusinessObjectStorageProvider`. Currently, this is done implicitly using the custom stream implementation `InMemoryReflectionBusinessObjectStream`.

Another solution would be to use a simpler object instead of a custom stream implementation. This objects exposes the `MemoryStream` and handles unlocking the object. This then requires that all `Stream` return values are changed to the new type and the logic `XmlReflectionBusinessObjectStorageProvider` is getting a little bit more verbose. The current locking complexity stays the same.

Since `InMemoryReflectionBusinessObjectStream` is 80% wrapper around `MemoryStream` and 20% unlocking and the unlocking part is not going anywhere I would prefer the `Stream` variant over a custom non-stream object.

MichaelKetting:
> Since the stream is consumed at the outermost points in `XmlReflectionBusinessObjectStorageProvider` it is not as simple as creating a decorator for `IReflectionBusinessObjectStorageProvider`. Without creating another indirection for the XML serialization the locking must be done in the `XmlReflectionBusinessObjectStorageProvider`. Currently, this is done implicitly using the custom stream implementation `InMemoryReflectionBusinessObjectStream`.

Hmm... oh, yes, thought the nesting was the other around.

Okay, we should just add a lock statement to the public members of XmlReflecitonBOStorageProvider  (and `protected CreateObject(...)` ) and be done with it. Given how XmlProvider is implemented, it's not threadsafe, either and if we do threadsafety, we should do it the right way. Completely strictly speaking, the BusinessObjects aren't threadsafe either, but that's one step too far right now.

a-ctor:
@MichaelKetting So replace `inmememorygroup`, `inmemoryobject`, and `inmemorystream` with a simple dictionary lookup (along the lines of `Dictionary<(Type, Guid), MemoryStream>`) in the `inmemory...provider`. And lock the class from the outside to prevent accidental multithreading?

Remotion/ObjectBinding/Sample/ReadFallbackReflectionBusinessObjectStorageProviderDecorator.cs
```diff
@@ -0,0 +1,69 @@
...
+      _innerBusinessObjectStorageProvider = innerBusinessObjectStorageProvider;
+      _readFallbackBusinessObjectStorageProvider = readFallbackBusinessObjectStorageProvider;
+    }
+
+    /// <inheritdoc />
+    public IReadOnlyCollection<Guid> GetObjectIDsForType (Type type)
+    {
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      return _readFallbackBusinessObjectStorageProvider.GetObjectIDsForType (type)
```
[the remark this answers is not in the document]
a-ctor: This would change the behavior of the function. e.g. two files in the directory -> method will return two IDs but if then written and stored in memory the method would only return the one that is now in memory and the one from the file system is lost.

Remotion/ObjectBinding/Sample/XmlReflectionBusinessObjectStorageProvider.cs
```diff
@@ -75,18 +70,20 @@ public BindableXmlObject GetObject (Type type, Guid id)
...
-      XmlSerializer serializer = GetXmlSerializer (concreteType);
-      using (FileStream stream = new FileStream (fileName, FileMode.Open, FileAccess.Read))
-      {
-        obj = (BindableXmlObject) serializer.Deserialize (stream);
+          obj = (BindableXmlObject) serializer.Deserialize (stream);
+        }
       }
+
       obj._id = id;
       AddToIdentityMap (obj);
```
MichaelKetting, line 88: The lock needs to cover `GetFromidentityMap` and `AddToIdentityMap` as well. that's why we also have to update `CreateObject`

Remotion/ObjectBinding/Sample/InMemoryReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Sample
+{
+  public class InMemoryReflectionBusinessObjectStorageProvider : IReflectionBusinessObjectStorageProvider
+  {
+    private readonly Dictionary<Tuple<Type, Guid>, MemoryStream> _reflectionBusinessObjectData = new Dictionary<Tuple<Type, Guid>, MemoryStream>();
```
MichaelKetting, line 27: Should we use a ValueTuple to make it better readable (i.e. use named items in the tuple)? Or should we use a nested dictionary?

Actually, we have an AutoInitDictionary, so we could go:
`AutoInitDictionary<Type,Dictionary<Guid,MemoryStream>>`

Accessing would then be `dict[type].TryGetValue(guid, out memoryStream)` 
or `dict[type].GetValueOrDefault(guid)`  (we have an extension method for this)

MichaelKetting (changes requested):

MichaelKetting (approved):

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+      var reflectionBusinessObjectStorageProvider = GetReflectionBusinessObjectStorageProviderForCurrentSession();
+      return reflectionBusinessObjectStorageProvider.GetWriteObjectStream (type, id);
+    }
+
+    private IReflectionBusinessObjectStorageProvider GetReflectionBusinessObjectStorageProviderForCurrentSession ()
+    {
+      var httpContext = _httpContextProvider.GetCurrentHttpContext();
+      var httpSessionStateBase = httpContext.Session;
+
+      var reflectionBusinessObjectStorageProvider = httpSessionStateBase?[s_sessionStateEntryName] as IReflectionBusinessObjectStorageProvider;
```
MichaelKetting, line 75: Shouldn't this be a hardcast?

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+      ArgumentUtility.CheckNotNull ("type", type);
+
+      var reflectionBusinessObjectStorageProvider = GetReflectionBusinessObjectStorageProviderForCurrentSession();
+      return reflectionBusinessObjectStorageProvider.GetWriteObjectStream (type, id);
+    }
+
+    private IReflectionBusinessObjectStorageProvider GetReflectionBusinessObjectStorageProviderForCurrentSession ()
+    {
+      var httpContext = _httpContextProvider.GetCurrentHttpContext();
+      var httpSessionStateBase = httpContext.Session;
```
MichaelKetting, line 73: httpSession

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+
+      var reflectionBusinessObjectStorageProvider = httpSessionStateBase?[s_sessionStateEntryName] as IReflectionBusinessObjectStorageProvider;
+      if (reflectionBusinessObjectStorageProvider == null)
+      {
+        reflectionBusinessObjectStorageProvider = _reflectionBusinessObjectStorageProviderFactory.CreateBusinessObjectStorageProvider();
+        if (httpSessionStateBase != null)
+          httpSessionStateBase[s_sessionStateEntryName] = reflectionBusinessObjectStorageProvider;
```
MichaelKetting, line 80: s_sessionKeyForStorageProvider

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+    private IReflectionBusinessObjectStorageProvider GetReflectionBusinessObjectStorageProviderForCurrentSession ()
+    {
+      var httpContext = _httpContextProvider.GetCurrentHttpContext();
+      var httpSessionStateBase = httpContext.Session;
+
+      var reflectionBusinessObjectStorageProvider = httpSessionStateBase?[s_sessionStateEntryName] as IReflectionBusinessObjectStorageProvider;
+      if (reflectionBusinessObjectStorageProvider == null)
+      {
+        reflectionBusinessObjectStorageProvider = _reflectionBusinessObjectStorageProviderFactory.CreateBusinessObjectStorageProvider();
+        if (httpSessionStateBase != null)
```
MichaelKetting, line 79: could we write this as `httpSession?[s_sessionKeyForStorageProvider] = reflectionBusinessObjectStorageProvider;` ? I.e. using null-propagation also for the write side.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
a-ctor: I have tried but this is not possible :/

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+      var reflectionBusinessObjectStorageProvider = GetReflectionBusinessObjectStorageProviderForCurrentSession();
+      return reflectionBusinessObjectStorageProvider.GetWriteObjectStream (type, id);
+    }
+
+    private IReflectionBusinessObjectStorageProvider GetReflectionBusinessObjectStorageProviderForCurrentSession ()
+    {
+      var httpContext = _httpContextProvider.GetCurrentHttpContext();
+      var httpSession = httpContext.Session;
+
+      var reflectionBusinessObjectStorageProvider = (IReflectionBusinessObjectStorageProvider) httpSession?[s_sessionKeyForStorageProvider];
```
MichaelKetting, line 75: since the null-coalese doesn't work, would it be better to do a session-null-check up front and return the `_reflectionBusinessObjectStorageProviderFactory.CreateBusinessObjectStorageProvider();` before continuing with the "main" code?

Yes, we would do the Create-call twice. On the other hand, we would make it clear that the first call is a special case instead of just a rather obscure guard (hey I'm not a bug, please believe me!) in the main code?

Thoughts?

MichaelKetting (changes requested):

Remotion/ObjectBinding/Sample/SessionStateReflectionBusinessObjectStorageProvider.cs
a-ctor: It is probably a bit cleaner if we check for null beforehand. My biggest issue with this code parts is that it is marked as "heuristically unreachable" because ReSharper/VisualStudio thinks that Session cannot be null. But sadly, we cannot get rid of this issue.

MichaelKetting: That should be solved once the NRT annoations are here.

MichaelKetting (approved):

PR #17, opened 2021-02-11
RM-7601 boc boolean value sets wrong html element when updating the description

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-7601

The element sequence was changed in RM-7058 with the introduction of the required-label.

### DIFF
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRenderer.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRenderer.cs
@@ -255,14 +255,14 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocBooleanValueImplementation.R
       scriptBuilder.Append (", ");
       scriptBuilder.Append ("$(this).parent().children('a')[0]");
       scriptBuilder.Append (", ");
-      scriptBuilder.Append ("$(this).parent().children('a').children('img').first()[0]");
+      scriptBuilder.Append ("$(this).parent().children('a').children('img')[0]");
       scriptBuilder.Append (", ");
       if (renderingContext.Control.ShowDescription)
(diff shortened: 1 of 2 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRenderer.cs
```diff
@@ -258,7 +258,7 @@ private void RegisterStarupScriptIfNeeded (BocBooleanValueRenderingContext rende
       scriptBuilder.Append ("$(this).parent().children('a').children('img').first()[0]");
       scriptBuilder.Append (", ");
       if (renderingContext.Control.ShowDescription)
-        scriptBuilder.Append ("$(this).parent().children('span').first()[0]");
+        scriptBuilder.Append ("$(this).parent().children('span').slice(1, 2)[0]");
```
MichaelKetting, line 261: The selector was simply changed to chose the second instead of the first item, reflecting the updated elements sequence. 

It would have been possible to also use the elements ID during selection but this would result in a larger change and was therefor not targeted for this bugfix.

iam-mholle: What's the reason behind choosing `.slice(1, 2)[0]` instead of simply `[1]`?

MichaelKetting: I did not realize that jquery has this option when working with the children() selector.

MichaelKetting: @iam-mholle it's now updated to use the shorter syntax consistently.

iam-mholle (approved):

PR #24, opened 2021-02-16
RM-7576 Fix issue where remaining browser processes are not closed when an exception is thrown

a-ctor:


Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -114,35 +117,79 @@ public static int GetParentProcessID ([NotNull] Process target)
     /// </item>
     /// </list>
     /// </remarks>
-    public static void GracefulProcessShutdown ([NotNull] Process process, int timeout)
+    public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan timeout)
```
a-ctor, line 120: @MichaelKetting I have converted the timeout parameter from an `int` to a `TimeSpan`. Is this okay or should I leave it as an `int`? If it should be left as an `int` it would probably make sense to also change the type of the timeout parameter on the overload.

MichaelKetting: Hmm... should leave the backport with int on both methods and then document a breaking change between v2.28 and v3.0

MichaelKetting (changes requested):
The rest of the diff is identical to develop?

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
a-ctor: The change from `int` to `TimeSpan` was not done in this commit but instead in RM-7050 so the breaking changes was already made before this issue

MichaelKetting: I'm sorry, I don't follow.

MichaelKetting (approved):

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
MichaelKetting: @a-ctor in RM-7050 (apparently?) the int parameter has been introduced in the current API. Plus, here we got a second method also with an int parameter. Now, on develop, this parameter was changed to TimeSpan. This should be documented as two breaking changes between (v2.28 and v3.0( and (v-somehting-old and v3.0)

PR #33, opened 2021-02-24
RM-7608 Migrate ListMenu.js

iam-mholle:


Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,315 @@
...
+        item.className = this._itemClassName;
+        if (itemInfo.Href != null)
+        {
+          if (itemInfo.Href.toLowerCase().indexOf ('javascript:') >= 0)
+          {
+            anchor.href = '#';
+            anchor.removeAttribute ('target');
+            anchor.setAttribute ('javascript', itemInfo.Href);
+            anchor.removeAttribute ('onclick');
+            anchor.onclick = () => { eval (<string>anchor.getAttribute('javascript')); return false; };
```
iam-mholle, line 157: 1. `anchor.getAttribute('javascript')` should have a null check - bug?
2. I replaced `this.getAttribute(...)` with `anchor.getAttribute(...)` here. Was this the right decision?

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,315 @@
...
+        if (itemInfo.RequiredSelection == this._requiredSelectionExactlyOne && selectionCount != 1)
+        {
+          isEnabled = false;
+        }
+        if (itemInfo.RequiredSelection == this._requiredSelectionOneOrMore && selectionCount < 1)
+        {
+          isEnabled = false;
+        }
+      }
+      var item = <Element>document.getElementById (itemInfo.ID);
```
iam-mholle, line 141: `item` should have a null check - bug?

Remotion/Web/ClientScript/Scripts/ListMenu.ts
MichaelKetting: @iam-mholle 
1. No, that's only "helpful" if we worry that someone manipulates our DOM after the script has been initialized.
2. {{this}} was chosen because the click-handler is on the anchor, so anchor is {{this}}. I didn't want to capture the anchor in the handler. There's also historical issue where DOM/Handler+Handler/DOM dependencies make a cyclic dependency that cannot be collected.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
MichaelKetting: No, this never be null so long as no one manipulates the DOM. These cases, I would simply make a Damit operator and be done with it in C#. @a-ctor how do we do this in TS?

a-ctor: Same way as in C# with a bang!

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -14,39 +14,97 @@
...
+    this._isDisabled = isDisabled;
+    this._href = href;
+    this._target = target;
+    this._diagnosticMetadata = diagnosticMetadata;
+    this._diagnosticMetadataForCommand = diagnosticMetadataForCommand;
+  }
 }
 
-function ListMenu() { }
+class ListMenu
```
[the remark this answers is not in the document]
iam-mholle: @MichaelKetting should the global methods call the class methods? just to be sure.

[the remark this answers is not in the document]
MichaelKetting: @iam-mholle 
What @a-ctor said is correct. We want
```
function ListMenu_Initialize(foo, bar)
{
  ListMenu.Initialize (foo, bar)
}
```
for now and do the cleanup later. We only need to do this for methods that we actually use from C# code. Function calls inside the class-functions should be rewritten to use the new static functions.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -14,39 +14,97 @@
...
+  public get Icon(): string { return this._icon; }
+  public get IconDisabled(): string { return this._iconDisabled; }
+  public get RequiredSelection(): number { return this._requiredSelection; }
+  public get IsDisabled(): boolean { return this._isDisabled; }
+  public get Href(): string { return this._href; }
+  public get Target(): string { return this._target; }
+  public get DiagnosticMetadata(): Dictionary<string | boolean> { return this._diagnosticMetadata; }
+  public get DiagnosticMetadataForCommand(): Dictionary<string | boolean> { return this._diagnosticMetadataForCommand; }
+
+  constructor(
```
a-ctor, line 67: I asked MK and he is okay with the TypeScript shortcut syntax for quick assigning constructor parameters:
```ts
constructor(
  public readonly id,
  public readonly category,
  ... ) { }
```

If the constructor sets additional state then the normal variant of declaring and assigning each field should be used

MichaelKetting: Reasoning for this is to keep all state declarations at one place so we can get a quick and complete overview of a class's state declarations.

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+    public readonly DiagnosticMetadata: Dictionary<string | boolean>,
+    public readonly DiagnosticMetadataForCommand: Dictionary<string | boolean>)
+  {
+  }
+}
+
+class ListMenu
+{
+  private static _listMenuInfos: Dictionary<ListMenu_MenuInfo> = {};
+  private static _itemClassName: string = 'listMenuItem';
```
MichaelKetting, line 46: This kind of thing should be updated to const when we do the cleanup. please add note to follow-up task for this. @a-ctor  known where / how

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+        item.className = this._itemClassName;
+        if (itemInfo.Href != null)
+        {
+          if (itemInfo.Href.toLowerCase().indexOf ('javascript:') >= 0)
+          {
+            anchor.href = '#';
+            anchor.removeAttribute ('target');
+            anchor.setAttribute ('javascript', itemInfo.Href);
+            anchor.removeAttribute ('onclick');
+            // TODO: nullcheck anchor.getAttribute ('javascript')
```
MichaelKetting, line 115: this will always have the javascript attribute. a check is only needed if we wouldn't get an exception otherwise. We do have a ton of Javascript that makes assumption on what other javascript code in the same class does or what our Renderer generates. we don't want to add guards for all possible bugs in the rendering of the DOM. a NullReferenceException is okay when our own code has a mismatch. What is important is that the line where the error occurs identifies the problem clearly.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+        if (itemInfo.RequiredSelection == this._requiredSelectionExactlyOne && selectionCount != 1)
+        {
+          isEnabled = false;
+        }
+        if (itemInfo.RequiredSelection == this._requiredSelectionOneOrMore && selectionCount < 1)
+        {
+          isEnabled = false;
+        }
+      }
+      // TODO: nullcheck
```
MichaelKetting, line 98: This, too is a case of renderer should produce correct marktup.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+          });
+
+        itemInfo.DiagnosticMetadataForCommand['data-is-disabled'] = isEnabled ? 'false' : 'true';
+
+        $(anchor).attr(itemInfo.DiagnosticMetadataForCommand);
+      }
+    }
+  }
+
+  public static OnKeyDown (event: JQueryKeyEventObject, $listMenu: JQuery): void
```
MichaelKetting, line 180: Is this only needed from within the own class or also from other places?

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+  private static _requiredSelectionExactlyOne: number = 1;
+  private static _requiredSelectionOneOrMore: number = 2;
+
+  public static Initialize ($listMenu: JQuery): void
+  {
+    ArgumentUtility.CheckNotNullAndTypeIsJQuery("$listMenu", $listMenu);
+
+    $listMenu.keydown (function (event)
+    {
+      ListMenu_OnKeyDown (event, $listMenu);
```
MichaelKetting, line 59: should use the static function, not the global function

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+      case 13: //enter
+      case 32: //space
+        {
+          event.preventDefault();
+          event.stopPropagation();
+
+          if (currentMenuItemIndex >= 0)
+          {
+            let $newMenuItem = $($menuItems[currentMenuItemIndex]);
+            ListMenu_UpdateFocus ($newMenuItem, $oldMenuItem);
```
MichaelKetting, line 225: should use the static function, not the global function

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -0,0 +1,300 @@
...
+
+          var $newMenuItem = $($menuItems[currentMenuItemIndex]);
+          ListMenu_UpdateFocus ($newMenuItem, $oldMenuItem);
+
+          return;
+        }
+    }
+  }
+
+  public static UpdateFocus ($newMenuItem: JQuery, $oldMenuItem: Nullable<JQuery>): void
```
MichaelKetting, line 267: Is this only needed from within the own class or also from other places?

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
a-ctor: Class members cannot be `const`, only function variables can be declared `const`. Or do you mean `readonly`? We do have a task to change `var`s into `let`s/`const`s but no task for `readonly` fields

MichaelKetting: {{readonly}} it is, then. Yes, please add a task for that.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
iam-mholle: > a check is only needed if we wouldn't get an exception otherwise

Well `eval(null)` doesn't throw, it just does nothing/returns `null`. That would make one ugly bug to find.

MichaelKetting: Yes, so it satisfies my condition for "check" -> ticket and todo with ticket number please.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
iam-mholle: I only found references to this function within this file, no idea about downstream projects though, that's why I left the global function as-is.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
iam-mholle: I only found references to this function within this file, no idea about downstream projects though, that's why I left the global function as-is.

MichaelKetting: we may assume they're unused.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
MichaelKetting: we may assume they're unsed.

MichaelKetting (approved):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
a-ctor: @MichaelKetting can I add the `readonly` refactoring to the `let`/`const` task or do you want to create a separate task for it?

MichaelKetting: add it. should fit in nicely.

a-ctor (approved):

PR #37, opened 2021-02-25
RM-7609 Convert the smaller JavaScript files in Core.Web.ClientScript to typescript

a-ctor:


Remotion/Web/ClientScript/Scripts/DatePicker.ts
```diff
@@ -87,11 +94,11 @@ function DatePicker_CloseDatePicker(setFocusOnTarget)
   current.DatePicker.remove();
   if (current.TargetDocument != null)
   {
-    $(current.TargetDocument).unbind('click', current.ClickHandler);
+    $(current.TargetDocument).unbind('click', current.ClickHandler!); // TODO RM-7647: DatePick.DatePicker_CloseDatePicker removes all click handlers from target element if no click handler was specified for the DatePicker
```
MichaelKetting, line 97: The assumtion was that `unbind` will only remove the specified `ClickHandler`, similiar to how C# events work. `current.ClickHandler` should not be null. I do see there's a mismatch with the bind-function. That's a bug.

Remotion/Web/ClientScript/Scripts/DatePicker.ts
```diff
@@ -118,14 +125,14 @@ function DatePicker_Create(datePickerID, button, target, src, width, height)
 
   var frame = window.document.createElement("iframe");
   datePicker.append($(frame));
-  var queryStringConcatenator = src.indexOf ('?') === '-1' ? '?' : '&';
+  var queryStringConcatenator = src.indexOf ('?') as unknown === '-1' ? '?' : '&'; // TODO RM-7646: IndexOf comparison in DatePicker.DatePicker_Create will always yield false due to type mismatch
```
MichaelKetting, line 128: bug

Remotion/Web/ClientScript/Scripts/typings/common.d.ts
```diff
@@ -31,3 +31,9 @@ type Success<T> = { success: true, value: T };
 type NotNullNorUndefined = string | number | boolean | symbol | bigint | object;
 type NotNull = NotNullNorUndefined | undefined;
 type NotUndefined = NotNullNorUndefined | null;
+
+// This is only here until SmartPage is translated to TypeScript
```
MichaelKetting, line 35: Should this have a todo for SmartPage task?

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DatePicker.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+
+class DatePickerFrame {
+  public static Calendar_SelectionChanged(value: string): void
+  {
+    window.parent.DatePicker.UpdateValue(value);
+  }
+}
+
+function DatePicker_ShowDatePicker (button: string, container: HTMLElement, target: HTMLInputElement, src: string, width: number, height: number): void
```
MichaelKetting, line 216: We only want those functions as global functions that are used from C# code. Is this the case for all functions?

And we only need those ex-global functions as public functions, that are used from the global functions (i.e. the C# code in Renderer and ControObjects. This may not be true for any and all global functions of re-motion JS files (see downstream-dependencies), but it is true for DatePicker.

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DatePicker.ts
a-ctor: The global functions are only used by C#.

I don't quite get what you are trying to say with the second paragraph. The global functions are getting removed in a future task and the C# files will use the `DatePicker.ShowDatePicker` syntax instead. Are you saying that we cannot remove the functions because they are used downstream?

MichaelKetting: No, I'm saying we only need those that are actually used in C# but the PR preserves all global functions, which I think contains unsed functions. -> please verify.

a-ctor: Ok I removed most of the methods. Since we are changing the name of the function we are introducing breaking changes. Do you want to create breaking change tickets for each of the methods?

MichaelKetting: Breaking Change: yes, please. Single ticket per type is sufficient (here, DatePicker / BocDateTimeValue). We should then list the changed/removed APIs in the ticket desription.

MichaelKetting: What about making the static functions private when they're not used outside the script?

MichaelKetting (commented):
See thread

Remotion/Web/ClientScript/Scripts/DatePicker.ts
a-ctor: `private` restricts access to the class itself so even the global functions could not access them.

MichaelKetting: Yes. I was referring to the functions used only by the class itself.

a-ctor: Which methods are not public API? All that are not used from C#?

MichaelKetting: That is correct. See
> And we only need those ex-global functions as public functions, that are used from the global functions (i.e. the C# code in Renderer and ControObjects. This may not be true for any and all global functions of re-motion JS files (see downstream-dependencies), but it is true for DatePicker.

a-ctor: Now I understand your previous comment 😄

MichaelKetting (approved):

PR #38, opened 2021-02-25
RM-7587: Upgrade Firefox to Version 85

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7587

### DIFF
--- a/Remotion/Web/Development.WebTesting/CoypuElementScopeExtensions.cs
+++ b/Remotion/Web/Development.WebTesting/CoypuElementScopeExtensions.cs
@@ -125,6 +125,14 @@ namespace Remotion.Web.Development.WebTesting
       {
         scope.ElementFinder.Options.Timeout = TimeSpan.Zero;
 
+        try
+        {
+          scope.Exists (Options.NoWait);
+        }
+        catch (Exception)
+        {
+          // Sometimes, the first scope.Exists() call fails after a switch to an IFrame. So we swallow that first exception and just try again.
+        }
         var exists = scope.Exists (Options.NoWait);
 
         // scope.Exists (...) does not work correctly in some circumstances, whereby we do the exist check via this workaround. See RM-6773 for details.
--- a/Remotion/Web/Development.WebTesting/WebTestConfigurationFactory.cs
+++ b/Remotion/Web/Development.WebTesting/WebTestConfigurationFactory.cs
@@ -57,7 +57,7 @@ namespace Remotion.Web.Development.WebTesting
     /// Represents the latest version of Firefox verified to be compatible with Selenium WebDriver.
     /// In order to achieve a stable testing environment a standalone Firefox with a matching GeckoDriver version should be used.
     /// </summary>
-    protected const string LatestTestedFirefoxVersion = "82";
+    protected const string LatestTestedFirefoxVersion = "85";
 
     /// <summary>
     /// Creates a new <see cref="IBrowserConfiguration"/> from app.config.


raffaelfoidl:
The test `AccessibilityBocDateTimeValueTest.DateOnly_ExpandedDatePicker` (`ObjectBinding.Web.IntegrationTests`) fails continuously (9 out of 9 times).

Exception and Stacktrace:
```c#
OpenQA.Selenium.NoSuchWindowException : Browsing context has been discarded
    at OpenQA.Selenium.Remote.RemoteWebDriver.UnpackAndThrowOnError(Response errorResponse)
    at OpenQA.Selenium.Remote.RemoteWebDriver.Execute(String driverCommandToExecute, Dictionary`2 parameters)
    at OpenQA.Selenium.Remote.RemoteWebDriver.FindElements(String mechanism, String value)
    at Remotion.Web.Development.WebTesting.Accessibility.AccessibilityAnalyzer.IsEmptyFrame() in C:\Build\28109e4c1456d59b\working\Remotion\Web\Development.WebTesting\Accessibility\AccessibilityAnalyzer.cs:line 241
    at Remotion.Web.Development.WebTesting.Accessibility.AccessibilityAnalyzer.InjectIntoIFrames(String source) in C:\Build\28109e4c1456d59b\working\Remotion\Web\Development.WebTesting\Accessibility\AccessibilityAnalyzer.cs:line 228
    at Remotion.Web.Development.WebTesting.Accessibility.AccessibilityAnalyzer.InjectAxeSource(String source) in C:\Build\28109e4c1456d59b\working\Remotion\Web\Development.WebTesting\Accessibility\AccessibilityAnalyzer.cs:line 250
    at Remotion.Web.Development.WebTesting.Accessibility.AccessibilityAnalyzer.GetAccessibilityResult(String cssSelector) in C:\Build\28109e4c1456d59b\working\Remotion\Web\Development.WebTesting\Accessibility\AccessibilityAnalyzer.cs:line 195
    at Remotion.Web.Development.WebTesting.Accessibility.ControlObjectExtensions.Analyze(ControlObject controlObject, AccessibilityAnalyzer analyzer) in C:\Build\28109e4c1456d59b\working\Remotion\Web\Development.WebTesting\Accessibility\ControlObjectExtensions.cs:line 38
    at Remotion.ObjectBinding.Web.IntegrationTests.BocDateTimeValue.AccessibilityBocDateTimeValueTest.DateOnly_ExpandedDatePicker() in C:\Build\28109e4c1456d59b\working\Remotion\ObjectBinding\Web.IntegrationTests\BocDateTimeValue\AccessibilityBocDateTimeValueTest.cs:line 108
```

* I have not been able to reproduce the error from within Visual Studio yet.
* I was able to reproduce the error with the nunit3 console runner (locally + on TC agent).
  * However, the error appeared only once or twice, all subsequent tests were successful.
  * Executing the test more often does not _seem_ to make a difference (tested up until `[Repeat (150)]`)
  * I used the following command to test it with the nunit console (based on the TC log):
```cmd
"<PathToNUnitConsole>\bin\net35\nunit3-console.exe" "C:\temp\Remotion\ObjectBinding\Web.IntegrationTests\bin\Debug\Remotion.ObjectBinding.Web.IntegrationTests.dll" --framework=NET-4.7.2 --test=Remotion.ObjectBinding.Web.IntegrationTests.BocDateTimeValue.AccessibilityBocDateTimeValueTest.DateOnly_ExpandedDatePicker --list-extensions
```
* I have yet to find a way to the make the test red again after a successful for meaningful debugging.
* A test run with FF 82 (which is currently in use in re-motion) was successful, so the culprit is most likely the new browser version.
* FF 86 is already out (a Jira ticket will probably be created soon). Is it worth a try to check if something has changed with the newer version or should we stick to v85 for now?

MichaelKetting:
>     * FF 86 is already out (a Jira ticket will probably be created soon). Is it worth a try to check if something has changed with the newer version or should we stick to v85 for now?
@raffaelfoidl  difficult to say. probably a coin toss at best.

MichaelKetting:
@raffaelfoidl 
I did google this for you and there seems to be a consens that this has been out there for 20 versions already:
https://github.com/mozilla/geckodriver/issues/1553
Has been closed because of inactivity.

https://stackoverflow.com/questions/50824371/c-sharp-selenium-firefox-error-exception-browsing-context-has-been-discarded
Has a random solution option for this.

Bugzilla has an open ticket for something google things is a match
https://bugzilla.mozilla.org/show_bug.cgi?id=1510099
https://bugzilla.mozilla.org/show_bug.cgi?id=1306896

TBH, I don't have a good idea right now. Maybe @Stefan-Ilic and @iam-mholle can offer a suggestion on how to make the code more robust. I think @iam-mholle did some work with Iframes in Firefox last year, so... Maybe it's just a retry-in-production-code-to-make-robust case?

As for the cause, there could just be some memory management optimization in a recent Firefox release that adversely affects this code.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -241,6 +243,35 @@ private bool IsEmptyFrame ()
...
+    /// </summary>
+    private bool FrameLoaded (IWebDriver driver)
+    {
+      driver.SwitchTo().DefaultContent();
+      var iframe = driver.FindElement(By.TagName("iframe"));
+      driver.SwitchTo().Frame(iframe);
+
+      return driver.FindElements(By.TagName("body")).Count > 0;
+    }
+
```
raffaelfoidl, line 274: @iam-mholle 
I don't really know how to implement this change such that it fits nicely into the existing infrastructure. That's why I have pushed it like this for now. I am not sure if this wait mechanism is "trivial enough" to warrant its co-existence alongside with `ScreenshotBocDateTimeValuePickerExtensions` (https://github.com/re-motion/Trunk-SVN2Git/pull/21/files).

I could also imagine creating a corresponding `WebDriverExtensions` class with a slightly more general version of this wait mechanism that is called accordingly from `ScreenshotBocDateTimeValuePickerExtensions` (so as not to cause invocation changes downstream) and `AccessibilityAnalyzer`.

Anyways, I am eager to learn your opinion on this.

MichaelKetting: I'll give @iam-mholle first pick at this but I'd like to add my 2c to it after he's done.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -241,6 +243,35 @@ private bool IsEmptyFrame ()
...
+      }
+    }
+
+    /// <summary>
+    /// Sometimes, due to the delayed generation of the compiled aspx file, the content of the DatePicker iframe is not fully loaded when the iframe is rendered.
+    /// </summary>
+    private bool FrameLoaded (IWebDriver driver)
+    {
+      driver.SwitchTo().DefaultContent();
+      var iframe = driver.FindElement(By.TagName("iframe"));
```
iam-mholle, line 269: This workaround only ensures that the *first* `iframe` is loaded. Imo `WaitUntilFrameIsVisible` should accept the frame element as parameter instead of querying anew (if possible).

iam-mholle (changes requested):
The iframe loading problems would definitly justify either extracting the wait logic into an extension method (but for what type? Element? 😢) or integrating them into own abstractions of the iframe navigation logic. Would definitely be a separate issue. For the scope of this PR however, I would accept the duplication of the workaround. Interested in @MichaelKetting 's opinion.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
raffaelfoidl: I was thinking it was fine because I am querying from within the loop (L231) and somehow thought that restricted to scope to the "current" `iframe`. But what you say makes sense - I am querying for the _first_ `iframe` over and over again, correct? In that case I will give it a try to wait based on the each "current" `iframe` of the loop.

MichaelKetting:
@iam-mholle 
Separate Ticket would be better. We should discuss if this feature needs to be completed before we do the Firefox thing. This would certainly result in a cleaner history and there's need to try and do Firefox quick and dirty because there's no external constraint. The issue also doesn't appear to be strictly related to the Firefox upgrade so fixing it in this PR isn't correct as far as tracking goes.

The method WaitUntilFrameIsVisible would be an extension method in Remotion.Web.Development, samewhere, where we collect these types of utility classes.

What's intersting is the remark regarding compiling the ASPX for the first time. If that's true, one alternative could be that we know about this particular situation and just request the ASPX manually first (and just once per test-run) and then open the IFrame with the already hot ASPX. Unfortunately, this is a very specific solution and, also unfortunately, downstream we have an entire application hosted inside an iframe, so who knows, this extension method might be helpful for other people as well.

raffaelfoidl:
> The issue also doesn't appear to be strictly related to the Firefox upgrade so fixing it in this PR isn't correct as far as tracking goes.

I don't know if that makes a difference or if I misunderstood your reasoning. However, I tested once again (without the proposed "wait-fix") for a few Firefox versions (82, 83, 84, 85, 86) by calling `Remove-Item -Recurse "$env:temp\Temporary ASP.NET Files\root"` in PS every time before executing the test. The test succeeded consistently for 82, 83, 84 and failed consistently for 85, 86. So wouldn't it be justified to argue that it's indeed related to the Firefox upgrade?

MichaelKetting:
@raffaelfoidl no, you got it in one. Fascinating, so yes, it's a timing issue triggered by FF 85 but at the same time, it's a basic raceconditon as well. We had this disucssion a little while ago, how to track such bugs. I think the consensus was that we want to track all bugs as separate issues (on which the browser upgrade then has a depends-on link) as long as they do not fix an API change. That way, we can easily document and track what issues have existed instead of finding things inside browser-upgrade-tickets. We still document in the bug-issue that this only happens with FF 85 (e.g. "iframe racecondition in FF 85" as the ticket summary) but we can make sure the code isn't suceptible to this problem even before the upgrade.

raffaelfoidl:
The changes related to the faulty behaviour discussed above were further examined and implemented in #47.
Hence, the corresponding change set is not part of this PR anymore.

iam-mholle (approved):

MichaelKetting (approved):

PR #45, opened 2021-03-05
RM-7693: Convert DropDownMenu.js Core.Web to TypeScript

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7693


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+  private static _nestedHoverClassName: string = 'nestedHover';
+  private static _currentMenu: Nullable<JQuery> = null;
+  private static _currentPopup: Nullable<HTMLDivElement> = null;
+  private static _currentStatusPopup: Nullable<HTMLDivElement> = null;
+  private static _currentLoadOperation: Nullable<Object> = null;
+  private static _menuItemIDPrefix: string = 'menuItem_';
+  private static _requiredSelectionAny: number = 0;
+  private static _requiredSelectionExactlyOne: number = 1;
+  private static _requiredSelectionOneOrMore: number = 2;
+  private static _itemClickHandler: Nullable<(e: JQueryEventObject) => boolean> = null;
```
a-ctor, line 82: We should introduce names for any free-floating function type. e.g. `type DropDownMenu_ItemClickHandler = ...`

This simplifies future refactorings if needed. You can assign function types that do not match exactly, for example, one that has less parameters.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+  private static _statusPopupRepositionTimer: Nullable<number> = null;
+  private static _blurTimer: Nullable<number> = null;
+  private static _updateFocus: boolean = true;
+
+  public static AddMenuInfo(menuInfo: DropDownMenu_MenuInfo): void
+  {
+    DropDownMenu._menuInfos[menuInfo.ID] = menuInfo;
+  }
+
+  public static BindOpenEvent (openTarget: JQuery, menuID: string, eventType: string, getSelectionCount: Nullable<() => number>, moveToMousePosition: boolean): void
```
a-ctor, line 96: free-floating function type

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+          .bind ("mouseover", function (evt) { openTarget.find ('.' + DropDownMenu._buttonClassName).addClass (DropDownMenu._nestedHoverClassName); })
+          .bind ("mouseout", function (evt) { openTarget.find ('.' + DropDownMenu._buttonClassName).removeClass (DropDownMenu._nestedHoverClassName); });
+    }
+  }
+
+  public static LoadFilteredMenuItems(
+      itemInfos: DropDownMenu_ItemInfo[],
+      loadMenuItemStatus: Nullable<(s: (i: DropDownMenu_ItemInfo[]) => void, e: () => void) => void>,
+      onSuccess: (i: DropDownMenu_ItemInfo[]) => void,
+      onError: () => void): void
```
a-ctor, line 132: Those and any following usages for inline types can then be replaced with their respective type.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+  {
+    if (loadMenuItemStatus !== null)
+    {
+      const currentLoadOperationID = new Object();
+      const onSuccessForLoadMenuItemStatus = function (itemStatusArray: DropDownMenu_ItemInfo[])
```
a-ctor, line 137: We should explicitly add the type annotation to handlers, because there can be similar ducking problems when the function type is changed.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+      {
+        if (DropDownMenu._currentLoadOperation !== currentLoadOperationID)
+          return;
+        DropDownMenu._currentLoadOperation = null;
+
+        const itemStatusMap = itemStatusArray.reduce(
+          function (map: Dictionary<any>, itemStatus)
```
a-ctor, line 144: Should be `Dictionary<DropDownMenu_ItemInfo>` and we should annotate the second parameter as well

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+      if (evt != null)
+      {
+        var bounds = context[0].getBoundingClientRect();
+        if (bounds.left <= evt.clientX &&
+          bounds.right >= evt.clientX &&
+          bounds.top <= evt.clientY &&
+          bounds.bottom >= evt.clientY)
+          event = evt;
+      }
+      DropDownMenu.OpenPopUp(menuID, context, getSelectionCount, event!);
```
a-ctor, line 223:
```suggestion
      DropDownMenu.OpenPopUp(menuID, context, getSelectionCount, event);
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+
+  private static OpenPopUp (menuID: string, context: JQuery, getSelectionCount: Nullable<() => number>, evt: Nullable<JQueryMouseEventObject>): void
+  {
+    ArgumentUtility.CheckNotNullAndTypeIsJQuery('context', context);
+
+    let selectionCount = -1;
+    if (getSelectionCount !== null)
+      selectionCount = getSelectionCount();
+
+    const menuInfo = DropDownMenu._menuInfos[menuID] as DropDownMenu_MenuInfo;
```
a-ctor, line 236:
```suggestion
    const menuInfo = DropDownMenu._menuInfos[menuID]!;
```

Prefer the bang over an `as` cast since. The `as` cast can be bad if the type changes because it can do stuff like: 

```ts
declare var test : number | string;
let a = test as string;
// a is now a string and number is gone
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+      {
+        clearTimeout(statusPopupTimer);
+        if (DropDownMenu._currentStatusPopup && DropDownMenu._currentStatusPopup === statusPopup)
+        {
+          if (statusPopup.childElementCount > 0 && (statusPopup.lastChild as HTMLElement).tagName.toUpperCase() !== 'IFRAME')
+            statusPopup.removeChild(statusPopup.lastChild!);
+          // Remove and re-add the element to force the screenreader (JAWS) to always announce the updated value immediately.
+          const statusPopupParent = statusPopup.parentElement;
+          statusPopupParent!.removeChild(statusPopup);
+          statusPopupParent!.appendChild(statusPopup);
```
a-ctor, line 271:
```suggestion
          const statusPopupParent = statusPopup.parentElement!;
          statusPopupParent.removeChild(statusPopup);
          statusPopupParent.appendChild(statusPopup);
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+
+    // position drop-down list
+    var top = clickEvent ? clickEvent.clientY : Math.max(0, space_top + referenceElement.outerHeight());
+    var left = clickEvent ? clickEvent.clientX : 'auto';
+    var right = clickEvent ? 'auto' : Math.max(0, $(window).width() - referenceElement.offset().left - referenceElement.outerWidth());
+
+    popUpDiv.css('top', top);
+    popUpDiv.css('bottom', 'auto');
+    popUpDiv.css('right', right as string);
+    popUpDiv.css('left', left as string);
```
a-ctor, line 488: This is a bug since css values cannot just have values assigned except for `0` otherwise a unit (e.g. `px` is needed). Additionally we should not type cast here but convert to a string.

```suggestion
    // TODO: Missing unit
    popUpDiv.css('right', "" + right);
    popUpDiv.css('left', "" + left);
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+      const menuPopup = DropDownMenu._currentPopup;
+      const $menuButton = $('a[aria-controls="' + menuPopup.id + '"]');
+      $menuButton.unbind('focus', DropDownMenu.BlurHandler);
+      $menuButton.unbind('blur', DropDownMenu.BlurHandler);
+      const menuButton = $menuButton[0];
+      menuButton.setAttribute('aria-expanded', 'false');
+      menuButton.removeAttribute('aria-controls');
+      if (updateFocus === DropDownMenu._updateFocus)
+        menuButton.focus();
+      // TODO: missing nullcheck?
```
a-ctor, line 539: I don't think that this warrants an own issue. We can assume that an element has a parent and just use `!` BANG

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+      menuPopup.parentElement!.removeChild(menuPopup);
+    }
+
+    if (DropDownMenu._currentStatusPopup !== null
+        && document.body.contains(DropDownMenu._currentStatusPopup))
+    {
+      const statusPopup = DropDownMenu._currentStatusPopup;
+      // Clear the role=alert before removing to item to prevent screenreaders (JAWS) from announcing the old value during removal.
+      statusPopup.removeAttribute('role');
+      // TODO: missing nullcheck?
```
a-ctor, line 549: Same here. Just use bang

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+    {
+      var iconPlaceholder = document.createElement('span');
+      iconPlaceholder.className = DropDownMenu._itemIconClassName;
+      anchor.appendChild(iconPlaceholder);
+    }
+
+    if (itemInfo.DiagnosticMetadata)
+    {
+      // Do not render empty diagnostic metadata attributes
+      $.each(itemInfo.DiagnosticMetadata, function (key, value)
```
a-ctor, line 643:
```suggestion
      $.each(itemInfo.DiagnosticMetadata, function (key: string, value: string | boolean)
```

Otherwise they are assumed as `any` because the jQuery type annotations.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+
+  private static Options_OnKeyDown(event: JQueryKeyEventObject, dropDownMenu: JQuery): void
+  {
+    ArgumentUtility.CheckNotNullAndTypeIsJQuery('dropDownMenu', dropDownMenu);
+
+    if (DropDownMenu._currentPopup == null)
+      return;
+
+    // TODO: bug?
+    var itemInfos = (DropDownMenu._menuInfos[dropDownMenu[0].id] as DropDownMenu_MenuInfo).ItemInfos;
```
a-ctor, line 755:
```suggestion
    var itemInfos = DropDownMenu._menuInfos[dropDownMenu[0].id]!.ItemInfos!;
```

We should not split up the same todo into multiple smaller parts. We can use a `!` at the end to do it once. This is then removed once the bug is fixed. Any usages do not need any further annotation

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+        DropDownMenu.ClosePopUp(DropDownMenu._updateFocus);
+        break;
+      case 40: // down arrow
+        event.preventDefault();
+        event.stopPropagation();
+        isSelectionUpdated = true;
+        oldIndex = currentItemIndex;
+        while (true) // skip separators
+        {
+          if (currentItemIndex < itemInfos!.length - 1) // TODO: nullcheck above
```
a-ctor, line 792:
```suggestion
          if (currentItemIndex < itemInfos.length - 1)
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+            break;
+          if ($(dropDownMenuItems[currentItemIndex]).children('a').length > 0) // skip separators
+            break;
+        }
+        break;
+      case 38: // up arrow
+        event.preventDefault();
+        event.stopPropagation();
+        isSelectionUpdated = true;
+        oldIndex = currentItemIndex >= 0 ? currentItemIndex : itemInfos!.length - 1; // TODO: nullcheck above
```
a-ctor, line 813:
```suggestion
        oldIndex = currentItemIndex >= 0 ? currentItemIndex : itemInfos.length - 1;
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+        while (true) // skip separators
+        {
+          if (currentItemIndex > 0)
+            currentItemIndex--;
+          else
+            currentItemIndex = itemInfos!.length - 1; // TODO: nullcheck above
```
a-ctor, line 819:
```suggestion
            currentItemIndex = itemInfos.length - 1;
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,928 @@
...
+  private static GetTarget (event: JQueryMouseEventObject, tagName: string): Node
+  {
+    var element = event.target as Node & Element;
+    while (element && element.tagName != tagName)
+      element = element.parentNode as Node & Element;
+    // more fun with IE, sometimes event.target is empty, just ignore it then
+    if (!element)
+      return [] as unknown as Node;
+    return element;
+  }
```
a-ctor, line 879:
```suggestion
  private static GetTarget (event: JQueryMouseEventObject, tagName: string): Element
  {
    var element: Nullable<Element> = event.target;
    while (element && element.tagName != tagName)
      element = element.parentNode as Nullable<Element>;
    // more fun with IE, sometimes event.target is empty, just ignore it then
    if (!element) // TODO: What is this? This should also fail by the caller.
      return [] as unknown as Element;
    return element; // TODO: Handle parent possible being null
  }
```

I don't know why this is like it is but the signature should be good, the code inside is just broken.

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Whoops, why is there no transpiler error (itemStatus should be annotated) like for normal methods?

a-ctor: Because a the supplied function can be converted to the implied function. `Dictionary<any>` can be reduced to `Dictionary<BodyList_InfoItem>`

iam-mholle: I meant the missing annotation on the second parameter

a-ctor: `reduce` can infer the parameter type from `itemStatusArray`.

```ts
interface Array<T> {
  reduce<U>(callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: T[]) => U, initialValue: U): U;
}
```

with `U` being the type we are mapping to (`Dictionary<DropDownMenu_ItemInfo>`) and `T` being the type of the array.

a-ctor (commented):
Looks good, MK can take a look now.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+        let previousVisibleItemIsSeparator = false;
+        for (var i = itemInfos.length - 1; i >= 0; i--)
+        {
+          const itemInfo = itemInfos[i];
+          const hasFilterOption = itemInfo.ID !== null;
+          const isSeparator = !hasFilterOption && itemInfo.Text === '-';
+          if (hasFilterOption)
+          {
+            // TODO: What if ID is null?
+            const itemStatus = itemStatusMap[itemInfo.ID!]!;
```
MichaelKetting, line 168: This is taken care of by hasFilterOption, see line 163

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+      const onSuccessForLoadMenuItemStatus = function (itemStatusArray: DropDownMenu_ItemInfo[])
+      {
+        if (DropDownMenu._currentLoadOperation !== currentLoadOperationID)
+          return;
+        DropDownMenu._currentLoadOperation = null;
+
+        const itemStatusMap = itemStatusArray.reduce(
+          function (map: Dictionary<DropDownMenu_ItemInfo>, itemStatus: DropDownMenu_ItemInfo)
+          {
+            // TODO: What if ID is null?
```
MichaelKetting, line 153: The C# type used for itemStatus as a non-null ID.
WebMenuItemProxy

-> declare itemStatus.ID as not-nullable

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+    {
+      const currentLoadOperationID = new Object();
+      const onSuccessForLoadMenuItemStatus = function (itemStatusArray: DropDownMenu_ItemInfo[])
+      {
+        if (DropDownMenu._currentLoadOperation !== currentLoadOperationID)
+          return;
+        DropDownMenu._currentLoadOperation = null;
+
+        const itemStatusMap = itemStatusArray.reduce(
+          function (map: Dictionary<DropDownMenu_ItemInfo>, itemStatus: DropDownMenu_ItemInfo)
```
MichaelKetting, line 151: ItemStatus should be a separate type, not DropDownMenu_ItemInfo. See WebMenuItemProxy

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+      var eventTarget = DropDownMenu.GetTarget (event, "LI");
+      $("li", ul).removeClass(DropDownMenu._itemSelectedClassName);
+      if (eventTarget.firstChild != null && eventTarget.firstChild.nodeName.toLowerCase() === 'a')
+      {
+        $(eventTarget).addClass (DropDownMenu._itemSelectedClassName);
+        (eventTarget.firstChild as HTMLElement).focus();
+      }
+    }).keydown (function (event)
+    {
+      // TODO: missing nullcheck?
```
MichaelKetting, line 398: follow-up task to figure out how to best solve this. _currentMenu can be null. it should never be null at this point.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+      DropDownMenu.Options_OnKeyDown (event, DropDownMenu._currentMenu!);
+    });
+
+    DropDownMenu._itemClickHandler = function(this: HTMLAnchorElement)
```
MichaelKetting, line 402: Is this function a declaration error in the original code and should be written as a private instance function with correct casing.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+    var space_right = $(window).width() - referenceElement.offset().left - referenceElement.width();
+
+    // position drop-down list
+    var top = clickEvent ? clickEvent.clientY : Math.max(0, space_top + referenceElement.outerHeight());
+    var left = clickEvent ? clickEvent.clientX : 'auto';
+    var right = clickEvent ? 'auto' : Math.max(0, $(window).width() - referenceElement.offset().left - referenceElement.outerWidth());
+
+    popUpDiv.css('top', top);
+    popUpDiv.css('bottom', 'auto');
+    // TODO: Missing unit
```
MichaelKetting, line 496: bug ticket. there might already be one for missing units

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+    anchor.setAttribute('tabindex', '-1');
+    if (isEnabled)
+    {
+      anchor.setAttribute('href', '#');
+    }
+    else
+    {
+      anchor.setAttribute('aria-disabled', 'true');
+    }
+    // TODO: missing nullcheck?
```
MichaelKetting, line 604: See earlier comment regarding _itemClickHandler. It cannot be null once we're done.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+          }
+        });
+
+      itemInfo.DiagnosticMetadataForCommand['data-is-disabled'] = isEnabled ? 'false' : 'true';
+
+      $(anchor).attr(itemInfo.DiagnosticMetadataForCommand);
+    }
+
+    var span = document.createElement('span');
+    span.innerHTML = itemInfo.Text!;
```
MichaelKetting, line 678: What happens if Text is null?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+
+  private static Options_OnKeyDown(event: JQueryKeyEventObject, dropDownMenu: JQuery): void
+  {
+    ArgumentUtility.CheckNotNullAndTypeIsJQuery('dropDownMenu', dropDownMenu);
+
+    if (DropDownMenu._currentPopup == null)
+      return;
+
+    // TODO: bug?
+    var itemInfos = DropDownMenu._menuInfos[dropDownMenu[0].id]!.ItemInfos!;
```
MichaelKetting, line 763: both values cannot be null when this code is reached. There's probably not much we can do to make this typesafe right now. Maybe a discriminated union down the road?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+    return isEnabled;
+  }
+
+  private static GetTarget (event: JQueryMouseEventObject, tagName: string): Element
+  {
+    var element: Nullable<Element> = event.target;
+    while (element && element.tagName != tagName)
+      element = element.parentNode as Nullable<Element>;
+    // more fun with IE, sometimes event.target is empty, just ignore it then
+    if (!element) // TODO: What is this? This should also fail by the caller.
```
MichaelKetting, line 882: I don't understand the question. See line 881.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+      return [] as unknown as Element;
+    return element; // TODO: Handle parent possible being null
```
MichaelKetting, line 884: `!element` is meant to not allow `element` in line 884 to be null

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+  private static CreateTitleDivGetter ($context: JQuery): DropDownMenu_TitleDivGetter
+  {
+    const context = $context[0];
+    const contextID = context.id;
+    if (contextID == null)
+    {
+      return function ()
+      {
+        // TODO: nullcheck?
+        return $(context.firstElementChild!);
```
MichaelKetting, line 468: how did we deal with not-supported dom-structures in other files? just ignore or add a dedicated exception?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+      };
+    }
+    else
+    {
+      return function ()
+      {
+        // TODO: shouldn't this be rewritten anyway? getElementById should only be called once here.
```
MichaelKetting, line 475: The developer probably didn't care about making it look nice.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,934 @@
...
+        // TODO: nullability
```
MichaelKetting, line 476: The first part does a null check and the second one returns the never-null firstchild. At least, that's the intention of the code.

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Is it only about tha lookz tho? ;)

a-ctor (approved):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Right, if the `null` check "fails", `null` is returned.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Nothing.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: @MichaelKetting  So I should remove the TODO and leave the `!`'s as they are now?

MichaelKetting: discuss with @a-ctor if a follow-up ticket has any chance of yielding a usable solution

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Why is `[]` returned if `!element`?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: https://re-motion.atlassian.net/browse/RM-7699 should this be added here?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Sorry, I don't understand

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: okay. am wondering if we should still use `|| ''` here to make this more explicit that a) null is properly handled and b) we intend that it's empty in this case.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @iam-mholle decide after checking with @a-ctor: if there's a good idea on how to improve it, we can add a todo with issue reference. otherwise, just drop the todo

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: ah, yes. I think this acted as a null-object for this particular scenario, allwing the callsite to just do NOPs in the IF-case. Sicne it's IE code and IE is dead, it's not relevant either way.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: yes, and the ticket reworded. I think it's okay to fix all the PX errors in one ticket.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: We already have precedent for the case when the DOM rendered by us does not match the Javascript and would thus trigger a null-exception. @a-ctor should know where we found those cases. I want to deal with these cases the same way whenever they pop up.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: We already have a "discriminated union" here: `DropDownMenu_ItemInfo[] | null`. We should just check for the nullablility here as well.

Another solution would be to remove the nullability and use an empty array instead of `null`.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: We have been ignoring cases as it would not make much sense to validate the DOM structure with each call. Ideally the JavaScript would create all the DOM element and store them in the class. This way we wouldn't have to worry about the DOM structure and simply access the elements through the class members.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor I was thinking more along the lines that the implementation doesn't actually allow the null-cases to be triggered by the current code. So, maybe there's a way to redesign the entire file to make this null-safe. The only question is, if this is even sensible because it's definitely too much effort to do this right now. => The dropdownmenu must be in the collection and must have ItemInfos, otherwise, the keydown-event could not have been fired in the first place. 

cc @iam-mholle

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor regarding "ideally" => only when viewing JavaScript alone. We don't do an SPA here, the JavaScript just extends the rendered features. To move "radnom" stuff into JavaScript would make for more complicted client/server interaction and it would also ad to the startup-load of the page.

iam-mholle: What's the decision here? Remove the `TODO` + leave `!`?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: I guess the todo should be removed and calling the `getElementById` twice is OK?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Result of a discussion with MK: Remove todo and leave with both `!`. Both should never be null in this context and this preserves the previous behavior of throwing an error if one of the targets is `null`.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Result of a discussion with MK: Yes, remove `TODO` and leave with a BANG

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: See comment below

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Result of a discussion with MK: we can remove the first `document.getElementById` but we should create an issue here to investigate if returning `null` is something that should be supported

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,937 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+interface DropDownMenuResources
+{
+  LoadFailedErrorMessage: string;
+  LoadingStatusMessage: string;
+}
+
+interface DropDownMenu_ItemStatus
```
MichaelKetting, line 24: move that one after ItemInfo

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -0,0 +1,937 @@
...
+
+  private static BlurHandler (): void
+  {
+    if (DropDownMenu._blurTimer !== null)
+      clearTimeout (DropDownMenu._blurTimer);
+
+    DropDownMenu._blurTimer = setTimeout (function () { DropDownMenu.ClosePopUp (!DropDownMenu._updateFocus); }, 50);
+  }
+
+  private static ItemClickHandler (this: HTMLAnchorElement): boolean
```
MichaelKetting, line 886: Move to after `OnHeadControlClick`

Consider renaming to `OnItemClick`

Since we a) move the method and b) changed the exact name anyway, we can also fix the name. And yes, I'm aware that there's a bunch of different conventions going on in this file.

MichaelKetting (changes requested):

MichaelKetting (approved):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: @a-ctor The fix you provided with `'' + right` does not work here.
The following JQuery types are defined:
image
If a string without `px` is passed in, the outcoming setting is wrong - in my case, the dropdown menu popup is stuck to the right side of the browser. If a number is passed in, the code is working fine. I chose to do `popUpDiv.css('right', right as number);` now, as this satisfies the transpiler and the code is still working fine.

cc @MichaelKetting @Stefan-Ilic

iam-mholle:
@MichaelKetting Ready to merge after https://github.com/re-motion/Framework/pull/45#discussion_r595185223 in [`a9ce5a3` (#45)](https://github.com/re-motion/Framework/pull/45/commits/a9ce5a328e743963fc6cf2c54b1e4fd294ef73a9) ?

PR #46, opened 2021-03-10
RM-7604: Convert BocList.js in ObjectBinding.Web to TypeScript

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RM-7604

BocAutoCompleteReferenceValue tests are failing on this branch, but I have no idea why. Please take a look, maybe you can find it


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+
+  //  Selects a row.
+  //  Adds the row to the _bocList_selectedRows array and increments _bocList_selectedRowsLength.
+  //  bocList: The BocList to which the row belongs.
+  //  rowBlock: The row to be selected.
+  //  isRowHighlightingEnabled: true to mark the selected rows via css-class
+  public static SelectRow(bocList: HTMLElement, rowBlock: BocList_RowBlock, isRowHighlightingEnabled: boolean): void
+  {
+    //  Add currentRow to list  
+    var selectedRows = BocList._selectedRows[bocList.id];
```
a-ctor, line 268: Here we can also just assume that the DOM structure is correct and use a `!` to override nullability here in the declaration of `selectedRows`. We can then remove the `if` checks again.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+    {
+      var rowBlock = selectedRows?.Rows[rowID];
+      if (rowBlock != null)
+      {
+        BocList.UnselectRow (bocList, rowBlock, isRowHighlightingEnabled);
+      }
+    }
+    
+    //  Start over with a new array
+    selectedRows?.Clear();
```
a-ctor, line 302:
```suggestion
    var selectedRows = BocList._selectedRows[bocList.id]!;
    for (var rowID in selectedRows.Rows)
    {
      var rowBlock = selectedRows.Rows[rowID];
      if (rowBlock != null)
      {
        BocList.UnselectRow (bocList, rowBlock, isRowHighlightingEnabled);
      }
    }
    
    //  Start over with a new array
    selectedRows.Clear();
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+
+  //  Unselects a row.
+  //  Removes the row frin the _bocList_selectedRows array and decrements _bocList_selectedRowsLength.
+  //  bocList: The BocList to which the row belongs.
+  //  rowBlock: The row to be unselected.
+  //  isRowHighlightingEnabled: true to update the row's css-class.
+  public static UnselectRow(bocList: HTMLElement, rowBlock: BocList_RowBlock, isRowHighlightingEnabled: boolean): void
+  {
+    //  Remove currentRow from list
+    var selectedRows = BocList._selectedRows[bocList.id];
```
a-ctor, line 313: Sames as above. Assert non-nullability using `!` and restore the old code without the ifs.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+    if (isRowHighlightingEnabled)
+      $(rowBlock.Row).removeClass(BocList.TrClassNameSelected);
+
+    if (selectedRows != undefined)
+    {
+      BocList.ClearSelectAllRowsSelector (selectedRows);
+    }
+  }
+
+  public static SetSelectAllRowsSelectorOnDemand(selectedRows: BocList_SelectedRows): void
```
a-ctor, line 331:
```suggestion
  public static SetSelectAllRowsSelectorOnDemand (selectedRows: BocList_SelectedRows): void
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+  {
+    if (selectedRows.DataRowCount == selectedRows.Length && selectedRows.DataRowCount > 0)
+      selectedRows.SelectAllSelectorControls?.each (function (this: JQuery) { this.checked = true; });
```
a-ctor, line 334:
```suggestion
      selectedRows.SelectAllSelectorControls!.each (function (this: JQuery) { this.checked = true; });
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+  }
+
+  public static ClearSelectAllRowsSelector(selectedRows: BocList_SelectedRows)
```
a-ctor, line 337:
```suggestion
  public static ClearSelectAllRowsSelector (selectedRows: BocList_SelectedRows)
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+  {
+    selectedRows?.SelectAllSelectorControls?.each (function (this: JQuery) { this.checked = false; });
```
a-ctor, line 339:
```suggestion
    selectedRows!.SelectAllSelectorControls!.each (function (this: JQuery) { this.checked = false; });
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+    });
+  }
+}
+
+function BocList_InitializeGlobals(): void
+{
+  BocList.InitializeGlobals();
+}
+
+function BocList_InitializeList(bocList: HTMLElement, selectRowSelectorControlName: string, selectAllSelectorControlName: string, selection: Dictionary<HTMLElement> | number, hasClickSensitiveRows: boolean, onSelectionChangedHandler: { (boclist: HTMLElement, flag: boolean): void }): void
```
a-ctor, line 691: The `selection` parameter should be `number` if I am not mistaken. This should then also be changed in the `BocList_SelectedRows` class.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+      }
+      else
+      {
+        return false;
+      }
+    });
+  }
+}
+
+function BocList_InitializeGlobals(): void
```
a-ctor, line 686: Check for each global function if they are used from C# code or not. If they are not we can remove them from here and make the class member `private`.

Remotion/ObjectBinding/Web.ClientScript/tsconfig.json
```diff
@@ -15,6 +15,8 @@
     "./Scripts/**/*.ts"
   ],
   "files": [
-    "../../Web/ClientScript/Scripts/typings/common.d.ts"
+    "../../Web/ClientScript/Scripts/typings/common.d.ts",
+    "../../Web/ClientScript/Scripts/Utilities.ts",
```
a-ctor, line 19: This temporary fix can go after rebasing on top of develop.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+  //  Applies the checked state of the title's selectorControl to all data rows' selectu=ion selectorControles.
+  //  bocList: The BocList to which the selectorControl belongs.
+  //  selectRowControlName: The name of the row selector controls.
+  //  isRowHighlightingEnabled: true to enable highting of the rows
+  public static OnSelectAllSelectorControlClick(bocList: HTMLElement, selectAllSelectorControl: JQuery, isRowHighlightingEnabled: boolean): void
+  {
+    var selectedRows = BocList._selectedRows[bocList.id];
+
+    if (selectedRows?.Selection != BocList.rowSelectionMultiple)
+      return;
```
a-ctor, line 352:
```suggestion
    var selectedRows = BocList._selectedRows[bocList.id]!;

    if (selectedRows.Selection != BocList.rowSelectionMultiple)
      return;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+    //  BocList_SelectRow will increment the length, therefor initialize it to zero.
+    if (selectAllSelectorControl.checked)
+      selectedRows.Length = 0;
+  
+    selectedRows.SelectRowSelectorControls?.each (function (this: JQuery)
```
a-ctor, line 357: We might be missing a null check here
```suggestion
    // TODO: Missing null check?
    selectedRows.SelectRowSelectorControls!.each (function (this: JQuery)
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+      {
+        if (pageNumberField.val().length > 0)
+          setTimeout(function () { pageNumberField.focus(); }, 0);
+  
+        pageNumberField.val(String(parseInt(pageIndexField.val(), 10) + 1));
+        return false;
+      }
+      else
+      {
+        pageIndexField.val (String(pageNumber - 1));
```
a-ctor, line 649: I actually do not know if `String(...)` works correctly because it is a constructor and should be used with `new`. I would just concat with an empty string.
```suggestion
        pageNumberField.val("" + (parseInt(pageIndexField.val(), 10) + 1));
        return false;
      }
      else
      {
        pageIndexField.val ("" + (pageNumber - 1));
```

a-ctor (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Stefan-Ilic: Is that really the TypeScript way to do it? To me it looks like a dumb JavaScript hack 😂

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,779 @@
...
+  public static _rowSelectionDisabled = 0;
+
+  public static _selectedRows: Dictionary<BocList_SelectedRows>;
+
+  //  The css class used for rows in their selected state.
+  public static TrClassNameSelected = 'selected';
+
+  public static _rowSelectionSingleCheckBox = 1;
+  public static _rowSelectionSingleRadioButton = 2;
+  public static rowSelectionMultiple = 3;
```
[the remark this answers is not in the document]
Stefan-Ilic: I made them public on purpose because they were `var`s without a scope before, meaning they were globally accessible. Is that incorrect?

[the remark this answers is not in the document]
a-ctor: They were publicly available, but actually private APIs. So we can move them around. Since we already moved them into the class any code that depends on them is broken anyway. `public` vs. `private` has no effect on the resulting JavaScript. `public` and `private` are TypeScript features that cause compile time exceptions. Any `private` member is still publicly available in JavaScript land

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: TypeScript is a superset of JavaScript so dumb hacks also work in TypeScript ;)

Converting to a string like this is also common in C#. You could use `.toString()` or a template string as alternative options.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Stefan-Ilic: @a-ctor The thing is, in the JavaScript version, there is a comment right below `selection` that statates
```
//  Associative Array: <SelectorControl ID>, <BocList_RowBlock>
```
However, numbers are being assigned to it as well, which has me confused. Do you think I should still just change it to a number?

a-ctor: If it works with just the `number` type we should remove the `Dictionary`. From my quick glance it is only compared to numbers and initialized to numbers. The comment might be outdated and probably can be removed. You can blame the comment to make sure.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,716 @@
...
+  public Rows: Dictionary<Nullable<BocList_RowBlock>>;
+  public SelectRowSelectorControls: Nullable<JQuery>;
+  public SelectAllSelectorControls: Nullable<JQuery>;
+  public DataRowCount: number;
+
+  constructor(selection: number)
+  {
+    this.Selection = selection;
+    this.Length = 0;
+    this.Rows = {}
```
a-ctor, line 33:
```suggestion
    this.Rows = {};
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,716 @@
...
+    this.SelectRowSelectorControls = null;
+    this.SelectAllSelectorControls = null;
+    this.DataRowCount = 0;
+  }
+
+  public Clear(): void
+  {
+    this.Length = 0;
+    this.Rows = {}
```
a-ctor, line 42:
```suggestion
    this.Rows = {};
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,716 @@
...
+    BocList.ClearSelectAllRowsSelector (selectedRows);
+  }
+
+  private static SetSelectAllRowsSelectorOnDemand (selectedRows: BocList_SelectedRows): void
+  {
+    if (selectedRows.DataRowCount == selectedRows.Length && selectedRows.DataRowCount > 0)
+      selectedRows.SelectAllSelectorControls!.each (function (this: JQuery) { this.checked = true; });
+  }
+
+  private static ClearSelectAllRowsSelector (selectedRows: BocList_SelectedRows)
```
a-ctor, line 328:
```suggestion
  private static ClearSelectAllRowsSelector (selectedRows: BocList_SelectedRows): void
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,716 @@
...
+
+function BocList_GetSelectionCount (bocListID: string): number
+{
+  return BocList.GetSelectionCount (bocListID)
+}
+
+function BocList_FixUpScrolling(bocList: JQuery): void
+{
+  BocList.FixUpScrolling(bocList);
+}
```
a-ctor, line 706: Those two do not need to be public if I am not mistaken.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,716 @@
...
+    }
+  
+    return false;
+  }
+
+  public static FixUpScrolling(bocList: JQuery): void
+  {
+    var tableBlock = bocList.children('div.bocListTableBlock').first();
+
+    var scrollTimer: number | null = null;
```
a-ctor, line 458:
```suggestion
    var scrollTimer: Nullable<number> = null;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,716 @@
...
+    {
+      // For Internet Explorer + JAWS 2018ff, the tabindex-attribute on the table root will break a table with a scrollable header part.
+      tableContainer.removeAttr('tabindex');
+      var captionLabelIDs = bocList.attr('aria-labelledby');
+      var lastCaptionLabelID = captionLabelIDs.split(' ').slice(-1)[0];
+      var captionElement = document.getElementById (lastCaptionLabelID);
+      var tableCaption = $ ('<span/>')
+        .attr ({
+          'tabindex' : 0,
+          'aria-label': captionElement?.innerText,
```
a-ctor, line 508:
```suggestion
      var captionElement = document.getElementById (lastCaptionLabelID)!;
      var tableCaption = $ ('<span/>')
        .attr ({
          'tabindex' : 0,
          'aria-label': captionElement.innerText,
```

a-ctor (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Stefan-Ilic: `BocList_GetSelectionCount` is used in `BocList.cs` Line 3694, but you're right about `BocList_FixUpScrolling`

a-ctor (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+
+  private static BindRowClickEventHandler(bocList: HTMLElement, row: HTMLElement, cell: string, selectorControl: JQuery): void
+  {
+    $(row).click(function (evt)
+    {
+      var hasSelectionChanged = BocList.OnRowClick (evt, bocList, row, selectorControl);
+      if (hasSelectionChanged)
+      {
+        var selectedRows = BocList._selectedRows[bocList.id];
+        selectedRows!.OnSelectionChanged (bocList, false);
```
MichaelKetting, line 166: consistency please. we should place the `!` in the same way when accessing the same members. @a-ctor do we have the bang mostly at the usage site or when getting the variable? I think it's probably best to add it after getting the variable instead of at the usage site since there could be multiple usages and we want to express that this variable isn't null to begin with.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+    BocList.ClearSelectAllRowsSelector (selectedRows);
+  }
+
+  private static SetSelectAllRowsSelectorOnDemand (selectedRows: BocList_SelectedRows): void
+  {
+    if (selectedRows.DataRowCount == selectedRows.Length && selectedRows.DataRowCount > 0)
+      selectedRows.SelectAllSelectorControls!.each (function (this: JQuery) { this.checked = true; });
+  }
+
+  private static ClearSelectAllRowsSelector (selectedRows: BocList_SelectedRows): void
```
MichaelKetting, line 328: the parameter is not nulalble so the first bang is extra.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+    var selectedRows = BocList._selectedRows[bocList.id]!;
+
+    if (selectedRows.Selection != BocList.rowSelectionMultiple)
+      return;
+    //  BocList_SelectRow will increment the length, therefor initialize it to zero.
+    if (selectAllSelectorControl.checked)
+      selectedRows.Length = 0;
+  
+    // TODO: Missing null check?
+    selectedRows.SelectRowSelectorControls!.each (function (this: JQuery)
```
MichaelKetting, line 349: No, this can't be null at this point. but we probably want a follow-up task to make the BocList_SelectedRows object more than a DTO by moving all the select-features onto that type. Please add a ticket for this and probably should add a TODO reference where we use Bang with selectedRows.Select...

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+      selectedRows.Length = 0;
+  
+    selectedRows.OnSelectionChanged (bocList, false);
+  }
+
+  //  Event handler for the selection selectorControl in a data row.
+  //  Sets the _bocList_isSelectorControlClick flag.
+  private static OnRowSelectorClick(): void
+  {
+    BocList._isRowSelectorClickExecuting = true; //TODO either the documentation is incorrect or the body of the method
```
MichaelKetting, line 378: Docs wrong. either way, the docs pretty much useless where and in the next function. not that doc-cleanup is a focus for the current task...

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+  
+    pageNumberField.bind('change', function () {
+  
+      var pageNumber = parseInt(pageNumberField.val(), 10);
+      if (isNaN (pageNumber) || !TypeUtility.IsInteger (pageNumber))
+      {
+        if (pageNumberField.val().length > 0)
+          setTimeout(function () { pageNumberField.focus(); }, 0);
+  
+        pageNumberField.val("" + (parseInt(pageIndexField.val(), 10) + 1));
```
MichaelKetting, line 636: we use single-quotes for JS strings

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+        return false;
+      }
+      else
+      {
+        pageIndexField.val ("" + (pageNumber - 1));
```
MichaelKetting, line 641: see above

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
MichaelKetting, line 636: In a previous PR, we had a cast to number instead of adding a blank string. Maybe, that was just because the other code was jquery. There is also a ticket for improving these things. Please add a TODO. (@a-ctor Ticket please)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+{
+  BocList.OnCommandClick()
+}
+
+function BocList_GetSelectionCount (bocListID: string): number
+{
+  return BocList.GetSelectionCount (bocListID)
+}
+
+function BocList_CreateFakeTableHead(tableContainer: JQuery, scrollableContainer: JQuery, bocList: JQuery): void
```
MichaelKetting, line 703: I think that one we don't need.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -0,0 +1,711 @@
...
+        return;
+  
+      BocList.FixHeaderSize(scrollableContainer);
+      BocList.FixHeaderPosition(tableContainer, scrollableContainer);
+      setTimeout(resizeHandler, resizeInterval);
+    };
+    resizeHandler();
+  }
+
+  public static CreateFakeTableHead(tableContainer: JQuery, scrollableContainer: JQuery, bocList: JQuery): void
```
MichaelKetting, line 496: private

MichaelKetting (changes requested):
There wasn't really that much to do in this review

@a-ctor

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: When getting the variable is the preferred place, since it needs less `!` when the variable is used multiple times. Additionally, if the variable assignment changes, there is not need to hunt for leftover `!` across all usages.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: This code is also jQuery but the case your are mentioning was `.css(name, value)` which had an overload for number. `.val` does not have an overload for number (at least the typings do not have one, jQuery converts any numbers to strings). So in this case we are stuck with converting to a `string` or adjusting the typings to add an overload with `number`.

MichaelKetting: Thanks. probably should do this consistently in the cleanup-PX task after the jquery migration

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Stefan-Ilic: @MichaelKetting Done, see [RM-7711](https://re-motion.atlassian.net/browse/RM-7711)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: @MichaelKetting So change to `'' + ` and create a cleanup issue after the jQuery migration?

MichaelKetting: The clenaup issue we already have. Change the quotes now.

Stefan-Ilic: What's the number of the clean up issue? I can't find it anywhere

Stefan-Ilic: For continuity, it's [RM-7699](https://re-motion.atlassian.net/browse/RM-7699)

MichaelKetting (approved):

PR #3, opened 2021-01-08
RM-7543 Log Browser and Webdriver version in WebTestHelper

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7543

PR #47, opened 2021-03-12
RM-7704: Race Condition in AccessibilityAnalyzer regarding IFrames in Firefox Version 85

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7704

### DIFF
--- a/Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
+++ b/Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
@@ -34,6 +34,8 @@ namespace Remotion.Web.Development.WebTesting.Accessibility
   /// </summary>
   public class AccessibilityAnalyzer
   {
+    private static readonly TimeSpan s_defaultMaximumTimeToWaitForFrame = TimeSpan.FromMilliseconds (3000);
+
     /// <summary>
     /// Creates an <see cref="AccessibilityAnalyzer"/> that uses an instance of <see cref="RemoteWebDriver"/> for both <see cref="IWebDriver"/>
     /// and <see cref="IJavaScriptExecutor"/>.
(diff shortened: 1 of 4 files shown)


Remotion/Web/Development.WebTesting/WebElementExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Waits until an IFrame is fully rendered.
+    /// </summary>
+    /// <param name="frame">The <see cref="IWebElement"/> representing the IFrame to wait for.</param>
+    /// <param name="timeout">A maximum timeout until the IFrame must be visible.</param>
+    public static void WaitUntilFrameIsVisible (this IWebElement frame, int timeout = 3000)
+    {
+      var webDriver = ((IWrapsDriver) frame).WrappedDriver;
+      var webDriverWait = new WebDriverWait (webDriver, TimeSpan.FromMilliseconds (timeout))
+                          {
+                              PollingInterval = TimeSpan.FromMilliseconds (timeout / 10)
```
iam-mholle, line 40: A lower interval might be alright here.

iam-mholle (approved):

MichaelKetting (changes requested):

raffaelfoidl:
> Follow-up task to update the 3 APIs that use an int timeout (WaitUntilVisible) to use a Timespan.

Follow-up task: https://re-motion.atlassian.net/browse/RM-7713. I assumed it's going to be backported, please disagree if that's not the case, then I'll adapt the ticket.

MichaelKetting:
> 
> 
> > Follow-up task to update the 3 APIs that use an int timeout (WaitUntilVisible) to use a Timespan.
> 
> Follow-up task: https://re-motion.atlassian.net/browse/RM-7713. I assumed it's going to be backported, please disagree if that's not the case, then I'll adapt the ticket.

Ticket's been updated re backport. Those would be breaking changes and are not triggered by new features so, no backport. please don't forget the breaking change ticket in v3.0 implementation.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -225,8 +225,10 @@ private bool AxeIsInjected ()
     /// </remarks>
     private void InjectIntoIFrames (string source)
     {
+      var maximumTimeToWaitForFrame = TimeSpan.FromMilliseconds (3000);
```
MichaelKetting, line 228: Okay, I'm afraid this would require us to actually pass the timeout from the public API all the way to down-here. We can define the actual default value once in here, but we should allow injecting a new value.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -225,8 +225,10 @@ private bool AxeIsInjected ()
...
       foreach (var frame in WebDriver.FindElements (By.TagName ("iframe")))
       {
+        frame.WaitUntilFrameIsVisible (maximumTimeToWaitForFrame);
```
MichaelKetting, line 231: how did we find out that there's two issues, once here and one in DateTimeValue? I don't remember which tests failed so I just assumed it was only in DateTimeValueExntesions that we had to add this guard (right now).

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
raffaelfoidl: The guard in DateTimeValueExtensions stems from some time ago and was Chrome-related. The actual reason this ticket exists is that `AccessibilityBocDateTimeValueTest.DateOnly_ExpandedDatePicker` fails with Firefox >= 85, hence we are waiting for the frame to render. The changes in `ScreenshotBocDateTimeValuePickerExtensions` from this PR are purely to avoid/eliminate code duplication because it's exactly the same workaround, just with a specifiable selector so that the new extension method can be used universally.

MichaelKetting (commented):
Looks good, please restructre history

#1 new extension method incl. overload with all changes in single commit
#2 fix bug
#3 remove code duplication

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -220,13 +220,17 @@ private bool AxeIsInjected ()
     /// <summary>
     /// Injects Axe into every iframe manually because axe-core cannot do it itself.
     /// </summary>
+    /// <param name="source">Source of axe.min.js to inject.</param>
+    /// <param name="maximumTimeToWaitForFrame">Maximum wait time after which an iframe must be visible such that Axe can be injected. Default is 3000 ms.</param>
     /// <remarks>
     /// Source: https://www.deque.com/blog/writing-automated-tests-accessibility/
     /// </remarks>
-    private void InjectIntoIFrames (string source)
+    private void InjectIntoIFrames (string source, TimeSpan? maximumTimeToWaitForFrame = null)
```
MichaelKetting, line 228: In the AccessibilityAnalyzer, we need to allow injecting the value into the public members. It's just that it's OK to define the default value in here. Although, the longer I think about it, the less I like it though, that we define the default in the private method. Maybe it would be best to make a static field for the default and apply it at the call sites when the public method is passed a null-value. 

=> each public API gets a nullable timeout parameter. When the parameter is null, we null-propagate the static-field as fallback.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
raffaelfoidl: Okay, it seems I misunderstood your previous comment a bit. I hope it's okay now.

Fyi, I have implemented your comment only tentatively in a new commit and will move the change to the second commit ("bug fix") if you okay it. (I think you wouldn't get a nice diff with only the new change if I had rebased it just now)

MichaelKetting (approved):
Looks good now :) Hope the final design ideas seemed like good ideas during implementation.

PR #49, opened 2021-03-17
RM-7605: Convert BocAutoCompleteReferenceValue.js and related JavaScript files in ObjectBinding.Web to TypeScript (Part 1)

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RM-7605


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+  {
+    ArgumentUtility.CheckTypeIsObject('command', command);
+    command.removeAttr('href');
+    command.removeAttr('onclick');
+    command.removeAttr('title');
+    command.removeAttr('target');
+    command.removeClass('hasIcon');
+  };
+}
+
```
MichaelKetting, line 213: ?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+type ResourceIdentifier = {
```
MichaelKetting, line 18: No public names without the type name as prefix. 

`BocReferenceValueBase_Resources`

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+  LoadIconFailedErrorMessage: string,
+  LoadDataFailedErrorMessage: string,
+  SearchStringForDropDownDoesNotMatchRegexMessage: string,
+  NoDataFoundMessage: string,
+  InternetExplorerScreenReaderLabelText: string
+}
+
+type SearchStringValidationInfo = {
```
MichaelKetting, line 26: BocReferenceValueBase_SearchStringValidationInfo

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+  ValidSearchStringRegex: string,
+  ValidSearchStringForDropDownRegex: string,
+  SearchStringForDropDownDoesNotMatchRegexMessage: string,
+  IgnoreSearchStringForDropDownUponValidInput: string
+}
+
+type CommandInfo = {
```
MichaelKetting, line 33: BocReferenceValueBase_CommandInfo

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+  href: Nullable<string>,
+  target: Nullable<string>,
+  onClick: Nullable<string>,
+  title: Nullable<string>
+}
+
+type IconContext = {
```
MichaelKetting, line 40: BocReferenceValueBase_IconContext

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+  businessObjectClass: Nullable<string>,
+  arguments: Nullable<string>
+}
+
+type ErrorHandler = (error: Sys.Net.WebServiceError) => void
```
MichaelKetting, line 45: BocReferenceValueBase_ErrorHandler

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+
+type IconInformation = {
```
MichaelKetting, line 47: BocReferenceValueBase_IconInformation

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+    {
+      var pageRequestManager = Sys.WebForms.PageRequestManager.getInstance();
+      if (pageRequestManager.get_isInAsyncPostBack())
+        return oldCommand;
+
+      var params: any = { businessObject: businessObject };
+      for (var propertyName in iconContext)
+        params[propertyName] = (iconContext as any)[propertyName];
+
+      Sys.Net.WebServiceProxy.invoke(
```
MichaelKetting, line 94: Use WebServiceProxyInvokeFunction<TResult, TUserContext> as type. See Utilities.WebServiceUtility.

Also, TODO for changing implementation to use WebServiceUtility.Execute instead of doing the calls manually here and in BocAutoCompleteReferenceValue.jquery.js. I'd almost say that we change it immediately but it's just a bit out of scope and we don't really have that much extra work just fixing the typings right now.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+    oldCommand.replaceWith(newCommand);
+
+    if (isIconUpdateEnabled)
+    {
+      var pageRequestManager = Sys.WebForms.PageRequestManager.getInstance();
+      if (pageRequestManager.get_isInAsyncPostBack())
+        return oldCommand;
+
+      var params: any = { businessObject: businessObject };
+      for (var propertyName in iconContext)
```
MichaelKetting, line 91: @a-ctor This is a simple property-copy from one type to another. Is there a better way to write this? Should we make a utility for this?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+    for (var i = 0; i < oldCommandAttributes.length; i++)
+    {
+      var value = oldCommandAttributes[i].nodeValue;
+      if (value != null && value != '')
+        newCommand.attr(oldCommandAttributes[i].nodeName, value);
+    }
+
+    for (var property in tempCommandInfo)
+    {
+      var value: Nullable<string> = "" + (tempCommandInfo as any)[property]; // TODO is there a better way to do this?
```
MichaelKetting, line 154: @a-ctor see todo, but given we're going to be dropping the command, we shouldn't invest here.

also, single quotes instead of double quotes.

MichaelKetting, line 154: While string + null = null, this is still really non-intutive in the end. We should not use string contact to mimic casting or auto-conversion.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+      var params: any = { businessObject: businessObject };
+      for (var propertyName in iconContext)
+        params[propertyName] = (iconContext as any)[propertyName];
+
+      Sys.Net.WebServiceProxy.invoke(
+          controlServiceUrl!,
+          'GetIcon',
+          false,
+          params,
+          function (result: string, eventArgs: Sys.EventArgs): void
```
MichaelKetting, line 99: `result` is not of type string byt of type `IconInformation`

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,223 @@
...
+
+type IconInformation = {
+  Url: string,
+  AlternateText: string,
+  ToolTip: string,
+  Width: number,
+  Height: number
+}
+
+class BocReferenceValueBase
```
MichaelKetting, line 55: This is from it's intention a base type for the two ReferenceValue-types. @a-ctor is there a way to express this in TypeScript? If so, we might want to consider a follow-up task to improve this code.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValue.ts
```diff
@@ -0,0 +1,105 @@
...
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+class BocReferenceValue {
```
MichaelKetting, line 18: open parens in next line, drop the blank lines

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValue.ts
```diff
@@ -0,0 +1,105 @@
...
+    ArgumentUtility.CheckNotNullAndTypeIsBoolean('isAutoPostBackEnabled', isAutoPostBackEnabled);
+    ArgumentUtility.CheckNotNullAndTypeIsBoolean('isIconUpdateEnabled', isIconUpdateEnabled);
+    if (isIconUpdateEnabled)
+      ArgumentUtility.CheckNotNullAndTypeIsString('controlServiceUrl', controlServiceUrl);
+    if (isIconUpdateEnabled)
+      ArgumentUtility.CheckNotNullAndTypeIsObject('iconContext', iconContext);
+    ArgumentUtility.CheckTypeIsObject('commandInfo', commandInfo);
+    ArgumentUtility.CheckNotNullAndTypeIsObject('resources', resources);
+
+    dropDownList.change(function () {
```
MichaelKetting, line 43: fix open-parens in file.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,323 @@
...
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+type Item = { UniqueIdentifier: string, DisplayName: string }
```
MichaelKetting, line 18: `BocAutoCompleteReferenceValue_Item`

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    BocAutoCompleteReferenceValue._itemBackUp = null;
+    BocAutoCompleteReferenceValue._isInvalidated = false;
+    BackupItemData(hiddenField.val(), textbox.val());
+    BocAutoCompleteReferenceValue._command = command;
+    BocAutoCompleteReferenceValue._commandBackUp = command;
+    BocAutoCompleteReferenceValue._selectListID = baseID + '_Results';
+    BocAutoCompleteReferenceValue._informationPopUpID = baseID + '_Information';
+
+    if (BrowserUtility.GetIEVersion() > 0) {
```
MichaelKetting, line 72: fix open-parens in file

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,323 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+type Item = { UniqueIdentifier: string, DisplayName: string }
+
+class BocAutoCompleteReferenceValue
+{
+
+  private static _itemBackUp: Nullable<Item>;
```
MichaelKetting, line 23: Those are instance fields

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    var hiddenField = <HTMLInputElement | null>document.getElementById(referenceValueHiddenFieldID);
+    if (hiddenField == null || hiddenField.value == nullValueString)
+      return 0;
+
+    return 1;
+  };
+}
+
+function Initialize(
```
MichaelKetting, line 298: this would be a static function on the class.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,323 @@
...
+      return searchString;
+    }
+  };
+
+  //  Returns the number of rows selected for the specified ReferenceValue
+  public static GetSelectionCount (referenceValueHiddenFieldID: string, nullValueString: string): number {
+    ArgumentUtility.CheckNotNullAndTypeIsString('referenceValueHiddenFieldID', referenceValueHiddenFieldID);
+    ArgumentUtility.CheckNotNullAndTypeIsString('nullValueString', nullValueString);
+
+    var hiddenField = <HTMLInputElement | null>document.getElementById(referenceValueHiddenFieldID);
```
MichaelKetting, line 290: Nullable<HtmlInputElement>

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
a-ctor: Class inheritance is done using `extends` instead of `:` but otherwise behaves similarly

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
a-ctor: `Object.assign(params, iconContext)`

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
MichaelKetting: sounds good. @Stefan-Ilic please make a follow-up ticket with TODO comment for this.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
MichaelKetting: @Stefan-Ilic please add a ticket + todo for this.
@a-ctor do we have other places where this is already needed / already a ticket for this?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
a-ctor: `tempCommandInfo` should have the type `Dictionary<unknown>`, allowing arbitrary properties to exist on the object. Since the property is not actually limited to the four properties that are currently assigned.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
a-ctor: There are some places where `jQuery.extends` is used which is the same functionality. But for this "raw" js usage we do not have an issue yet

MichaelKetting: okay, so those issues get fixed during the jquery migration anyway. we might want to fix this issue together with the query stuff, just because it's the same usecase, but I'm not tied to it.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
Stefan-Ilic: I'm afraid I don't know what `?` exactly means, please take a moment to appreciate this comment from my point of view

image

MichaelKetting: There's a bunch ob empty lines starting right at the ? and going all the way down to the end of the file. I do admit, I thought it would enough of a hint to point right at the issue without spelling it out. My bad.

Stefan-Ilic: I'm afraid spelling these things out is necessary probably more often than not, especially without the proper context. I would have understood if you had made a multi line comment showing most of the empty lines (those are there btw because it's very much a work in progress after all, I didn't review it myself yet because I'm working on the other files)

Nonetheless, thanks for clarifying that bit, now I get it

MichaelKetting: Ah, multi-line comments, that's a thing? Cool. The comment overview instead of looking at the sources, there is often an embarassing lack of context. Github could do better in this type of situation.

Stefan-Ilic: yes the multi line thing actually helps a lot

hZntfql633

Stefan-Ilic: my cursor looks like it had too much coffee but you get the gist

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
Stefan-Ilic: Is this okay?:

```js
var value: Nullable<string> = String(tempCommandInfo[property]);
```

MichaelKetting: @a-ctor ?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1512,12 +1577,12 @@
...
 
       var popUpDiv = popUp.children ('div');
-      var contentHeight = Math.max(0, Math.max(popUpDiv.children().map(function () { return this.offsetHeight + this.offsetTop; }).get()));
+      var contentHeight = Math.max(0, Math.max(parseInt(String(popUpDiv.children().map(function (this: JQuery) { return this.offsetHeight + this.offsetTop; }).get()))));
 
       var contentWidth = popUp.data('popUpContentWidth');
       if (!isVisibe)
       {
-        contentWidth = Math.max(0, Math.max(popUpDiv.children().map(function () { return this.offsetWidth + this.offsetLeft; }).get()));
+        contentWidth = Math.max(0, Math.max(parseInt(String(popUpDiv.children().map(function (this: JQuery) { return this.offsetWidth + this.offsetLeft; }).get()))));
```
a-ctor, line 1585: These are both issues and we should for now do a `as any` cast. The conversion to string and parsing back to number has a different behavior from the previous code. But the previous code is already broken since an array is passed to `Math.max` (without a spread operator). 

This has the following wonderful behvior:
```js
Math.max( [] ); // 0
Math.max( [34] ); // 34
Math.max( [34, 12] ); // NaN
```

This should be a bug issue

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1486,11 +1537,15 @@
         var documentWidth = number($(window).width());
         var documentHeight = number($(window).height());
 
-        var space = new Object();
-        space.top = offset.top - scrollTop;
-        space.bottom = documentHeight - ((offset.top + height) - scrollTop);
-        space.left = offset.left - scrollLeft;
-        space.right = documentWidth - ((offset.left + width) - scrollLeft);
+        var space: Space = {
```
a-ctor, line 1540: That type is not quite correct because the `spaceVertical`, `spaceHorizontal`, and `space` are actually never null. For now just do an `any` type or save all the properties in variables and create the type from that. This way we can remove the nullability from the `Space` type.

a-ctor (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
Stefan-Ilic: Does that work? C# only calls static members and methods and does not save an instance of the class to operate on it as far as I'm aware

MichaelKetting: That's not correct. The initialize function got all the parameters and did a ton of closures and thus passed the values along when intantiating the closure. Now, a bunch of statics is populated with information that's specfic to individual calls of Initialize. Best case, it works but the data is just invalid after each invocation and the final invocations leaves said invalid data behind. Worst case: something gets captured and a field is lazy evaluated and then the subsequent calls have wrong state.

Either way: the expressed semantics would be wrong.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
Stefan-Ilic: Thanks for the nice examples, I'll create a ticket right away

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,211 @@
...
+    ArgumentUtility.CheckTypeIsString('businessObject', businessObject);
+
+    var newCommand = $('<a/>');
+
+    var tempCommandInfo: Dictionary<unknown> = {
+      href: null,
+      target: null,
+      onClick: null,
+      title: null
+    }
```
a-ctor, line 125: The initialization is not needed here and might cause problem in certain situations. We can also type the dictionary to `Nullable<string>`

```suggestion
    var tempCommandInfo: Dictionary<Nullable<string>> = {};
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,211 @@
...
+    for (var i = 0; i < oldCommandAttributes.length; i++)
+    {
+      var value = oldCommandAttributes[i].nodeValue;
+      if (value != null && value != '')
+        newCommand.attr(oldCommandAttributes[i].nodeName, value);
+    }
+
+    for (var property in tempCommandInfo)
+    {
+      var value: Nullable<string> = String(tempCommandInfo[property]);
```
a-ctor, line 153: We can keep the original code here but need to use `const` instead of `var` to make TypeScript happy.

```suggestion
      const value = tempCommandInfo[property];
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValueBase.ts
```diff
@@ -0,0 +1,211 @@
...
+      {
+        tempCommandInfo.href = String.format(commandInfo.href, businessObject);
+      }
+    }
+
+    var oldCommandAttributes = oldCommand[0].attributes;
+
+    for (var i = 0; i < oldCommandAttributes.length; i++)
+    {
+      var value = oldCommandAttributes[i].nodeValue;
```
a-ctor, line 146: Use `const` here since we need to use it below.

```suggestion
      const value = oldCommandAttributes[i].nodeValue;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocReferenceValue.ts
```diff
@@ -0,0 +1,115 @@
...
+  public static Initialize (
+    dropDownList: JQuery,
+    command: JQuery,
+    nullValueString: string,
+    isAutoPostBackEnabled: boolean,
+    isIconUpdateEnabled: boolean,
+    controlServiceUrl: string,
+    iconContext: BocReferenceValueBase_IconContext,
+    commandInfo: BocReferenceValueBase_CommandInfo,
+    resources: BocReferenceValueBase_Resources): void
```
a-ctor, line 29:
```suggestion
    dropDownList: JQuery,
    command: Nullable<JQuery>,
    nullValueString: string,
    isAutoPostBackEnabled: boolean,
    isIconUpdateEnabled: boolean,
    controlServiceUrl: string,
    iconContext: BocReferenceValueBase_IconContext,
    commandInfo: Nullable<BocReferenceValueBase_CommandInfo>,
    resources: BocReferenceValueBase_Resources): void
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,352 @@
...
+    combobox: JQuery, textbox: JQuery, hiddenField: JQuery, button: JQuery, command: JQuery, controlServiceUrl: string,
+    completionSetCount: number, dropDownDisplayDelay: number, dropDownRefreshDelay: number, selectionUpdateDelay: number,
+    searchStringValidationInfo: BocReferenceValueBase_SearchStringValidationInfo,
+    nullValueString: string,
+    isAutoPostBackEnabled: boolean,
+    searchContext: boolean,
+    isIconUpdateEnabled: boolean,
+    iconContext: BocReferenceValueBase_IconContext,
+    commandInfo: BocReferenceValueBase_CommandInfo,
+    resources: BocReferenceValueBase_Resources)
```
a-ctor, line 40:
```suggestion
  constructor (
    baseID: Nullable<string>,
    combobox: JQuery, textbox: JQuery, hiddenField: JQuery, button: JQuery, command: Nullable<JQuery>, controlServiceUrl: string,
    completionSetCount: number, dropDownDisplayDelay: number, dropDownRefreshDelay: number, selectionUpdateDelay: number,
    searchStringValidationInfo: BocReferenceValueBase_SearchStringValidationInfo,
    nullValueString: string,
    isAutoPostBackEnabled: boolean,
    searchContext: object,
    isIconUpdateEnabled: boolean,
    iconContext: BocReferenceValueBase_IconContext,
    commandInfo: Nullable<BocReferenceValueBase_CommandInfo>,
    resources: BocReferenceValueBase_Resources)
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,352 @@
...
+// 
+
+type BocAutoCompleteReferenceValue_Item = { UniqueIdentifier: string, DisplayName: string }
+
+class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes BocReferenceValue and BocAutoCompleteReferenceValue inherit from BocReferenceValueBase
+{
+  private _itemBackUp: Nullable<BocAutoCompleteReferenceValue_Item>;
+  private _isInvalidated: boolean;
+  private _command: JQuery;
+  private _commandBackUp: JQuery;
```
a-ctor, line 25:
```suggestion
  private _command: Nullable<JQuery>;
  private _commandBackUp: Nullable<JQuery>;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,352 @@
...
+    combobox: JQuery, textbox: JQuery, hiddenField: JQuery, button: JQuery, command: JQuery, controlServiceUrl: string,
+    completionSetCount: number, dropDownDisplayDelay: number, dropDownRefreshDelay: number, selectionUpdateDelay: number,
+    searchStringValidationInfo: BocReferenceValueBase_SearchStringValidationInfo,
+    nullValueString: string,
+    isAutoPostBackEnabled: boolean,
+    searchContext: boolean,
+    isIconUpdateEnabled: boolean,
+    iconContext: BocReferenceValueBase_IconContext,
+    commandInfo: BocReferenceValueBase_CommandInfo,
+    resources: BocReferenceValueBase_Resources): void
```
a-ctor, line 337:
```suggestion
  public static Initialize(
    baseID: Nullable<string>,
    combobox: JQuery, textbox: JQuery, hiddenField: JQuery, button: JQuery, command: Nullable<JQuery>, controlServiceUrl: string,
    completionSetCount: number, dropDownDisplayDelay: number, dropDownRefreshDelay: number, selectionUpdateDelay: number,
    searchStringValidationInfo: BocReferenceValueBase_SearchStringValidationInfo,
    nullValueString: string,
    isAutoPostBackEnabled: boolean,
    searchContext: object,
    isIconUpdateEnabled: boolean,
    iconContext: BocReferenceValueBase_IconContext,
    commandInfo: Nullable<BocReferenceValueBase_CommandInfo>,
    resources: BocReferenceValueBase_Resources): void
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,352 @@
...
+        }
+        out.Value = actualItem;
+      }
+      finally
+      {
+        this._isInvalidated = false;
+      }
+    });
+
+    function RestoreCommand(autocomplete: BocAutoCompleteReferenceValue)
```
a-ctor, line 249: We should not modify the existing functions but instead use the `this` parameter. Since those functions also use the parameters of the function it gets a bit tricky since we cannot convert them to methods.

Instead we can convert the methods into variables and assign lambdas to them. e.g. `const RestoreCommand = () => { ... }`. This allows them to use `this` instead of introducing a new parameter. Since those methods are used above we need to moved them above their usage.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,352 @@
...
+      {
+        extraParams: searchContext,
+        isAutoPostBackEnabled: isAutoPostBackEnabled,
+        nullValue: nullValueString, // the hidden field value indicating that no value has been selected
+        searchStringValidationParams:
+        {
+          inputRegex: new RegExp(searchStringValidationInfo.ValidSearchStringRegex),
+          dropDownTriggerRegex: new RegExp(searchStringValidationInfo.ValidSearchStringForDropDownRegex),
+          dropDownTriggerRegexFailedMessage: searchStringValidationInfo.SearchStringForDropDownDoesNotMatchRegexMessage,
+          getDropDownSearchString: function (this: BocAutoCompleteReferenceValue, searchString: string)
```
a-ctor, line 126: Be careful when adding a `this: xxx` parameter to a function. TypeScript cannot validate if the correct `this` is provided. In this case, `this` will not be what you expect it to be. Instead a lambda expression should be used to capture the current `this` context.

a-ctor (changes requested):
`BocAutoCompleteReferenceValue` still needs a bit of work, the other files just have minor typing mismatches.

Remotion/ObjectBinding/Web.ClientScript/Scripts/typings/jquery/jquery.d.ts
```diff
@@ -2331,7 +2331,7 @@ interface JQuery {
      * @param eventType A string containing one or more DOM event types, such as "click" or "submit," or custom event names.
      * @param handler A function to execute each time the event is triggered.
      */
-    bind(eventType: string, handler: (eventObject: JQueryEventObject) => any): JQuery;
+    bind(eventType: string, handler: (eventObject: JQueryEventObject, ...eventHandlerArgs: any[]) => any): JQuery;
```
Stefan-Ilic, line 2334: @a-ctor I chased the bugs for a while and it finally seems that this was it. It doesn't say anywhere in the official docs that this is supposed to be the signature but it's apparently hidden away in the examples of the `bind` [docs](https://api.jquery.com/bind/):

```js
$( "p" ).bind( "myCustomEvent", function( e, myName, myValue ) {
  $( this ).text( myName + ", hi there!" );
  $( "span" )
    .stop()
    .css( "opacity", 1 )
    .text( "myName = " + myName )
    .fadeIn( 30 )
    .fadeOut( 1000 );
  });
```

I deleted the second argument anywhere where it was used since it didn't match the signature provided in this file and it looks like that's what caused the builds to fail

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,353 @@
...
+    {
+      if (this._isInvalidated)
+        return;
+
+        this._isInvalidated = true;
+
+      hiddenField.val(nullValueString);
+      UpdateCommand(nullValueString);
+      //Do not fire change-event
+    }).updateResult((e: any, item: BocAutoCompleteReferenceValue_Item, out: any) => //TODO what is this
```
a-ctor, line 244: The type of `out` is `{ Value: Nullable<BocAutoCompleteReferenceValue_Item> }` but we should create a separate type for it. Maybe `BocAutoCompleteReferenceValue_UpdateResult` and use it here as type. And `e` is of type `JQueryEventObject`.

a-ctor (changes requested):

a-ctor (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -807,11 +852,11 @@
             }
         };
 
-        function parse(data) {
+        function parse(data: string | string[]) {
```
MichaelKetting, line 855: Where do we get data as `string[]`?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -807,11 +852,11 @@
...
             var parsed = [];
-            var rows = data.split("\n");
+            var rows = (data as string).split("\n");
             for (var i = 0; i < rows.length; i++) {
-                var row = $.trim(rows[i]);
+                var row: string | string[] = $.trim(rows[i]);
```
MichaelKetting, line 859: same question, how does row become a `string[]`?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -858,27 +903,27 @@
         isAutoPostBackEnabled: false,
         extraParams: {},
         // re-motion: changed selectFirst from boolean field to function
-        selectFirst: function(inputValue, searchTerm) { return true; },
-        formatItem: function(row) { return { html : row[0], class : null, isAnnotation : false }; },
+        selectFirst: function(inputValue: string, searchTerm: string) { return true; },
+        formatItem: function(row: JQuery) { return { html : row[0], class : null, isAnnotation : false }; },
```
MichaelKetting, line 907: why is row a `JQuery` object instead of string.

@a-ctor I'm starting to wonder if we don't want to do a follow-up task where we replace a ton of those strings with value types to add some kind of typesafety.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -31,10 +31,53 @@
...
+    isAutoPostBackEnabled: boolean,
+    extraParams: Dictionary<string | number> ,
+    params: Dictionary<any>,
+    searchStringValidationParams: { inputRegex: { test: (x: any) => boolean }, dropDownTriggerRegex: { test: (x: any) => boolean }, getDropDownSearchString: (x: string) => string, dropDownTriggerRegexFailedMessage: string },
+    noDataFoundMessage: string,
+    selectFirst: (val: any, q: Nullable<string>) => any,
+    max: number,
+    handleRequestError: (err: Sys.Net.WebServiceError) => void,
+    parse: (result: any) => any,
+    formatResult: (row: string[], row2?: string) => any
```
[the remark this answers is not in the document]
MichaelKetting: `data` and `value` should probably get an actual type. Needs checking what type this would be, but I think we already have them defined in the BocAutoCompleteReferenceValue.ts file

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1031,17 +1076,17 @@
...
             .appendTo(innerDiv)
-            .mouseover(function (event) {
-                if (target(event).nodeName && target(event).nodeName.toUpperCase() === 'LI')
+            .mouseover(function (event: JQueryMouseEventObject) {
+                if ((target(event) as Node).nodeName && (target(event) as Node).nodeName.toUpperCase() === 'LI')
                 {
-                  active = $("li", list).removeClass(CLASSES.ACTIVE).index(target(event));
-                  var activeItem = $(target (event));
+                  active = $("li", list).removeClass(CLASSES.ACTIVE).index(target(event) as JQuery);
+                  var activeItem = $(target (event) as JQuery);
```
MichaelKetting, line 1083: why do we have to JQuery-cast the value we want to pass into the JQuery 'constructor'?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1554,7 +1605,7 @@
         //Reset height if content has changed to calculate new height
         popUp.css ({ height : 'auto' });
       }
-      var elementHeight = popUp.outerHeight();
+      var elementHeight: number | string = popUp.outerHeight();
```
MichaelKetting, line 1608: string or number? Why is this needed?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1486,11 +1531,15 @@
         var documentWidth = number($(window).width());
         var documentHeight = number($(window).height());
 
-        var space = new Object();
-        space.top = offset.top - scrollTop;
-        space.bottom = documentHeight - ((offset.top + height) - scrollTop);
-        space.left = offset.left - scrollLeft;
-        space.right = documentWidth - ((offset.left + width) - scrollLeft);
+        var space: any = {
```
MichaelKetting, line 1534: this should be a type

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1471,9 +1516,9 @@
         }
     };
 
-    $.Autocompleter.calculateSpaceAround = function(element) {
+    $.Autocompleter.calculateSpaceAround = function(element: JQuery): any {
         // re-motion: make sure CSS values are allways numbers, IE can return 'auto'
-        function number(num) {
+        function number(num: any) {
```
MichaelKetting, line 1521: isn't this typed to `number|string`?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1444,7 +1489,7 @@
         };
     };
 
-    $.Autocompleter.Selection = function(field, start, end, focusInputAfterSelection) {
+    $.Autocompleter.Selection = function(field: JQuery & HTMLElement, start: number, end: number, focusInputAfterSelection: boolean) {
```
MichaelKetting, line 1492: how can the `field` type be both jquery and HTML?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1362,11 +1407,11 @@
         };
     };
 
-    $.Autocompleter.InformationPopUp = function(options, input) {
+    $.Autocompleter.InformationPopUp = function(options: Options, input: HTMLElement) {
 
         var needsInit = true;
-        var element = null;
-        var repositionTimer = null;
+        var element: Nullable<JQuery> = null;
```
MichaelKetting, line 1413: this should get a TODO to refactor the nullablity

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -974,18 +1019,18 @@
...
-        var listItems,
+        var listItems: JQuery,
     active = -1,
-    data = null,
+    data: any = null,
     term = "",
     needsInit = true,
-    element,
-    list;
+    element: Nullable<JQuery>,
```
MichaelKetting, line 1032: needs a TODO to refactor the nullablity

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,355 @@
...
+    ArgumentUtility.CheckNotNullAndTypeIsString('nullValueString', nullValueString);
+    ArgumentUtility.CheckNotNullAndTypeIsBoolean('isAutoPostBackEnabled', isAutoPostBackEnabled);
+    ArgumentUtility.CheckNotNullAndTypeIsObject('searchContext', searchContext);
+    ArgumentUtility.CheckNotNullAndTypeIsBoolean('isIconUpdateEnabled', isIconUpdateEnabled);
+    if (isIconUpdateEnabled)
+      ArgumentUtility.CheckNotNullAndTypeIsObject('iconContext', iconContext);
+    ArgumentUtility.CheckTypeIsObject('commandInfo', commandInfo);
+    ArgumentUtility.CheckNotNullAndTypeIsObject('resources', resources);
+
+    const BackupItemData = (uniqueIdentifier: string, displayName: string) =>
```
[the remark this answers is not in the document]
Stefan-Ilic: I don't have the sufficient experience with JavaScript/TypeScript to engage in such a conversation, please ask @a-ctor, who told me to do it like this in the first place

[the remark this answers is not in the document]
Stefan-Ilic: As a reference, this is what he wrote as a comment, maybe that's even enough:

> We should not modify the existing functions but instead use the this parameter. Since those functions also use the parameters of the function it gets a bit tricky since we cannot convert them to methods.

> Instead we can convert the methods into variables and assign lambdas to them. e.g. const RestoreCommand = () => { ... }. This allows them to use this instead of introducing a new parameter. Since those methods are used above we need to moved them above their usage.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -31,10 +31,53 @@
 // Significantly modified for re-motion
 // ************************************************
 
+interface JQueryStatic { Autocompleter: any; }
```
MichaelKetting, line 34: @Stefan-Ilic , updated with @a-ctor: We're going to keep the jquery.js file in this branch and move the typescript version to a new PR @a-ctor will use to do his part-2 of the file. the intention has always been to invest the minimal effort needed to get typescript to work (just sprinkle some annotations on it). Since we don't merge all files into a single typescript file right away, we can defer changing the jquery.js file to typescript. This allows us to do a single, cohesive update of the file instead of dealing with all those half-truths in the typesystem.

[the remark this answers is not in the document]
Stefan-Ilic: I don't think I fully understand. Would you like me to change every invocation of `$.Autocompleter` so that it's not a static member of JQueryStatic  anymore? Would you like me to change the name `Autocompleter` to something else?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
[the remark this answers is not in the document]
MichaelKetting: @Stefan-Ilic I understand where @a-ctor is coming from (or I think I do) but I don't think is is necessary. Most usages already access fields based on `this` so calling functions based on `this` should be the same. What remains are one or two usage sites where we have nested closures. there, we could just capture `this` in a variable and call the function based on this variable.

This design allows us to keep using instance functions, does not require moving code blocks around and is generally more consistent with the rest of the migration.

What I did notice (but I don't think we should change this right now) that the anonynous functions added to the jquery object (see `updateResult`, `invalidateResult`, ...) could perhaps also be changed to instance functions which are then simply added as a delegate to the jquery object. this would perhaps result in an even cleaner design since it creates better defined chunks of code instead of this jquery-spagetti we have right now.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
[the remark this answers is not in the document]
MichaelKetting: No, it's a naming thing. You added a globally accessible type that is specific to BocAutoCompleteReferenceValue but uses a generic name. Therefor, someone else could wish to introduce a type of the same name and run into problems.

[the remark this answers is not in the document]
Stefan-Ilic: But it needs to be named `JQueryStatic` so we can access `Autocompleter` on `$`. Perhaps I'm still misunderstanding you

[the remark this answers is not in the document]
MichaelKetting: Ah, okay. @a-ctor I'd like some more details on this.

@Stefan-Ilic so, maybe it's just the 'regular' types, then, like `Options` in line 36 that need renaming.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
Stefan-Ilic: It was, and then I reverted it because @a-ctor suggested so

> That type is not quite correct because the spaceVertical, spaceHorizontal, and space are actually never null. For now just do an any type or save all the properties in variables and create the type from that. This way we can remove the nullability from the Space type.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
Stefan-Ilic: Alright, then I will revert the file and disregard all comments concerning it

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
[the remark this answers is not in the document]
Stefan-Ilic: I will make the changes you requested, but I do hope you understand how frustrating it is to repeatedly work in and revert conflicting change requests from two different reviewers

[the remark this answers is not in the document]
MichaelKetting: @a-ctor, @Stefan-Ilic is correct about the missing alignment. This type and it's constraints is simply unfortunate. In the end we want the most amount of statically typed code and try to avoid closures when we can write clean code in a differnt way.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
MichaelKetting: > 
> 
> Alright, then I will revert the file and disregard all comments concerning it

Please don't forget to create a backup of the file in a separate branch. @a-ctor will probably want to start his work off yours :)

MichaelKetting: > 
> 
> I also thought the intention is to do the minimum work to make TypeScript work and simply sprinkle annotations on it, which is why many of your comments regarding this file confused me, since they seemed to contradict that notion, at least to me

Yeah, there was just way more work done and way less typing for the work involved to make sense any more. But mostly, with the option to just not create the bundled file right now, a lot of the constraints went out the window.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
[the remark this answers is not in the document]
a-ctor: There are two problems in this file: 
 a) parameters are used in the helper functions and 
 b) due to the class syntax local variables cannot be accessed in the helper functions without capturing.

Here are some possible solutions:

```js
const ClearError = () =>
{
  if (textbox.hasClass('error'))
  {
    textbox.removeClass ('error');
    textbox.trigger ('hideInformationPopUp');
  }
}

// Usage:
ClearError(); // must be after the declaration
```

Creating closures solves both problems, but has the downside of needing to move the declarations before its usage. This was the current solution I came up with.

```js
// this.textbox has to be set in the constructure

// Definition
private ClearError()
{
  if (this.textbox.hasClass('error'))
  {
    this.textbox.removeClass ('error');
    this.textbox.trigger ('hideInformationPopUp');
  }
}

// Usage:
this.ClearError();
// but if it needs to be passed as a delegate
passDel(this.ClearError.bind(this))
```

Another solution would be to convert the helper methods into instance methods and set any parameters that are used in those methods as fields. But whenever the method is passed as a delegate the `this` parameter must be fixed in the method using `.bind(this)`. Otherwise `this` is not set when the method is called.

```js
// this.textbox has to be set in the constructure

// Definition
ClearError = () =>
{
  if (this.textbox.hasClass('error'))
  {
    this.textbox.removeClass ('error');
    this.textbox.trigger ('hideInformationPopUp');
  }
};

// Usage:
this.ClearError();
```

Probably the best solution would be to do the previous solution but declare the helper functions as closures. This has the benefit that the function already captures `this` so you can just use it without any `.apply`. Parameters still need to be set to instance fields, but you also no do not have to reorder members.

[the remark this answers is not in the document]
MichaelKetting: After rereading, yes, we have a couple (~4) captured parameters that would have to be converted to fields, but the closures I was originally referring to, those don't get used via delegates and should be easily convertible to instance functions. 
* BackupItemData
* UpdateCommand
* GetDropDownSearchStringForValidInput 
* (I hope I got all)

Thanks for the reminder regarding the delagates, that would just be syntax clenaup anway and create more noise than it would fix right now. Not saying that we won't do it later.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,356 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+type BocAutoCompleteReferenceValue_Item = { UniqueIdentifier: string, DisplayName: string };
+
+type BocAutoCompleteReferenceValue_UpdateResult = { Value: Nullable<BocAutoCompleteReferenceValue_Item> }
+
+class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes BocReferenceValue and BocAutoCompleteReferenceValue inherit from BocReferenceValueBase
+{
+  private _itemBackUp: Nullable<BocAutoCompleteReferenceValue_Item>;
```
MichaelKetting, line 24: type: Nullable<BocAutoCompleteReferenceValue_ItemBackup>

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,356 @@
...
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+
+type BocAutoCompleteReferenceValue_Item = { UniqueIdentifier: string, DisplayName: string };
```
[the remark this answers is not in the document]
MichaelKetting: We might be able to use Item instead of ItemBackup, but right now, I'm not sure if this will be a good design so it's easier to keep them separate for now and only unify once we know more about the required shapes.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,356 @@
...
+        informationPopUpID: this._informationPopUpID,
+        dropDownButtonID: button.attr('id'),
+        inputAreaClass: 'content',
+        // this can be set to true/removed once the problem is fixed that an empty textbox still selects the first element, making it impossible to clear the selection
+        selectFirst: function (inputValue: string, searchTerm: string)
+        {
+          return inputValue.length > 0;
+        },
+        dataType: 'json',
+        parse: function (data: any)
```
MichaelKetting, line 185: data can be assigned a type. I think it's an array of `row`, and row is of type BocAutoCompleteReferenceValue_Item

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,356 @@
...
+    {
+      if (this._isInvalidated)
+        return;
+
+        this._isInvalidated = true;
+
+      hiddenField.val(this._nullValueString);
+      this.UpdateCommand(this._nullValueString, errorHandler);
+      //Do not fire change-event
+    }).updateResult((e: JQueryEventObject, item: BocAutoCompleteReferenceValue_Item, out: BocAutoCompleteReferenceValue_UpdateResult) =>
```
MichaelKetting, line 242: type BocAutoCompleteReferenceValue_Item stays here

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,356 @@
...
+        isAutoPostBackEnabled: this._isAutoPostBackEnabled,
+        nullValue: this._nullValueString, // the hidden field value indicating that no value has been selected
+        searchStringValidationParams:
+        {
+          inputRegex: new RegExp(searchStringValidationInfo.ValidSearchStringRegex),
+          dropDownTriggerRegex: new RegExp(searchStringValidationInfo.ValidSearchStringForDropDownRegex),
+          dropDownTriggerRegexFailedMessage: searchStringValidationInfo.SearchStringForDropDownDoesNotMatchRegexMessage,
+          getDropDownSearchString: (searchString: string) =>
+          {
+            return searchStringValidationInfo.IgnoreSearchStringForDropDownUponValidInput ? this.GetDropDownSearchStringForValidInput(searchString) : searchString;
```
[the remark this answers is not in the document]
a-ctor: Since a lambda is used to defined the function `this` from the outer scope is captured. The containing function is the constructor so `this` will be the instance passed to the constructor.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,356 @@
...
+      if (combobox.attr('aria-describedby') !== undefined)
+      {
+        textbox.attr('aria-describedby', combobox.attr('aria-describedby'));
+        combobox.removeAttr('aria-describedby');
+      }
+
+      combobox = textbox;
+    }
+
+    const ClearError = () =>
```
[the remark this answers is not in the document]
Stefan-Ilic: If `ClearError` becomes an instance method, `textbox` must become an instance variable and `clearRequestError` passed to `this._textbox.autocomplete` must become a lambda, otherwise it looks like it can't capture the `this` context, but if we do that, it is possible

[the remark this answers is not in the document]
MichaelKetting: Yes, the instance-field for `textbox` would beneeded. `clearRequestError` I don't yet understand. quick session before/after standup?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
MichaelKetting: ah yes. Okay, then we can't use a separate type for Item and ItemBackup and instead have to do everything with Item. This might discover some issues where not all relevant information is copied when creating the backup. in this case, we preservice the current behavior by assigning NULL/FALSE to the uninitialzied fields and add a bug-ticket + TODO comment for this.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -0,0 +1,343 @@
...
+            row.IsAnnotation = row.UniqueIdentifier === nullValueString;
+
+            return {
+              data: row,
+              value: row.UniqueIdentifier,
+              result: row.IsAnnotation ? '' : row.DisplayName
+            };
+          });
+        },
+        formatItem: function (item: JQuery) //What we display on input box
```
MichaelKetting, line 185: This review comment is still unresolved:
> I think this should be BocAutoCompleteReferenceValue_Item, not JQuery

Also, interesting question: why does this compile? does the JQuery type have `any` quality?

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
Stefan-Ilic: Yes, I overlooked this comment, thanks! I believe the JQuery type has some type of "catch-all" member

MichaelKetting (approved):

PR #50, opened 2021-03-18
Feature/rm 7674 remove type script global functions use class members instead

a-ctor:


Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -1281,19 +1281,3 @@ class SmartPage_Context
...
-{
-  SmartPage_Context.Instance!.OnStartUp(isAsynchronous, isDirty);
-}
-
-function SmartPage_PageRequestManager_pageLoaded(sender: Sys.WebForms.PageRequestManager, args: Sys.WebForms.PageLoadedEventArgs): void
-{
-  SmartPage_Context.Instance!.PageRequestManager_pageLoaded(sender, args);
-}
-
-function RenderThisHtml(theHtml: string): void
```
MichaelKetting, line 1296: this needs a separate breaking change doc. ages ago, we added this as a utility function to deal with ActiveX.

Remotion/Web/ClientScript/Scripts/ExecutionEngine.ts
```diff
@@ -253,29 +253,29 @@ class WxePage_Context
 
     WxePage_Context._instance = instance;
   }
-}
 
-function WxePage_OnLoading(hasSubmitted: boolean, isCached: boolean, isAsynchronous: boolean): void
-{
-  WxePage_Context.GetInstance()!.OnLoading(hasSubmitted, isCached, isAsynchronous);
-}
+  public static OnLoading (hasSubmitted: boolean, isCached: boolean, isAsynchronous: boolean): void
```
MichaelKetting, line 257: Question: we pass the function name in C#
`SmartPage.RegisterClientSidePageEventHandler(...,"WxeContext.OnLoading") and then use `fct = eval(functionName);` in JavaScript to get the function for this string. Does this also work for dotted function names?

MichaelKetting (changes requested):

Remotion/Web/Core/UI/SmartPageImplementation/SmartPageInfo.cs
```diff
@@ -96,7 +96,7 @@ public void RegisterClientSidePageEventHandler (SmartPageEvents pageEvent, strin
     {
       ArgumentUtility.CheckNotNullOrEmpty ("key", key);
       ArgumentUtility.CheckNotNullOrEmpty ("function", function);
-      if (! Regex.IsMatch (function, @"^([a-zA-Z_][a-zA-Z0-9_]*)$"))
+      if (! Regex.IsMatch (function, @"^([a-zA-Z_][a-zA-Z0-9_]*(\.[a-zA-Z_][a-zA-Z0-9_]*)*)$"))
```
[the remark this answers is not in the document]
MichaelKetting: Okay, so we can supply dotted functions. See other comment/question on this.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
a-ctor: You mean a separate breaking change ticket for the removal of `RenderThisHtml`? Where was this function even used? I could not find any usages in JS or C#

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -412,15 +412,15 @@ class SmartPage_Context
     var pageRequestManager = this.GetPageRequestManager();
     if (pageRequestManager != null)
     {
-      pageRequestManager.remove_pageLoaded(SmartPage_PageRequestManager_pageLoaded);
-      pageRequestManager.add_pageLoaded(SmartPage_PageRequestManager_pageLoaded);
+      pageRequestManager.remove_pageLoaded(this.PageRequestManager_pageLoaded.bind(this));
```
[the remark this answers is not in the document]
a-ctor: `this.PageRequestManager_pageLoaded` is a member function so the `this` parameter must set accordingly. Just passing the method into the handler might cause a wrong `this` to be assigned. `.bind` creates a new function where the this parameter is explicitly set to the current `this` parameter.

Lambda expressions in JavaScript do basically the same thing, they capture the current `this`. `() => this.Test()` is equivalent to `(function() { this.Test() }).bind(this)`

Remotion/Web/ClientScript/Scripts/ExecutionEngine.ts
a-ctor: With `eval("WxeContext.OnLoading")` it does, with something like `window["WxeContext.OnLoading"]` it wouldn't

MichaelKetting: ah, okay. Could you please consider if there's more elegant ways to write this event dispatching logic and if those would require the global function syntax? If there's none, we can keep the new logic.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
MichaelKetting: Exactly. This function wasn't used by re-motion but downstream. It may well be that the last usages have been removed by now anyway, but we need to do a breaking change doc. And given that it stood appart from the rest of the APIs, a separate ticket is helpful.

Remotion/Web/ClientScript/Scripts/ExecutionEngine.ts
a-ctor: I don't think that there is a more elegant way to do this. But it would probably be a good idea to get rid of the `eval` and use a custom resolver that looks up the function name from the window object.

MichaelKetting: As discussed, we should keep the gobal-function requirement here. It keeps future code simpler and we can always change our minds later on.

MichaelKetting (approved):

PR #51, opened 2021-03-19
Feature/rm 7706 create TypeScript utility class for complex jQuery constructs

a-ctor:


Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -200,6 +200,52 @@ class ArgumentUtility
     ArgumentUtility.CheckNotNull(name, value);
     ArgumentUtility.CheckTypeIsJQueryObject(name, value);
   };
+
+  // Resolves any provided css selector, checks the element for null and returns it.
+  public static ResolveCssSelectorOrElement<TElement extends Element> (selectorOrElement: CssSelectorOrElement<TElement>, context?: ParentNode): TElement
```
MichaelKetting, line 205: Um, you just added this to the existing ArgumentUtility class.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -200,6 +200,52 @@ class ArgumentUtility
...
+  {
+    if (TypeUtility.IsString(selectorOrElement))
+    {
+      const queryContext = context ? context : document;
```
MichaelKetting, line 209: A stylistic quirk: `window.document`. I try to avoid using single-word members, so I use `window.document` instead of `document`. The dot in the reference is just something that helps me when reading code that refers to global state.

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: Naming: `ResolveSingleElementOrCssSelector` and `ResolveUnboundedElementsOrCssSelector` or perhaps `ResolveMultipleElementsOrCssSelector`.  Reasoning: making it obvious if we're going for the single or the multiple variant.

There's still the question if `Resolve` is the best term for this. Do we resolve the element? I know we resolve the selector. But it just feels odd.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -200,6 +200,52 @@ class ArgumentUtility
...
+      return selectorOrElement;
+    }
+    else
+    {
+      throw ('Error: The type of parameter "selectorOrElements" is "' + (typeof selectorOrElement) + '" but "string" or "object" was expected.');
+    }
+  }
+
+  // Resolves any provided css selector, and returns the found elements as an array.
+  public static ResolveCssSelectorOrElements<TElement extends Element> (selectorOrElements: CssSelectorOrElements<TElement>, context?: ParentNode): TElement[]
```
MichaelKetting, line 228: Question (also affects the naming issue from above): Do we a) actually have a case where we want to pass a list of elements instead of a css selector? Maybe the either/or variant is only for single-valued results.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -200,6 +200,52 @@ class ArgumentUtility
...
+      return resolvedElement;
+    }
+    else if (TypeUtility.IsObject(selectorOrElement))
+    {
+      ArgumentUtility.CheckNotNull("selectorOrElement", selectorOrElement);
+      return selectorOrElement;
+    }
+    else
+    {
+      throw ('Error: The type of parameter "selectorOrElements" is "' + (typeof selectorOrElement) + '" but "string" or "object" was expected.');
```
MichaelKetting, line 223: Should there be a special case for NULL?

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: Do you want the element resolving on another utility? I put it in the `ArgumentUtility` because it is used at the beginning of a method together with other argument checks:

```js
function test(selectorOrElement, key) {
  const element = ArgumentUtility.ResolveCssSelectorOrElement(selectorOrElement);
  ArgumentUtility.CheckIsString("key", key)
}
```

I know that it looks a little bit out of place, but since it also does argument checking I think it is a better solution than to have a Resolve utility and need to do argument checks after.

a-ctor: I also do not like the naming 100%. It resolves the css selector part, otherwise (except from a null check) it returns the specified element.

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: I am not yet sure if we have a case where multiple elements are selected. Most occurrences will be single-element only but there might be some that do have multiple elements. Therefore, I would add it for now and remove it when it is not used after jQuery has been removed

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: `null` is handled in the `object` case. It is important to remember that in JS `typeof undefined === "undefined"` but `typeof null === "object"`. So `null` is just a special case of object. If a value is `=== typeof "string"` it cannot be `null`. In this case the null value is therefore handled in the second case.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -363,4 +409,15 @@ class WebServiceUtility
         }
       });
   };
+}
+
+class LayoutUtility {
```
iam-mholle, line 414: What about `.outerHeight(true)`? This would be `.offsetHeight` with the margin, right?

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: Resolving and argument checks should be separate tasks. Resolving should be a separate utility. => 2 calls at the start of the method. But then the resolve method could perhaps live without the argument check given we use typescript.

The argument check also tells you the name of the argument which failed, so this information would then be missing.

MichaelKetting: Naming: we probably need to discuss some options to find the right one.

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: okay. So basically this oculd mean we do a global update after we 're done to cleanup and use the "right" name for what's left over?

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: I was thinking in terms of argument checking  / providing a clear error message that something's NULL. It may be that `(typeof NULL)` already prints something kind of OK, but the argument utility is more precise.

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: Yes we can do that. And if the multiple element resolver method is not used we can remove it then.

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: Naming: I would suggest `ElementResolverUtility` with `ResolveSingleElementCssSelector` and `ResolveMultipleElementCssSelector` (note the missing "or" - otherwise the method it seems like it needs to resolve the element and not the css selector)

> But then the resolve method could perhaps live without the argument check given we use typescript.

I don't think I understand what you mean.

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: I think I did not explain my point well enough. A `null` value won't ever be handled in this `else` branch. This is because `null` has the type `object` and is already handled in the `else if (TypeUtility.IsObject(selectorOrElements))` case. The `ArgumentUtility` is used there to check for `null` values.

MichaelKetting: Ah, okay. And since the Resolve-method already does not accept nullable as long as you're typesafe, there's no issue here.

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: Hmm...How about we just drop the CssSelector and make this `ResolveSingle` and `ResolveMultiple`. The `Element` we already have in the type so that's taken care of.

MichaelKetting: > I don't think I understand what you mean.

As long as you use TypeScript, you can't pass a nullable argument value into the `Resolve` method. Therefor, we may not need to use argument checks inside the `Resolve` methods. We do use argument checks at the call sites because we get the css selectors passed down from c# and want to make sure nothing slips through.

a-ctor: I have left the arg checks in the method, since they can catch some usages errors. Normally, the calling method does a null check before using the resolver, thus never causing the resolve method to throw. This prevents the resolve method from returning a nullable element, which would need an additional argument check on the call side.

so now:
```js
function test(selectorOrElement) {
  ArgumentUtility.CheckNotNull("selectorOrElement", selectorOrElement);
  const element = ElementResolve.ResolveSingle(selectorOrElement); // is always below the arg checks
  ...
}
```

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -363,4 +363,75 @@ class WebServiceUtility
...
+      const queryContext = context ? context : window.document;
+      const resolvedElement = queryContext.querySelector(selectorOrElement) as TElement;
+      if (!resolvedElement)
+        throw ('Error: Cannot find an element specified by selector "' + selectorOrElement + '".');
+
+      return resolvedElement;
+    }
+    else if (TypeUtility.IsObject(selectorOrElement))
+    {
+      ArgumentUtility.CheckNotNull("selectorOrElement", selectorOrElement);
```
MichaelKetting, line 384: Please move the argumentcheck to the beginning of the method. NULL's never an option.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -363,4 +363,75 @@ class WebServiceUtility
...
+    else
+    {
+      throw ('Error: The type of parameter "selectorOrElements" is "' + (typeof selectorOrElement) + '" but "string" or "object" was expected.');
+    }
+  }
+
+  // Resolves any provided css selector, and returns the found elements as an array.
+  public static ResolveMultiple<TElement extends Element> (selectorOrElements: CssSelectorOrElements<TElement>, context?: ParentNode): TElement[]
+  {
+    if (TypeUtility.IsString(selectorOrElements))
```
MichaelKetting, line 396: Please move the argumentcheck to the beginning of the method. NULL's never an option.

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: Added my review comment before I saw this mention.  Don't think my review contradicts the latest comment.

MichaelKetting (approved):

PR #52, opened 2021-03-19
RM-7666: Remove JQuery from ListMenu.ts

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7666

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/ListMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/ListMenu.ts
@@ -52,23 +52,35 @@ class ListMenu
   private static readonly _requiredSelectionExactlyOne: number = 1;
   private static readonly _requiredSelectionOneOrMore: number = 2;
 
-  public static Initialize ($listMenu: JQuery): void
+  public static Initialize (listMenuOrSelector: CssSelectorOrElement<HTMLElement>): void
   {
-    ArgumentUtility.CheckNotNullAndTypeIsJQuery("$listMenu", $listMenu);
+    ArgumentUtility.CheckNotNull ('listMenuOrSelector', listMenuOrSelector);
 
-    $listMenu.keydown (function (event)
+    const listMenu = ElementResolverUtility.ResolveSingle (listMenuOrSelector);
+
+    listMenu.addEventListener('keydown', function (event)
     {
-      ListMenu.OnKeyDown (event, $listMenu);
+      ListMenu.OnKeyDown (event, listMenu);
     });
   }
 
-  public static AddMenuInfo (listMenu: HTMLElement, menuInfo: ListMenu_MenuInfo): void
+  public static AddMenuInfo (listMenuOrSelector: CssSelectorOrElement<HTMLElement>, menuInfo: ListMenu_MenuInfo): void
   {
+    ArgumentUtility.CheckNotNull ('listMenuOrSelector', listMenuOrSelector);
+    ArgumentUtility.CheckNotNullAndTypeIsObject ('menuInfo', menuInfo);
+
+    const listMenu = ElementResolverUtility.ResolveSingle (listMenuOrSelector);
+
     ListMenu._listMenuInfos[listMenu.id] = menuInfo;
   }
 
-  public static Update (listMenu: HTMLElement, getSelectionCount: Nullable<() => number>): void
+  public static Update (listMenuOrSelector: CssSelectorOrElement<HTMLElement>, getSelectionCount: Nullable<() => number>): void
   {
+    ArgumentUtility.CheckNotNull ('listMenuOrSelector', listMenuOrSelector);
+    ArgumentUtility.CheckTypeIsFunction ('getSelectionCount', getSelectionCount);
+
+    const listMenu = ElementResolverUtility.ResolveSingle (listMenuOrSelector);
+
     var menuInfo = ListMenu._listMenuInfos[listMenu.id];
     if (menuInfo == null)
       return;
@@ -99,8 +111,8 @@ class ListMenu
       }
 
       var item = document.getElementById (itemInfo.ID) as HTMLElement;
-      let anchor = $ (item).children().eq (0)[0] as HTMLAnchorElement;
-      var icon = $ (anchor).children().eq (0)[0] as HTMLImageElement;
+      let anchor = item.children[0] as HTMLAnchorElement;
+      var icon = anchor.children[0] as HTMLImageElement;
       if (isEnabled)
       {
         if (icon != null && icon.nodeType == 1)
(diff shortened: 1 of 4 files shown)


Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,71 @@ class ListMenu
 
       if (itemInfo.DiagnosticMetadata)
       {
+        // TODO: merge loops?
```
a-ctor, line 162: We can merge the loops on both occasions, JS is single threaded so the behavior is not changed at all.

Just make sure to but the seconds loop content into the `else` since the first loop removes elements that would not be iterated in the second loop.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,71 @@ class ListMenu
...
           {
             delete itemInfo.DiagnosticMetadata[key];
           }
-        });
+        }
 
-        $(item).attr(itemInfo.DiagnosticMetadata);
+        for (const [key, value] of Object.entries(itemInfo.DiagnosticMetadata))
+        {
+          item.setAttribute(key, value as string);
```
a-ctor, line 174: `as string` here is not quite right. Yes `setAttribute` is converting implicitly to a string so it won't matter if the input is a string or not, but in this case it is better to do an explict conversion instead. e.g. `"" + value`

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,71 @@ class ListMenu
...
+    var menuItems = listMenu.querySelectorAll('a');
 
     var oldMenuItemIndex = -1;
 
-    var selectedMenuItem = document.activeElement;
+    var selectedMenuItem = document.activeElement as HTMLAnchorElement;
     if (selectedMenuItem != null && TypeUtility.IsDefined (selectedMenuItem.tagName) && selectedMenuItem.tagName.toUpperCase() === 'A')
     {
-      oldMenuItemIndex = $menuItems.index(selectedMenuItem);
+      oldMenuItemIndex = Array.from(menuItems).indexOf(selectedMenuItem);
```
a-ctor, line 209: We should move the `Array.from` up to the declaration. While this causes a little bit of overhead if the `else` branch is taken, it simplifies future extensions/refactoring.

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
iam-mholle: Is the delete even needed here? I do not really see the purpose, imo the condition could be inverted and the delete removed.

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
a-ctor: Yes since we no longer need to pass the object to a jQuery method we also do not need to "clean" the object of any empty values.

a-ctor (approved):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -52,23 +52,33 @@ class ListMenu
...
     {
-      ListMenu.OnKeyDown (event, $listMenu);
+      ListMenu.OnKeyDown (event, listMenu);
     });
   }
 
-  public static AddMenuInfo (listMenu: HTMLElement, menuInfo: ListMenu_MenuInfo): void
+  public static AddMenuInfo (listMenuOrSelector: CssSelectorOrElement<HTMLElement>, menuInfo: ListMenu_MenuInfo): void
   {
+    ArgumentUtility.CheckNotNull ('listMenuOrSelector', listMenuOrSelector);
```
MichaelKetting, line 69: `menuInfo` should also have an argument check to stay consistent

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,59 @@ class ListMenu
...
-        // Do not render empty diagnostic metadata attributes
-        $.each(itemInfo.DiagnosticMetadata, function (key, value)
+        for (const [key, value] of Object.entries (itemInfo.DiagnosticMetadata))
         {
-          if (value === "" || value === null)
+          // Do not render empty diagnostic metadata attributes
+          if (value !== "" && value !== null)
           {
-            delete itemInfo.DiagnosticMetadata[key];
+            item.setAttribute (key, '' + value);
```
MichaelKetting, line 167: new code should not rely on concat-to-convert workarounds unless this is really the official an correct way to do this?
@a-ctor

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,59 @@ class ListMenu
...
-
         itemInfo.DiagnosticMetadataForCommand['data-is-disabled'] = isEnabled ? 'false' : 'true';
 
-        $(anchor).attr(itemInfo.DiagnosticMetadataForCommand);
+        for (const [key, value] of Object.entries (itemInfo.DiagnosticMetadataForCommand))
+        {
+          // Do not render empty diagnostic metadata attributes
+          if (value !== "" && value !== null)
+          {
+            anchor.setAttribute (key, '' + value);
```
MichaelKetting, line 181: see previous review remark

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,59 @@ class ListMenu
 
       if (itemInfo.DiagnosticMetadata)
       {
-        // Do not render empty diagnostic metadata attributes
-        $.each(itemInfo.DiagnosticMetadata, function (key, value)
+        for (const [key, value] of Object.entries (itemInfo.DiagnosticMetadata))
         {
-          if (value === "" || value === null)
+          // Do not render empty diagnostic metadata attributes
```
MichaelKetting, line 164: we can drop the comment. this was only needed for the original way to do this. now, it's rather obvious.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -149,65 +159,59 @@ class ListMenu
...
-              delete itemInfo.DiagnosticMetadataForCommand[key];
-            }
-          });
-
         itemInfo.DiagnosticMetadataForCommand['data-is-disabled'] = isEnabled ? 'false' : 'true';
 
-        $(anchor).attr(itemInfo.DiagnosticMetadataForCommand);
+        for (const [key, value] of Object.entries (itemInfo.DiagnosticMetadataForCommand))
+        {
+          // Do not render empty diagnostic metadata attributes
```
MichaelKetting, line 178: drop comment

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
a-ctor: There is no such thing as an "official" conversion. There are three options:

 - `'' + value`
 - `value.toString()`
 - `String(value)`

All of them produce the same result, with the exception that `.toString` throws on `null`/`undefined`. So depending on if you want to throw an exception I would either use `'' + ...` or `.toString`

MichaelKetting: toString(), please. it feels the least hacky and TypeScript should make sure we know if it's nullable

MichaelKetting (approved):

PR #55, opened 2021-03-22
RM-7665: Remove JQuery from DropDownMenu.ts

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7665


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -634,38 +637,44 @@ class DropDownMenu
...
           delete itemInfo.DiagnosticMetadata![key];
         }
-      });
+      }
 
-      $ (item).attr (itemInfo.DiagnosticMetadata);
+      for (const [key, value] of Object.entries (itemInfo.DiagnosticMetadata))
+      {
+        item.setAttribute(key, value as string);
+      }
```
a-ctor, line 651: Same procedure as last PR a-ctor?
Same procedure as every PR iam-mholle.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -634,38 +637,44 @@ class DropDownMenu
...
+        }
+      }
 
       itemInfo.DiagnosticMetadataForCommand['data-is-disabled'] = isEnabled ? 'false' : 'true';
 
-      $(anchor).attr(itemInfo.DiagnosticMetadataForCommand);
+      for (const [key, value] of Object.entries (itemInfo.DiagnosticMetadataForCommand))
+      {
+        anchor.setAttribute(key, value as string);
+      }
```
a-ctor, line 670: déjà vu

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -707,9 +716,14 @@ class DropDownMenu
     return false;
   }
 
-  public static OnKeyDown(event: JQueryKeyEventObject, dropDownMenu: JQuery, getSelectionCount: Nullable<DropDownMenu_SelectionCountGetter>, hasDedicatedDropDownMenuElement: boolean): void
+  public static OnKeyDown(event: KeyboardEvent, dropDownMenuOrSelector: CssSelectorOrElement<HTMLAnchorElement>, getSelectionCount: Nullable<DropDownMenu_SelectionCountGetter>, hasDedicatedDropDownMenuElement: boolean): void
   {
-    ArgumentUtility.CheckNotNullAndTypeIsJQuery('dropDownMenu', dropDownMenu);
+    ArgumentUtility.CheckNotNull ('event', event);
```
a-ctor, line 721:
```suggestion
    ArgumentUtility.CheckNotNullAndTypeIsObject ('event', event);
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -756,21 +770,19 @@ class DropDownMenu
...
-    var itemInfos = DropDownMenu._menuInfos[dropDownMenu[0].id]!.ItemInfos!;
+    var itemInfos = DropDownMenu._menuInfos[dropDownMenu.id]!.ItemInfos!;
     if (itemInfos.length === 0)
       return;
 
     var oldIndex;
     var isSelectionUpdated = false;
-    var dropDownMenuItems = $(DropDownMenu._currentPopup).find('ul').children();
-    var currentItemIndex = $('.' + DropDownMenu._itemSelectedClassName, DropDownMenu._currentPopup).index();
+    var dropDownMenuItems = Array.from(DropDownMenu._currentPopup.querySelectorAll('ul')).flatMap((el: HTMLUListElement) => Array.from(el.children));
```
a-ctor, line 784: Flatmap takes an enumerable so we don't have to convert to array first if I am not mistaken.

```suggestion
    var dropDownMenuItems = Array.from(DropDownMenu._currentPopup.querySelectorAll('ul')).flatMap(el => el.children);
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -783,7 +795,7 @@ class DropDownMenu
         event.stopPropagation();
         if (currentItemIndex >= 0)
         {
-          var itemAnchor = $(dropDownMenuItems[currentItemIndex]).children('a');
+          var itemAnchor = dropDownMenuItems[currentItemIndex].querySelector ('a')!;
```
a-ctor, line 798: `.children` requires a `:scope >` prefix to only search for children. `querySelector` and `querySelectorAll` would otherwise go through all the descendants. See notes in the handbook to `.children("a")`

```suggestion
          var itemAnchor = dropDownMenuItems[currentItemIndex].querySelector (':scope > a')!;
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -812,7 +824,7 @@ class DropDownMenu
 
           if (oldIndex === currentItemIndex)
             break;
-          if ($(dropDownMenuItems[currentItemIndex]).children('a').length > 0) // skip separators
+          if (dropDownMenuItems[currentItemIndex].querySelectorAll ('a').length > 0) // skip separators
```
a-ctor, line 827:
```suggestion
          if (dropDownMenuItems[currentItemIndex].querySelectorAll (':scope > a').length > 0) // skip separators
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +842,23 @@ class DropDownMenu
 
           if (oldIndex === currentItemIndex)
             break;
-          if ($(dropDownMenuItems[currentItemIndex]).children('a').length > 0)
+          if (dropDownMenuItems[currentItemIndex].querySelectorAll ('a').length > 0)
```
a-ctor, line 845:
```suggestion
          if (dropDownMenuItems[currentItemIndex].querySelectorAll (':scope > a').length > 0)
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +842,23 @@ class DropDownMenu
...
             break;
         }
         break;
     }
 
     if (isSelectionUpdated)
     {
-      $("li", DropDownMenu._currentPopup).removeClass (DropDownMenu._itemSelectedClassName);
+      DropDownMenu._currentPopup.querySelectorAll("li").forEach((el: HTMLLIElement) => el.classList.remove (DropDownMenu._itemSelectedClassName));
```
a-ctor, line 853:
```suggestion
      DropDownMenu._currentPopup.querySelectorAll("li").forEach(el => el.classList.remove (DropDownMenu._itemSelectedClassName));
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +842,23 @@ class DropDownMenu
...
       if (currentItemIndex >= 0 && currentItemIndex < itemInfos.length)
       {
-        var dropDownMenuItem = $(dropDownMenuItems[currentItemIndex]);
-        if (dropDownMenuItem.children('a').length > 0) // skip separators
+        var dropDownMenuItem = dropDownMenuItems[currentItemIndex];
+        if (dropDownMenuItem.querySelectorAll('a').length > 0) // skip separators
```
a-ctor, line 857:
```suggestion
        if (dropDownMenuItem.querySelectorAll(':scope > a').length > 0) // skip separators
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +842,23 @@ class DropDownMenu
...
         {
-          dropDownMenuItem.addClass(DropDownMenu._itemSelectedClassName);
-          var nonJQueryAnchor = dropDownMenuItem.children ('a')[0];
-          // do not use jQuery focus() here because jQuery causes the focus/blur events to fire in the wrong order (first focus, then blur)
-          nonJQueryAnchor.focus();
+          dropDownMenuItem.classList.add (DropDownMenu._itemSelectedClassName);
+          var anchor = dropDownMenuItem.querySelector ('a')!;
```
a-ctor, line 860:
```suggestion
          var anchor = dropDownMenuItem.querySelector (':scope > a')!;
```

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
```diff
@@ -266,7 +266,7 @@ private void RegisterEventHandlerScripts (DropDownMenuRenderingContext rendering
           : renderingContext.Control.GetSelectionCount);
       string hasDedicatedDropDownMenuElement = renderingContext.Control.Mode == MenuMode.DropDownMenu ? "true" : "false";
       string script = string.Format (
-          "$('#{0}').keydown( function(event){{ DropDownMenu.OnKeyDown(event, $('#{0}'), {1}, {2}); }} );",
+          "$('#{0}').keydown( function(event){{ DropDownMenu.OnKeyDown(event, '#{0}', {1}, {2}); }} );",
```
a-ctor, line 269: I still smell some jQuery 🐕

a-ctor: You can use a regex like `\$|jQuery` to catch most of the remaining jQuery usages that might still linger in C# code

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -105,38 +105,41 @@ class DropDownMenu
...
 
-      var allButMenuButton = openTarget.find ('a[href]').not(menuButton);
-      allButMenuButton
-          .bind ("mouseover", function (evt) { openTarget.find ('.' + DropDownMenu._buttonClassName).addClass (DropDownMenu._nestedHoverClassName); })
-          .bind ("mouseout", function (evt) { openTarget.find ('.' + DropDownMenu._buttonClassName).removeClass (DropDownMenu._nestedHoverClassName); });
+      var allButMenuButton = Array.from(openTarget.querySelectorAll ('a[href]')).filter((el: Element) => menuButton !== el);
+      allButMenuButton.forEach((el: Element) => el.addEventListener ("mouseover", function (evt) { openTarget.querySelectorAll ('.' + DropDownMenu._buttonClassName).forEach ((el: Element) => el.classList.add (DropDownMenu._nestedHoverClassName)); }));
+      allButMenuButton.forEach((el: Element) => el.addEventListener ("mouseout", function (evt) { openTarget.querySelectorAll ('.' + DropDownMenu._buttonClassName).forEach ((el: Element) => el.classList.remove (DropDownMenu._nestedHoverClassName)); }));
```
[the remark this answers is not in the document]
iam-mholle: I guess you mean `el: Element` in this case?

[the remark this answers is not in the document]
a-ctor: yes

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Getting errors without it...

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
iam-mholle: How did I miss that 😂

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Interesting because my IDE does not show me an error and I also get no error when testing in the TypeScript playground. Did it not compile or just show a red squigglies?

iam-mholle: Did not compile. Please check it out, have a look. I'm confused.

a-ctor: Ok my bad, we need to do the `Array.from`, flatMap also supports returning only one element instead of an array (aka a map). This is what happens if we do not use the `Array.from` it maps to a `NodeList[]` instead of applying the flatmap

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -105,38 +105,51 @@ class DropDownMenu
...
+    var menuButton: HTMLAnchorElement;
+
+    if (openTarget.tagName.toLowerCase() === 'a')
+    {
+      menuButton = openTarget as HTMLAnchorElement;
+    }
+    else
+    {
+      menuButton = Array.from(openTarget.querySelectorAll<HTMLAnchorElement>('a[href]')).slice(-1)[0];
+    }
```
a-ctor, line 132: I meant expanding to an `if-else` in order to write a more understandable .last()

```suggestion
    if (openTarget.tagName.toLowerCase() === 'a')
    {
      menuButton = openTarget as HTMLAnchorElement;
    }
    else
    {
      const menuButtons = openTarget.querySelectorAll<HTMLAnchorElement> ('a[href]');
      menuButton = menuButtons[menuButtons.length - 1];
    }
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -756,21 +770,19 @@ class DropDownMenu
...
+    var itemInfos = DropDownMenu._menuInfos[dropDownMenu.id]!.ItemInfos!;
     if (itemInfos.length === 0)
       return;
 
     var oldIndex;
     var isSelectionUpdated = false;
-    var dropDownMenuItems = $(DropDownMenu._currentPopup).find('ul').children();
-    var currentItemIndex = $('.' + DropDownMenu._itemSelectedClassName, DropDownMenu._currentPopup).index();
+    var dropDownMenuItems = Array.from (DropDownMenu._currentPopup.querySelectorAll ('ul')).flatMap (el => Array.from (el.children));
+    var currentItemIndex = Array.from (document.querySelectorAll ('.' + DropDownMenu._itemSelectedClassName)).indexOf (DropDownMenu._currentPopup);
```
a-ctor, line 785: Now that I look at it a second time, I think we have a problem here. `.index()` takes the first element in the jQuery object and determines its index regarding its parent. So what we want to do here is find any selected item and determine its index or -1 if no item is selected.

Something like:
```js
const selectedElement = DropDownMenu._currentPopup.querySelector('.' + DropDownMenu._itemSelectedClassName);
const selectedElementIndex = selectedElement ? Array.from(selectedElement .parentNode.children).indexOf(selectedElement) : -1;

```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -368,38 +381,39 @@ class DropDownMenu
...
     div.className = 'DropDownMenuOptions';
     div.id = menuOptionsID;
-    div.setAttribute('role', 'menu');
-    div.setAttribute('tabindex', '-1');
-    if (menuButton.attr('aria-labelledby') !== undefined)
-      div.setAttribute('aria-labelledby', menuButton.attr('aria-labelledby'));
+    div.setAttribute ('role', 'menu');
+    div.setAttribute ('tabindex', '-1');
+    if (menuButtonAnchor.getAttribute ('aria-labelledby') !== undefined)
+      div.setAttribute ('aria-labelledby', menuButtonAnchor.getAttribute('aria-labelledby')!);
```
a-ctor, line 393:
```suggestion
      div.setAttribute ('aria-labelledby', menuButtonAnchor.getAttribute ('aria-labelledby')!);
```

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Makes sense

a-ctor (approved):

a-ctor (approved):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -208,27 +227,25 @@ class DropDownMenu
...
-    $('#' + menuID).data(DropDownMenu._button_timestampDataKey, new Date().getTime());
+    document.getElementById (menuID)!.dataset[DropDownMenu._button_timestampDataKey] = new Date().getTime().toString();
 
     if (DropDownMenu._itemClicked)
     {
       DropDownMenu._itemClicked = false;
       return;
     }
-    if (DropDownMenu._currentMenu != null && DropDownMenu._currentMenu[0] !== context[0])
+    if (DropDownMenu._currentMenu != null && DropDownMenu._currentMenu !== context)
```
MichaelKetting, line 239: context is never null, so we could simplify this check.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -317,15 +332,14 @@ class DropDownMenu
     statusPopup.setAttribute('aria-atomic', 'true');
     // Do not set role=alert before it is needed prevent an alert-update during 'normal' showing of menu.
     //statusPopup.setAttribute('role', 'alert');
-    if (menuButton.attr('aria-labelledby') !== undefined)
-      statusPopup.setAttribute('aria-labelledby', menuButton.attr('aria-labelledby'));
+    if (menuButton.getAttribute ('aria-labelledby') !== undefined) // TODO: should be compared with null
```
MichaelKetting, line 335: todo needs a ticket

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -423,9 +437,10 @@ class DropDownMenu
       if (DropDownMenu._repositionTimer)
         clearTimeout (DropDownMenu._repositionTimer);
 
-      if (DropDownMenu._currentPopup && DropDownMenu._currentPopup == div && $(div).is (':visible'))
+      const isDivVisible = (div.offsetWidth > 0 || div.offsetHeight > 0);
```
MichaelKetting, line 440: @a-ctor the intention of this code was to test via CSS / layout information, not height/width. I did not do a deep dive, are offsetWidth/Height 0 when the element is set to display:none ?

Remotion/Web/Development.WebTesting.ControlObjects/DropDownMenuControlObjectBase.cs
```diff
@@ -62,7 +62,7 @@ public bool IsOpen ()
     public void Close ()
     {
       if (IsOpen())
-        Context.Browser.Driver.ExecuteScript ("DropDownMenu_ClosePopUp()", Scope);
+        Context.Browser.Driver.ExecuteScript ("DropDownMenu_ClosePopUp(true)", Scope);
```
MichaelKetting, line 65: isn't that a semantic change, passing true instead of false/undefined?

MichaelKetting: And yes, it may be a valid bugfix, but this should then be a separate ticket.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -725,8 +738,8 @@ class DropDownMenu
         event.stopPropagation();
         if (dropDownMenu !== DropDownMenu._currentMenu)
         {
-          DropDownMenu.OnClick (dropDownMenu, dropDownMenu[0].id, getSelectionCount, null);
-          event.keyCode = 40; // always act as if the down-arrow was used when opening the drop down menu.
+          DropDownMenu.OnClick (dropDownMenu, dropDownMenu.id, getSelectionCount, null);
+          (event.keyCode as any) = 40; // always act as if the down-arrow was used when opening the drop down menu.
```
MichaelKetting, line 742: wouldn't this be number?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -745,8 +758,8 @@ class DropDownMenu
...
```
MichaelKetting, line 762: see 742

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -208,27 +227,25 @@ class DropDownMenu
     }
   }
 
-  private static OnClick(context: JQuery, menuID: string, getSelectionCount: Nullable<DropDownMenu_SelectionCountGetter>, evt: Nullable<JQueryMouseEventObject>): void
+  private static OnClick (context: HTMLElement, menuID: string, getSelectionCount: Nullable<DropDownMenu_SelectionCountGetter>, evt: Nullable<MouseEvent>): void
   {
-    ArgumentUtility.CheckNotNullAndTypeIsJQuery('context', context);
-
-    $('#' + menuID).data(DropDownMenu._button_timestampDataKey, new Date().getTime());
+    document.getElementById (menuID)!.dataset[DropDownMenu._button_timestampDataKey] = new Date().getTime().toString();
```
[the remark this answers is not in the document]
a-ctor: It would probably be a good idea to normalize all the data-attribute getting/setting. Should I create an issue?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Yes they are set to `0`.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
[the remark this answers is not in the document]
MichaelKetting: @a-ctor yes, please. The ticket should contain a note that we have to decide if different spellings for the attribute on the TS and C# side are actually a good thing. Right now, there's just 5 or so `data-` usages and those are part of the client/server contract.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor since it popped up here and there've been a couple other more complex steps in this file or one of the others (e.g. applying of a dictioanry of attributes): should we introduce utilities for these type of logic blocks? They do appear to be sufficiently complicated to "get right".

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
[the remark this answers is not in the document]
a-ctor: https://re-motion.atlassian.net/browse/RM-7729

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Removing code duplication is probably a good idea. The visibility check could be put into `LayoutHelper.IsVisible(element)`. For other utilities we would need to create a new class to make the name fit.

MichaelKetting: The best solution would probably be to create the utility methods as soon as the need arises, but in a separate PR. It's also okay if we start with TODO comments and then update the code later on. The third option would be to create them in the branch as separate commits and just resolve merge conflicts in the utility when they arise. We should aim to prioritize de-blocking and actively track code duplications where they cannot be avoided.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -105,38 +105,52 @@ class DropDownMenu
...
+      menuButton = menuButtons[menuButtons.length - 1];
+    }
+
+    menuButton.setAttribute ('aria-haspopup', 'menu');
+    menuButton.setAttribute ('aria-expanded', 'false');
+
+    openTarget.addEventListener (
         eventType,
-        function (evt: JQueryMouseEventObject)
+        function (evt)
```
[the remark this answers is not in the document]
iam-mholle: `Event` could be added, but it is inferred here anyway. As eventType is no string known at "compile-time", no more specific type can be added here.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: So every utility method gets a separate ticket, fixed in a separate PR? Then I'll add a ticket for `LayoutHelper.IsVisible(element)` and insert a todo here.

Remotion/Web/Development.WebTesting.ControlObjects/DropDownMenuControlObjectBase.cs
iam-mholle: You're right, that wasn't the intention. There is no bug associated, I'll change this to `false` without further actions needed.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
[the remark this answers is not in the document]
MichaelKetting: so we still need to do a cast to `MouseEvent` further down. Okay, doesn't help us, then.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Yes. changed. FYI it's a deprecated API: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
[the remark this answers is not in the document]
iam-mholle: Resolving this discussion as no change from my part is needed, please re-open if I'm wrong.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: Could you please make a ticket for this (don't need a todo). We need to schedule the cleanup before the API goes the way of the Dodo.

@iam-mholle , @a-ctor We also use this API in the support-versions, do we have any indication when we might expect the API to die so we know when/if we need to fix the support branches and roll out an update?

Remotion/Web/Development.WebTesting.ControlObjects/DropDownMenuControlObjectBase.cs
MichaelKetting: @iam-mholle please create ticket (RMTEAM) for us to investigate if this actually a webtest bug. Thanks!

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Nevermind, this was valid with jquery (`attr(...)` returns `undefined` if no value is present), `getAttribute` returns `null` however.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: It been deprecated for a long time, the `code` property should be used instead. I don't think that the property is ever going to be removed, but any browser might remove it at any time

MichaelKetting: Then we just try to update this sometime this year and do not do a backport.

iam-mholle: So this does not get changed in this PR? Is there an existing issue or should I create one?

MichaelKetting: @iam-mholle correct, we keep keyCode in this PR. Please create a ticket.

iam-mholle: Added https://re-motion.atlassian.net/browse/RM-7733

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -423,9 +437,11 @@ class DropDownMenu
       if (DropDownMenu._repositionTimer)
         clearTimeout (DropDownMenu._repositionTimer);
 
-      if (DropDownMenu._currentPopup && DropDownMenu._currentPopup == div && $(div).is (':visible'))
+      // TODO RM-7730 replace with LayoutHelper.IsVisible
```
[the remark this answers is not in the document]
a-ctor: I have created PR #65 to add the utility method

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -434,100 +450,99 @@ class DropDownMenu
...
 
-  private static ApplyPosition (popUpDiv: JQuery, clickEvent: Nullable<JQueryMouseEventObject>, referenceElement: JQuery): void
+  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement): void
   {
-    var space_top = Math.round(referenceElement.offset().top - $(document).scrollTop());
-    var space_bottom = Math.round($(window).height() - referenceElement.offset().top - referenceElement.height() + $(document).scrollTop());
-    var space_left = referenceElement.offset().left;
-    var space_right = $(window).width() - referenceElement.offset().left - referenceElement.width();
+    var space_top = Math.round (LayoutUtility.offset (referenceElement).top - window.pageYOffset);
+    var space_bottom = Math.round (document.documentElement.clientHeight - LayoutUtility.offset (referenceElement).top - referenceElement.clientHeight + window.pageYOffset);
```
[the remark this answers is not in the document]
a-ctor: Yes all the `clientHeight` and `clientWidth` properties that were created from a `.height()` need to be converted to a call to the `LayoutHelper`. The only exception are `window` and `document` as they get special treatment when checking the height/width

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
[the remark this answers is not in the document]
MichaelKetting: merged

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -756,21 +770,20 @@ class DropDownMenu
...
     if (itemInfos.length === 0)
       return;
 
     var oldIndex;
     var isSelectionUpdated = false;
-    var dropDownMenuItems = $(DropDownMenu._currentPopup).find('ul').children();
-    var currentItemIndex = $('.' + DropDownMenu._itemSelectedClassName, DropDownMenu._currentPopup).index();
+    var dropDownMenuItems = Array.from (DropDownMenu._currentPopup.querySelectorAll ('ul')).flatMap (el => Array.from (el.children));
+    var currentItem = DropDownMenu._currentPopup.querySelector ('.' + DropDownMenu._itemSelectedClassName);
+    var currentItemIndex = currentItem ? Array.from (currentItem.parentNode!.children).indexOf (currentItem) : -1;
```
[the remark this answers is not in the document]
a-ctor: Yes this is possible using `.findIndex (e => e.class.contains (_itemSelectedClassName))`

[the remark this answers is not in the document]
MichaelKetting: Thanks! 

 @iam-mholle I think the suggestion + getting ul.childrent directly would express the code's intent more cleanly. Thoughts?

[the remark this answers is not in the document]
iam-mholle: Changed in [`3b0386e` (#55)](https://github.com/re-motion/Framework/pull/55/commits/3b0386ecc12f8c05ec991031bab620f44c4dd809), definitely more clear (if the solution is still correct 😀)

MichaelKetting (approved):

PR #80, opened 2021-04-09
RM-7670 remove jQuery from remaining file types

a-ctor:


Remotion/Data/DomainObjects.Web.Test/Performance/Form.aspx
```diff
@@ -140,7 +140,7 @@
     prm.add_pageLoading(Page_OnBeginAsyncLoading);
     prm.add_endRequest(Page_OnEndAsyncRequest);
 
-    $(window).load(Page_OnEndSyncRequest);
+    $(window).load(Page_OnEndSyncRequest); // TODO: what is this? .load(string) is the actually signature
```
MichaelKetting, line 143: This executes the Page_OnEndSyncRequest function when the window.load event is fired. 

https://api.jquery.com/load-event/#load-handler
jQuery doc says `load (function(){...})` works, so passing an existing function by handle also works.

Remotion/ObjectBinding/Web.Test/TestForm.aspx
```diff
@@ -14,49 +14,60 @@
...
+    AsyncPostBackTimeout="3600" />
+  <input id="StringField" type="text" />
+  <input id="IntField" type="text" />
+  <input type="button" value="ASP.NET Ajax" onclick="DoAspNetAjaxCall()" />
+  <input type="button" value="jQuery Ajax" onclick="DoJQueryAjaxCall()" />
+  <div id="Result">
+  </div>
+  </form>
+<script type="text/javascript">
     function DoAspNetAjaxCall()
```
MichaelKetting, line 37: why move this into the body? I think splitting this would be the choice that better follows the existing design. Keep the declarations up top and move the invocations to the end of the page.

Remotion/ObjectBinding/Web.Test/TestForm.aspx
```diff
@@ -14,49 +14,60 @@
...
-        stringValue: $("#StringField").val(),
+        stringValue: document.getElementById("StringField").value,
         intValue: parseInt(intValueAsString)
       };
       executingRequest = Sys.Net.WebServiceProxy.invoke("TestService.asmx", "DoStuff", false, params,
                                           function (result, context, methodName)
                                           {
                                             executingRequest = null;
-                                            $("#Result").text(result);
+                                            document.getElementById("Result").textContent = result;
```
MichaelKetting, line 51: in the other scripts, we're using innerText, here it's textContent, why?

Remotion/ObjectBinding/Web.Test/TestForm.aspx
```diff
@@ -14,49 +14,60 @@
...
-      var intValueAsString = $("#IntField").val();
+      var intValueAsString = document.getElementById("IntField").value;
       intValueAsString = intValueAsString == '' ? 0 : intValueAsString;
       var params = {
-        stringValue: $("#StringField").val(),
+        stringValue: document.getElementById("#StringField").value,
         intValue: parseInt(intValueAsString)
       };
-      $.ajax({
+      $.ajax({ // TODO keep this?
```
MichaelKetting, line 70: okay, the entire test is set to up test the jQuery and ASP.NET XmlHttpRequest APIs. Since we're keeping a jquery dependency in the code base, we can keep invoking jquery.ajax here, just then switch it to the noconflict variant. We do drop all the other jquery stuff, besides the ajax call.

In theory, the entire test may become redundant with dropping jquery, but since it doesn't hurt us on the test page, I want to keep it in case we need to do some testing in the future.

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
```diff
@@ -153,46 +153,45 @@ class SmartFocus
...
       {
-        var isEnabledFilter = function (this: HTMLElement)
+        function isEnabledFilter(el: HTMLElement)
         {
-          var _this = $ (this);
-          if (_this.is ('button'))
+          if (el.tagName === 'BUTTON')
             return true;
-          if (_this.is ('input[type=button]'))
+          if (el.tagName === 'INPUT' && (el as HTMLInputElement).type.toLowerCase() === "button")
```
MichaelKetting, line 163: In line 169, it's `el.matches(...)`, here it's separate checks. Why not use `el.matches(...)` at the other spots, too?

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
```diff
@@ -153,46 +153,45 @@ class SmartFocus
...
             return true;
-          if (_this.is ('input[type=submit]'))
+          if (el.tagName === 'INPUT' && (el as HTMLInputElement).type.toLowerCase() === "submit")
             return true;
-          if (_this.is ('a'))
+          if (el.tagName === 'A')
             return true;
-          if (!_this.is ('a') && _this.is (':enabled') && !_this.is ('input[type=hidden]'))
+          if (el.tagName !== 'INPUT' && el.matches (':enabled') && !el.matches ('input[type=hidden]'))
```
MichaelKetting, line 169: actually, I think this is a semantic change here and in the next IF.

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
```diff
@@ -202,7 +201,7 @@ class SmartFocus
     return undefined as unknown as boolean;
   }
 
-  public static SetFocus (element: JQuery): void
+  private static SetFocus (element: HTMLElement): void
```
MichaelKetting, line 204: have to keep that one public, at least for now. There's a downstream workaround based on being able to replace / override this function.
```suggestion
  public static SetFocus (element: HTMLElement): void
```

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -922,7 +910,6 @@ class SmartPage_Context
         message +
         '</td></tr></table>';
       statusMessageWindow.appendChild (statusMessageBlock);
-      $(statusMessageBlock).iFrameShim({ top: '0px', left: '0px', width: '100%', height: '100%' });
```
MichaelKetting, line 925: Do we still have usages of iFrameShim? Need to remember to update or delete iFrameShim when we do jQuery noConflict.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -1224,23 +1207,11 @@ class SmartPage_Context
...
       }
-      //setTimeout (function ()
-    //{
-      var html = $ ('html');
-      html.removeClass('SmartPageBusy');
-
-        // Needed in IE8, Firefox 23, Chrome 4
-        // Does not work in Safari 4 for Windows
-        // Does not work in IE10, IE11
-        var cursorBackUp = html.css ('cursor');
```
MichaelKetting, line 1237: Why did this code get deleted?

MichaelKetting (changes requested):

Remotion/Web/Core/Utilities/IScriptUtility.cs
```diff
@@ -39,16 +39,16 @@ public interface IScriptUtility
...
-    /// Registers the <paramref name="eventHandler"/> for the element identified by the <paramref name="jquerySelector"/>. 
+    /// Registers the <paramref name="eventHandler"/> for the element identified by the <paramref name="cssSelector"/>. 
     /// Call <see cref="RegisterJavaScriptInclude"/> before calling this method.
     /// </summary>
     /// <param name="control">The <see cref="IControl"/> for which the <paramref name="eventHandler"/> is registered.</param>
-    /// <param name="jquerySelector">The element-selector in jquery-syntax.</param>
+    /// <param name="cssSelector">The element-selector as a css selector.</param>
     /// <param name="eventHandler">The eventhandler, with the following signatur: <c>function (element)</c>.</param>
-    void RegisterResizeOnElement (IControl control, string jquerySelector, string eventHandler);
+    void RegisterResizeOnElement (IControl control, string cssSelector, string eventHandler);
```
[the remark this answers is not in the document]
a-ctor: There are already two breaking change tickets. Are they okay or am I missing something?

Remotion/ObjectBinding/Web.Test/TestForm.aspx
a-ctor: I looked up the difference and there seems to be no difference when setting the properties, but `textContent` is semantically more accurate when setting the text content of a element. `innerText` removes invisible elements and such when getting the property (a rendered version of `textContent`) and this is sometimes desired.

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
a-ctor: The performance part of my brain would not let me. I changed the if statements to avoid overusing `.matches` and used it when there is no simple way converting the css selector to js or if the if-statement would get overcrowded.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
a-ctor: Only in jquery.js so we should be fine

Remotion/Web/Core/Utilities/IScriptUtility.cs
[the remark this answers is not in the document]
MichaelKetting: ah, yes. looks good.

Remotion/ObjectBinding/Web.Test/TestForm.aspx
MichaelKetting: ah, I.C. So, jQuery text() behaves more like textContent? I never new that one. However, I don't have problems with leading whitespace, so the question is, should we stick with innerText to have everything consistent and do a separate task for reviewing/cleaning up or accept that it's inconsistent right now?

a-ctor: Since `textContent` is available on `Node` instead of `HTMLElement` it seems to be the better choice (when setting - getting always depends on the usage). But I don't think it is worth fixing, we only have 3 usages of `innerText` that would need to be fixed.

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
MichaelKetting: Ah, I.C. Performance is always helpful. I'll wait for the updated version when the other stuff's fixed, but maybe we should just resolve the other two `matches` to lower level lookups, too, since it's better performance and keeps things consistent within the IF-statement. Without symmetry, you always wonder if there's a special reason to it, beyond 'here it read better this way'.

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
```diff
@@ -153,46 +153,45 @@ class SmartFocus
...
+          if (el.tagName === 'INPUT' && (el as HTMLInputElement).type.toLowerCase() === "button")
             return true;
-          if (_this.is ('input[type=submit]'))
+          if (el.tagName === 'INPUT' && (el as HTMLInputElement).type.toLowerCase() === "submit")
             return true;
-          if (_this.is ('a'))
+          if (el.tagName === 'A')
             return true;
-          if (!_this.is ('a') && _this.is (':enabled') && !_this.is ('input[type=hidden]'))
+          if ((el as HTMLInputElement).disabled === false && !(el.tagName === 'INPUT' && (el as HTMLInputElement).type.toLowerCase() === "hidden"))
```
MichaelKetting, line 169:
```suggestion
          if (el.tagName !== 'A' && (el as HTMLInputElement).disabled === false && !(el.tagName === 'INPUT' && (el as HTMLInputElement).type.toLowerCase() === "hidden"))
```

I know it's redundant, the reasoning is to make the individual cases easier to reason over -- you have everything you need to understand a single IF in one line instead of needing to check what other exit conditions there hav been.

Remotion/Web/ClientScript/Scripts/SmartNavigation.ts
```diff
@@ -153,46 +153,45 @@ class SmartFocus
...
             return true;
-          if (!_this.is ('a') && _this.is ('*[tabindex]'))
+          if (el.hasAttribute ('tabindex'))
```
MichaelKetting, line 171:
```suggestion
          if (el.tagName !== 'A' && el.hasAttribute ('tabindex'))
```

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #83, opened 2021-04-19
RM-7679: Upgrade Firefox to Version 86

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7679


Stefan-Ilic (approved):
Just for continuity's sake: Apparently, there are some DateTime test(s) that fail sporadically, but that behavior isn't exclusive to this branch, and we decided to upgrade Firefox anyway because it doesn't seem to be connected to the upgrade. Alrighty then

MichaelKetting (approved):

PR #84, opened 2021-04-19
RM-7760: Fix DropDownMenu CSS rules being overridden by BocList rules

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7760

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
@@ -400,6 +400,22 @@ tr.bocListDataRow td.bocListDataCell span.CommandText
 
 
 /* Style to override the BocList's anchor style for drop down menus rendered inside a cell. */
+td.bocListDataCell div ul.DropDownMenuOptions li a
+{
+  display: block;
+}
+
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled.selected a,
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled a:active,
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled a:focus,
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled a:hover
+{
+  background-color: #ECEEEC;
+  border-style: none;
+  padding: 2px;
+  color: gray;
+}
+
 td.bocListDataCell div span.DropDownMenuSelect a.DropDownMenuLabel,
 td.bocListDataCell div span.DropDownMenuSelect:focus a.DropDownMenuLabel,
 td.bocListDataCell div span.DropDownMenuSelect:hover a.DropDownMenuLabel


MichaelKetting (changes requested):
I've checked with customizations downstream - I don't think we can do it this way due to the breaking change this would involve. 

Instead, we have to fix the style for DropDownMenu (and possibly also for BocAutoCompleteReferenceValue) from within the BocList.css. There's already precedence for this in line 400ff. We should, however, limit the fixes to the required minimum, i.e. identify which parts actually override/break settings on the nested controls and only fix those.

MichaelKetting:
> 
> 
> I've checked with customizations downstream - I don't think we can do it this way due to the breaking change this would involve.
> 
> Instead, we have to fix the style for DropDownMenu (and possibly also for BocAutoCompleteReferenceValue) from within the BocList.css. There's already precedence for this in line 400ff. We should, however, limit the fixes to the required minimum, i.e. identify which parts actually override/break settings on the nested controls and only fix those.

There's one other possible option: all anchor elements in BocList that are targeted as a command have a rendered onclick-event either on themselves or in their `span`/`div` parents that contain the text `BocList.OnCommandClick();` or `BocList_OnCommandClick();` (re-motion v2.x). Maybe this could be used as a heurisitc? Still, for now I think the brute-force variant might be better (cheaper, less danger of breaking changes)

iam-mholle:
@MichaelKetting 

> Instead, we have to fix the style for DropDownMenu [...] from within the BocList.css.

Did that now.

> ...and possibly also for BocAutoCompleteReferenceValue...

Is there a testsite where I could see the style of it in a BocList?

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -400,6 +400,22 @@ tr.bocListDataRow td.bocListDataCell span.CommandText
...
+}
+
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled.selected a,
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled a:active,
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled a:focus,
+td.bocListDataCell div ul.DropDownMenuOptions li.DropDownMenuItemDisabled a:hover
+{
+  background-color: #ECEEEC;
+  border-style: none;
+  padding: 2px 2px 2px 2px;
```
MichaelKetting, line 415: I may be missing somehting (only looked at the DropDownMenu.css), but line 287ff 
`ul.DropDownMenuOptions li.DropDownMenuItemDisabled a`
says that padding uses `1px 3px 3px 3px`, here it's 2px?

iam-mholle: Imo a padding of `2px` is more correct anyway, as non-disabled anchors have a padding of `1px` + the anchor with a width of `1px`, so the disabled one needs a padding of `2px`. The question is rather why the other file sets `1px 3px 3px 3px`.

Anyway, the line should be changed to 
```suggestion
  padding: 2px;
```

MichaelKetting: Very true. But given that I don't know rihgt now, what the bug is, plase use the original value from DropDownMenu.css to keep things consistent, even if it's strange/wrong. Otherwise, we will eventually confronted with the question, why there's a difference and what's the correct version.

iam-mholle: That would need other CSS rule additions in this file. Using the proposed values leads to style induced movement of the anchor element when hovering.

MichaelKetting: ah, okay. you win :)

MichaelKetting (approved):
Please don't forget to update the jira ticket with the caused-by link.

PR #89, opened 2021-04-22
RM-7708: Upgrade Chrome to Version 89

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7708

### DIFF
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocReferenceValueControlObjectTest.cs
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocReferenceValueControlObjectTest.cs
@@ -223,8 +223,8 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
 
       Assert.That (options[15].ItemID, Is.EqualTo (fValue));
       Assert.That (options[15].Index, Is.EqualTo (16));
-      
-      if (Helper.BrowserConfiguration.IsChromium())
+
+      if (Helper.BrowserConfiguration.IsEdge())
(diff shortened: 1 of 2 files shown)


iam-mholle (approved):

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocReferenceValueControlObjectTest.cs
```diff
@@ -223,8 +223,8 @@ public void TestGetOptionDefinitions ()
 
       Assert.That (options[15].ItemID, Is.EqualTo (fValue));
       Assert.That (options[15].Index, Is.EqualTo (16));
-      
-      if (Helper.BrowserConfiguration.IsChromium())
+
+      if (Helper.BrowserConfiguration.IsEdge())
```
MichaelKetting, line 227: Hmm... with the double dependency on Chrome and Edge, this might be a back-and-forth change, isn't it?

raffaelfoidl: My plan would have been to first merge Chrome. Then, after rebasing the Edge branch, this `if` can be disposed of entirely.

This `IsEdge()` check is only still in place here because otherwise, nightly builds would fail for Edge (since it is not upgraded to 89 yet on this branch).

MichaelKetting: ah, okay, that's fine.

MichaelKetting (approved):

PR #91, opened 2021-04-26
RM-7768 Walk through the whole interface hierarchy when searching for service methods in a type

a-ctor:


Remotion/Web/Core/Services/WebServiceFactory.cs
```diff
@@ -98,11 +100,31 @@ private Type GetAndCheckCompiledType<T> (string virtualPath)
...
+      var visitedTypes = new HashSet<Type>();
+      var typesToVisit = new Queue<Type>();
+      typesToVisit.Enqueue (type);
+
+      var serviceMethods = new List<Tuple<string, IReadOnlyCollection<string>>>();
+
+      while (typesToVisit.Count > 0)
+      {
+        var nextType = typesToVisit.Dequeue();
+        if (!visitedTypes.Add (nextType))
```
MichaelKetting, line 112: from a typical-code perspective, this would usually be written with an inverted condition. that way, there's both no `not` and no `continue` *) in the code. I know that it would remove the "it's an guard condition", but this might not be relvant when the rest of the loop code becomes simpler to read. 

*) we try to avoid break and contiune in loops so long as the result code isn't unreadable because we don't have to think about the exit condition while reading the loop.

=> so I'd like to see how the other version looks like before going with continue.

Remotion/Web/UnitTests/Core/Services/WebServiceFactoryTests/CreateScriptService.cs
```diff
@@ -42,14 +42,22 @@ private interface IValidScriptService
       void MethodWithParameters (int param1, bool param2);
     }
 
+    private interface IValidInheritedScriptService : IValidScriptService
+    {
```
MichaelKetting, line 46: I know that from a whitebox approach, this is probably sufficient. I think a version with methods and one without methods would still be helpful to provide sufficient coverage in the event of...odd changes to the code.

Remotion/Web/UnitTests/Core/Services/WebServiceFactoryTests/CreateScriptService.cs
```diff
@@ -133,5 +141,26 @@ public void Test_VirtualPathCannotBeCompiled ()
           () => _webServiceFactory.CreateWebService<IInvalidInterface> ("~/VirtualServicePath"),
           Throws.InvalidOperationException.And.Message.EqualTo ("Web service '~/VirtualServicePath' could not be compiled."));
     }
+
+    [Test]
+    public void Test_InterfaceWithParent ()
```
MichaelKetting, line 146: Naming convention:
Test_DerivedInterfaceWithMethodsOnBaseType

Extra tests:
Test_DerivedInterfaceWithMethodsOnBaseTypeAndDerivedType
Test_DerivedInterfaceWithMethodsOnMultipleBaseTypes

MichaelKetting (changes requested):

Remotion/Web/Core/Services/WebServiceFactory.cs
```diff
@@ -98,11 +100,31 @@ private Type GetAndCheckCompiledType<T> (string virtualPath)
...
+      var visitedTypes = new HashSet<Type>();
+      var typesToVisit = new Queue<Type>();
+      typesToVisit.Enqueue (type);
+
+      var serviceMethods = new List<Tuple<string, IReadOnlyCollection<string>>>();
+
+      while (typesToVisit.Count > 0)
+      {
+        var nextType = typesToVisit.Dequeue();
+        if (visitedTypes.Add (nextType))
```
MichaelKetting, line 112: yeah, I think it looks easier to follow now.

Remotion/Web/UnitTests/Core/Services/WebServiceFactoryTests/CreateScriptService.cs
```diff
@@ -42,14 +42,45 @@ private interface IValidScriptService
...
+    private interface IInvalidInheritedScriptServiceWithMethodsOnBase : IInvalidScriptServiceWithMissingScriptMethodAttribute
+    {
+    }
+
+    private interface IInvalidInheritedScriptServiceWithMethodsOnBaseAndDerived : IInvalidScriptServiceWithMissingScriptMethodAttribute
+    {
+      void InheritedScriptMethodWithoutScriptMethodAttribute ();
+    }
+
+    private interface IInvalidInheritedScriptServiceWithMethodsOnMultipleBases : IInvalidInheritedScriptServiceWithMethodsOnBase
```
MichaelKetting, line 72: Oh, okay, we've got a mis-communication here: I meant multiple-interitance, not a nested hierarchy.

MichaelKetting (changes requested):

PR #96, opened 2021-05-04
RM-7740: Remove _Assembly usages from unit tests

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7740

### DIFF
--- a/Remotion/Mixins/UnitTests/Core/CodeGeneration/AttributeBasedMetadataImporterTest.cs
+++ b/Remotion/Mixins/UnitTests/Core/CodeGeneration/AttributeBasedMetadataImporterTest.cs
@@ -180,63 +180,68 @@ namespace Remotion.Mixins.UnitTests.Core.CodeGeneration
     public void GetMetadataForMixedType ()
     {
       var classContext1 = ClassContextObjectMother.Create(typeof (object));
-
-      var attribute1 = ConcreteMixedTypeAttribute.FromClassContext (classContext1, new Type[0]);
-
-      var typeMock = MockRepository.GenerateMock<_Type> ();
-      typeMock.Expect (mock => mock.GetCustomAttributes (typeof (ConcreteMixedTypeAttribute), false)).Return (new[] { attribute1 });
-      typeMock.Replay ();
-
+      var type = typeof (ClassWithConcreteMixedTypeAttribute);
       var importer = new AttributeBasedMetadataImporter ();
-      var result = importer.GetMetadataForMixedType (typeMock);
-      Assert.That (result, Is.EqualTo (classContext1));
 
-      typeMock.VerifyAllExpectations ();
+      var result = importer.GetMetadataForMixedType (type);
+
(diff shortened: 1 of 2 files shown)


Remotion/Mixins/UnitTests/Core/CodeGeneration/AttributeBasedMetadataImporterTest.cs
```diff
@@ -180,63 +180,46 @@ public void GetMetadataForMixinType_NoAttribute ()
...
     [Test]
     public void GetIdentifierForMixinType ()
     {
-      var typeMock = MockRepository.GenerateMock<_Type> ();
       var identifier = new ConcreteMixinTypeIdentifier (typeof (object), new HashSet<MethodInfo>(), new HashSet<MethodInfo>());
-      var attribute = ConcreteMixinTypeAttribute.Create (identifier);
+      var type = typeof (ClassWithConcreteMixinTypeAttribute);
+      var importer = new AttributeBasedMetadataImporter ();
 
-      typeMock.Expect (mock => mock.GetCustomAttributes (typeof (ConcreteMixinTypeAttribute), false)).Return (new[] { attribute });
```
MichaelKetting, line 218: Given that the test is asserting the attribute inheritance via the expectation, we should add a test that uses a derived type without attribute to verify that we don't inherit the attribute. Yes, the attribute is already defined as `Inherited=false`, but it may be that this code here is the only place that verifies the correct behavior. I would probably also accept a pointer to another piece of code that deals with inheritance, but I think it's quicker to just add a test here for the two attribute types.

MichaelKetting (changes requested):

Remotion/Mixins/UnitTests/Core/CodeGeneration/AttributeBasedMetadataImporterTest.cs
```diff
@@ -368,5 +370,27 @@ private Type CreateTypeWithFakeWrappers ()
...
+    private class ClassWithConcreteMixedTypeAttribute
+    {
+    }
+
+    [ConcreteMixinType (new object[] { typeof (object), new object[0], new object[0] })]
+    private class ClassWithConcreteMixinTypeAttribute
+    {
+    }
+
+    private class DerivedClassWithConcreteMixedTypeAttribute : ClassWithConcreteMixedTypeAttribute
```
MichaelKetting, line 388:
```suggestion
    private class DerivedClassWithConcreteMixedTypeAttributeOnBaseClass : ClassWithConcreteMixedTypeAttribute
```

Naming: the original name was ambigous, we don't know from the class name that the attribute would be on the base.
We could use `DerivedClassForBaseClassWith...Attribute` or the suggestion from the inline edit. I think the second option from the explanation might be more correct but probably less readable? Dealer's choice. The second class also needs a rename then.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #101, opened 2021-05-11
RM-7769: Add support for multiple files downloaded in parallel

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7769

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
@@ -346,6 +346,24 @@ Unmatched files in the download directory (will be cleaned up by the infrastruct
           Throws.Nothing);
     }
 
+    [Test]
+    [Category ("LongRunning")]
+    [Repeat (200)] // This test tests flaky behavior, therefore needs multiple runs.
+    public void TestDownload_HandleMultipleFiles ()
+    {
+      var home = Start();
+
+      var button = home.Scope.FindId ("body_DownloadMultipleFiles");
+      button.Click();
+
+      Assert.That (
+          () => Helper.BrowserConfiguration.DownloadHelper.HandleDownloadWithExpectedFileName (c_sampleXmlFileName),
+          Throws.Nothing);
+      Assert.That (
+          () => Helper.BrowserConfiguration.DownloadHelper.HandleDownloadWithExpectedFileName (c_sampleTxtFileName),
+          Throws.Nothing);
+    }
+
     private WebFormsTestPageObject Start ()
     {
       return Start<WebFormsTestPageObject> ("FileDownloadTest.aspx");
--- a/Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
+++ b/Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
@@ -32,6 +32,8 @@ namespace Remotion.Web.Development.WebTesting.DownloadInfrastructure
   /// </summary>
   public class DownloadedFileFinder
   {
+    // TODO RM-7837: This class and related helpers should be reworked and simplified.
+
     private class FileInformationTuple
     {
       private readonly DateTime _lastWriteTimeUtc;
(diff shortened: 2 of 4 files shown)


Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -101,28 +101,33 @@ public FileInformationTuple (DateTime lastWriteTimeUtc, long length)
...
-              fileInfoOfFoundPartialFile.Length);
+          states.AddRange (fileInfoOfFoundPartialFile.Select (i => new PartialFileState (i.Key, i.Value.LastWriteTimeUtc, i.Value.Length)));
           currentStateTimeout = downloadUpdatedTimeout;
         }
-        else if (PartialFileWasFoundInPreviousIteration (partialFileStateOfCurrentDownload)
-                 && PartialFileWasUpdated (partialFileStateOfCurrentDownload)
-                 && TryGetFileInformation (partialFileStateOfCurrentDownload.GetPartialFile(), out var fileInfoOfPartialFile))
+        else if (PartialFileDetected (states)
+                 && PartialFileWasUpdated (states)
+                 && TryGetFileInformations (states.Select(s => s.GetPartialFile()), out var fileInfosOfPartialFiles))
```
a-ctor, line 119:
```suggestion
                 && TryGetFileInformations (states.Select (s => s.GetPartialFile()), out var fileInfosOfPartialFiles))
```

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -101,28 +101,33 @@ public FileInformationTuple (DateTime lastWriteTimeUtc, long length)
...
         {
-          partialFileStateOfCurrentDownload.UpdatePartialFileLastWriteAccessUtc (fileInfoOfPartialFile.LastWriteTimeUtc);
-          partialFileStateOfCurrentDownload.UpdatePartialFileLength (fileInfoOfPartialFile.Length);
+          foreach (var state in states)
+          {
+            var partialFileName = state.GetPartialFile();
+            if (fileInfosOfPartialFiles.TryGetValue(partialFileName, out var fileInfo)
```
a-ctor, line 124:
```suggestion
            if (fileInfosOfPartialFiles.TryGetValue (partialFileName, out var fileInfo)
```

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -171,23 +186,31 @@ private List<string> GetNewFiles (IEnumerable<string> filesInDownloadDirectoryBe
           .ToList();
     }
 
-    private bool TryGetFileInformation (string fileName, out FileInformationTuple fileInformation)
+    private bool TryGetFileInformations (IEnumerable<string> fileNames, out IDictionary<string, FileInformationTuple> fileInformations)
+    {
+      fileInformations = fileNames.Distinct()
+          .Select(n => new { FileName = n, FileInformation = GetFileInformation(n) })
```
a-ctor, line 192:
```suggestion
          .Select (n => new { FileName = n, FileInformation = GetFileInformation(n) })
```

a-ctor (changes requested):

a-ctor (approved):

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -171,23 +186,31 @@ private List<string> GetNewFiles (IEnumerable<string> filesInDownloadDirectoryBe
           .ToList();
     }
 
-    private bool TryGetFileInformation (string fileName, out FileInformationTuple fileInformation)
+    private bool TryGetFileInformations (IEnumerable<string> fileNames, out IDictionary<string, FileInformationTuple> fileInformations)
```
MichaelKetting, line 189: Given that the dictionary is never null and the only result is an empty or a filled dictionary, this feels like it violates the intent of the try-pattern. I see how it helps with the conditions but it makes the code harder to reason over. For instance, when looking at this method alone, it's not clear why this isn't done as a simple Get-method. 

I haven't played with the code, but I think it can be simplified more / can be made easier to read by further restructuring. 

Caveat: Where does the idea about a complete rewrite factor in to all this?

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
```diff
@@ -346,6 +346,24 @@ public void TestDownload_HandleZipFileDownload ()
           Throws.Nothing);
     }
 
+    [Test]
+    [Category ("LongRunning")]
+    [Repeat (200)] // This test tests flaky behavior, therefore needs multiple runs.
```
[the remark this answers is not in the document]
iam-mholle: The logic now is not flaky, the issue itself is flaky behavior. To reproduce the issue consistently I had to Repeat the test many times. The logic now works as intended, of that I'm sure. Keep in mind that this is "Repeat 200 times until failure" and not "Retry 200 times until success".

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -101,28 +101,33 @@ public FileInformationTuple (DateTime lastWriteTimeUtc, long length)
...
-                 && TryGetFileInformation (partialFileStateOfCurrentDownload.GetPartialFile(), out var fileInfoOfPartialFile))
+        else if (PartialFileDetected (states)
+                 && PartialFileWasUpdated (states)
+                 && TryGetFileInformations (states.Select (s => s.GetPartialFile()), out var fileInfosOfPartialFiles))
         {
-          partialFileStateOfCurrentDownload.UpdatePartialFileLastWriteAccessUtc (fileInfoOfPartialFile.LastWriteTimeUtc);
-          partialFileStateOfCurrentDownload.UpdatePartialFileLength (fileInfoOfPartialFile.Length);
+          foreach (var state in states)
+          {
+            var partialFileName = state.GetPartialFile();
```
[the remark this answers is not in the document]
iam-mholle: This is actually a simple getter https://github.com/re-motion/Framework/blob/4444e406c69d651c4ce25b78a287c106973b4332/Remotion/Web/Development.WebTesting/DownloadInfrastructure/PartialFileState.cs#L61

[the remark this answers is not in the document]
MichaelKetting: Okay, not sure I like it but we shouldn't rewrite this right now, of course.

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
iam-mholle: > Caveat: Where does the idea about a complete rewrite factor in to all this?

@MichaelKetting It doesn't. This PR fixes the current problem with "minimal effort" as a rewrite is planned anyway, as we talked about ~a week ago. Please let me know if the rewrite should happen immediately.

MichaelKetting: @iam-mholle Mis-communication: meant, if we change this to GetFileInformations instead of Try..., is this just busiwork until we do the rework or is it a helpful code cleanup?

iam-mholle: I don't think it would make the code more understandable, rather increase the complexity by adding more nesting as the condition could no longer be written the way it is written now.

I actually also don't think it necessarily violates the `Try...` pattern (in my mind at least 🤷🏼‍♂️) as the boolean return value only indicates if the operation was successful, and that the `out` parameter value can be used if the return value is `true`. If the value is `false`, the `out` parameter value could be `null`, but could also be an instance of some type. Either way, it shouldn't be used from a semantic point of view.

MichaelKetting: @iam-mholle okay, and once we actually do the rewrite, all this wonderful stuff will be deleted anyway and re-written with some other patterns, correct?

iam-mholle: > @iam-mholle okay, and once we actually do the rewrite, all this wonderful stuff will be deleted anyway and re-written with some other patterns, correct?

That's what I would expect.

MichaelKetting: Okay, then please add a ticket for the rewrite, add a todo to the class do document that this will be rewritten and we're done.

MichaelKetting (changes requested):

MichaelKetting (approved):

iam-mholle:
Linked issue closed as won't fix.

PR #103, opened 2021-05-20
RM-7829: Enhance SharedSource/Core/Reflection/TypeExtensions.cs

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7829

### DIFF
--- a/SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
+++ b/SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
@@ -1,0 +1,152 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+
+using System;
+using System.Reflection;
+using NUnit.Framework;
+using Remotion.Reflection;
+
+#nullable enable
+// ReSharper disable once CheckNamespace
+namespace Remotion.UnitTests.Reflection
+{
+  [TestFixture]
+  public class TypeExtensionsTest
+  {
+    [Test]
+    public void GetAssemblyQualifiedNameSafe_WithSystemType_ReturnsAssemblyQualifiedName ()
+    {
+      var typeArgument = typeof (string);
+
+      Assert.That (typeArgument.AssemblyQualifiedName, Is.Not.Null);
+      Assert.That (typeArgument.GetAssemblyQualifiedNameSafe(), Does.StartWith ("System.String,"));
+    }
+
+    [Test]
+    public void GetAssemblyQualifiedNameSafe_WithGenericTypeParameter_ReturnsTypeName ()
+    {
+      var genericTypeParameters = typeof (CustomType<>).GetTypeInfo().GenericTypeParameters;
+
+      Assert.That (genericTypeParameters[0].AssemblyQualifiedName, Is.Null);
+      Assert.That (genericTypeParameters[0].FullName, Is.Null);
+      Assert.That (genericTypeParameters[0].GetAssemblyQualifiedNameSafe(), Is.EqualTo ("T"));
+    }
+
+    [Test]
+    public void GetAssemblyQualifiedNameChecked_WithSystemType_ReturnsAssemblyQualifiedName ()
+    {
+      var typeArgument = typeof (CustomType<string>).GetGenericArguments();
+
+      Assert.That (typeArgument[0].AssemblyQualifiedName, Is.Not.Null);
+      Assert.That (typeArgument[0].GetAssemblyQualifiedNameChecked(), Does.StartWith ("System.String,"));
+    }
+
+    [Test]
+    public void GetAssemblyQualifiedNameChecked_WithGenericTypeParameter_Throws ()
+    {
+      var genericTypeArgumentsParameters = typeof (CustomType<>).GetTypeInfo().GenericTypeParameters;
+
+      Assert.That (genericTypeArgumentsParameters[0].AssemblyQualifiedName, Is.Null);
+      Assert.That (
+          () => genericTypeArgumentsParameters[0].GetAssemblyQualifiedNameChecked(),
+          Throws.InvalidOperationException.With.Message.EqualTo ("Type 'T' does not have an assembly qualified name."));
+    }
+
+    [Test]
+    public void GetFullNameSafe_WithSystemType_ReturnsFullName ()
+    {
+      var typeArgument = typeof (string);
+
+      Assert.That (typeArgument.FullName, Is.Not.Null);
+      Assert.That (typeArgument.GetFullNameSafe(), Is.EqualTo ("System.String"));
+    }
+
+    [Test]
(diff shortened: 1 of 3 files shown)


SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+    {
+      var genericTypeArgumentsParameters = typeof (CustomType<>).GetTypeInfo().GenericTypeParameters;
+      const string expectedTypeArgumentName = "T";
+
+      Assert.That (genericTypeArgumentsParameters, Has.Length.EqualTo (1));
+      Assert.That (genericTypeArgumentsParameters[0].AssemblyQualifiedName, Is.Null);
+      Assert.That (
+          () => genericTypeArgumentsParameters[0].GetAssemblyQualifiedNameChecked(),
+          Throws.InvalidOperationException.With.Message.EqualTo (
+              string.Format ("Type '{0}' does not have an assembly qualified name.", expectedTypeArgumentName)));
```
MichaelKetting, line 74:
```suggestion
          Throws.InvalidOperationException.With.Message.EqualTo ("Type 'T' does not have an assembly qualified name."));
```
In cases such as this, I usually avoid string.format, opting for a completed string instead. The recommendation is always to have the expectations with as little logic as possible. string.format is only chosen if the actual value either changes with each test run or is otherwise highly volatile or dependent on external dependencies (e.g. localized strings that get updated regularily or are reused a lot and would thus result in many updates if changed). It also makes the test easier to read.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+namespace Remotion.UnitTests.Reflection
+{
+  [TestFixture]
+  public class TypeExtensionsTest
+  {
+    [Test]
+    public void GetAssemblyQualifiedNameSafe_SystemType_AssemblyQualifiedName ()
+    {
+      var typeArgument = typeof (string);
+      const string expectedAssemblyQualifiedNamePrefix = "System.String,";
```
MichaelKetting, line 34: Since we don't use the string more than once, it would be preferred to inline the value. That way, there's less indirection when trying to understand the test. This typically improves readability and makes copy&paste errors more obvious, e.g. when there's more than one assertion with a string, for instance when testing AQN and FullName in the same test.

I think I we should do this in all the tests here.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+
+#nullable enable
+// ReSharper disable once CheckNamespace
+namespace Remotion.UnitTests.Reflection
+{
+  [TestFixture]
+  public class TypeExtensionsTest
+  {
+    [Test]
+    public void GetAssemblyQualifiedNameSafe_SystemType_AssemblyQualifiedName ()
```
MichaelKetting, line 31: Namings:
GetAssemblyQualifiedNameSafe_**With**SystemType_**Returns**AssemblyQualifiedName

MichaelKetting (changes requested):

raffaelfoidl:
@iam-mholle

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
raffaelfoidl: I agree and have implemented your suggestion, but see the comment regarding `string.format` for my reasoning.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
raffaelfoidl: I think this was also partly elaborated in the last meeting. My intention behind the layout with the constant and the `string.format` was just that you see - especially for when the value is not volatile or some other developer might stumble upon the test at a later point in time - at one glance which parts of the expectation can (in some way) influenced by the test author (in this case the (generic) type with `CustomType`). So you would see the metaphoric screws to adjust, even without making oneself too familiar with the intrinsics of the test.

But, of course, one has gauge whether that's worth it or how one would want to value that. In any way, your suggestions make the test undeniably clearer, more easily comprehensible and more concise. I have updated the code accordingly.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+    {
+      var typeArgument = typeof (string);
+      const string expectedAssemblyQualifiedNamePrefix = "System.String,";
+
+      Assert.That (typeArgument.AssemblyQualifiedName, Is.Not.Null);
+      Assert.That (typeArgument.GetAssemblyQualifiedNameSafe(), Does.StartWith (expectedAssemblyQualifiedNamePrefix));
```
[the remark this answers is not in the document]
MichaelKetting: >  I opted against doing so because I did not want to "mirror" the implementation of GetAssemblyQualifiedNameSafe too much, but rather test based on the understanding of what an assembly qualified name is.
Yes, that's the reasoning I usually employ in such cases as well. Here, it's just that not the actual intended case. We don't want to emulate an AQN but rather show that we return THE AssemblyQualifiedName for the type used in the test, as opposted to the FullName or the Name. So we don't care about the details, and as long as we can assume it as a given that AQN, FullName and Name will be different, there's no need to be more explicit when ensuring that the correct property is used. If we could mock the type, we could be even more explicit, of course.

> So, if I understand correctly, it's fine to leave it as it is?
Yup.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
MichaelKetting: Thanks. And yes, there's this dichotomy between the SUT code and test code. The SUT must be clean in terms of duplication, obvious in the way the result is constructed, etc. The test must be obvious in terms of the expectation and obviousness as to where the error is. Once that is achieved, fixing the string shouldn't be difficult in the event that ever changes. 

But yes, if we needed different types based on the compilation target, etc, things would be very different indeed.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
MichaelKetting: I can see that and it's probably a matter of style. re-motion does employ the context words in the test fixtures we've written since xUnit Patterns has been published so the mental expectation does exist. So there's the mental expectation to see them and especially the Returns/Throws information is very important since it helps differentiate if the test name is simply done slopily (skipping the Returns information) or this actually being the returned value.

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
```diff
@@ -0,0 +1,152 @@
...
+    public void GetFullNameChecked_WithSystemType_ReturnsFullName ()
+    {
+      var typeArgument = typeof (string);
+
+      Assert.That (typeArgument.FullName, Is.Not.Null);
+      Assert.That (typeArgument.GetFullNameChecked(), Is.EqualTo ("System.String"));
+    }
+
+    [Test]
+    public void GetFullNameChecked_WithGenericTypeParameter_Throws ()
```
MichaelKetting, line 97:
```suggestion
    public void GetFullNameChecked_WithGenericTypeParameter_ThrowsInvalidOperationException ()
```

It's not mandatory for simple cases, but good practice to give the type of the exception

SharedSource/UnitTests/Reflection/TypeExtensionsTest.cs
```diff
@@ -0,0 +1,152 @@
...
+    public void GetNamespaceChecked_WithTypeWithNamespace_ReturnsNamespace ()
+    {
+      var typeWithNamespace = typeof (string);
+
+      Assert.That (typeWithNamespace.Namespace, Is.Not.Null);
+      Assert.That (typeWithNamespace.GetNamespaceChecked(), Is.EqualTo ("System"));
+    }
+
+    [Test]
+    public void GetNamespaceChecked_WithTypeWithoutNamespace_ReturnsThrows ()
```
MichaelKetting, line 135:
```suggestion
    public void GetNamespaceChecked_WithTypeWithoutNamespace_ThrowsInvalidOperationException ()
```

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #106, opened 2021-05-25
RM-7775: Re-implement CallContextStorageProvider based on AsyncLocal<T>

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7775


Remotion/Core/UnitTests/Context/SafeContextIntegrationTest.cs
```diff
@@ -24,7 +24,7 @@ namespace Remotion.UnitTests.Context
   [TestFixture]
   public class SafeContextIntegrationTest
   {
-    [Test]
+    [Test, Ignore ("Testing via CallContext is no longer possible.")]
```
iam-mholle, line 27: @MichaelKetting what should happen to those tests when asserting via CallContext queries is no longer possible?

MichaelKetting: @iam-mholle The basic test logic can remain (unless it's redundant with the unit tests?). As for the rest, it would be helpful if we could set up threaded and async tests to show that a) there's no cross talk between new threads and b) correct behavior exists for async calls. If necesarry, we can split this and only do the threaded tests for now to show the values are isolated between threadpool-threads.

Remotion/Core/UnitTests/Context/CallContextStorageProviderTest.cs
```diff
@@ -20,6 +20,7 @@
 
 namespace Remotion.UnitTests.Context
 {
+  [Obsolete]
```
MichaelKetting, line 23: that test fixture can go bye-bye given the production code is (meant to) throwing exceptions when called.

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// <inheritdoc />
+    public void SetData (string key, object? value)
+    {
+      ArgumentUtility.CheckNotNull ("key", key);
+
+#if DEBUG && NETFRAMEWORK
+      if (value is System.Runtime.Remoting.Messaging.ILogicalThreadAffinative)
+        throw new PlatformNotSupportedException ("Remoting is no longer supported.");
```
MichaelKetting, line 57:
```suggestion
        throw new NotSupportedException ("Remoting is not supported.");
```

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+  [TestFixture]
+  public class AsyncLocalStorageProviderTest
+  {
+    private AsyncLocalStorageProvider _provider;
+
+    [SetUp]
+    public void SetUp ()
+    {
+      _provider = new ();
```
MichaelKetting, line 32:
```suggestion
      _provider = new();
```

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -0,0 +1,161 @@
...
+    public void FreeData_WithoutValue ()
+    {
+      _provider.FreeData ("Foo");
+      Assert.That (_provider.GetData ("Foo"), Is.Null);
+    }
+
+    [Test]
+    public void SetData_FromOtherThread_NotAccessible ()
+    {
+      const string foo = "Foo";
```
MichaelKetting, line 81: Why start here to use a const instead of an inline value as before. We should have the test suite use a consistent form.

If you choose to use a const everywhere, I think `key` would be better to demonstrate in the test what the value means. Somehow, if it's a random string, it's kind of obvious that this is a the key. If it's a variable/const named `foo`, the semantics are much harder to deduce. Instead, at least for me, naming the variable `foo` means 'just some unused variable, please ignore' isntead of 'this is the key and it's important'.

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+#if NETFRAMEWORK && DEBUG
+
+    [Test]
+    public void SetData_WithClassImplementingILogicalThreadAffinative_ThrowsPlatformNotSupportedException ()
+    {
+      var data = new ClassImplementingILogicalThreadAffinative();
+      Assert.That (
+          () => _provider.SetData ("_", data),
+          Throws.InstanceOf<PlatformNotSupportedException>()
```
MichaelKetting, line 151: As you like: if the message fits in the same line as the exception, I don't do a linebreak. I don't mind if it's there, it's just not something I typically do myself.

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+          if (!args.ThreadContextChanged)
+            return;
+
+          if (args.CurrentValue == null)
+            return;
+
+          s_data!.Value = new Hashtable();
+        });
+
+    private static Hashtable Data => s_data.Value ??= new Hashtable();
```
MichaelKetting, line 40: do the tests cover line 37 and line 40? And possibly the other two conditions here?

MichaelKetting (changes requested):

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
iam-mholle: Yes I think so. Have I missed something?

MichaelKetting: I honestly don't know. Will do a coverage test to make sure.

MichaelKetting: Okay, 100% coverage, looks good!

Remotion/Core/Core/Context/CallContextStorageProvider.obsolete.cs
```diff
@@ -16,33 +16,28 @@
...
   /// </summary>
-  [ImplementationFor (typeof (ISafeContextStorageProvider), Position = 1)]
+  [Obsolete ("CallContext is deprecated in .NET Core. Use AsyncLocalStorageProvider instead. (Version 3.0.0-alpha.14)", true)]
   public class CallContextStorageProvider : ISafeContextStorageProvider
   {
     public object GetData (string key)
     {
-      ArgumentUtility.CheckNotNull ("key", key);
-      return CallContext.GetData (key);
+      throw new NotSupportedException();
```
MichaelKetting, line 30: Please add the obsolete message from the attribute to the exceptions. This covers the odd case when there's been no re-compile.

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+    private static readonly AsyncLocal<Hashtable> s_data = new (
+        args =>
+        {
+          if (!args.ThreadContextChanged)
+            return;
+
+          if (args.CurrentValue == null)
+            return;
+
+          s_data!.Value = new Hashtable();
```
MichaelKetting, line 37: Say, except for using an anonymous function directly in the inline ctor, `s_data` would not be nulalble, correct? would it be cleaner if we used a named method instead?

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Threading;
+using NUnit.Framework;
+using Remotion.Context;
+
+namespace Remotion.UnitTests.Context
+{
+  [TestFixture]
+  public class AsyncLocalStorageProviderTest
+  {
+    private const string c_testKey = "Foo";
```
MichaelKetting, line 27: +1

MichaelKetting (changes requested):

Remotion/Core/UnitTests/Core.UnitTests.csproj
```diff
@@ -57,7 +57,7 @@
     <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
     <ConfigurationOverrideFile>
     </ConfigurationOverrideFile>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>DEBUG;TRACE;NETFRAMEWORK;</DefineConstants>
```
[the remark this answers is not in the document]
iam-mholle: I chose `NETFRAMEWORK` here, just like in the last RMTP tasks because thats the predefined preprocessor symbol for full framework when using SDK style projects. Imo this should stay and the occurrences in RM-7800 should be changed. (Refer to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation )

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
iam-mholle: That's true and a nice find.

Remotion/Core/UnitTests/Core.UnitTests.csproj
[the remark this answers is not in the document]
MichaelKetting: Ah, I C.

That's fine, it was just a hipshot for me. Could you please sync with @a-ctor  and @raffaelfoidl to use `NETFRAMEWORK` consistently despite what's written in the Jira tickets?

MichaelKetting (approved):

PR #113, opened 2021-05-28
RM-7806: Disable unit tests that depend on AssemblyCompiler

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7806

### DIFF
--- a/Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
+++ b/Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
@@ -30,6 +30,9 @@ namespace Remotion.UnitTests.Reflection.TypeDiscovery.AssemblyFinding
 {
   [TestFixture]
   [Serializable]
+#if !NETFRAMEWORK
+  [Ignore ("TODO RM-7808: Integrate the RoslynCodeDomProvider and renable the AssemblyCompiler tests")]
+#endif
   public class AssemblyFinderIntegrationTest
   {
     private const string c_testAssemblySourceDirectoryRoot = @"Reflection\TypeDiscovery\TestAssemblies";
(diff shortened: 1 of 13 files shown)


a-ctor (approved):

Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
```diff
@@ -30,6 +30,7 @@ namespace Remotion.UnitTests.Reflection.TypeDiscovery.AssemblyFinding
 {
   [TestFixture]
   [Serializable]
+  [Ignore ("TODO RM-7808: Integrate the RoslynCodeDomProvider and renable the AssemblyCompiler tests")]
```
MichaelKetting, line 33: Please add a preprocessor `!NETFRAMEWORK` to those ignores since the code works in net framework.

a-ctor (changes requested):
I have asked @MichaelKetting how the tests should be ignores since we have some tests that are ignored using pre-processor directives and others that use Ignore.

For this PR we should use both. Currently, the tests would also be ignored in .NET Framework which we do not want. So the ignore attributes should be wrapped into a pre-processor directive like such:

```c#
#if !NETFRAMEWORK
    [Ignore ("TODO RM-7808: Integrate the RoslynCodeDomProvider and renable the AssemblyCompiler tests")]
#endif
```

A simple search and replace should do the trick

MichaelKetting (changes requested):

Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
```diff
@@ -30,7 +30,9 @@ namespace Remotion.UnitTests.Reflection.TypeDiscovery.AssemblyFinding
 {
   [TestFixture]
   [Serializable]
-  [Ignore ("TODO RM-7808: Integrate the RoslynCodeDomProvider and renable the AssemblyCompiler tests")]
+  #if !NETFRAMEWORK
```
a-ctor: The `NETFRAMEWORK` pre-processor directive needs to be added to all the projects. Both in the `DEBUG` and the `RELEASE` configuration using the `DefineConstants` property

a-ctor (changes requested):
The pre-processor directives need to be added to the

a-ctor (approved):

MichaelKetting (approved):

PR #114, opened 2021-05-28
RM-7746: Replace JavaScriptSerializer with System.Text.JSON equivalent

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7746

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericPageTestCaseFactoryBase`.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericPageTestCaseFactoryBase`.cs
@@ -16,13 +16,11 @@
 // 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
-using System.Web.Script.Serialization;
+using System.Text.Json;
 using JetBrains.Annotations;
 using NUnit.Framework;
 using Remotion.Utilities;
-using Remotion.Web.Development.WebTesting.Utilities;
 using Remotion.Web.Development.WebTesting.WebDriver;
 
 namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure
@@ -86,9 +84,8 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure
       if (string.IsNullOrWhiteSpace (text))
         throw new InvalidOperationException ("The generic test page did not provide any test information.");
 
-      var serializer = new JavaScriptSerializer();
-      serializer.RegisterConverters (new[] { GenericTestParameterConverter.Instance });
-      var dto = serializer.Deserialize<GenericTestPageParameterDto> (text);
+      var serializerOptions = new JsonSerializerOptions { Converters = { GenericTestParameterConverter.Instance } };
+      var dto = JsonSerializer.Deserialize<GenericTestPageParameterDto> (text, serializerOptions);
       if (dto == null)
         throw new InvalidOperationException ("The generic test page output is not in a correct format.");
 
--- a/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
@@ -1,69 +1,59 @@
-﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// This file is part of the re-motion Core Framework (www.re-motion.org)
 // Copyright (c) rubicon IT GmbH, www.rubicon.eu
-// 
-// The re-motion Core Framework is free software; you can redistribute it 
-// and/or modify it under the terms of the GNU Lesser General Public License 
-// as published by the Free Software Foundation; either version 2.1 of the 
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
 // License, or (at your option) any later version.
-// 
-// re-motion is distributed in the hope that it will be useful, 
-// but WITHOUT ANY WARRANTY; without even the implied warranty of 
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 // GNU Lesser General Public License for more details.
-// 
+//
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
-// 
+//
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using System.Web.Script.Serialization;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Remotion.Utilities;
 
 // ReSharper disable once CheckNamespace
 
 namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure
 {
-  public class GenericTestParameterConverter : JavaScriptConverter
+  public class GenericTestParameterConverter : JsonConverter<GenericTestPageParameterDto>
   {
-    public static readonly GenericTestParameterConverter Instance = new GenericTestParameterConverter();
-
-    /// <inheritdoc />
-    public override IEnumerable<Type> SupportedTypes { get; } = new[] { typeof (GenericTestPageParameterDto) };
+    public static readonly GenericTestParameterConverter Instance = new();
 
     private GenericTestParameterConverter ()
     {
     }
 
-    /// <inheritdoc />
-    public override object Deserialize (IDictionary<string, object> dictionary, Type type, JavaScriptSerializer serializer)
+    public override GenericTestPageParameterDto Read (ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-      var status = (GenericTestPageStatus) (int) dictionary["status"];
-      var parameters = GetDeserializedParameterDictionary (dictionary["parameters"]);
+      var container = Assertion.IsNotNull (JsonSerializer.Deserialize<Dictionary<string, JsonElement>> (ref reader, options));
+      var status = (GenericTestPageStatus) container["status"].GetInt32();
+      var parameters = container["parameters"].EnumerateObject()
+          .ToDictionary (
+              p => p.Name,
+              p => new GenericTestPageParameter (p.Name, p.Value.EnumerateArray().Select (i => i.GetString()).ToArray()));
 
-      return new GenericTestPageParameterDto (status, parameters);
+      return new (status, parameters);
     }
 
-    /// <inheritdoc />
-    public override IDictionary<string, object> Serialize (object obj, JavaScriptSerializer serializer)
+    public override void Write (Utf8JsonWriter writer, GenericTestPageParameterDto value, JsonSerializerOptions options)
     {
-      var information = (GenericTestPageParameterDto) obj;
-      return new Dictionary<string, object>
-             {
-                 { "status", (int) information.Status },
-                 { "parameters", information.Parameters.ToDictionary (pair => pair.Key, pair => pair.Value.Arguments) }
-             };
-    }
+      var container = new Dictionary<string, object>
+                      {
+                          { "status", (int) value.Status },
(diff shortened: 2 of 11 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericPageTestCaseFactoryBase`.cs
```diff
@@ -86,9 +84,8 @@ protected void PrepareTest ([NotNull] GenericPageTestMethodAttribute attribute,
       if (string.IsNullOrWhiteSpace (text))
         throw new InvalidOperationException ("The generic test page did not provide any test information.");
 
-      var serializer = new JavaScriptSerializer();
-      serializer.RegisterConverters (new[] { GenericTestParameterConverter.Instance });
-      var dto = serializer.Deserialize<GenericTestPageParameterDto> (text);
+      var serializerOptions = new JsonSerializerOptions { Converters = { GenericTestParameterConverter.Instance } };
+      var dto = JsonSerializer.Deserialize<GenericTestPageParameterDto> (text, serializerOptions);
```
raffaelfoidl, line 88: The documentation for exceptions thrown for the `Deserialize<T>` method differ a bit between `System.Text.Json` and the old `JavaScriptSerializer` ([Js](https://docs.microsoft.com/en-us/dotnet/api/system.web.script.serialization.javascriptserializer.deserialize?view=netframework-4.8#System_Web_Script_Serialization_JavaScriptSerializer_Deserialize__1_System_String_), [Text.Json](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-5.0#System_Text_Json_JsonSerializer_Deserialize__1_System_String_System_Text_Json_JsonSerializerOptions_)).

The reasons for throwing an exception will probably boil down to being approximately the same (maybe except the things with `__type`?). But just to be sure, since it's only for internal usage, I guess it's okay to have different exception types thrown in case somethings goes bad?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
```diff
@@ -1,69 +1,108 @@
...
-    public override IDictionary<string, object> Serialize (object obj, JavaScriptSerializer serializer)
+    public override bool CanConvert (Type typeToConvert)
     {
-      var information = (GenericTestPageParameterDto) obj;
-      return new Dictionary<string, object>
-             {
-                 { "status", (int) information.Status },
-                 { "parameters", information.Parameters.ToDictionary (pair => pair.Key, pair => pair.Value.Arguments) }
-             };
+      return typeToConvert == typeof (GenericTestPageParameterDto);
```
raffaelfoidl, line 95: This should be redundant:
_Override the CanConvert method only if necessary. The default implementation returns true when the type to convert is of type T. Therefore, converters that support only type T don't need to override this method (...)._ ([Source](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#steps-to-follow-the-basic-pattern))

Except, of course, we want it to be as explicit as possible.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
```diff
@@ -1,69 +1,108 @@
...
+
+      if (reader.TokenType != JsonTokenType.PropertyName)
+        throw CreateExceptionForExpectedTokenType (JsonTokenType.PropertyName, reader);
+
+      if (reader.GetString() != "parameters")
+        throw CreateExceptionForExpectedTokenStringValue ("parameters", reader);
+
+      reader.Read();
+
+      var dictionary = JsonSerializer.Deserialize<Dictionary<string, IList<string>>> (ref reader);
```
raffaelfoidl, line 61: Is it maybe useful to pass the `JsonSerializerOptions` given by the `options` parameter into this method or plain unnecessary?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
```diff
@@ -1,69 +1,108 @@
...
-      return ((IDictionary<string, object>) serializedParameterDictionary)
-          .ToDictionary (pair => pair.Key, pair => new GenericTestPageParameter (pair.Key, CastToStringArray (pair.Value)));
+      return new ($"Expecting '{expected}', got '{reader.TokenType}' at index {reader.TokenStartIndex}.");
     }
 
-    private string[] CastToStringArray (object x)
+    private JsonException CreateExceptionForExpectedTokenStringValue (string expected, Utf8JsonReader reader)
     {
-      return ((ArrayList) x).Cast<string>().ToArray();
+      return new ($"Expecting value '{expected}', got '{reader.GetString()}' at index {reader.TokenStartIndex}.");
```
raffaelfoidl, line 105: Is this correct?
The docs for `reader.GetString()` state: _Reads the next JSON token value from the source, unescaped, and transcoded as a string._

So when you, for instance, make the `reader.GetString() != "status"` check in L45, throw the exception and call `GetString()` in `CreateExceptionForExpectedTokenStringValue` again - won't you get an incorrect value (the next token value) in `got '...'`?

Maybe I am misinterpreting this (I didn't execute it to check), but for such reasons, I personally would prefer not to pass `reader` into the exception construction methods, but only simple data - in order to reduce the risk of inducing side-effects (even though it might be necessary to store the result of a `reader`-invocation into a variable, then).

a-ctor (changes requested):

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
iam-mholle: I don't think the `GenericTestParameterDto` will ever be nested within itself, but it still is best practice because of other possible options set. I changed it.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericPageTestCaseFactoryBase`.cs
iam-mholle: I do think so, I'll leave this open for the lead review.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
iam-mholle: That's an interesting point. Actually only `.Read()` advances the position in the stream, the various `.Get...` methods only extract the "next" (in fact current) value. So the output in the exception messages would be valid.

It still is a good point to only pass simple values to the exception factory methods. I'll leave this open for the lead review.

raffaelfoidl: Yes, I was not sure if the `.Get...` maybe could be doing an implicit read (and didn't look at the implementation, so I just asked to be sure). But if that's not the case it's nice :)

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
```diff
@@ -1,69 +1,108 @@
...
 
     private GenericTestParameterConverter ()
     {
     }
 
-    /// <inheritdoc />
-    public override object Deserialize (IDictionary<string, object> dictionary, Type type, JavaScriptSerializer serializer)
+    public override GenericTestPageParameterDto Read (ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+      if (reader.TokenType != JsonTokenType.StartObject)
```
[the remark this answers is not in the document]
iam-mholle: I hope the changes I made in [`feb64ac` (#114)](https://github.com/re-motion/Framework/pull/114/commits/feb64ac37782ea417d78d2bd892465aa7733575a) are alright for you.

a-ctor (approved):

raffaelfoidl (approved):

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
```diff
@@ -1,69 +1,105 @@
...
-    public override IEnumerable<Type> SupportedTypes { get; } = new[] { typeof (GenericTestPageParameterDto) };
+    public static readonly GenericTestParameterConverter Instance = new();
 
     private GenericTestParameterConverter ()
     {
     }
 
-    /// <inheritdoc />
-    public override object Deserialize (IDictionary<string, object> dictionary, Type type, JavaScriptSerializer serializer)
+    public override GenericTestPageParameterDto Read (ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
```
MichaelKetting, line 35: I didn't actually play with the code, but couldn't we have done something like 
`dto = JsonSerializer.Deserialize<Dictionary<string, object>> (ref reader, options);`
and then
`var status = (int)dto[status]`
`var parameters = (IDictionary<string, IList<string>)dto[parameters]`

to have less of gritty details? The code appears to be correct and very safe, it just feels like there's a lot of work for 'just test infrastructure' here.

MichaelKetting (commented):
I'd like to discuss my question and then go with approve/request changes.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
iam-mholle: @MichaelKetting  Doing this would result in a `Dictionary<string, JsonElement>` typed as `Dictionary<string, object>`, the dictionary values would still need some kind of extraction/manipulation.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericPageTestCaseFactoryBase`.cs
iam-mholle: @MichaelKetting we do not want any exception, still there is no graceful way of dealing with an error at this point, as no sensible fallback mechanism can be provided that wouldn't hide the original error making it even harder to fix it. But that's just my opinion, I'm open for suggestions.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
MichaelKetting: @iam-mholle okay, so we make it `Dictionary<string, JsonElement>`? Still appears to be the more compact, less detail-y way of doing this?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericPageTestCaseFactoryBase`.cs
MichaelKetting: @iam-mholle okay, I'm confused. I figure it's an exception with the old code and an exception with the new code?

iam-mholle: It is.

I guess I misunderstood the "dealing with the exception" part in the following.

> since this is just our test code, we a) never want to haven an expcetion thrown anyway and b) **will deal with the exceptions if and when they happen**

I thought you mean exception handling in the source code, not handling the error by fixing it when it causes failing tests.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/GenericTestParameterConverter.cs
iam-mholle: Refer to [`cc9bd8e` (#114)](https://github.com/re-motion/Framework/pull/114/commits/cc9bd8e68ef58232f36161b59e024fe98aa0ec66) for the updated implementation.

MichaelKetting: I think it looks better. In theory, we could keep the original code structure with the helper methods, but I don't mind either way.

For serialize: did you try doing the dictionary-to-json approach from the original code base?

iam-mholle: That would work too, I guess its easier to understand and maintain, I just liked the other approach a bit more but I don't mind changing it.

MichaelKetting (approved):

MichaelKetting (approved):
Looks good. To me, the object-based approach is definitely better for readability/maintainability.

PR #115, opened 2021-06-01
RM-7809: Replace usages of AppDomain.BaseDirectory with AppContext.Bas…

iam-mholle:
…eDirectory

https://re-motion.atlassian.net/browse/RM-7809


iam-mholle:
https://github.com/re-motion/Framework/blob/2f987774434766346be0b5ae2af00a8d0d8d523c/Remotion/Core/Core/Reflection/TypeDiscovery/AssemblyFinding/SearchPathRootAssemblyFinder.cs#L30-L43

There still are mentions of `AppDomain.BaseDirectory` here, but I decided to leave them as they are, as it still makes sense and I guess this class will be refactored soon anyway.

MichaelKetting (approved):
Reasoning about the docs was helpful, yes. Thanks for spelling it out, helped me not to think too much about it :)

PR #116, opened 2021-06-04
RM-7794: Refactor QueryConfiguration to allow testing without app domains

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7794

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
@@ -23,8 +23,9 @@ using Remotion.Data.DomainObjects.Persistence;
 using Remotion.Data.DomainObjects.Queries.Configuration;
 using Remotion.Data.DomainObjects.UnitTests.Factories;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain;
-using Remotion.Development.UnitTesting;
 using Remotion.Development.UnitTesting.Configuration;
+using Remotion.Utilities;
+using Rhino.Mocks;
 using File = System.IO.File;
 
 namespace Remotion.Data.DomainObjects.UnitTests.Queries.Configuration
@@ -34,11 +35,31 @@ namespace Remotion.Data.DomainObjects.UnitTests.Queries.Configuration
   {
     private StorageGroupBasedStorageProviderDefinitionFinder _storageProviderDefinitionFinder;
 
+    private string _tempQueriesFilePath;
+    private string _tempQueriesDirectory;
+
+    public override void OneTimeSetUp ()
+    {
+      base.OneTimeSetUp();
+
+      var queriesFile = Path.Combine (TestContext.CurrentContext.TestDirectory, "queries.xml");
+      _tempQueriesDirectory = Path.GetTempPath();
+      _tempQueriesFilePath = Path.Combine (_tempQueriesDirectory, "queries.xml");
+      File.Copy (queriesFile, _tempQueriesFilePath, true);
+    }
+
     public override void SetUp ()
     {
       base.SetUp ();
 
       _storageProviderDefinitionFinder = new StorageGroupBasedStorageProviderDefinitionFinder (DomainObjectsConfiguration.Current.Storage);
+    }
+
+    public override void TestFixtureTearDown ()
+    {
+      File.Delete (_tempQueriesFilePath);
+
+      base.TestFixtureTearDown();
     }
 
     [Test]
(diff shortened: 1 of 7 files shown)


Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -126,98 +128,77 @@ public void GetDefaultQueryFilePath_BaseDirectory ()
 #endif
     public void GetDefaultQueryFilePath_WithRelativeSearchPath ()
     {
-      AppDomainSetup setup = AppDomain.CurrentDomain.SetupInformation;
-      setup.ApplicationBase = Path.GetPathRoot (AppContext.BaseDirectory);
-      setup.DynamicBase = Path.GetTempPath ();
-      setup.PrivateBinPath = Path.GetFullPath (AppContext.BaseDirectory).Substring (setup.ApplicationBase.Length); // make a relative path
+      IAppContextProvider provider = MockRepository.GenerateStub<IAppContextProvider>();
+      provider.Stub (_ => _.BaseDirectory).Return (Path.GetPathRoot (AppContext.BaseDirectory));
```
raffaelfoidl, line 132: Copy `TestContext.CurrentContext.TestDirectory\queries.xml` into a previously created temporary folder and use this temp folder as `BaseDirectory`.

Creating directory, copy file and removing directory afterwards may be done only once in fixture setup and teardown.

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -126,98 +128,77 @@ public void GetDefaultQueryFilePath_BaseDirectory ()
...
+      provider.Stub (_ => _.RelativeSearchPath).Return (Path.GetFullPath (AppContext.BaseDirectory).Substring (provider.BaseDirectory.Length)); // make a relative path
```
raffaelfoidl, line 133: analogously, use the temp directory instead of `AppContext.BaseDirectory`

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -399,5 +380,22 @@ public void Load_ProviderFromUndefinedStorageGroup ()
...
+      AppDomainSetup setup = AppDomain.CurrentDomain.SetupInformation;
+      setup.ApplicationBase = @"C:\Base\";
+      setup.DynamicBase = Path.GetTempPath ();
+      setup.PrivateBinPath = @"Custom\Relative;Search\Path";
+
+      new AppDomainRunner (setup, delegate
+      {
+        Assert.That (AppDomain.CurrentDomain.BaseDirectory, Is.EqualTo (@"C:\Base\"));
+        Assert.That (AppDomain.CurrentDomain.RelativeSearchPath, Is.EqualTo (@"Custom\Relative;Search\Path"));
+      }).Run ();
```
raffaelfoidl, line 397: superflious whitespace in `Run ()`

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -399,5 +380,22 @@ public void Load_ProviderFromUndefinedStorageGroup ()
           queries["QueryFromUndefinedStorageGroup"].StorageProviderDefinition,
           Is.SameAs(DomainObjectsConfiguration.Current.Storage.DefaultStorageProviderDefinition));
     }
+
+#if NETFRAMEWORK
+    [Test]
+    public void CodeRunInSeparateAppDomain_AppDomainSetupSetsPropertiesCorrectly ()
```
raffaelfoidl, line 386: Move this test to test fixture for `IAppContextProvider` implementation.

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -399,5 +380,22 @@ public void Load_ProviderFromUndefinedStorageGroup ()
...
+    {
+      AppDomainSetup setup = AppDomain.CurrentDomain.SetupInformation;
+      setup.ApplicationBase = @"C:\Base\";
+      setup.DynamicBase = Path.GetTempPath ();
+      setup.PrivateBinPath = @"Custom\Relative;Search\Path";
+
+      new AppDomainRunner (setup, delegate
+      {
+        Assert.That (AppDomain.CurrentDomain.BaseDirectory, Is.EqualTo (@"C:\Base\"));
+        Assert.That (AppDomain.CurrentDomain.RelativeSearchPath, Is.EqualTo (@"Custom\Relative;Search\Path"));
```
raffaelfoidl, line 396: Assert against `IAppContextProvider` implementation instead of AppDomain.
(1 test for `NETFRAMEWORK`, 1 for `!NETFRAMEWORK`)

Remotion/Data/DomainObjects/Queries/AppContextProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Remotion.Data.DomainObjects.Queries
+{
+  /// <summary>
+  /// Represents the default implementation of the <see cref="IAppContextProvider"/> interface.
+  /// </summary>
+  public class AppContextProvider : IAppContextProvider
+  {
+    public string BaseDirectory => AppContext.BaseDirectory;
+
+    public string RelativeSearchPath => AppDomain.CurrentDomain.RelativeSearchPath;
```
raffaelfoidl, line 12: We need different implementations for L12 for full framework and .NET 5.

In the case of .NET 5, `null`, `""` or `"."`.
Have a look, without strong consideration of implications on the functionalities, at semantics entailed by the different options.

Interface + Implementation need to move to `Remotion/Core/Core/Utilities`.

raffaelfoidl (commented):
Live-Paired-Review

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -126,98 +147,77 @@ public void GetDefaultQueryFilePath_BaseDirectory ()
...
 
-      new AppDomainRunner (setup, delegate (object[] args)
-      {
-        QueryConfiguration configuration = new QueryConfiguration ();
-        Assert.That (!File.Exists (Path.Combine (AppContext.BaseDirectory, "queries.xml")));
-        Assert.That (configuration.GetDefaultQueryFilePath (), Is.EqualTo (Path.Combine ((string) args[0], "queries.xml")));
-      }, AppContext.BaseDirectory).Run();
+      QueryConfiguration configuration = new QueryConfiguration (providerStub);
+      Assert.That (!File.Exists (Path.Combine (providerStub.BaseDirectory, "queries.xml")));
+      Assert.That (configuration.GetDefaultQueryFilePath (), Is.EqualTo (Path.Combine (_tempQueriesDirectory, "queries.xml")));
```
raffaelfoidl, line 156: Unfortunately, I could not remember clearly if we had said the assertion in line 155 (`!File.Exists`) was still necessary or not. Did I forget to write it to the review or did we conclude it's fine?

MichaelKetting: That line can go the way of the [Dodo ](https://en.wikipedia.org/wiki/Dodo). It's only been there to document that the file actually exists where we think it exists. Maybe it was sensible in the old test code, it's definitely not in the new test code.

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
```diff
@@ -126,98 +147,75 @@ public void GetDefaultQueryFilePath_BaseDirectory ()
 #endif
     public void GetDefaultQueryFilePath_WithRelativeSearchPath ()
     {
-      AppDomainSetup setup = AppDomain.CurrentDomain.SetupInformation;
-      setup.ApplicationBase = Path.GetPathRoot (AppContext.BaseDirectory);
-      setup.DynamicBase = Path.GetTempPath ();
-      setup.PrivateBinPath = Path.GetFullPath (AppContext.BaseDirectory).Substring (setup.ApplicationBase.Length); // make a relative path
+      IAppContextProvider providerStub = MockRepository.GenerateStub<IAppContextProvider>();
+      providerStub.Stub (_ => _.BaseDirectory).Return (Path.GetPathRoot (_tempQueriesDirectory));
+      providerStub.Stub (_ => _.RelativeSearchPath).Return (Path.GetFullPath (_tempQueriesDirectory).Substring (providerStub.BaseDirectory.Length)); // make relative path
```
MichaelKetting, line 152: We should not use stubbed values when setting up other stubs. Yes, sometimes it's convenient, and yes, there's code with this pattern in the code base, but it's still not a great practice. 

Reasoning: it can theoretically result in some weird bugs if there's an ordering issue in the mock setup. Plus, it may imply a logic duplication with the production code, which should also be avoided in tests.

Remotion/Data/DomainObjects/Queries/Configuration/QueryConfiguration.cs
```diff
@@ -51,6 +56,10 @@ public QueryConfiguration ()
       _properties.Add (_queryFilesProperty);
     }
 
+    public QueryConfiguration () : this(new AppContextProvider())
+    {
+    }
+
     public QueryConfiguration (params string[] configurationFiles) : this()
```
MichaelKetting, line 63: We need an overload with IAppContextProvider

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Queries/Configuration/QueryConfiguration.cs
raffaelfoidl: Maybe I am misunderstanding you, but there is an overload with `IAppContextProvider`, defined in lines 43-57?

MichaelKetting: We are missing the overload for the ctor that accepts configuration files. 

And while we're at it (and the pattern is a bit unfortunate here), but in theory, all overloads for ctors should converge to a single implementation unless there's a reason against it. Nominally, this change would be out of scope of the current task, unless the code needed to introduce IAppContextProvider creates more noise/less maintainable code than what we had before. Said differently, if the new ctor parameter requires we refactor the ctors first to have one implemention and several overloads, that's okay.

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/QueryConfigurationTest.cs
raffaelfoidl: Changed in c6ec9d3cb9825d5cee347bbe71b1eb6bed142966.

Remotion/Data/DomainObjects/Queries/Configuration/QueryConfiguration.cs
raffaelfoidl: Again, I'm hoping I got you right. I refactored the constructor(s) in b58204f27459d9d476d97917bc3e60a8f91dee96. IMO, it does not really cluster the change set or make noisy (given I interpreted you correctly).

Remotion/Data/DomainObjects/Queries/Configuration/QueryConfiguration.cs
```diff
@@ -63,6 +64,18 @@ public QueryConfiguration (params string[] configurationFiles) : this()
...
 
+    public QueryConfiguration (IAppContextProvider provider) : this (provider, new string[0])
+    {
+    }
+
+    public QueryConfiguration (params string[] configurationFiles) : this (new AppContextProvider(), configurationFiles)
+    {
+    }
+
+    public QueryConfiguration () : this (new AppContextProvider(), new string[0])
```
MichaelKetting, line 75: Looks good. Just switch the order of the ctor methods, please: the empty ctor first, then the other overloads and finally the full implementation. (that's the typically used order with ctors and, I think, other overloads, too.)

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #121, opened 2021-06-09
RM-7846 Fix incorrect DatePicker positioning when the root node is scrolled

a-ctor:
https://re-motion.atlassian.net/browse/RM-7846

### DIFF
--- a/Remotion/ObjectBinding/Web.Test/StandardMode.Master
+++ b/Remotion/ObjectBinding/Web.Test/StandardMode.Master
@@ -3,7 +3,7 @@
 <%@ Import Namespace="System" %>
 
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" ID="Html" runat="server">
 <head runat="server">
   <meta http-equiv="X-UA-Compatible" content="IE=Edge">
   <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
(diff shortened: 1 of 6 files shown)


iam-mholle (changes requested):
As communicated per dm, DatePicker popup can overflow to the right, I think it should snap to the right viewport border in this case, as for all other directions.

iam-mholle (approved):
Works great
stay there

Remotion/ObjectBinding/Web.Test/IndividualControlTests/Form.aspx
```diff
@@ -44,7 +44,26 @@
       </View>
       
       <BottomControls>
-         <asp:UpdatePanel ID="StackUpdatePanel" runat="server" UpdateMode="Conditional">
+        <asp:CheckBox ID="MainContentScrollableCheckBox" runat="server" AutoPostBack="True" Text="Main content scrollable" OnCheckedChanged="MainContentScrollableCheckBox_OnCheckedChanged"/>
```
MichaelKetting, line 47: I'd like to do a pair-review for the ASPX stuff to see if there're sensible improvement options.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #130, opened 2021-06-15
RM-7597: Add NRT support in security

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7597

### DIFF
--- a/Remotion/Security/Core/Metadata/AccessTypeReflector.cs
+++ b/Remotion/Security/Core/Metadata/AccessTypeReflector.cs
@@ -123,7 +123,7 @@ namespace Remotion.Security.Metadata
       return instanceMethods.Cast<MethodInfo>();
     }
 
-    private bool FindSecuredMembersFilter (MemberInfo member, object filterCriteria)
+    private bool FindSecuredMembersFilter (MemberInfo member, object? filterCriteria)
     {
       return Attribute.IsDefined (member, typeof (DemandPermissionAttribute), true);
     }
@@ -132,7 +132,7 @@ namespace Remotion.Security.Metadata
     {
       foreach (var methodInfo in methodInfos)
       {
-        var values = _permissionReflector.GetRequiredMethodPermissions (methodInfo.DeclaringType, MethodInfoAdapter.Create(methodInfo));
+        var values = _permissionReflector.GetRequiredMethodPermissions (methodInfo.DeclaringType!, MethodInfoAdapter.Create(methodInfo));
         foreach (Enum value in values)
         {
           EnumValueInfo accessType = _enumerationReflector.GetValue (value, cache);
--- a/Remotion/Security/Core/Metadata/EnumerationReflector.cs
+++ b/Remotion/Security/Core/Metadata/EnumerationReflector.cs
@@ -52,25 +52,25 @@ namespace Remotion.Security.Metadata
       Dictionary<Enum, EnumValueInfo> enumValueInfos = new Dictionary<Enum, EnumValueInfo> ();
       for (int i = 0; i < values.Count; i++)
       {
-        Enum value = (Enum) values[i];
+        Enum? value = (Enum?) values[i];
         enumValueInfos.Add (value, GetValue (value, cache));
       }
 
       return enumValueInfos;
     }
 
-    public EnumValueInfo GetValue (Enum value, MetadataCache cache)
+    public EnumValueInfo GetValue (Enum? value, MetadataCache cache)
     {
       ArgumentUtility.CheckNotNull ("value", value);
       ArgumentUtility.CheckNotNull ("cache", cache);
 
-      EnumValueInfo info = cache.GetEnumValueInfo (value);
+      EnumValueInfo? info = cache.GetEnumValueInfo (value);
       if (info == null)
       {
         string name = value.ToString ();
         info = new EnumValueInfo (TypeUtility.GetPartialAssemblyQualifiedName (value.GetType ()), name, Convert.ToInt32 (value));
-        FieldInfo fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public);
-        PermanentGuidAttribute attribute = (PermanentGuidAttribute) Attribute.GetCustomAttribute (fieldInfo, typeof (PermanentGuidAttribute), false);
+        FieldInfo? fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public);
+        PermanentGuidAttribute? attribute = (PermanentGuidAttribute?) Attribute.GetCustomAttribute (fieldInfo, typeof (PermanentGuidAttribute), false);
         if (attribute != null)
           info.ID = attribute.Value.ToString ();
 
--- a/Remotion/Security/Core/Metadata/EnumerationReflector.cs
+++ b/Remotion/Security/Core/Metadata/EnumerationReflector.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using Remotion.Reflection;
 using Remotion.Utilities;
@@ -52,14 +53,14 @@ namespace Remotion.Security.Metadata
       Dictionary<Enum, EnumValueInfo> enumValueInfos = new Dictionary<Enum, EnumValueInfo> ();
       for (int i = 0; i < values.Count; i++)
       {
-        Enum? value = (Enum?) values[i];
+        Enum value = (Enum) values[i]!;
         enumValueInfos.Add (value, GetValue (value, cache));
       }
 
       return enumValueInfos;
     }
 
-    public EnumValueInfo GetValue (Enum? value, MetadataCache cache)
+    public EnumValueInfo GetValue (Enum value, MetadataCache cache)
     {
       ArgumentUtility.CheckNotNull ("value", value);
       ArgumentUtility.CheckNotNull ("cache", cache);
@@ -69,7 +70,8 @@ namespace Remotion.Security.Metadata
       {
         string name = value.ToString ();
         info = new EnumValueInfo (TypeUtility.GetPartialAssemblyQualifiedName (value.GetType ()), name, Convert.ToInt32 (value));
-        FieldInfo? fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public);
+        // TODO RM-7868: Add debug notnull assertion
+        FieldInfo fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public)!;
         PermanentGuidAttribute? attribute = (PermanentGuidAttribute?) Attribute.GetCustomAttribute (fieldInfo, typeof (PermanentGuidAttribute), false);
         if (attribute != null)
           info.ID = attribute.Value.ToString ();
--- a/Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
+++ b/Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
@@ -22,10 +22,13 @@ namespace Remotion.Security.Metadata
 {
   public class LocalizationFileNameStrategy
   {
+    // TODO RM-7869: Add missing argument checks
+
     public string GetLocalizationFileName (string metadataFilename, CultureInfo culture)
     {
       string baseFilename = Path.GetFileNameWithoutExtension (metadataFilename);
-      string? basePath = Path.GetDirectoryName (metadataFilename);
+      // TODO RM-7870: Add notnull assertion
+      string basePath = Path.GetDirectoryName (metadataFilename)!;
       string baseFilePath = Path.Combine (basePath, baseFilename);
 
       if (string.IsNullOrEmpty (culture.Name))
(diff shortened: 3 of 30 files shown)


iam-mholle:
[`3a05d95` (#130)](https://github.com/re-motion/Framework/pull/130/commits/3a05d95ef94bd9d39302dabb7c97af37ed650de4) and [`7dec4c2` (#130)](https://github.com/re-motion/Framework/pull/130/commits/7dec4c2d6c926e95ec1a5f8660aec05319b2bf62) should be separate issues in the end, right?

Remotion/Core/Core/Utilities/StringUtility.cs
```diff
@@ -250,7 +250,7 @@ public static string NullToEmpty (string? str)
       return str;
     }
 
-    public static string? EmptyToNull (string str)
+    public static string? EmptyToNull (string? str)
```
MichaelKetting, line 253: Separate ticket (and PR, if possible), nullbug as suggested

SharedSource/Core/Utilities/EqualityUtility.cs
```diff
@@ -229,7 +229,7 @@ public static bool EqualsEquatableValue<T> (T a, object? b)
         return false;
     }
 
-    public static bool NotNullAndSameType<T> (T a, T? b)
+    public static bool NotNullAndSameType<T> (T a, [NotNullWhen (true)] T? b)
```
MichaelKetting, line 232: Separte ticket and PR, nullbug as suggested.

Remotion/Security/Core/Metadata/ReflectionBasedMemberResolver.cs
```diff
@@ -144,9 +144,9 @@ private static bool TypeHasMember (Type type, string methodName, BindingFlags bi
       return existingMembers.Length > 0;
     }
 
-    private static bool IsSecuredMethod (MemberInfo memberInfo, object filterCriteria)
+    private static bool IsSecuredMethod (MemberInfo memberInfo, object? filterCriteria)
     {
-      string memberName = (string) filterCriteria;
+      var memberName = (string?) filterCriteria;
```
MichaelKetting, line 149: I think filterCriteria would not be allowed to be null in our usecase?It's just the signature of FindMembers that makes it nullable. Should we add a debug-nullguard?

Remotion/Security/Core/Metadata/StatePropertyReflector.cs
```diff
@@ -58,25 +58,25 @@ public StatePropertyInfo GetMetadata (PropertyInfo property, MetadataCache cache
       if (!property.PropertyType.IsEnum)
       {
         throw new ArgumentException (
-            string.Format ("The type of the property '{0}' in type '{1}' is not an enumerated type.", property.Name, property.DeclaringType.GetFullNameSafe()),
+            string.Format ("The type of the property '{0}' in type '{1}' is not an enumerated type.", property.Name, property.DeclaringType!.GetFullNameSafe()),
```
MichaelKetting, line 61: Did we use null-bang or null-propagation for exceptions in earlier PRs? I know we use null-propagation for logging and when rethrowing an exception, but what about the message of initial exceptions?

Remotion/Security/Core/Metadata/AccessTypeReflector.cs
```diff
@@ -132,7 +132,7 @@ private void AddAccessTypesFromAttribute (IEnumerable<MethodInfo> methodInfos, D
     {
       foreach (var methodInfo in methodInfos)
       {
-        var values = _permissionReflector.GetRequiredMethodPermissions (methodInfo.DeclaringType, MethodInfoAdapter.Create(methodInfo));
+        var values = _permissionReflector.GetRequiredMethodPermissions (methodInfo.DeclaringType!, MethodInfoAdapter.Create(methodInfo));
```
MichaelKetting, line 135: Do we have a ticket about guarding DeclaringType get-access?

Remotion/Security/Core/Metadata/EnumerationReflector.cs
```diff
@@ -52,7 +52,8 @@ public EnumerationReflector ()
       Dictionary<Enum, EnumValueInfo> enumValueInfos = new Dictionary<Enum, EnumValueInfo> ();
       for (int i = 0; i < values.Count; i++)
       {
-        Enum value = (Enum) values[i];
+        // TODO: Add notnull assertion
```
MichaelKetting, line 55: TODO-solving now or with ticket? (see all other TODOs)

MichaelKetting (changes requested):
I'm done except for the TODOs.

Remotion/Security/Core/Metadata/StatePropertyReflector.cs
iam-mholle: Null bangs were used

Remotion/Security/Core/Metadata/EnumerationReflector.cs
iam-mholle: I would add tickets, just wanted to know if you agree with the todos first.

Remotion/Security/Core/Metadata/AccessTypeReflector.cs
iam-mholle: There is [RM-7811](https://re-motion.atlassian.net/browse/RM-7811) for the Mixins project, but it could be made more generic to add guarded extension methods for DeclaringType, just like for FullName.

Remotion/Security/Core/Metadata/EnumerationReflector.cs
MichaelKetting: okay, I'll check

Remotion/Security/Core/Metadata/EnumerationReflector.cs
```diff
@@ -52,7 +52,8 @@ public EnumerationReflector ()
...
+        Enum value = (Enum) values[i]!;
```
MichaelKetting, line 56: Could you check, please, I think we can change `IList` in line 50 to `Enum[]` . That way, we don't need a not-null assertion or a bang. Not sure if doing this change belongs in the current PR or should be a separate task (read, we could just fix it if you like).

Remotion/Security/Core/Metadata/EnumerationReflector.cs
```diff
@@ -64,13 +65,14 @@ public EnumValueInfo GetValue (Enum value, MetadataCache cache)
...
-      EnumValueInfo info = cache.GetEnumValueInfo (value);
+      EnumValueInfo? info = cache.GetEnumValueInfo (value);
       if (info == null)
       {
         string name = value.ToString ();
         info = new EnumValueInfo (TypeUtility.GetPartialAssemblyQualifiedName (value.GetType ()), name, Convert.ToInt32 (value));
-        FieldInfo fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public);
-        PermanentGuidAttribute attribute = (PermanentGuidAttribute) Attribute.GetCustomAttribute (fieldInfo, typeof (PermanentGuidAttribute), false);
+        // TODO: Add notnull assertion
+        FieldInfo fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public)!;
```
MichaelKetting, line 74: that's a debug-notnull at most. The value cannot be null given how .net works. Ticket or update

Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
```diff
@@ -25,7 +25,8 @@ public class LocalizationFileNameStrategy
     public string GetLocalizationFileName (string metadataFilename, CultureInfo culture)
     {
       string baseFilename = Path.GetFileNameWithoutExtension (metadataFilename);
-      string basePath = Path.GetDirectoryName (metadataFilename);
+      // TODO: Add notnull assertion
+      string basePath = Path.GetDirectoryName (metadataFilename)!;
```
MichaelKetting, line 29: todo with possible null-bug-guard via Assert-not-null since it should check that the metadataFilename is correctly formatted as a full path. Also, this is a public method w/o arg-checks. not nice.

Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
```diff
@@ -37,7 +38,8 @@ public string GetLocalizationFileName (string metadataFilename, CultureInfo cult
     public string[] GetLocalizationFileNames (string metadataFilename)
     {
       string baseFileName = Path.GetFileNameWithoutExtension (metadataFilename);
-      string basePath = Path.GetDirectoryName (metadataFilename);
+      // TODO: Add notnull assertion
+      string basePath = Path.GetDirectoryName (metadataFilename)!;
```
MichaelKetting, line 42: see line 29

Remotion/Security/Core/Metadata/LocalizingMetadataConverter.cs
```diff
@@ -78,18 +78,21 @@ private LocalizedName[] GetLocalizedNames (MetadataCache cache, CultureInfo cult
 
     private LocalizedName CreateLocalizedNameFromClassInfo (SecurableClassInfo classInfo, string text)
     {
-      return new LocalizedName (classInfo.ID, classInfo.Name, classInfo.Description);
+      // TODO: Add debug notnull assertions
```
MichaelKetting, line 81: those lazy initialzied Info-properties are not the best concept as far as null-safety is concerned. for now, we can leave the bang given (the LocalizedName does argchecks anyway and a follow-up ticket/todo to document this should be prevented via the typesystem. (e.g. builders) or by some other refactoring.

Remotion/Security/Core/Metadata/MetadataCache.cs
```diff
@@ -102,10 +102,11 @@ private PropertyInfo NormalizeProperty (PropertyInfo property)
       if (property.DeclaringType == property.ReflectedType)
         return property;
       else
-        return property.DeclaringType.GetProperty (property.Name);
+        // TODO: Add debug notnull assertion
```
MichaelKetting, line 105: notnull-assertion.  (not debugnotnull, since the property info could theoretically be passed from uncontrolled external code.

Remotion/Security/Core/Metadata/MetadataToXmlConverter.cs
```diff
@@ -98,7 +98,8 @@ private void AppendEnumValueInfoAttributes (XmlDocument document, EnumValueInfo
       XmlAttribute enumValueIDAttribute = document.CreateAttribute ("id");
       enumValueIDAttribute.Value = enumValueInfo.ID;
 
-      EnumWrapper enumWrapper = EnumWrapper.Get(enumValueInfo.Name, enumValueInfo.TypeName);
+      // TODO: Add debug notnull assertions
+      EnumWrapper enumWrapper = EnumWrapper.Get(enumValueInfo.Name!, enumValueInfo.TypeName);
```
MichaelKetting, line 102: Same rule as with LocalizedName in other file. leave the bang (arg checks exist) and link the follow-up task/todo from before.

Remotion/Security/Core/ThreadPrincipalProvider.cs
```diff
@@ -33,11 +33,12 @@ public ThreadPrincipalProvider ()
 
     public ISecurityPrincipal GetPrincipal ()
     {
-      IIdentity identity = Thread.CurrentPrincipal.Identity;
+      // TODO: Add notnull assertions
+      IIdentity identity = Thread.CurrentPrincipal?.Identity!;
```
MichaelKetting, line 37: assert-not-null for both hops.

Remotion/Security/Core/ThreadPrincipalProvider.cs
```diff
@@ -33,11 +33,12 @@ public ThreadPrincipalProvider ()
...
       if (!identity.IsAuthenticated)
         return s_nullSecurityPrincipal;
 
-      return new SecurityPrincipal (identity.Name, null, null, null);
+      return new SecurityPrincipal (identity.Name!, null, null, null);
```
MichaelKetting, line 41: probably want to do an assert-not-null here since we don't control the IIdentity implementation. and we had an issue with someone implementing IIdentity the wrong way .

MichaelKetting (changes requested):

Remotion/Security/Core/Metadata/EnumerationReflector.cs
iam-mholle: Implicit conversion does not work at compile type, explicit cast throws at runtime "Unable to cast object of type '<enumtype>[]' to type 'System.Enum[]'".

The following would work:
```csharp
var values = Enum.GetValues (type).Cast<Enum>().ToArray();

Dictionary<Enum, EnumValueInfo> enumValueInfos = new Dictionary<Enum, EnumValueInfo> ();
for (int i = 0; i < values.Length; i++)
{
  // TODO: Add notnull assertion
  Enum value = (Enum) values[i]!;
  enumValueInfos.Add (value, GetValue (value, cache));
}
```

MichaelKetting: Hmm... yeah, don't think this is worth it (extra runtime cost and code busy work to get pseudo-typesafety). probably should just stick with the bang and be done with it.

iam-mholle: Alright, then I'll just remove the TODO and leave the bang.

Remotion/Security/Core/Metadata/AccessTypeReflector.cs
iam-mholle: @MichaelKetting should I do this?

MichaelKetting: Huh, I thought I wrote an answer: yes, please (unless it starts to look like it's getting too big for a single task).

MichaelKetting (approved):

PR #135, opened 2021-06-18
RM-7840: Replace Rhino.Mocks with Moq in Security

derlistigelurch:


Remotion/Security/UnitTests/ObjectSecurityStrategyTests/HasAccessWithCaching_ObjectSecurityStrategyTest.cs
```diff
@@ -16,164 +16,163 @@
...
       var cache = new Cache<ISecurityPrincipal, AccessType[]>();
-      var strategy = ObjectSecurityStrategy.CreateWithCustomCache (_securityContextFactoryStub, cache);
+      var strategy = ObjectSecurityStrategy.CreateWithCustomCache (_securityContextFactoryStub.Object, cache);
 
       _securityProviderMock
-          .Expect (_ => _.GetAccess (null, null))
-          .IgnoreArguments()
-          .Throw (new InvalidOperationException ("Should not be called."));
+          .Setup  (_ => _.GetAccess (It.IsAny<ISecurityContext>(), It.IsAny<ISecurityPrincipal>()))
+          .Throws  (new InvalidOperationException ("Should not be called."))
```
iam-mholle, line 168:
```suggestion
          .Setup (_ => _.GetAccess (It.IsAny<ISecurityContext>(), It.IsAny<ISecurityPrincipal>()))
          .Throws (new InvalidOperationException ("Should not be called."))
```

iam-mholle (changes requested):

iam-mholle (approved):

Remotion/Security/UnitTests/SecurityClientTests/SecurityClientTestHelper.cs
```diff
@@ -17,44 +17,42 @@
...
-      SetupResult.For (_stubPrincipalProvider.GetPrincipal()).Return (_userStub);
-
-      _securableObject = new SecurableObject (_mockObjectSecurityStrategy);
+      _mockSecurityProvider = new Mock<ISecurityProvider> (MockBehavior.Strict);
+      _mockPermissionReflector = new Mock<IPermissionProvider> (MockBehavior.Strict);
+      _mockObjectSecurityStrategy = new Mock<IObjectSecurityStrategy> (MockBehavior.Strict);
+      _mockFunctionalSecurityStrategy = new Mock<IFunctionalSecurityStrategy> (MockBehavior.Strict);
+      _mockMemberResolver = new Mock<IMemberResolver> (MockBehavior.Strict);
+      _userStub = new Mock<ISecurityPrincipal>();
+      _userStub.Setup (_ => _.User).Returns ("user").Verifiable();
```
MichaelKetting, line 46: Why did the tooling make this verifyable?

Remotion/Security/UnitTests/SecurityClientTests/SecurityClientTestHelper.cs
```diff
@@ -17,44 +17,42 @@
...
+      _stubPrincipalProvider = new Mock<IPrincipalProvider>();
+      _stubPrincipalProvider.Setup (_ => _.GetPrincipal()).Returns (_userStub.Object).Verifiable();
```
MichaelKetting, line 48: Same here

MichaelKetting (changes requested):

Remotion/Security/UnitTests/SecurityClientTests/SecurityClientTestHelper.cs
derlistigelurch: The original line was `SetupResult.For (_userStub.User).Return ("user");`
`SetupResult.For` is converted to `Expect (...)`, which is later converted to `Setup (...).Verifiable (...)`.
So it does not matter if `SetupResult.For` contains a `Stub` or a `Mock` it will always be verifiable. However in contrast to a `Mock` a `Stub` is not automatically verified in a converted `VerifyAll` statement, so even though the converted `Stub` is verifiable it will not be verfified.

MichaelKetting: okay...I think we once discussed this as Expect is always acting as a Stub if there's a stub in there. So, the conversion could take that into account. Anyhow... we can drop the Verifyable here and below.

MichaelKetting (approved):

PR #139, opened 2021-06-24
RM-7868: Add DebugNotNull check for fieldInfo in Remotion.Security.Metadata.EnumerationReflector.GetValue

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7868

### DIFF
--- a/Remotion/Security/Core/Metadata/EnumerationReflector.cs
+++ b/Remotion/Security/Core/Metadata/EnumerationReflector.cs
@@ -70,8 +70,8 @@ namespace Remotion.Security.Metadata
       {
         string name = value.ToString ();
         info = new EnumValueInfo (TypeUtility.GetPartialAssemblyQualifiedName (value.GetType ()), name, Convert.ToInt32 (value));
-        // TODO RM-7868: Add debug notnull assertion
-        FieldInfo fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public)!;
+        FieldInfo? fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public);
+        Assertion.DebugIsNotNull (fieldInfo, "Field '{0}' was not found on type '{1}'.", name, value.GetType());
         PermanentGuidAttribute? attribute = (PermanentGuidAttribute?) Attribute.GetCustomAttribute (fieldInfo, typeof (PermanentGuidAttribute), false);
         if (attribute != null)
           info.ID = attribute.Value.ToString ();


iam-mholle (approved):

Remotion/Security/Core/Metadata/EnumerationReflector.cs
```diff
@@ -70,8 +70,8 @@ public EnumValueInfo GetValue (Enum value, MetadataCache cache)
       {
         string name = value.ToString ();
         info = new EnumValueInfo (TypeUtility.GetPartialAssemblyQualifiedName (value.GetType ()), name, Convert.ToInt32 (value));
-        // TODO RM-7868: Add debug notnull assertion
-        FieldInfo fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public)!;
+        FieldInfo? fieldInfo = value.GetType ().GetField (name, BindingFlags.Static | BindingFlags.Public);
+        Assertion.DebugIsNotNull (fieldInfo, "fieldInfo must not be null");
```
MichaelKetting, line 74:
```suggestion
        Assertion.DebugIsNotNull (fieldInfo, "Field '{0}' was not found on type '{0}'.", name, value.GetType());
```

MichaelKetting (changes requested):

Remotion/Security/Core/Metadata/EnumerationReflector.cs
raffaelfoidl: @MichaelKetting `Assertion.DebugIsNotNull` does not have an overload taking a format string with arguments.

In 958ed4fb2f5c1c9de577f64b18a338e7b29d5520, I used `Assertion.DebugIsNotNull (fieldInfo, string.Format ("Field '{0}' was not found on type '{1}'.", name, value.GetType()))`, because string interpolation does not seem to be used in this project. Also ok like this?

MichaelKetting: okay, so we have to add an overload, please. it's not a good thing to do either type of string-manipulation here because this could would be executed for relase-builds, too (I think) and thus result in cpu overhead.

raffaelfoidl: Makes sense. Separate ticket or within this task?

MichaelKetting: separate ticket / PR please

raffaelfoidl: -> [RM-7883](https://re-motion.atlassian.net/browse/RM-7883])

MichaelKetting: This one waits for the rebase on the other task so we can drop the string.Format.

raffaelfoidl: Having merged RM-7883 and rebased, I have implemented your suggestion from above.

MichaelKetting (approved):

PR #140, opened 2021-06-24
RM-7870: The return values of Path.GetDirectoryName must be checked for null in Remotion.Security.Metadata.LocalizationFileNameStrategy

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7870

### DIFF
--- a/Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
+++ b/Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
@@ -30,8 +30,11 @@ namespace Remotion.Security.Metadata
       ArgumentUtility.CheckNotNull ("culture", culture);
 
       string baseFilename = Path.GetFileNameWithoutExtension (metadataFilename);
-      // TODO RM-7870: Add notnull assertion
-      string basePath = Path.GetDirectoryName (metadataFilename)!;
+
+      string? basePath = Path.GetDirectoryName (metadataFilename);
+      if (basePath == null)
+        throw new ArgumentException ("The metadata filename must denote a path that is not a root directory.", "metadataFilename");
+
       string baseFilePath = Path.Combine (basePath, baseFilename);
 
       if (string.IsNullOrEmpty (culture.Name))
@@ -45,8 +48,11 @@ namespace Remotion.Security.Metadata
       ArgumentUtility.CheckNotNullOrEmpty ("metadataFilename", metadataFilename);
 
       string baseFileName = Path.GetFileNameWithoutExtension (metadataFilename);
-      // TODO RM-7870: Add notnull assertion
-      string basePath = Path.GetDirectoryName (metadataFilename)!;
+
+      string? basePath = Path.GetDirectoryName (metadataFilename);
+      if (basePath == null)
+        throw new ArgumentException ("The metadata filename must denote a path that is not a root directory.", "metadataFilename");
+
       string searchPattern = baseFileName + ".Localization.*xml";
 
       if (basePath == string.Empty)


Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
```diff
@@ -27,8 +28,9 @@ public class LocalizationFileNameStrategy
     public string GetLocalizationFileName (string metadataFilename, CultureInfo culture)
     {
       string baseFilename = Path.GetFileNameWithoutExtension (metadataFilename);
-      // TODO RM-7870: Add notnull assertion
-      string basePath = Path.GetDirectoryName (metadataFilename)!;
+      string? basePath = Path.GetDirectoryName (metadataFilename);
+      Assertion.IsNotNull (basePath);
```
iam-mholle, line 32: Please add a message, maybe something of the sort `nameof(metadataFilename) + " must contain a directory name."`? @MichaelKetting

iam-mholle (changes requested):

Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
raffaelfoidl: See 79d79dad6011aaf8d60c7e29251a69c5e3a363b3. Went with "non-root directory" due to respective docs.

MichaelKetting: Thanks to the acutal code and discussion and need to inject the parameter value, I realized that a) the method is public, b) this is a parameter, and c) the paramter's value is actually invalid if this error occurs. So, we should make this into a straight forward argument exception.

Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
```diff
@@ -27,8 +28,9 @@ public class LocalizationFileNameStrategy
     public string GetLocalizationFileName (string metadataFilename, CultureInfo culture)
     {
       string baseFilename = Path.GetFileNameWithoutExtension (metadataFilename);
-      // TODO RM-7870: Add notnull assertion
-      string basePath = Path.GetDirectoryName (metadataFilename)!;
+      string? basePath = Path.GetDirectoryName (metadataFilename);
+      Assertion.IsNotNull (basePath, nameof (metadataFilename) + " must denote a path that is not a root directory.");
```
MichaelKetting, line 32: See discussion.

MichaelKetting (changes requested):

Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
```diff
@@ -30,8 +30,13 @@ public string GetLocalizationFileName (string metadataFilename, CultureInfo cult
...
 
       string baseFilename = Path.GetFileNameWithoutExtension (metadataFilename);
-      // TODO RM-7870: Add notnull assertion
-      string basePath = Path.GetDirectoryName (metadataFilename)!;
+
+      string? basePath = Path.GetDirectoryName (metadataFilename);
+      if (basePath == null)
+      {
+        throw new ArgumentException ("The metadata filename must denote a path that is not a root directory.", "metadataFilename");
+      }
```
MichaelKetting, line 38:
```suggestion
        throw new ArgumentException ("The metadata filename must denote a path that is not a root directory.", "metadataFilename");
```

For single-line IFs, we can (and usually should) sckip the branches to improve readdability/compactness of the code.

Remotion/Security/Core/Metadata/LocalizationFileNameStrategy.cs
```diff
@@ -45,8 +50,13 @@ public string[] GetLocalizationFileNames (string metadataFilename)
...
 
       string baseFileName = Path.GetFileNameWithoutExtension (metadataFilename);
-      // TODO RM-7870: Add notnull assertion
-      string basePath = Path.GetDirectoryName (metadataFilename)!;
+
+      string? basePath = Path.GetDirectoryName (metadataFilename);
+      if (basePath == null)
+      {
+        throw new ArgumentException ("The metadata filename must denote a path that is not a root directory.", "metadataFilename");
+      }
```
MichaelKetting, line 58:
```suggestion
        throw new ArgumentException ("The metadata filename must denote a path that is not a root directory.", "metadataFilename");
```

MichaelKetting (changes requested):

MichaelKetting (approved):

iam-mholle (approved):

PR #142, opened 2021-06-24
RM-7854 The returned value of validator.ToString() in DefaultValidatorFormatter.Format should be checked for null

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7854

### DIFF
--- a/Remotion/Validation/Core/Implementation/DefaultValidatorFormatter.cs
+++ b/Remotion/Validation/Core/Implementation/DefaultValidatorFormatter.cs
@@ -31,8 +31,9 @@ namespace Remotion.Validation.Implementation
     {
       ArgumentUtility.CheckNotNull ("validator", validator);
 
-      // TODO RM-7854: Assert that the returned value is not null.
-      return validator.ToString()!;
+      string? formattedValidator = validator.ToString();
+      Assertion.DebugIsNotNull (formattedValidator, "formattedValidator must not be null");
+      return formattedValidator;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


Remotion/Validation/Core/Implementation/DefaultValidatorFormatter.cs
```diff
@@ -31,8 +31,9 @@ public string Format (IPropertyValidator validator, Func<Type, string> typeNameF
     {
       ArgumentUtility.CheckNotNull ("validator", validator);
 
-      // TODO RM-7854: Assert that the returned value is not null.
-      return validator.ToString()!;
+      string? formattedValidator = validator.ToString();
```
iam-mholle, line 34:
```suggestion
      var formattedValidator = validator.ToString();
```

Remotion/Validation/Core/Implementation/DefaultValidatorFormatter.cs
```diff
@@ -31,8 +31,9 @@ public string Format (IPropertyValidator validator, Func<Type, string> typeNameF
...
+      Assertion.DebugIsNotNull (formattedValidator, "formattedValidator must not be null");
```
iam-mholle, line 35:
```suggestion
      Assertion.DebugIsNotNull (formattedValidator, "formattedValidator must not be null.");
```

iam-mholle (approved):

Remotion/Validation/Core/Implementation/DefaultValidatorFormatter.cs
MichaelKetting, line 35:
```suggestion
      Assertion.IsNotNull (formattedValidator, "ToString() of validator type '{0}' returned null." validator.GetType());
```

Reasoning: validator is user-extensible, so you can't guarantee against somebody shooting at feet.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #160, opened 2021-07-29
RM-7905 Replace images with svg sprite

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7905


Remotion/ObjectBinding/Web.Test/Experimental/ControlLayout-v1.htm
```diff
@@ -92,7 +93,8 @@
       box-sizing: border-box;
       -moz-box-sizing: border-box;
       -webkit-box-sizing: border-box;
-      background-image: url('../res/Remotion.ObjectBinding.Web/Themes/NovaBlue/Image/AutoCompleteReferenceValueDropDownButtonFace.gif');
+      background-image: url('../res/Remotion.ObjectBinding.Web/Themes/NovaBlue/Image/sprite.svg#AutoCompleteReferenceValueDropDownButtonFace');
```
MichaelKetting, line 96: I'm sorry, I just realized I missed something in the review of the SVG conversion:
* AutoCompleteReferenceValueDropDownButtonFace
* DropDownMenuButton
are already sprites, so they need to be broken up into individual images before they can be integrated into a sprite-svg.

And 
* DropDownMenuRequestInProgressIndicator
* AutoCompleteReferenceValueRequestInProgressIndicator
are animated GIFs. I don't think they still work

Please fix either as a bugfix-task before this PR or at the start of this PR.

If the animated GIFs don't work in SVG, we probably have to keep them as GIF for now, and get back to this later -> follow-Up ticket.

MichaelKetting: okay, sorry, update:
It looks like the svg-background actually works with the merged sprites and the dependencies on them are only inside the CSS, we don't need to change the files for the old theme. I mean, we could, as a separate follow-up Task, but we don't have to.

We still need to fix the animated GIFs, though. It does seem that they work in Chrome/Edge but not in Firefox.

SecurityManager/Clients.Web/res/Themes/NovaGray/Image/sprite.svg
```diff
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg">
+<svg height="16" width="16" xmlns="http://www.w3.org/2000/svg">
```
MichaelKetting, line 1: I think we should use a CSS file for this. We don't know if there's ever going to be some other change to the image but keeping things consistent makes things easier to maintain.

MichaelKetting (changes requested):
Animated Gifs
Doc-Comments
SecurityManager SVG

Remotion/ObjectBinding/Web.Test/Experimental/ControlLayout-v1.htm
MichaelKetting: Bugreport: I meant create a re-motion ticket for this with a caused-by for the current task.

MichaelKetting (changes requested):

SecurityManager/Clients.Web/res/Themes/NovaGray/Html/Style.css
```diff
@@ -259,4 +265,10 @@ td.buttonCell span.DropDownMenuSelect
 td.buttonCell span.DropDownMenuSelect span.DropDownMenuButton
 {
   left: 0;
+}
+
+table.bocListTable tbody.bocListTableBody td.bocListDataCell a img.Icon
```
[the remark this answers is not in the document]
iam-mholle: It would rather be `img.Icon[src$=#MyImage]`, but yes, it would work.

The above CSS is for the following element:
image

Just tell me how I should simplify it, or if I should change it to the version you mentioned, that would be alright.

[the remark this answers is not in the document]
MichaelKetting: Ah okay. So, there's two different places that use Icon.
one in the BocListDataCell, the other in the conditionCell.
image

The Icon in the BocListDataCell will likely have a predefined default size evntually but for now, this works. and yes, simplify please. we don't need the whole path for this.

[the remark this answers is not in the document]
iam-mholle: Changed it in [`f6f9bf3` (#160)](https://github.com/re-motion/Framework/pull/160/commits/f6f9bf379d7c922c610d27d949e0021fd8eac5cd), please also review the change removing `.buttonCell`, to make the rule apply to `.conditionCell` too.

Remotion/ObjectBinding/Web.Test/Experimental/ControlLayout-v1.htm
iam-mholle: Added https://re-motion.atlassian.net/browse/RM-7972

MichaelKetting (approved):
Looks good. Can we do a quick live demo tomorrow before you merge? I'd like to take a quick look through Objectbinding.Web.Test  and SecurityManager.Clients.Web.Test.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -276,6 +290,11 @@ th.bocListTitleCell > a[href]:active
   border: solid 1px #569CE0;
   background-color: #C5E0FC;
 }
+th.bocListTitleCell .bocListSortingOrder img
+{
+  height: 8px;
+  width: 13px;
+}
```
MichaelKetting, line 297: Blank line
```suggestion

th.bocListTitleCell .bocListSortingOrder img
{
  height: 8px;
  width: 13px;
}
```

Fascinating, that the original did not work.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
iam-mholle: When editing a row, the sorting icon is no longer an anchor.

MichaelKetting (approved):

PR #165, opened 2021-08-06
RM-7971 Fix race condition in AsyncLocalStorageProviderTest

a-ctor:
https://re-motion.atlassian.net/browse/RM-7971

### DIFF
--- a/Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
+++ b/Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
@@ -15,6 +15,7 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 //
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using NUnit.Framework;
 using Remotion.Context;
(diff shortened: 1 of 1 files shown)


Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -102,42 +103,37 @@ public void SetData_FromOtherThread_DoesNotOverride ()
...
+              else
+              {
+                result = _provider.GetData (c_testKey);
+                completed = true;
+              }
+
+              Thread.MemoryBarrier();
               resetEvent.Set();
             });
         resetEvent.WaitOne();
```
MichaelKetting, line 133: Say, should we add a timeout to the WaitOne and fail the test if it's reached so the code can't hang the build if there's a thread-starvation bug or just environment condition? And if if so, new ticket (please create and schedule) or right here?

MichaelKetting (approved):
Looks good, please check comment.

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
a-ctor: I don't think that such a case will ever come up but I've added a timeout of 3sec onto the wait.

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -102,42 +103,39 @@ public void SetData_FromOtherThread_DoesNotOverride ()
...
               resetEvent.Set();
             });
-        resetEvent.WaitOne();
-      }
-      finally
-      {
-        ThreadPool.SetMaxThreads (maxWorkerThreads, maxCompletionPortThreads);
-        ThreadPool.SetMinThreads (minWorkerThreads, minCompletionPortThreads);
+
+        if (!resetEvent.WaitOne (3 * 1000))
```
MichaelKetting, line 134:
```suggestion
        if (!resetEvent.WaitOne (TimeSpan.FromSeconds (3)))
```

Yes, I know, highly unlikely. But a hanging build on the build server trumps all the 'unlikelies' we can see and easily prevent. The problem is, first you need to notice, then you need to be able to attach a debugger before someone kills it, and then you have a chance to figure it out. Unfortunately, the in-general scenario of the hanging build without obivous cause isn't just theory, so going the extra inch is usually worth the effort. Especially if the lead reviewer is usually the one to find the hanging build early in the morning and then gets to have some fun.

MichaelKetting (approved):
Please update the syntax and we're done.

PR #168, opened 2021-08-09
RM-7906: Replace old GIFs with SVGs

iam-mholle:


iam-mholle:
@MichaelKetting please also check the attached missing icons in https://re-motion.atlassian.net/browse/RM-7969.

MichaelKetting:
@iam-mholle 
Alert Icon and Progress Indicator seem to be missing in RM-7969?

MichaelKetting (changes requested):
I've added suggested versions for 9 images to the missing-icons ticket. please include them and in return, add the BocList MoveXxx-icons also to the missing icons list. We should get an official word on those.

MichaelKetting (changes requested):
The drop-down array of the BocAutoComplete is broken. 
The pagination arrows in BocList are broken
The sort arrows in BocList are broken

I think we should update the css to make sure the correct image sizes are used.

iam-mholle:
> The drop-down array of the BocAutoComplete is broken.
> The pagination arrows in BocList are broken
> The sort arrows in BocList are broken
> 
> I think we should update the css to make sure the correct image sizes are used.

@MichaelKetting  This way the layout will be broken anyway as the sizes are different, and all css files will be touched twice. Just checking if I should proceed still.

MichaelKetting:
@iam-mholle Naive thought: the changes to make this work should be minimal. And yes, we would need to touch them twice, but probably not at the same place. I think it would be okay if things don't line up perfectly but are displaying the correct and mostly complete image.

MichaelKetting (approved):

MichaelKetting:
> 
>      @MichaelKetting This way the layout will be broken anyway as the sizes are different, and all css files will be touched twice. Just checking if I should proceed still.

> @iam-mholle Naive thought: the changes to make this work should be minimal. And yes, we would need to touch them twice, but probably not at the same place. I think it would be okay if things don't line up perfectly but are displaying the correct and mostly complete image.

@iam-mholle 
Don't get the confused-smily-question.

PR #169, opened 2021-08-09
RM-7911: Create style for new theme for BocTextValue, BocMultilineTextValue

FlorianLautner:


FlorianLautner:
# MultilineText
## Chrome_MultilineText_new
Chrome_MultilineText_new
## Chrome_MultilineText_old
Chrome_MultilineText_old
## Edge_MultilineText_new
Edge_MultilineText_new
## Edge_MultilineText_old
Edge_MultilineText_old
## Firefox_MultilineText_new
Firefox_MultilineText_new
## Firefox_MultilineText_old
Firefox_MultilineText_old

FlorianLautner:
# Text
## Chrome_Text_new
Chrome_Text_new
## Chrome_Text_old
Chrome_Text_old
## Edge_Text_new
Edge_Text_new
## Edge_Text_old
Edge_Text_old
## Firefox_Text_new
Firefox_Text_new
## Firefox_Text_old
Firefox_Text_old

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -1,15 +1,27 @@
...
+
 span.bocMultilineTextValue
 {
   width: 15em;
-  min-height: 4.5em;
+  min-height: var(--min-height);
   position: relative;
   display: inline-block;
   vertical-align: middle;
   overflow: hidden;
```
MichaelKetting, line 16: okay, I've just doodled a bit. 

We should remove the overflow (test things still work), make an outline that's 0.2em (perhaps 0.3em, is tbd) and uses the border blue. For readonly, we also use the outline and background color.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -1,15 +1,27 @@
 /* Style applied to the BocMultilineTextValue. */
 
+:root
+{
+  --min-height: 4.5em;
```
MichaelKetting, line 5: Each control should be isolated. Refining root-properties such as this would bleed into other css files, correct?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
MichaelKetting, line 16: When optimzing and there is time, we should check if overflow is still needed. We need to check behavior with long readonly-texts when dropping overflow. Or, possibly, keep it for the readonly-view.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -29,37 +41,38 @@ span.bocMultilineTextValue .content
   right: 6px;
```
MichaelKetting, line 41: why is this still needed?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -29,37 +41,38 @@ span.bocMultilineTextValue .content
...
 
-.mozilla span.bocMultilineTextValue .content
+span.bocMultilineTextValue textarea
 {
-  right: 8px;
-  bottom: calc(0.5em + 3px);
+  width: 100%;
+  height: 100%;
+  display: inline-block;
+  min-height: var(--min-height);
```
MichaelKetting, line 52: I'll be needing an explanation/defails for this.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -29,37 +41,38 @@ span.bocMultilineTextValue .content
...
 }
 
-.bocMultilineTextValue.readOnly .content
+.mozilla span.bocMultilineTextValue .content
 {
-  position: static;
+  /* Ensures that the outline of the textarea is displayed correctly. */
+  bottom: var(--mozilla-offset);
```
MichaelKetting, line 58: Removing overflow would take care of this and no browser-specifics would be preferred.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -29,37 +41,38 @@ span.bocMultilineTextValue .content
...
 }
 
-.bocMultilineTextValue.readOnly .content span[tabindex]
+.mozilla span.bocMultilineTextValue .content textarea
 {
-  /* Ensures that the outline is rendered as a block and outside of the space reserved for diacritic marks. */
-  display: inline-block;
+  /* Ensures that the outline of the textarea is displayed correctly. */
+  min-height: calc(var(--min-height) - var(--mozilla-offset));
```
MichaelKetting, line 64: see above

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -29,37 +41,38 @@ span.bocMultilineTextValue .content
...
 }
 
-.bocMultilineTextValue.readOnly .content span[tabindex]:first-child
+.bocMultilineTextValue.readOnly .content
```
MichaelKetting, line 67: the css should be ordered based on their nesting depth, so, please move up.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -27,11 +34,7 @@ span.bocTextValue .content
   left: 0;
   right: 6px;
```
MichaelKetting, line 35: see BocMultilineTextValue. please also review other remarks and apply here as needed.

MichaelKetting (changes requested):
I've got a couple of suggestions/thoughts. please check them out and it's possible we'll need to do this live.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -45,48 +45,28 @@ span.bocTextValue .content
...
 {
-  /* Ensures that empty readonly values are not collapsed and can therefor visualize the keyboard focus */
-  width: calc(100% - 2px);
+  /* Ensures that empty readonly values are not collapsed and can therefore visualize the keyboard focus */
+  width: 100%;
   min-height: 1.5em;
-  padding-top: 0;
-  padding-bottom: 0;
-  vertical-align: -0.2em;
+  vertical-align: middle;
```
MichaelKetting, line 53: veritcal-align is tricky. I'm not sure if it's a) correct and b) how it works for multiline-text in readonly.
also, what happens when we remove it?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -45,48 +45,28 @@ span.bocTextValue .content
...
+  height: 1.2em;
+  line-height: 1em;
 }
 
 span.bocTextValue textarea
 {
   width: 100%;
-  height: 100%;
+  /* Forcing assignment of height from parent and overriding height set in code-behind. */
+  height: 100% !important;
```
MichaelKetting, line 68: Please link the place in the code-behind that's responsible for this. if the code-behind determines the height based on the number of rows, then the code-behind is correct. If it's just some fixed value, we should move it to css in both themes.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -45,48 +45,28 @@ span.bocTextValue .content
...
-  padding-top: 0;
-  padding-bottom: 0;
-  vertical-align: -0.2em;
+  vertical-align: middle;
 }
 
 span.bocTextValue input
 {
   width: 100%;
+  min-height: inherit;
```
MichaelKetting, line 59: why inherit?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -7,14 +7,19 @@ span.bocTextValue
...
 }
 
+span.bocTextValue .content,
+span.bocTextValue input,
+span.bocTextValue textarea
+{
+  box-sizing: border-box;
+}
+
+
```
MichaelKetting, line 19: extra blank line

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -7,36 +7,31 @@ span.bocMultilineTextValue
...
 span.bocMultilineTextValue .content
 {
   position: absolute;
   display: block;
-  left: 0;
-  right: 6px;
-  width: auto;
-  top: 0;
-  bottom: 0.5em;
+  width: inherit;
```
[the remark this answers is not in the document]
FlorianLautner: width:inherit is not necessary if replaced with width:100% (is this better?). height:auto was now replaced by height:100% in order to fill the space of the parent so that the textarea is neither smaller nor larger than the parent since without it, it wouldn't fit exactly (larger in Firefox, smaller in Chrome/Edge), display: block alone is not enough.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -7,36 +7,31 @@ span.bocMultilineTextValue
...
   height: auto;
-}
-
-.mozilla span.bocMultilineTextValue .content
-{
-  right: 8px;
-  bottom: calc(0.5em + 3px);
+  top: 0;
```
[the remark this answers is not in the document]
FlorianLautner: top and bottom have been removed and replaced by height:100% since it is needed anyway (see above).

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -24,14 +29,9 @@ span.bocTextValue
...
-  right: 6px;
-  width: auto;
-}
-
-.mozilla span.bocTextValue .content
-{
-  right: 8px;
+  width: inherit;
+  min-height: inherit;
+  height: inherit;
```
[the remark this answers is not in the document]
FlorianLautner: min-height:inherit is needed; for height and width, 100% does the job as well, so inherit was removed.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
FlorianLautner: With a bit of a workaround I managed to get rid of the vertical-align. By introducing another pseudo-class ':empty:before' and insertig a 'zero width space' character (scales with font size) there in order to ensure that the readonly value is not collapsed entirely as well as setting the width to 100% in the actual ':empty' pseudo-class, the readonly value does not collapse, is centered, and the tab-focus surrounds the entire readonly span. I'm not sure if this solution really is a good idea, though.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
FlorianLautner: If I am not mistaken, then it should be this line (and the one above): [https://github.com/re-motion/Framework/blob/develop/Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocTextValueUserControl.ascx#L116](url). Unless the code is auto-generated, the values for height (6em and 5em) are hard-coded and could be moved to css, I assume.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
FlorianLautner: Not necessary, 100% works too. One of the two is needed, though, to fill the parent.

MichaelKetting: Then `100%` it is. `inhert` doesn't mean 'copy the px count but instead, it's a reapply-the-parent's-style-here-because-my-current-style-was-overridden-and-I-want-the-parent-back.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
MichaelKetting: This 5em are user-defined and should not be overruled.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
MichaelKetting: okay, I'll admit, didn't check it out. Kudos for the creativity, but we should stick with the v-align. v-align isn't bad if it's necessary.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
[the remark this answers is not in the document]
MichaelKetting: sounds good.

MichaelKetting (commented):
see comments

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
FlorianLautner: I kept the `:empty:before` version for now since it both scales with font size changes and it let me remove more quirky bits of `.bocTextValue.readOnly .content span[tabindex]:empty` as well as having it properly sized and aligned without properties like `min-height: 1.5em; padding-top: 0; padding-bottom: 0;` (while some of them would still be needed for v-align). It's not necessary anymore to set these so that the size and complexity of the CSS is further reduced.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -7,31 +7,31 @@ span.bocTextValue
...
 
-.bocTextValue.disabled
+span.bocTextValue .content,
+span.bocTextValue input,
+span.bocTextValue textarea
 {
+  box-sizing: border-box;
+  width: 100%;
+  height: 100%;
 }
```
MichaelKetting, line 30: I just re-checked with the styleguide / screenshots and looked at the live styles: the designers would like to have a 0.3-0.5em padding between the textbox border and the text.
image
I think we should be golden by adding a padding-left and padding-right of 0.4em and setting the height of the input's parent the to min-height + 0.8em. Possibly, it would be a good idea to use calc for the height, that way, we can take the intended height and add 2*padding to it and thus make it clear where the dimensions come from. For textarea, we have to use padding all around. 
The padding/height is to be globally defined, probably also part of the global input[text,passowrd] element style.

Please make sure the text is properly centered vertically in the textbox.

The same styling applies to readonly elements, probably will need their own considerations about layout.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -41,52 +41,22 @@ span.bocTextValue .content
...
-  /* Ensures that diacritic marks can be rendered without being cut off. */
-  padding-top: 0.2em;
-}
-
-.bocTextValue.readOnly .content span[tabindex]:last-child
-{
-  /* Ensures that diacritic marks can be rendered without being cut off. */
-  padding-bottom: 0.2em;
+  /* Ensures that the outline is rendered outside of the space reserved for diacritic marks. */
+  padding: 0.2em 0 0.2em 0;
```
MichaelKetting, line 46: I think this one goes bye-bye once we do the 0.4em padding

MichaelKetting (changes requested):
This certainly looks like it's the beginning of a beutiful stylesheet.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
FlorianLautner: Since padding is now global for input, textarea, and readonly elements, the value has been set to 0.3em instead of 0.4 to avoid scroll bars for BocMultilineTextValue-textareas since the padding is applied there as well (~2px too large if 0.4em). In general, it still doesn't look _exactly_ like the designers' concept, however it is way closer now than before.

MichaelKetting: I'm not married to 0.4 vs 0.3, but I don't get the scrollbar bit. If the textarea is not high enough, we add height.

Could you do a screenshot-compare of the designer-provided template and our control, please? If we use the Roboto-font and make sure the textsize in EM alines to the one in PX, we should be able to determine if 0.3em or 0.4em better matches the distances. It doesn't need to be pixel-perfect but the ratio should be close and I've never seen the two side by side..

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -12,18 +12,18 @@
...
-
   --remotion-themed-border-radius: 2px;
   --remotion-themed-border-width: 1px;
   --remotion-themed-border: var(--remotion-themed-border-width) solid var(--color-standard-border);
   --remotion-themed-border-focus: var(--remotion-themed-border-width) solid var(--color-primary);
   --remotion-themed-dropdown-scrollbar-border-radius: 10px;
   --remotion-themed-dropdown-scrollbar-width: 10px;
   --remotion-themed-scrollbar-width: 14px;
+  --remotion-themed-input-height: 1.7em;
+  --remotion-themed-padding: 0.3em;
```
MichaelKetting, line 26: Please update. I want the input-height to use calc with 2*padding and the lineheight. I think we can make the lineheight a stand-aline value and then use 2*padding+lineheight in the usages further down in Common-CSS. At least, that's how I figure things would work best. If not, please explain.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
```diff
@@ -7,36 +7,31 @@ span.bocMultilineTextValue
   position: relative;
   display: inline-block;
   vertical-align: middle;
-  overflow: hidden;
```
MichaelKetting, line 10: can we do the min-height as 3*lineheight + 2*padding?

I'm not sure what x should be. it doesn't have to match the current height, instead it should be 3 lines + padding. Normally, I used 1.2em lineheight.

MichaelKetting (changes requested):
please update style.css to remove overrides for padding
please check if you can easily adjust the vertical alignment of the input so the baseline remains aligned in the form (aligned with the text in the label, etc) but the borders of the input nove up a bit so the text on the baseline is visually centered.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
FlorianLautner: Since the size of the textarea can simply be adjusted if necessary, I now went with the `0.4em` as initially suggested.

Comparison/screenshots with temporarily applied `Roboto` font:

# Facelift
## Firefox
### Currently
new_firefox

### With Roboto
new_firefox_roboto

### With Roboto and temporarily adjusted FromGrid cell padding to imitate intended design
new_firefox_roboto_cell_changed


## Chrome
### Currently
new_chrome

### With Roboto
new_chrome_roboto

### With Roboto and temporarily adjusted FromGrid cell padding to imitate intended design
new_chrome_roboto_cell_changed


# Zeplin
## Example 1
zeplin1

## Example 2
zeplin2

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
FlorianLautner: I introduced a new `--remotion-themed-line-height` variable with `1.2em` and used it to calculated `--remotion-themed-min-height` with `--remotion-themed-line-height` _+ 2 *_ `--remotion-themed-padding` (of `0.4em` as mentioned above).

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocMultilineTextValue.css
FlorianLautner: I used _3 * lineheight + 3 * padding_ since the result looked better than only _2 * padding_. Other than that, I changed the lineheight to `1.2em`, as suggested.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -114,4 +142,11 @@ select.remotion-themed::-webkit-scrollbar-thumb,
 {
   background-color: var(--color-primary);
   border-radius: var(--remotion-themed-dropdown-scrollbar-border-radius);
+}
+
+.remotion-themed.readOnly .content span
```
MichaelKetting, line 147: This one is too specific. `.content span` is a very specific implementation thingy

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -3,90 +3,65 @@
...
-  min-height: 0;
-  /* Ensures that the outline is visible in Firefox. */
-  overflow: visible;
+  resize: none;
+  /* Otherwise the textarea is slighty offset in Firefox since a margin is added there. */
+  margin: 0;
 }
 
-.bocTextValue.disabled
+.bocTextValue.readOnly span
```
MichaelKetting, line 18: .bocTextValue.readOnly .content 
or
.bocTextValue.readOnly .content span

MichaelKetting (changes requested):

MichaelKetting (approved):
when setting up the commits, could you please make one commit for common.css and one for the textvalue files?

PR #170, opened 2021-08-10
RM-7976 Add support for WebString

a-ctor:
https://re-motion.atlassian.net/browse/RM-7976

### DIFF
--- a/Remotion/Web/Core/Utilities/HtmlUtility.cs
+++ b/Remotion/Web/Core/Utilities/HtmlUtility.cs
@@ -16,38 +16,38 @@
 // 
 using System;
 using System.Text.RegularExpressions;
-using System.Web;
 using System.Web.UI;
 using JetBrains.Annotations;
+using Remotion.Obsolete;
 using Remotion.Utilities;
 
 namespace Remotion.Web.Utilities
 {
   public class HtmlUtility
   {
+    [Obsolete ("HtmlUtility.Format(string, params string[]) is obsolete. (Version 3.0.0)")]
     public static string Format (string htmlFormatString, params object[] nonHtmlParameters)
     {
       string[] htmlParameters = new string[nonHtmlParameters.Length];
       for (int i = 0; i < nonHtmlParameters.Length; ++i)
-        htmlParameters[i] = HtmlEncode (nonHtmlParameters[i].ToString());
+        htmlParameters[i] = WebString.CreateFromText (nonHtmlParameters[i].ToString()).ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
       return string.Format (htmlFormatString, (object[]) htmlParameters);
     }
 
+#if !NETFRAMEWORK
+    [Obsolete (
+        "HtmlUtility.HtmlEncode(string) is obsolete. Use WebString.CreateFromText(string).ToString(WebStringEncoding.HtmlWithTransformedLineBreaks) instead. (Version 3.0.0)",
+        DiagnosticId = ObsoleteDiagnosticIDs.HtmlUtility)]
+#endif
     public static string HtmlEncode (string nonHtmlString)
     {
-      string html = HttpUtility.HtmlEncode (nonHtmlString);
-      if (html != null)
-      {
-        html = html.Replace ("\r\n", "<br />");
-        html = html.Replace ("\n", "<br />");
-        html = html.Replace ("\r", "<br />");
-      }
-      return html;
+      return WebString.CreateFromText (nonHtmlString).ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
     }
 
+    [Obsolete ("HtmlUtility.HtmlEncode(string, HtmlTextWriter) is obsolete. Use WebString.CreateFromText(string).Write(HtmlTextWriter) instead. (Version 3.0.0)")]
     public static void HtmlEncode (string nonHtmlString, HtmlTextWriter writer)
     {
-      writer.Write (HtmlEncode (nonHtmlString));
+      WebString.CreateFromText (nonHtmlString).Write (writer);
     }
 
     private static readonly Regex s_stripHtmlTagsRegex = new Regex ("<.*?>", RegexOptions.Compiled);
--- a/Remotion/Web/Core/WebString.cs
+++ b/Remotion/Web/Core/WebString.cs
@@ -1,0 +1,239 @@
+﻿using System;
+using System.IO;
+using System.Web;
+using System.Web.UI;
+using JetBrains.Annotations;
+using Remotion.Utilities;
+
+namespace Remotion.Web
+{
+  /// <summary>
+  /// The <see cref="WebString"/> structure represents encoded or unencoded HTML strings that are rendered onto a web page.
+  /// Its <see cref="Type"/> determines if and how a <see cref="WebString"/> is encoded before being rendered.
+  /// </summary>
+  public readonly struct WebString : IEquatable<WebString>
+  {
+    private static readonly string[] s_newlineSplitSeparators = { "\r\n", "\r", "\n" };
+
+    /// <summary>
+    /// Creates a <see cref="WebString"/> from the specified <paramref name="html"/>.
+    /// The value will be rendered without being encoded and thus must be a safe string.
+    /// </summary>
+    /// <param name="html">The HTML from which a <see cref="WebString"/> is created.</param>
+    /// <returns>The newly created <see cref="WebString"/> containing the <paramref name="html"/>.</returns>
+    public static WebString CreateFromHtml ([CanBeNull] string html)
+    {
+      return new WebString (html, WebStringType.Encoded);
+    }
+
+    /// <summary>
+    /// Creates a <see cref="WebString"/> from the specified <paramref name="text"/>.
+    /// The value will be encoded before being rendered.
+    /// </summary>
+    /// <param name="text">The text from which a <see cref="WebString"/> is created.</param>
+    /// <returns>The newly created <see cref="WebString"/> containing the <paramref name="text"/>.</returns>
+    public static WebString CreateFromText ([CanBeNull] string text)
+    {
+      return new WebString (text, WebStringType.PlainText);
+    }
+
+    /// <summary>
+    /// Implements the equality operator for <see cref="WebString"/>s. The operator is implemented the same way as the <see cref="Equals(WebString)"/> method.
+    /// </summary>
+    /// <param name="left">The first value to be compared for equality.</param>
+    /// <param name="right">The second value to be compared for equality.</param>
+    /// <returns>
+    ///   <see langword="true" /> if both <paramref name="left"/> and <paramref name="right"/> have the same string value
+    ///   and the same type. Otherwise, <see langword="false" />.
+    /// </returns>
+    public static bool operator == (WebString left, WebString right)
+    {
+      return left.Equals (right);
+    }
+
+    /// <summary>
+    /// Implements the inequality operator for <see cref="WebString"/>s. The operator is implemented the same way as the <see cref="Equals(WebString)"/> method.
+    /// </summary>
+    /// <param name="left">The first value to be compared for inequality.</param>
+    /// <param name="right">The second value to be compared for inequality.</param>
+    /// <returns>
+    ///   <see langword="true" /> if both <paramref name="left"/> and <paramref name="right"/> are <see langword="null" /> do not have the same string value
+    ///   or type. Otherwise, <see langword="false" />.
+    /// </returns>
+    public static bool operator != (WebString left, WebString right)
+    {
+      return !left.Equals (right);
+    }
+
+    [CanBeNull]
+    private readonly string _value;
+    private readonly WebStringType _type;
+
+    private WebString ([CanBeNull] string value, WebStringType type)
+    {
+      _value = value;
+      _type = type;
+    }
+
+    /// <summary>
+    /// Gets the type of string represented by this <see cref="WebString"/>.
+    /// </summary>
+    /// <value>The type of string represented by this <see cref="WebString"/>.</value>
+    public WebStringType Type => _type;
+
+    /// <summary>
+    /// Tests whether the <see cref="WebString"/>s value is <see langword="null" /> or an empty string.
+    /// </summary>
+    public bool IsEmpty => string.IsNullOrEmpty (_value);
+
+    /// <summary>
+    /// Gets the raw string value of the <see cref="WebString"/> or an empty string if there is no value.
+    /// </summary>
+    [NotNull]
+    public string GetValue () => _value ?? string.Empty;
+
+    /// <summary>
+    /// Writes the <see cref="WebString"/>'s value to the specified <paramref name="writer"/>.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the value will be appended to. Must not be <see langword="null" />.</param>
+    public void Write ([NotNull] HtmlTextWriter writer)
+    {
+      ArgumentUtility.CheckNotNull (nameof (writer), writer);
+
+      switch (_type)
+      {
+        case WebStringType.PlainText:
+          var value = GetValue();
+          var parts = value.Split (s_newlineSplitSeparators, StringSplitOptions.None);
+
+          for (var i = 0; i < parts.Length; i++)
+          {
+            HttpUtility.HtmlEncode (parts[i], writer);
+            if (i < parts.Length - 1)
+              writer.WriteBreak();
+          }
+
+          break;
+        case WebStringType.Encoded:
+          writer.Write (GetValue());
+          break;
+        default:
+          throw new NotSupportedException ($"The WebStringType '{_type}' is not supported.");
+      }
+    }
+
+    /// <summary>
+    /// Adds the specified markup <paramref name="attribute"/> to the <paramref name="writer"/> using the <see cref="WebString"/> value.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <remarks>
+    /// Renders an attribute with an empty value if the <see cref="WebString"/> has no value.
+    /// </remarks>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The attribute that is to be added.</param>
+    public void AddAttribute ([NotNull] HtmlTextWriter writer, HtmlTextWriterAttribute attribute)
+    {
+      ArgumentUtility.CheckNotNull (nameof (writer), writer);
+
+      switch (_type)
+      {
+        case WebStringType.PlainText:
+          writer.AddAttribute (attribute, GetValue(), fEncode: true);
+          break;
+        case WebStringType.Encoded:
+          writer.AddAttribute (attribute, GetValue(), fEncode: false);
+          break;
+        default:
+          throw new NotSupportedException ($"The WebStringType '{_type}' is not supported.");
+      }
+    }
+
+    /// <summary>
+    /// Adds the specified markup <paramref name="attribute"/> to the <paramref name="writer"/> using the <see cref="WebString"/> value.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <remarks>
+    /// Renders an attribute with an empty value if the <see cref="WebString"/> has no value.
+    /// </remarks>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The name of the attribute that is to be added. Must not be <see langword="null" /> or empty.</param>
+    public void AddAttribute ([NotNull] HtmlTextWriter writer, [NotNull] string attribute)
+    {
+      ArgumentUtility.CheckNotNull (nameof (writer), writer);
+      ArgumentUtility.CheckNotNullOrEmpty (nameof (attribute), attribute);
+
+      switch (_type)
+      {
+        case WebStringType.PlainText:
+          writer.AddAttribute (attribute, GetValue(), fEndode: true);
+          break;
+        case WebStringType.Encoded:
+          writer.AddAttribute (attribute, GetValue(), fEndode: false);
+          break;
+        default:
+          throw new NotSupportedException ($"The WebStringType '{_type}' is not supported.");
+      }
+    }
+
+    /// <inheritdoc />
+    public bool Equals (WebString other)
+    {
+      return GetValue() == other.GetValue() && _type == other._type;
+    }
+
+    /// <inheritdoc />
+    public override bool Equals (object obj)
+    {
+      return obj is WebString other && Equals (other);
+    }
+
+    /// <inheritdoc />
+    public override int GetHashCode ()
+    {
+      unchecked
+      {
+        return (GetValue().GetHashCode() * 397) ^ (int) _type;
+      }
+    }
+
+    /// <summary>
+    /// Returns a HTML encoded string that represents the current <see cref="WebString"/>.
+    /// </summary>
+    /// <returns>A HTML encoded string that represents the current <see cref="WebString"/>.</returns>
+    public override string ToString ()
+    {
+      return ToString (WebStringEncoding.Html);
+    }
+
+    /// <summary>
+    /// Returns a string that represents the current <see cref="WebString"/> in the specified <paramref name="encoding"/>.
+    /// </summary>
+    /// <returns>A string that represents the current <see cref="WebString"/> in the specified <paramref name="encoding"/>.</returns>
+    public string ToString (WebStringEncoding encoding)
+    {
+      var value = GetValue();
+      if (Type == WebStringType.Encoded)
+        return value;
+
+      switch (encoding)
+      {
+        case WebStringEncoding.Html:
+          return HttpUtility.HtmlEncode (value);
+        case WebStringEncoding.HtmlWithTransformedLineBreaks:
+          using (var stringWriter = new StringWriter())
+          {
+            using (var htmlTextWriter = new HtmlTextWriter (stringWriter))
+            {
+              Write (htmlTextWriter);
+              return stringWriter.ToString();
+            }
+          }
+        case WebStringEncoding.Attribute:
+          return HttpUtility.HtmlAttributeEncode (value);
+        default:
+          throw new ArgumentOutOfRangeException (nameof (encoding), encoding, null);
+      }
+    }
+  }
+}
--- a/Remotion/Web/Core/WebStringEncoding.cs
+++ b/Remotion/Web/Core/WebStringEncoding.cs
@@ -1,0 +1,23 @@
+﻿namespace Remotion.Web
+{
+  /// <summary>
+  /// Defines the different encodings that a <see cref="WebString"/> can be encoded in.
+  /// </summary>
+  public enum WebStringEncoding
+  {
+    /// <summary>
+    /// An unencoded <see cref="WebString"/> will be HTML encoded.
+    /// </summary>
+    Html,
+    
+    /// <summary>
+    /// An unencoded <see cref="WebString"/> will be HTML encoded and line break characters will be replaced by &lt;br /&gt; tags.
+    /// </summary>
+    HtmlWithTransformedLineBreaks,
+    
+    /// <summary>
+    /// An unencoded <see cref="WebString"/> will be HTML attribute encoded.
+    /// </summary>
+    Attribute
+  }
+}
--- a/Remotion/Web/Core/WebStringType.cs
+++ b/Remotion/Web/Core/WebStringType.cs
@@ -1,0 +1,18 @@
+﻿namespace Remotion.Web
+{
+  /// <summary>
+  /// Defines the kind of a <see cref="WebString"/> which determines if it is encoded when rendered.
+  /// </summary>
+  public enum WebStringType
+  {
+    /// <summary>
+    /// The <see cref="WebString"/> contains plain text, which will be encoded before being rendered.
+    /// </summary>
+    PlainText,
+    
+    /// <summary>
+    /// The <see cref="WebString"/> contains already encoded text, which will be rendered as is.
+    /// </summary>
+    Encoded
+  }
+}
--- a/Remotion/Web/UnitTests/Core/Utilities/HtmlUtilityTest.cs
+++ b/Remotion/Web/UnitTests/Core/Utilities/HtmlUtilityTest.cs
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
+#pragma warning disable REMOTION0001
 using System;
 using NUnit.Framework;
 using Remotion.Web.Utilities;
--- a/Remotion/Web/UnitTests/Core/WebStringTest.cs
+++ b/Remotion/Web/UnitTests/Core/WebStringTest.cs
@@ -1,0 +1,269 @@
+﻿using System;
+using System.IO;
+using System.Web.UI;
+using NUnit.Framework;
+
+namespace Remotion.Web.UnitTests.Core
+{
+  [TestFixture]
+  public class WebStringTest
+  {
+    [Test]
+    public void CreateFromHtml ()
+    {
+      var stringValue = "aoe   \" & ' < > é \r \n \r\n";
+      var webString = WebString.CreateFromHtml (stringValue);
+
+      Assert.That (webString.Type, Is.EqualTo (WebStringType.Encoded));
+      Assert.That (webString.GetValue(), Is.EqualTo (stringValue));
+    }
+
+    [Test]
+    public void CreateFromHtml_WithNullValue_IsSameAsEmptyString ()
+    {
+      Assert.That (WebString.CreateFromHtml (null), Is.EqualTo (WebString.CreateFromHtml (string.Empty)));
+    }
+
+    [Test]
+    public void CreateFromText ()
+    {
+      var stringValue = "aoe   \" & ' < > é \r \n \r\n";
+      var webString = WebString.CreateFromText (stringValue);
+
+      Assert.That (webString.Type, Is.EqualTo (WebStringType.PlainText));
+      Assert.That (webString.GetValue(), Is.EqualTo (stringValue));
+    }
+
+    [Test]
+    public void CreateFromText_WithNullValue_IsSameAsEmptyString ()
+    {
+      Assert.That (WebString.CreateFromText (null), Is.EqualTo (WebString.CreateFromText (string.Empty)));
+    }
+
+    [Test]
+    public void IsEmpty ()
+    {
+      Assert.That (new WebString().IsEmpty, Is.True);
+      Assert.That (WebString.CreateFromText ("").IsEmpty, Is.True);
+      Assert.That (WebString.CreateFromText (" ").IsEmpty, Is.False);
+      Assert.That (WebString.CreateFromText ("test").IsEmpty, Is.False);
+    }
+
+    [Test]
+    public void Write_NullPlainTextWebString_RendersEmptyOutput ()
+    {
+      var webString = WebString.CreateFromText (null);
+
+      var renderedString = ExecuteWithHtmlTextWriter (webString.Write);
+      Assert.That (renderedString, Is.EqualTo (string.Empty));
+    }
+
+    [Test]
+    public void Write_EmptyPlainTextWebString_RendersEmptyOutput ()
+    {
+      var webString = WebString.CreateFromText (string.Empty);
+
+      var renderedString = ExecuteWithHtmlTextWriter (webString.Write);
+      Assert.That (renderedString, Is.EqualTo (string.Empty));
+    }
+
+    [Test]
+    public void Write_PlainTextWebString_RendersEncodedOutputWithTransformedLineBreaks ()
+    {
+      var stringValue = "aoe   \" & ' < > é \r \n \r\n";
+      var webString = WebString.CreateFromText (stringValue);
+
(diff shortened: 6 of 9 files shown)


Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+      var webString = WebString.CreateFromRawAttribute (stringValue);
+
+      Assert.That (webString.Type, Is.EqualTo (WebStringType.Attribute));
+      Assert.That (webString.GetRawValue(), Is.EqualTo (stringValue));
+    }
+
+    [Test]
+    public void CreateFromRawAttribute_WithNullValue_ThrowsArgumentNullException ()
+    {
+      Assert.That (() => WebString.CreateFromRawAttribute (null), Throws.ArgumentNullException);
```
MichaelKetting, line 25: Nice attention to detail! Just for reference purposes: typically, we don't bother with testing arg-null and arg-empty cases.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+      Assert.That (webString.GetRawValue(), Is.EqualTo (stringValue));
+    }
+
+    [Test]
+    public void IsEmpty ()
+    {
+      Assert.That (new WebString().IsEmpty, Is.True);
+      Assert.That (((WebString)"").IsEmpty, Is.True);
+      Assert.That (((WebString)" ").IsEmpty, Is.False);
+      Assert.That (((WebString)"test").IsEmpty, Is.False);
```
MichaelKetting, line 76:
```suggestion
      Assert.That (((WebString) "").IsEmpty, Is.True);
      Assert.That (((WebString)"  ").IsEmpty, Is.False);
      Assert.That (((WebString) "test").IsEmpty, Is.False);
```

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+      Assert.That (renderedString, Is.EqualTo (stringValue));
+    }
+
+    [Test]
+    public void Write_Attribute_ThrowsInvalidOperationException ()
+    {
+      var stringValue = "My <b>test</a>";
+      var webString = WebString.CreateFromRawAttribute (stringValue);
+
+      Assert.That (() => ExecuteWithHtmlTextWriter (webString.Write), Throws.InvalidOperationException);
```
MichaelKetting, line 105: Please assert ex message too. Ex-messages that we write outselves and are meant to comminucate important informaton / bug-details are usually part of the tested contract.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    [Test]
+    public void AddAttribute_TextWithAttributeEnum ()
+    {
+      var stringValue = "My <b>test</a>";
```
MichaelKetting, line 111: I think we should add a double quote in each of the attribute tests.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    [Test]
+    public void AddAttribute_HtmlWithAttributeEnum_ThrowsInvalidOperationException ()
+    {
+      var stringValue = "My <b>test</a>";
+      var webString = WebString.CreateFromRawHtml (stringValue);
+
+      Func<string> testLambda = () => ExecuteWithHtmlTextWriter (writer => webString.AddAttribute (writer, HtmlTextWriterAttribute.Class));
+      Assert.That (testLambda, Throws.InvalidOperationException);
```
MichaelKetting, line 131: we never have the test func outside of the Assert,That.

Also, I would prefer to have only one operation in the test func instead of the entire helper logic.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+      var stringValue = "My <b>test</a>";
+      var webString = WebString.CreateFromString (stringValue);
+
+      var renderedString = ExecuteWithHtmlTextWriter (
+          writer =>
+          {
+            webString.AddAttribute (writer, HtmlTextWriterAttribute.Class);
+            writer.RenderBeginTag ("a");
+            writer.WriteEndTag ("a");
```
MichaelKetting, line 119: So we actually need to do the closing tag?

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+      var stringValue = "My <b>test</a>";
+      var webString = WebString.CreateFromRawHtml (stringValue);
+
+      Func<string> testLambda = () => ExecuteWithHtmlTextWriter (writer => webString.AddAttribute (writer, HtmlTextWriterAttribute.Class));
+      Assert.That (testLambda, Throws.InvalidOperationException);
+    }
+
+    [Test]
+    public void AddAttribute_AttributeWithAttributeEnum ()
```
MichaelKetting, line 135: Should test with an enum that encodes and one that doesnt.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,210 @@
...
+    public void IsEmpty ()
+    {
+      Assert.That (new WebString().IsEmpty, Is.True);
+      Assert.That (((WebString)"").IsEmpty, Is.True);
+      Assert.That (((WebString)" ").IsEmpty, Is.False);
+      Assert.That (((WebString)"test").IsEmpty, Is.False);
+    }
+
+    [Test]
+    public void Write_Text ()
```
MichaelKetting, line 80: Please add result qualifier to test methods
Write_Text_EncodesOutput / ..._DoesNotEncodeOutput

Remotion/Web/UnitTests/Core/WebStringTest.cs
MichaelKetting, line 135:
```suggestion
    public void AddAttribute_WithTypeIsAttribute_WithNotEncodingAttribute_DoesNotEncodeOutput ()
```

MichaelKetting (changes requested):
There's one oversight in the spec and a couple of details...

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,208 @@
...
+  /// </summary>
+  public readonly struct WebString : IEquatable<WebString>
+  {
+    /// <summary>
+    /// Creates a <see cref="WebString"/> from the specified <paramref name="rawAttributeValue"/>.
+    /// The value will be rendered without being encoded and thus must be a safe string.
+    /// </summary>
+    /// <param name="rawAttributeValue">The raw attribute value from which a <see cref="WebString"/> is created.</param>
+    /// <returns>The newly created <see cref="WebString"/> containing the <paramref name="rawAttributeValue"/>.</returns>
+    public static WebString CreateFromRawAttribute (string rawAttributeValue)
```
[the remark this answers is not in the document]
a-ctor: Do we need to mention that here? It seems kind of out of scope on the create methods. Wouldn't it be a better fit on the `Write` or `AddAttribute` methods?

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// <summary>
+    /// Gets the type of string represented by this <see cref="WebString"/>.
+    /// </summary>
+    /// <value>The type of string represented by this <see cref="WebString"/>.</value>
+    public WebStringType Type => _type;
+
+    /// <summary>
+    /// Tests whether the <see cref="WebString"/>s value is <see langword="null" /> or an empty string.
+    /// </summary>
+    public bool IsEmpty => string.IsNullOrEmpty (_value);
```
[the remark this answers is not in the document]
a-ctor: Since we provide the Value as non-nullable we do not have a difference between the null and empty. Since we do not support null in any way

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the value will be appended to.</param>
+    /// <exception cref="InvalidOperationException">The <see cref="WebString"/>'s type is not compatible with the operation.</exception>
+    public void Write (HtmlTextWriter writer)
+    {
+      switch (_type)
+      {
+        case WebStringType.Text:
+          writer.Write (HttpUtility.HtmlEncode (GetRawValue()));
```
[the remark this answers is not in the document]
a-ctor: I am quite lost with the second comment. Are you worried about `HtmlUtility.HtmlEncode` potentially allocating the string multiple times?

Since we are not in span-land and `TextWriter` does not support writing only a part of a string, the only problem I see is that the method might create unnecessary copies when a string contains multiple line endings.

Remotion/Web/UnitTests/Core/WebStringTest.cs
a-ctor: In another comment you said that we are missing the `fEncode: true`. What should the encode behavior with type `Text`?

Because we do not need another set of tests if we tell it to always encode.

Remotion/Web/Core/WebString.cs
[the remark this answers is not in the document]
MichaelKetting: hmm... it just felt opportune since we do discuess that the string is being rendered. so, what happens to the empty string.

Remotion/Web/Core/WebString.cs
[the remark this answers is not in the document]
MichaelKetting: and creating default() and creating a WebString with string.empty is semantically equivalent.

Remotion/Web/Core/WebString.cs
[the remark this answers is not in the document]
MichaelKetting: > the only problem I see is that the method might create unnecessary copies when a string contains multiple line endings.
Yes and now. The implementation in HtmlUtility does multplie string-search&replace, so creating substrings would be okay. there's also the (extremely questionable) alternative of rendering each char separately, but that would be future perf-optimization if we actually need it, given that most strings don't contain line breaks anyway.

Remotion/Web/UnitTests/Core/WebStringTest.cs
MichaelKetting: in the review-comment-view on github, it's a bit context-less.

Here, the relevant part is, that there's two kinds of enum values, those that auto-encode and those that don't. we need to show in the test that the correct behavior happens in both cases. It's okay to have a few extra tests if it just coveres the relevant permutations explicitly. Encoding can be tricky, we don't want to think about it.

Remotion/Web/Core/WebString.cs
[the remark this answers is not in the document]
a-ctor: yes it is

a-ctor:
@MichaelKetting We might want to rethink allowing `null` values in the `WebString`. As of now we don't and this differs from `HttpUtility.HtmlEncode` which can handle `null` values. This makes migration harder has you have to keep this in mind when migrating because now you will get an `ArgumentException` if you feed in `null`.

MichaelKetting:
> 
> 
> @MichaelKetting We might want to rethink allowing `null` values in the `WebString`. As of now we don't and this differs from `HttpUtility.HtmlEncode` which can handle `null` values. This makes migration harder has you have to keep this in mind when migrating because now you will get an `ArgumentException` if you feed in `null`.

so, allow null during creation and defualt to empty for `RawValue` getter?

Remotion/Web/Core/ObsoleteDiagnosticIDs.cs
```diff
@@ -0,0 +1,25 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+
+namespace Remotion.Web
+{
+  internal static class ObsoleteDiagnosticIDs
+  {
+    public const string HtmlUtilityDiagID = "REMOTION+0001";
```
MichaelKetting, line 23:
```suggestion
    public const string HtmlUtility = "REMOTION+0001";
```

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -29,25 +28,28 @@ public static string Format (string htmlFormatString, params object[] nonHtmlPar
     {
       string[] htmlParameters = new string[nonHtmlParameters.Length];
       for (int i = 0; i < nonHtmlParameters.Length; ++i)
-        htmlParameters[i] = HtmlEncode (nonHtmlParameters[i].ToString());
+        htmlParameters[i] = WebString.CreateFromText (nonHtmlParameters[i].ToString() ?? string.Empty).ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
```
MichaelKetting, line 31: Why, just why may ToString() return null. This is just...seriously?

Anyhow, since you requested to allow null-input, this solved itself.

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -29,25 +28,28 @@ public static string Format (string htmlFormatString, params object[] nonHtmlPar
...
       return string.Format (htmlFormatString, (object[]) htmlParameters);
     }
 
+#if !NETFRAMEWORK
+    [Obsolete (
+        "HtmlUtility.Encode(string) is obsolete. Use WebString.ToString(WebStringEncoding.HtmlWithTransformedLineBreaks) instead. (Version 3.0.0)",
```
MichaelKetting, line 37:
```suggestion
        "HtmlUtility.HtmlEncode(string) is obsolete. Use WebString.CreateFromText(string).ToString(WebStringEncoding.HtmlWithTransformedLineBreaks) instead. (Version 3.0.0)",
```

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -29,25 +28,28 @@ public static string Format (string htmlFormatString, params object[] nonHtmlPar
...
-      }
-      return html;
+      return WebString.CreateFromText (nonHtmlString).ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
     }
 
+#if !NETFRAMEWORK
+    [Obsolete (
+        "HtmlUtility.Encode(string, HtmlTextWriter) is obsolete. Use WebString.Write(HtmlTextWriter) instead. (Version 3.0.0)",
+        DiagnosticId = ObsoleteDiagnosticIDs.HtmlUtilityDiagID)]
+#endif
```
MichaelKetting, line 49: This one we can make just obsolete. There's no usage and a straight-forward conversion.

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+  /// </summary>
+  public readonly struct WebString : IEquatable<WebString>
+  {
+    private static readonly string[] s_newlineSplitSeparators = { "\r\n", "\r", "\n" };
+
+    /// <summary>
+    /// Creates a <see cref="WebString"/> from the specified <paramref name="html"/>.
+    /// The value will be rendered without being encoded and thus must be a safe string.
+    /// </summary>
+    /// <param name="html">The HTML from which a <see cref="WebString"/> is created. Must not be <see langword="null" />. Must not be <see langword="null" />.</param>
```
MichaelKetting, line 22: you're going to update this anyway, but there's two of the "Must not be null" infos.

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// <summary>
+    /// Gets the type of string represented by this <see cref="WebString"/>.
+    /// </summary>
+    /// <value>The type of string represented by this <see cref="WebString"/>.</value>
+    public WebStringType Type => _type;
+
+    /// <summary>
+    /// Tests whether the <see cref="WebString"/>s value is <see langword="null" /> or an empty string.
+    /// </summary>
+    public bool IsEmpty => string.IsNullOrEmpty (_value);
```
MichaelKetting, line 85: After the ctor

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+    {
+      _value = value;
+      _type = type;
+    }
+
+    /// <summary>
+    /// Gets the raw string value of the <see cref="WebString"/> or an empty string in case the value is <see langword="null" />.
+    /// </summary>
+    [NotNull]
+    public string GetValue ()
```
MichaelKetting, line 97: I'm happy with using an expression here.

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+    public bool IsEmpty => string.IsNullOrEmpty (_value);
+
+    private WebString ([CanBeNull] string value, WebStringType type)
+    {
+      _value = value;
+      _type = type;
+    }
+
+    /// <summary>
+    /// Gets the raw string value of the <see cref="WebString"/> or an empty string in case the value is <see langword="null" />.
```
MichaelKetting, line 94:
```suggestion
    /// Gets the raw string value of the <see cref="WebString"/> or an empty string if there is no value.
```

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+    {
+      return _value ?? string.Empty;
+    }
+
+    /// <summary>
+    /// Writes the <see cref="WebString"/>'s value to the specified <paramref name="writer"/>.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the value will be appended to. Must not be <see langword="null" />.</param>
+    /// <exception cref="InvalidOperationException">The <see cref="WebString"/>'s type is not compatible with the operation.</exception>
```
MichaelKetting, line 107:
```suggestion
```

a-ctor:
> 
> 
> > @MichaelKetting We might want to rethink allowing `null` values in the `WebString`. As of now we don't and this differs from `HttpUtility.HtmlEncode` which can handle `null` values. This makes migration harder has you have to keep this in mind when migrating because now you will get an `ArgumentException` if you feed in `null`.
> 
> so, allow null during creation and defualt to empty for `RawValue` getter?

For example, yes. We could also return `null` when calling `GetValue()` and use an empty string in `.ToString()`. This seems to be the most natural thing in my opinion.

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -29,25 +28,28 @@ public static string Format (string htmlFormatString, params object[] nonHtmlPar
     {
       string[] htmlParameters = new string[nonHtmlParameters.Length];
       for (int i = 0; i < nonHtmlParameters.Length; ++i)
-        htmlParameters[i] = HtmlEncode (nonHtmlParameters[i].ToString());
```
MichaelKetting, line 32: Please make obsolete. there's no usage.

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -29,25 +28,28 @@ public static string Format (string htmlFormatString, params object[] nonHtmlPar
...
+        "HtmlUtility.Encode(string, HtmlTextWriter) is obsolete. Use WebString.Write(HtmlTextWriter) instead. (Version 3.0.0)",
+        DiagnosticId = ObsoleteDiagnosticIDs.HtmlUtilityDiagID)]
+#endif
     public static void HtmlEncode (string nonHtmlString, HtmlTextWriter writer)
     {
-      writer.Write (HtmlEncode (nonHtmlString));
+      WebString.CreateFromText (nonHtmlString).Write (writer);
     }
 
     private static readonly Regex s_stripHtmlTagsRegex = new Regex ("<.*?>", RegexOptions.Compiled);
```
MichaelKetting, line 55: please make obsolete. there's no usage.

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+            if (i < parts.Length - 1)
+              writer.WriteBreak();
+          }
+
+          break;
+        case WebStringType.Encoded:
+          writer.Write (GetValue());
+          break;
+        default:
+          throw new ArgumentOutOfRangeException();
```
MichaelKetting, line 130: NotSupportedException, please add the _type to the message.

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+      switch (_type)
+      {
+        case WebStringType.Unencoded:
+          writer.AddAttribute (attribute, GetValue(), fEncode: true);
+          break;
+        case WebStringType.Encoded:
+          writer.AddAttribute (attribute, GetValue(), fEncode: false);
+          break;
+        default:
+          throw new ArgumentOutOfRangeException();
```
MichaelKetting, line 154: see `Write`

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+      switch (_type)
+      {
+        case WebStringType.Unencoded:
+          writer.AddAttribute (attribute, GetValue(), fEndode: true);
+          break;
+        case WebStringType.Encoded:
+          writer.AddAttribute (attribute, GetValue(), fEndode: false);
+          break;
+        default:
+          throw new ArgumentOutOfRangeException();
```
MichaelKetting, line 178: See `Write`

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+      }
+    }
+
+    /// <summary>
+    /// Adds the specified markup <paramref name="attribute"/> to the <paramref name="writer"/> using the <see cref="WebString"/> value.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The attribute that is to be added.</param>
+    /// <exception cref="InvalidOperationException">The <see cref="WebString"/>'s type is not compatible with the operation.</exception>
```
MichaelKetting, line 140:
```suggestion
```

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+      }
+    }
+
+    /// <summary>
+    /// Adds the specified markup <paramref name="attribute"/> to the <paramref name="writer"/> using the <see cref="WebString"/> value.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The name of the attribute that is to be added.</param>
+    /// <exception cref="InvalidOperationException">The <see cref="WebString"/>'s type is not compatible with the operation.</exception>
```
MichaelKetting, line 164:
```suggestion
```

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /// <summary>
+    /// Adds the specified markup <paramref name="attribute"/> to the <paramref name="writer"/> using the <see cref="WebString"/> value.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The attribute that is to be added.</param>
+    /// <exception cref="InvalidOperationException">The <see cref="WebString"/>'s type is not compatible with the operation.</exception>
+    public void AddAttribute ([NotNull] HtmlTextWriter writer, HtmlTextWriterAttribute attribute)
```
MichaelKetting, line 141: Need to documetn what happens when WebString is empty. in the future, we might want to circle back and sckip empty attributes but this is out-of-scope right now because too many moving parts.

Remotion/Web/Core/WebStringType.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+  /// <summary>
+  /// Defines the kind of a <see cref="WebString"/> which determines if it is encoded when rendered.
+  /// </summary>
+  public enum WebStringType
+  {
+    /// <summary>
+    /// The <see cref="WebString"/> contains unencoded text, which will be encoded before being rendered.
+    /// </summary>
+    Unencoded,
```
MichaelKetting, line 11: `PlainText` feels better here. Thoughts?

Remotion/Web/Core/WebStringEncoding.cs
```diff
@@ -0,0 +1,23 @@
...
+  /// </summary>
+  public enum WebStringEncoding
+  {
+    /// <summary>
+    /// An unencoded <see cref="WebString"/> will be HTML encoded.
+    /// </summary>
+    Html,
+    
+    /// <summary>
+    /// An unencoded <see cref="WebString"/> will be HTML encoded but newlines will be replaced by a &lt;br /&gt; tag.
```
MichaelKetting, line 14:
```suggestion
    /// An unencoded <see cref="WebString"/> will be HTML encoded and line break characters will be replaced by  &lt;br /&gt; tags.
```

Remotion/Web/UnitTests/Core/Utilities/HtmlUtilityTest.cs
```diff
@@ -23,36 +23,6 @@ namespace Remotion.Web.UnitTests.Core.Utilities
...
-    {
-      Assert.That (HtmlUtility.HtmlEncode ("a\rb"), Is.EqualTo ("a<br />b"));
-    }
-
-    [Test]
-    public void HtmlEncode_LineFeedNewLine ()
-    {
-      Assert.That (HtmlUtility.HtmlEncode ("a\r\nb"), Is.EqualTo ("a<br />b"));
-    }
-
```
MichaelKetting, line 55: why deleted?

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,190 @@
...
+
+      var renderedString = ExecuteWithHtmlTextWriter (
+          writer =>
+          {
+            webString.AddAttribute (writer, "class");
+            writer.RenderBeginTag ("a");
+          });
+      Assert.That (renderedString, Is.EqualTo ($@"<a class=""{stringValue}"">"));
+    }
+
```
MichaelKetting, line 146: Write/AddAttribute tests for empty and default

MichaelKetting (changes requested):

MichaelKetting:
> 
> 
> > > @MichaelKetting We might want to rethink allowing `null` values in the `WebString`. As of now we don't and this differs from `HttpUtility.HtmlEncode` which can handle `null` values. This makes migration harder has you have to keep this in mind when migrating because now you will get an `ArgumentException` if you feed in `null`.
> > 
> > 
> > so, allow null during creation and defualt to empty for `RawValue` getter?
> 
> For example, yes. We could also return `null` when calling `GetValue()` and use an empty string in `.ToString()`. This seems to be the most natural thing in my opinion.

No, we should unify empty+null. Having null-strings when they do not have a semantic value is one of the most horrible design constructs we have to deal with.

Remotion/Web/UnitTests/Core/Utilities/HtmlUtilityTest.cs
a-ctor: Since we are covering those cases in the `WebStringTest` and I thought it would be better to remove them now than to remove them when we remove `HtmlUtility`.

MichaelKetting: Nah, I'd leave them. They don't hurt us and they must never become red anyway or we'd have a problem.

Remotion/Web/Core/WebString.cs
a-ctor: Empty strings will render an attribute with an empty value. Since we convert `null` to an empty string we always render an empty value.

Remotion/Web/Core/WebStringType.cs
a-ctor: `PlainText` is better

Remotion/Web/Core/WebString.cs
MichaelKetting: yeah, I checked now in the code:
null makes just the attribute-name.
empty makes attribute-name + `=""`

Either way, no need to support null and the decision to just not render a blank attribute will be made later on.

There's also the annoying bit about when an attribute without value is even valid html. (html yes, xhtml no). It's best to keep semantics simple.

Remotion/Web/Core/ObsoleteDiagnosticIDs.cs
```diff
@@ -0,0 +1,25 @@
...
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+
+namespace Remotion.Web
+{
+  internal static class ObsoleteDiagnosticIDs
```
MichaelKetting, line 21: Oh, I just noticed. I meant we put this in Core.Core, so, yes, it's public. Perhaps a subnamesace? We could misuse the Obsolete-namespace for this.

Remotion/Web/Core/Utilities/HtmlUtility.cs
MichaelKetting: The obsolete is missing for stripHtmlTags

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -25,29 +24,35 @@ namespace Remotion.Web.Utilities
...
+      return WebString.CreateFromText (nonHtmlString).ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
     }
 
+#if NETFRAMEWORK
+    [Obsolete ("HtmlUtility.HtmlEncode(string, HtmlTextWriter) is obsolete. Use WebString.CreateFromText(string).Write(HtmlTextWriter) instead. (Version 3.0.0)")]
+#else
+    [Obsolete (
+        "HtmlUtility.HtmlEncode(string, HtmlTextWriter) is obsolete. Use WebString.CreateFromText(string).Write(HtmlTextWriter) instead. (Version 3.0.0)",
+        DiagnosticId = ObsoleteDiagnosticIDs.HtmlUtility)]
+#endif
```
MichaelKetting, line 52:
```suggestion
```

No usages downstream, don't need the graceful obsolete.

Remotion/Web/Core/Utilities/HtmlUtility.cs
```diff
@@ -25,29 +24,35 @@ namespace Remotion.Web.Utilities
...
-      {
-        html = html.Replace ("\r\n", "<br />");
-        html = html.Replace ("\n", "<br />");
-        html = html.Replace ("\r", "<br />");
-      }
-      return html;
+      return WebString.CreateFromText (nonHtmlString).ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
     }
 
+#if NETFRAMEWORK
```
MichaelKetting, line 46:
```suggestion
```

MichaelKetting (commented):
Partial review

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,238 @@
...
+
+    [CanBeNull]
+    private readonly string _value;
+    private readonly WebStringType _type;
+
+    /// <summary>
+    /// Gets the type of string represented by this <see cref="WebString"/>.
+    /// </summary>
+    /// <value>The type of string represented by this <see cref="WebString"/>.</value>
+    public WebStringType Type => _type;
```
MichaelKetting, line 76: Move down

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,238 @@
...
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <remarks>
+    /// Renders an attribute with an empty value if the <see cref="WebString"/> has no value.
+    /// </remarks>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The name of the attribute that is to be added.</param>
+    public void AddAttribute ([NotNull] HtmlTextWriter writer, string attribute)
+    {
+      ArgumentUtility.CheckNotNull (nameof (writer), writer);
```
MichaelKetting, line 163: Oops, `attribute` needs a not-null-or-empty check and docs.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,254 @@
...
+    {
+      var stringValue = "aoe   \" & ' < > é \r \n \r\n";
+      var webString = WebString.CreateFromHtml (stringValue);
+
+      var renderedString = ExecuteWithHtmlTextWriter (webString.Write);
+      Assert.That (renderedString, Is.EqualTo (stringValue));
+    }
+
+    [Test]
+    public void AddAttribute_NullUnencodedWebString_RendersEmptyAttribute ()
```
MichaelKetting, line 109: Update those `Unencoded` with `PlainText`?

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,254 @@
...
+          writer =>
+          {
+            webString.AddAttribute (writer, HtmlTextWriterAttribute.Class);
+            writer.RenderBeginTag ("a");
+          });
+      Assert.That (renderedString, Is.EqualTo ("<a class=\"\">"));
+    }
+
+    [Test]
+    public void AddAttribute_UnencodedWebStringWithAttributeEnum_EncodesAttributeValue ()
```
MichaelKetting, line 137: unify `EncodesAttributeValue` and `EncodesOutput` in method names.

Hmmm, also use `RendersEncodedOutput` or `RenderesEncodedAttributeValue` to keep in line with other test methods in this file?

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,254 @@
...
+          writer =>
+          {
+            webString.AddAttribute (writer, HtmlTextWriterAttribute.Class);
+            writer.RenderBeginTag ("a");
+          });
+      Assert.That (renderedString, Is.EqualTo ("<a class=\"aoe   &quot; &amp; &#39; &lt; > é \r \n \r\n\">"));
+    }
+
+    [Test]
+    public void AddAttribute_UnencodedWebStringWithAttributeEnumThatHasEncodeDefault_EncodeOutput ()
```
MichaelKetting, line 152: Tests in line 137 and 152 appear to have identical code? Am I missing something?

MichaelKetting (changes requested):

Remotion/Web/Core/Utilities/HtmlUtility.cs
a-ctor: And what should I do with our 12 usages in remotion?

MichaelKetting: ah, crap. missed that one. okay, we leave thsi for now

Remotion/Web/Core/WebString.cs
```diff
@@ -0,0 +1,239 @@
...
+    /// <summary>
+    /// Adds the specified markup <paramref name="attribute"/> to the <paramref name="writer"/> using the <see cref="WebString"/> value.
+    /// The value is escaped if necessary, depending on the <see cref="Type"/> of the <see cref="WebString"/>.
+    /// </summary>
+    /// <remarks>
+    /// Renders an attribute with an empty value if the <see cref="WebString"/> has no value.
+    /// </remarks>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the attribute will be added. Must not be <see langword="null" />.</param>
+    /// <param name="attribute">The name of the attribute that is to be added. Must not be <see langword="null" /> or empty.</param>
+    public void AddAttribute ([NotNull] HtmlTextWriter writer, string attribute)
```
MichaelKetting, line 161:
```suggestion
    public void AddAttribute ([NotNull] HtmlTextWriter writer, [NotNull] string attribute)
```

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,239 @@
...
+          writer =>
+          {
+            webString.AddAttribute (writer, "class");
+            writer.RenderBeginTag ("a");
+          });
+      Assert.That (renderedString, Is.EqualTo ("<a class=\"aoe   &quot; &amp; &#39; &lt; > é \r \n \r\n\">"));
+    }
+
+    [Test]
+    public void AddAttribute_EncodedWebStringWithAttributeString_RendersUnencodedAttributeValue ()
```
MichaelKetting, line 182: Are we mssing the test with CreateFromHtml for Attribute.Class and Attribute.Cols ?

MichaelKetting (changes requested):

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -0,0 +1,269 @@
...
+          writer =>
+          {
+            webString.AddAttribute (writer, HtmlTextWriterAttribute.Cols);
+            writer.RenderBeginTag ("a");
+          });
+      Assert.That (renderedString, Is.EqualTo ($"<a cols=\"{stringValue}\">"));
+    }
+
+    [Test]
+    public void AddAttribute_EncodedWebStringWithAttributeString_RendersEncodedAttributeValue ()
```
MichaelKetting, line 197: should be enum, not string.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #171, opened 2021-08-11
RM-7921 Create styles for new theme for WebButton

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7921

### DIFF
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
@@ -2,12 +2,16 @@
 {
   --color-primary: #0C6594;
   --color-primary-highlight: #218FC9;
+  --color-primary-text: white;
   --color-highlight: #DEF3FE;
   --color-contrast-separator: #C4C4C4;
   --color-standard-separator: #E1E1E1;
   --color-contrast: #F2F2F2;
   --color-contrast-background: #F9F9F9;
   --color-standard-border: #979797;
+  --color-disabled-text: #979797;
+  --color-disabled-border: #979797;
+  --color-disabled-background: white;
   --color-indicator-error: #DA2222;
   --color-indicator-warning: #DFAF00;
   --color-indicator-success: #4AD439;
@@ -142,4 +146,14 @@ select.remotion-themed::-webkit-scrollbar-thumb,
 {
   background-color: var(--color-primary);
   border-radius: var(--remotion-themed-dropdown-scrollbar-border-radius);
+}
+
+button.remotion-themed:disabled img,
+.remotion-themed.disabled img,
+.remotion-themed button:disabled img,
+.remotion-themed .disabled img,
+.remotion-themed img.disabled
+{
+  opacity: 0.5;
+  filter: grayscale(100%);
 }
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
@@ -2,108 +2,50 @@
 
 button.webButton
 {
-  background-color: Transparent;
-  cursor: hand;
+  background-color: var(--color-standard-background);
+  border: var(--remotion-themed-border-width) solid var(--color-primary);
+  border-radius: var(--remotion-themed-border-radius);
+  padding: var(--remotion-themed-padding) calc(3 * var(--remotion-themed-padding));
+  box-sizing: border-box;
   cursor: pointer;
-  text-decoration: none !important;
-  border: none;
-  position: relative;
-  display: inline-block;
-  vertical-align: middle;
-  padding: 0px 2px;
-  -moz-padding-start: 0;
-  -moz-padding-end: 0;
-  outline: none;
 }
 
-button.webButton::-moz-focus-inner
-{ 
-  border: 0;
+button.webButton:hover,
+button.webButton:focus
+{
+  background-color: var(--color-highlight);
+}
+
+button.webButton[disabled]
+{
+  background-color: var(--color-disabled-background);
+  color: var(--color-disabled-text);
+  border-color: var(--color-disabled-border);
+  cursor: unset;
 }
 
 button.webButton span.buttonBody
 {
-  border: solid 1px #AAAAAA;
-  font-weight: bold;
-  background-image: url(../Image/sprite.svg#ButtonBackground);
-  background-size: 1px 20px;
-  background-color: #E6E5E7;
-  background-position: center top;
-  background-repeat: repeat-x;
-  
-  display: block;
-  position: relative;
-  top: 0px;
-  left: -2px;
-  
-  width: 100%;
-  white-space: nowrap;
-
-  padding-top: 2px;
-  padding-bottom: 2px;
-  padding-left: 1px;
-  padding-right: 1px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
 }
 
-button.webButton span.buttonBody:first-child::before
+button.webButton.primary
 {
-  content: "";
-  height: 1.25em;
-  width: 0;
-  display: inline-block;
-  vertical-align: text-bottom;
-}
-
-button.webButton span.buttonBody img
-{
-  vertical-align: bottom;
+  background-color: var(--color-primary);
+  color: var(--color-primary-text);
   border: none;
 }
 
-button.webButton:focus span.buttonBody,
-button.webButton:hover span.buttonBody
+button.webButton.primary:hover,
+button.webButton.primary:focus
 {
-  background-image: url(../Image/sprite.svg#ButtonBackgroundHover);
-  background-size: 1px 20px;
-  background-color: #DEECFC;
-  border: 1px solid #7CB2E6;
+  background-color: var(--color-primary-highlight);
 }
 
-button.webButton:focus span.buttonBody
+button.webButton.secondary
 {
-  text-decoration: underline;
-}
-
-button.webButton:active span.buttonBody
-{
-  background-image: url(../Image/sprite.svg#ButtonBackgroundActive);
-  background-size: 1px 20px;
-  background-color: #C5E0FC;
-  border: 1px solid #569CE0;
-}
-
-button.webButton[disabled] span.buttonBody,
-button.webButton[disabled]:hover span.buttonBody,
-button.webButton[disabled]:active span.buttonBody
-{
-  color: Gray;
-  border: solid 1px #BCBCBC;
-  background: none;
-  background-color: #EFEFF0;
-  cursor: default;
-}
-
-button.webButton[disabled] span.buttonBody img
-{
-  opacity: 0.4;
-  filter: progid:DXImageTransform.Microsoft.BasicImage(opacity=0.40),
-          progid:DXImageTransform.Microsoft.BasicImage(grayscale=1);
-}
-
-.mozilla button.webButton span.buttonBody
-{
-  left: 0;
-  width: unset;
-  padding-left: 3px;
-  padding-right: 3px;
+  background-color: transparent;
+  border: none;
 }


Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,57 @@
 
 button.webButton
```
a-ctor, line 3: In terms of variable usage and styling: we should introduce new variables for all the stlying we do on the button. border color, hover colors, ...

These are defined here and then we have one class for `primary`, `secondary`, ... that sets the variables to the new colors. This way we do not have to repeat the actual css properties and it is possible for outer controls/consumers to restyle only the web button

a-ctor (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,57 @@
...
 
-button.webButton span.buttonBody:first-child::before
+button.webButton:focus
 {
-  content: "";
-  height: 1.25em;
-  width: 0;
-  display: inline-block;
-  vertical-align: text-bottom;
+  text-decoration: underline;
```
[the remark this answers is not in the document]
MichaelKetting: Yes, you need to be able to identify the focus without relying on contrast. If you have a siginifact change in the coloring, e.g. something akin to a light/dark inversion, that works. Just using a slighlty brighter or dark color, not so much. It may be, that in the end we'll do a shadow-outline when we're done, instead of the underline, but for now, the underline is the cheapest and least controversial update to the styleguide to get the desired effect.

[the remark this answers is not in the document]
a-ctor: The browser does a shadow outline for a button element. Since we restyle a button element we get that effect if you highlight it with tab. Then it will get a dark shadow and an underline.

One highlight should be enough right? So I would remove the underline

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
MichaelKetting: Ah, that's new. IE didn't do it like that. Okay then, no underline.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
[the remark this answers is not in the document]
MichaelKetting: Ah, that's new. IE didn't do it like that. Okay then, no underline.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,57 @@
...
-  outline: none;
-}
-
-button.webButton::-moz-focus-inner
-{ 
-  border: 0;
+  background-color: transparent;
+  border: 1px solid var(--color-primary);
+  border-radius: 2px;
+  padding: 0.6em;
```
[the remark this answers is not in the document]
iam-mholle: They are more streched in the screenshots, still the width can be increased by the developer using the button control. What's @MichaelKetting 's opinion on this?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,57 @@
...
 
-button.webButton:focus span.buttonBody
+button.webButton.secondary
 {
-  text-decoration: underline;
+  border: none;
 }
 
-button.webButton:active span.buttonBody
+button.webButton[disabled]
```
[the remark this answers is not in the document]
iam-mholle: That's what I would've done, but I think we should order rules by nesting depth. I'll wait for @MichaelKetting 's response.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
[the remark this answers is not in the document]
MichaelKetting: We should use 3*0.4em for the left/right margin and 0.4em for the top/bottom. See https://github.com/re-motion/Framework/pull/169

In the end, the button should be like the following image and (probably) align with the textbox. Very likely, we will want to use the same property that defines the padding for textbox.
image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,54 @@
...
-button.webButton[disabled]:active span.buttonBody
+button.webButton span.buttonBody
 {
-  color: Gray;
-  border: solid 1px #BCBCBC;
-  background: none;
-  background-color: #EFEFF0;
-  cursor: default;
+  display: flex;
+  align-items: center;
```
MichaelKetting, line 48: The vertical alignment is more complicated. only the on-the-line portion of the text should be vertically centered. I do not know if we have a baseline option with flexgrif or if we need to use other means.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,54 @@
 
 button.webButton
 {
-  background-color: Transparent;
-  cursor: hand;
+  background-color: transparent;
+  border: 1px solid var(--color-primary);
```
MichaelKetting, line 6: border radius/border-width should be takable from the remotion theme defaults.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,54 @@
...
-  width: 0;
-  display: inline-block;
-  vertical-align: text-bottom;
-}
-
-button.webButton span.buttonBody img
-{
-  vertical-align: bottom;
+  background-color: var(--color-primary);
+  color: white;
```
MichaelKetting, line 22: We should introduce `--color-primary-text` for this.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,54 @@
...
 
-button.webButton:active span.buttonBody
+button.webButton[disabled]
 {
-  background-image: url(../Image/ButtonBackgroundActive.gif);
-  background-color: #C5E0FC;
-  border: 1px solid #569CE0;
+  background-color: transparent;
+  color: var(--color-standard-border);
+  border: 1px solid var(--color-standard-border);
```
MichaelKetting, line 41: We need to introduce `--color-disabled-border` for this.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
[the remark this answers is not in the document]
MichaelKetting: sounds okay?

MichaelKetting (changes requested):
please rebase

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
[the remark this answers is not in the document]
iam-mholle: Which way? The suggested one or https://github.com/re-motion/Framework/pull/169#discussion_r685722847 ?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
iam-mholle: Which would be equivalent to `--remotion-themed-border`?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
iam-mholle: Well the on-the-line portion of the text is vertically centered now, as well as the image, or am I missing something?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,105 +2,54 @@
 
 button.webButton
 {
-  background-color: Transparent;
-  cursor: hand;
+  background-color: transparent;
+  border: var(--remotion-themed-border-width) solid var(--color-primary);
+  border-radius: var(--remotion-themed-border-radius);
+  padding: 0.4em 1.2em;
```
MichaelKetting, line 8: one more thing: please check that setting the width on the button control or setting the width va css on button-element works correctly (i.e. horizontally centers the text)

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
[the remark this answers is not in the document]
MichaelKetting: The one @a-ctor made. 
We have an extra css class and completely new set of rules and they should be grouped together. Plus, it's only going to be used in the commit further down the line and you both want it like that. Sprinkling the new alternate in between the other items doesn't feel like it's helping readability.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
MichaelKetting: I think so? "disabled" is special as it's meant to make a quite uniform gray-style. It will be important to be able to distinguish between just a gray border and a disabled element, so there is also the grayed-out background when there's color and and the gray text-color. Only thing that' not becoming gray (typically) is white background.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
MichaelKetting: image

It doesn't look that way to me?

iam-mholle: My (not so scientific) measurements:
image

Pretty centered imo, even though its not perfect.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,108 +2,55 @@
...
```
MichaelKetting, line 8: the paddings get updated with the css properties from the RM-7911 BocTextValue PR. to be reviewed next.
```suggestion
  padding: calc(3 * var(--remotion-themed-padding)) calc(3 * var(--remotion-themed-padding));
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
iam-mholle: Added properties in [`3f60642` (#171)](https://github.com/re-motion/Framework/pull/171/commits/3f6064273d44ba50e94ed957dfa8a51abcbe75b0), please check.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,108 +2,55 @@
...
-  vertical-align: text-bottom;
-}
-
-button.webButton span.buttonBody img
-{
-  vertical-align: bottom;
-  border: none;
+  background-color: transparent;
+  color: var(--color-standard-border);
+  border: var(--remotion-themed-border);
```
MichaelKetting: for those two, we need properties with `--color-disabled-border` and `--color-disabled-text`.  please also check with RM-7911 if there's overlap.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,108 +2,55 @@
...
-  width: 0;
-  display: inline-block;
-  vertical-align: text-bottom;
-}
-
-button.webButton span.buttonBody img
-{
-  vertical-align: bottom;
-  border: none;
+  background-color: transparent;
```
MichaelKetting, line 21: maybe we should make a `--color-standard-background`?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,108 +2,55 @@
...
 
-button.webButton:active span.buttonBody
+button.webButton[disabled] span.buttonBody img
 {
-  background-image: url(../Image/sprite.svg#ButtonBackgroundActive);
-  background-size: 1px 20px;
-  background-color: #C5E0FC;
-  border: 1px solid #569CE0;
+  opacity: 0.5;
+  filter: grayscale(100%);
```
MichaelKetting, line 37: disabled images will recur in the styles. should we make common properties and/or do a re-motion themed option? Basically, we could go with re-motion themed image that is nested inside parent that has either the [disabled] attribute or the `.disabled` class applied.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
MichaelKetting: looks good.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
iam-mholle: It's already defined: `--color-standard-background: #FFFFFF;`

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
iam-mholle: This might be implemented later, but currently is out of scope.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
MichaelKetting: okay, update: we should use white (okay) but we should use --color-standard-background here and in line 5 (button.webbutton). Thus, only the secondary button has a transparent background.

Reasoning: we use white as background for all frameed elements, so the standard button should also get a white background with it's border. Works better with the properyt consistency as well.

 And yes, maybe the name "secondary" isn't great and should be changed in rm-7928. Would it be better to move those eitems with primary and secondary button from the current changeset to a code snippet in the rm-7928 task?

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
MichaelKetting: I just saw this on an input-value screenshop: the browser natively makes the background light gray instead of white (or we do it, don't know). We should consider using a light gray for the --color-disabled-background. Of cource, this could be decided later on, so long as we already have a css property for this.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -114,4 +117,15 @@ select.remotion-themed::-webkit-scrollbar-thumb,
...
   border-radius: var(--remotion-themed-dropdown-scrollbar-border-radius);
-}
\ No newline at end of file
+}
+
+.remotion-themed[disabled] img,
+.remotion-themed.disabled img,
+.remotion-themed [disabled] img,
+.remotion-themed .disabled img,
+.remotion-themed img[disabled],
```
MichaelKetting, line 126:
```suggestion
```
You can't mark an image as disabled, only input, button and select

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -114,4 +117,15 @@ select.remotion-themed::-webkit-scrollbar-thumb,
 {
   background-color: var(--color-primary);
   border-radius: var(--remotion-themed-dropdown-scrollbar-border-radius);
-}
\ No newline at end of file
+}
+
+.remotion-themed[disabled] img,
```
MichaelKetting, line 122:
```suggestion
button.remotion-themed:disabled img,
```
you can't nest img inside input or select or textarea, so it's just button:disabled.

There is no `disabled` attribute as a general concept in HTML. ASP.NET does it, but this is invalid markup.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -2,108 +2,50 @@
...
 }
 
-.mozilla button.webButton span.buttonBody
+button.webButton.secondary
 {
-  left: 0;
-  width: unset;
-  padding-left: 3px;
-  padding-right: 3px;
+  background-color: transparent;
```
MichaelKetting, line 49:
```suggestion
```
https://github.com/re-motion/Framework/pull/171#discussion_r690276050
> okay, update: we should use white (okay) but we should use --color-standard-background here and in line 5 (button.webbutton). **Thus, only the secondary button has a transparent background.**

MichaelKetting (approved):
done...

please make separate commits for common.css and webbutton.css.

a-ctor (approved):

PR #172, opened 2021-08-11
RM-7951 Create Shared project for ObjectBinding.Web.Development.WebTesting.TestSite

a-ctor:
https://re-motion.atlassian.net/browse/RM-7951


Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/TestResourceType.cs
```diff
@@ -0,0 +1,25 @@
...
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using Remotion.Web;
+
+namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite.Shared
+{
+  public static class TestResourceType
```
MichaelKetting, line 21: should we move that to testsite.infrastructure project?

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/ObjectBinding.Web.Development.WebTesting.TestSite.csproj
```diff
@@ -166,6 +134,8 @@
       <Link>Remotion LGPL.licenseheader</Link>
     </Content>
     <Content Include="packages.config" />
+    <None Include="Images\Remotion.ObjectBinding.Sample.Job.gif" />
+    <None Include="Images\Remotion.ObjectBinding.Sample.Person.gif" />
```
MichaelKetting, line 138: those images should be moved to shared project, folder res/Image and the paths updated where the image is used.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Sample/ReflectionBusinessObjectWebUIService.cs
```diff
@@ -35,6 +35,13 @@ public ToolTipBasedHelpInfo (string toolTip)
       }
     }
 
+    private readonly string _basePath;
+
+    public ReflectionBusinessObjectWebUIService (string basePath)
```
MichaelKetting, line 40: We should be able to solve this using the resourceurlfactory.
Inject the ResourceUrlFactory, use
```
var targetType = ((BindableObjectClass) obj.BusinessObjectClass).TargetType;
x.GetResourceUrl (targetType, ResourceType.Image, targetType + ".gif");

MichaelKetting (changes requested):
please remove log.txt

Remotion/ObjectBinding/Sample/ReflectionBusinessObjectWebUIService.cs
a-ctor: Yes but in one project the files are not yet at the correct location since the ObjectBinding.Web.Test project has not been converted yet. This change can be undone in once the other project has been converted.

It would also be possible to support both explicitly providing a path and defaulting to the resource path

MichaelKetting: Ah, yes. option 2, then, please. And a comment that his should be cleaned up when the other project is done.
> It would also be possible to support both explicitly providing a path and defaulting to the resource path

PR #180, opened 2021-08-19
RM-7856 Upgrade Firefox to Version 89

derlistigelurch:


Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
```diff
@@ -44,6 +44,7 @@ public void SetUp ()
     }
 
     [Test]
+    [Ignore ("Test does not work properly after Firefox browser upgrade (Version 89).")]
```
MichaelKetting, line 47: Hmm... that's actually just firefox, isn't it? so we need to do this ignore conditional for only-firefox.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #184, opened 2021-08-23
RM-7916 Create Styles for new theme for ListMenu

derlistigelurch:


Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
```diff
@@ -2,103 +2,55 @@
...
 td.listMenuRow
 {
-  padding-left: 0;
-  padding-right: 0;
-  padding-top: 1px;
-  padding-bottom: 1px;
-  vertical-align: middle;
-  white-space: nowrap;
+  display: flex;
+  width: 100%;
```
iam-mholle, line 12: Is this really needed?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
```diff
@@ -2,103 +2,55 @@
...
+  gap: var(--remotion-themed-padding);
+  padding-bottom: var(--remotion-themed-padding);
```
iam-mholle, line 14: Wouldn't `margin-bottom` make more sense here?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
```diff
@@ -2,103 +2,55 @@
...
-span.listMenuItemDisabled a
-{
-  display: inline-block;
-}
-
 span.listMenuItem a img,
 span.listMenuItemDisabled a img
 {
-  vertical-align: middle;
+  gap: var(--remotion-themed-padding);
```
iam-mholle, line 40: Is this needed? Shouldn't the gap be set on the container element?

iam-mholle (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
derlistigelurch: You are right. The gap isn't needed here anyway because it's already set on the container element.

iam-mholle (approved):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
```diff
@@ -2,103 +2,53 @@
...
-
 span.listMenuItem a img,
 span.listMenuItemDisabled a img
 {
-  vertical-align: middle;
+  height: var(--programm-icon-size);
+  width: var(--programm-icon-size);
 }
 
-span.listMenuItem a span,
```
MichaelKetting, line 94: please add a rule for img+span:empty as display:hidden. Otherwise, an image without text looks strange.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
```diff
@@ -2,103 +2,53 @@
...
 td.listMenuRow
 {
-  padding-left: 0;
-  padding-right: 0;
-  padding-top: 1px;
-  padding-bottom: 1px;
-  vertical-align: middle;
-  white-space: nowrap;
+  display: flex;
+  gap: var(--remotion-themed-padding);
```
MichaelKetting, line 12:
```suggestion
  gap: calc(2*var(--remotion-themed-padding));
```

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
derlistigelurch: Where does this behaviour occur?

MichaelKetting: Okay, just realized that the sample is not configured correctly. You have to disable the text-span explicitly if you don't want it. Just leaving it empty isn't working out.  Could you please update the test page instead to set `Style=Icon`?
image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -31,6 +31,8 @@
   --remotion-themed-min-height: calc(var(--remotion-themed-line-height) + 2 * var(--remotion-themed-padding));
   --remotion-themed-multiline-min-height: calc(2 * var(--remotion-themed-line-height) + 2 * var(--remotion-themed-padding));
   --remotion-themed-control-width: 16em;
+  --default-icon-size: 16px;
+  --default-text-color: black;
```
MichaelKetting, line 35: That's --color-standard-text with #000000. please add in line 21 before background.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #191, opened 2021-08-27
RM-7917 Create styles for new theme for BocEnumValue

FlorianLautner:
Currently, I fail to find a way to center the currently selected option in a single-line select in Firefox. In Chrome, it is by default centered, which doesn't make it viable to add padding since even though it's then centered in Firefox, it is off in Chrome.

grafik

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
@@ -4,7 +4,6 @@ span.bocEnumValue,
 div.bocEnumValue.readOnly
 {
   display: inline-flex;
-  align-items: center;
   width: var(--remotion-themed-control-width);
   min-height: var(--remotion-themed-min-height);
 }
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
@@ -3,112 +3,10 @@
 span.bocEnumValue,
 div.bocEnumValue.readOnly
 {
+  display: inline-flex;
+  align-items: center;
   width: var(--remotion-themed-control-width);
   min-height: var(--remotion-themed-min-height);
-  position: relative;
-  display: flex;
-  overflow: hidden;
-}
-
-div.bocEnumValue.readOnly span
-{
-  /*box-sizing: border-box;*/
-  display: inline-flex;
-  align-items: center;
-}
-
-/*div.bocEnumValue table
-{
-  height: 100%;
-  box-sizing: border-box;
-}*/
-
-div.bocEnumValue td,
-div.bocEnumValue td span
-{
-  box-sizing: border-box;
-  display: inline-flex;
-  align-items: center;
-  margin-right: 0.4em;
-}
-
-div.bocEnumValue input
-{
-  box-sizing: border-box;
-  /* Has to be set explicitely at the moment, since otherwise a top  
-     margin is applied I have not able to determine the source of. */
-  margin: 0 0.4em 0 0;
-}
-
-.bocEnumValue.readOnly span
-/*div.bocEnumValue.readOnly*/
-{
-  /*padding: var(--remotion-themed-padding);*/
-  box-sizing: border-box;
-  
-}
-
-span.bocEnumValue.readOnly span,
-div.bocEnumValue.readOnly span
-{
-  /*padding: var(--remotion-themed-padding);*/
-  min-height: var(--remotion-themed-min-height);
-}
-
-
-.bocEnumValue.disabled
-{
-}
-
-span.bocEnumValue select
-{
-  box-sizing: border-box;
-  width: 100%;
-  min-height: calc(var(--remotion-themed-min-height));
-  /*gap: 10em;*/
-
-}
-
-span.bocEnumValue option
-{
-  box-sizing: border-box;
-  padding: var(--remotion-themed-padding);
-  /*padding: calc(var(--remotion-themed-padding) * 0.75);*/
-  /*padding: calc(var(--remotion-themed-padding) * 0.5);*/
-  border-radius: calc(var(--remotion-themed-border-radius));
-  /*margin: calc(var(--remotion-themed-padding));*/
-  /*gap: 2em;*/
-  /*line-height: 3em;*/
-}
-
-
-span.bocEnumValue option:hover
-{
-  background-color: var(--color-highlight);
-  /*flex: 4;*/
-}
-
-span.bocEnumValue.disabled option
-{
-  background-color: var(--color-disabled-background);
-  text-decoration: none;
-  color: var(--color-disabled-text);
-}
-
-
-
-
-
-span.bocEnumValue select[size]
-{
-  /*height: auto;*/
-}
-
-span.bocEnumValue select[size="1"]
-{
-  /*height: 1.68em;*/
-  /*height: var(--remotion-themed-line-height);*/
-  /*height: auto;*/
 }
 
 .bocEnumValue td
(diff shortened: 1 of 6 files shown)


FlorianLautner:
The issue of off-centered labels of selected options in Firefox is now fixed. However, another issues persists in Firefox which slightly changes the height of the `td` containing the `select ` after an empty `option` has been selected vs. a non-empty `option` without any apparent reason.

## non-empty `option`
grafik

## empty `option`
grafik

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -231,9 +248,14 @@ private ListControl GetListControl (BocEnumValueRenderingContext renderingContex
         radioButton.InputAttributes.Add (HtmlTextWriterAttribute2.AriaRequired, HtmlAriaRequiredAttributeValue.True);
       }
 
+
+
```
derlistigelurch, line 252:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -231,9 +248,14 @@ private ListControl GetListControl (BocEnumValueRenderingContext renderingContex
...
       return listControl;
     }
 
+
+
+
```
derlistigelurch, line 258:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -152,7 +166,7 @@ protected override void AddDiagnosticMetadataAttributes (RenderingContext<IBocEn
 
       renderingContext.Writer.AddAttribute (DiagnosticMetadataAttributesForObjectBinding.NullIdentifier, renderingContext.Control.NullIdentifier);
     }
-
+
```
derlistigelurch, line 169:
```suggestion

```

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -129,6 +130,19 @@ public void Render (BocEnumValueRenderingContext renderingContext)
...
+      //if (!listControl.Enabled && WebControl.DisabledCssClass.Equals ("aspNetDisabled"))
+      //if (innerControl.ClientID.Contains("body_DataEditControl_DisabledUnboundMarriageStatusField") && !listControl.Enabled)
+      //if (!listControl.Enabled && WebControl.DisabledCssClass.Equals("aspNetDisabled"))
+      //if (!listControl.Enabled)
+      //{
+      //    listControl.
+      //    listControl.Attributes.Add("disabled", "disabled");
+      //}
+
+
```
derlistigelurch, line 145: Since these lines are no longer used they can be deleted right?

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -106,6 +106,7 @@ public void Render (BocEnumValueRenderingContext renderingContext)
       bool isControlWidthEmpty = renderingContext.Control.Width.IsEmpty && string.IsNullOrEmpty (renderingContext.Control.Style["width"]);
       Label label = GetLabel (renderingContext);
       ListControl listControl = GetListControl (renderingContext);
+
```
derlistigelurch, line 109:
```suggestion
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,92 @@
...
+  text-decoration: none;
 }
 
-.bocEnumValue.readOnly span[tabindex]:last-child
+.bocEnumValue.readOnly span
 {
-  /* Ensures that diacritic marks can be rendered without being cut off. */
-  padding-bottom: 0.2em;
+  box-sizing: border-box;
+  padding: var(--remotion-themed-padding) 0;
```
derlistigelurch, line 78: I think it's better to not use the shorthand property if the property has more than one value.
```suggestion
  padding-top: var(--remotion-themed-padding);
  padding-bottom: var(--remotion-themed-padding);
  padding-left: 0;
  padding-right: 0;
```

derlistigelurch (changes requested):

derlistigelurch (changes requested):
In person review

derlistigelurch (approved):

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -167,10 +167,13 @@ private ListControl GetListControl (BocEnumValueRenderingContext renderingContex
 
       var isRadioButtonList = renderingContext.Control.ListControlStyle.ControlType == ListControlType.RadioButtonList;
       var isDropDownList = renderingContext.Control.ListControlStyle.ControlType == ListControlType.DropDownList;
+      var isListBox = renderingContext.Control.ListControlStyle.ControlType == ListControlType.ListBox;
```
MichaelKetting, line 170: two things;
this change gets to be be moved into a stand alone bugfix in https://re-motion.atlassian.net/browse/RM-7332 on develop. Here in the branch, we accept that the disabled behavior doesn't work until we have rebased.

And I looked again into the asp.net source code and found a cleaner solution: in `ListControlStyle.Create()`, we create a derived version of ListBox (name: `CompatibleListBox`, nested type in Style where the `SupportsDisabledAttribute` is overridden to return `true`.) this way, things are fixed closer at the root-cause instead of worked-around.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,92 @@
...
-     Using line-height instead of padding works better with BocEnumValue. When using padding, the text is not aligned with the label. */
-  /*line-height: 1.5em;*/
-  /* Ensures that the outline is rendered outside of the space reserved for diacritic marks. */
-  display: inline-block;
+  padding: calc(var(--remotion-themed-padding) * 0.75);
+  border-radius: var(--remotion-themed-border-radius);
 }
 
-.bocEnumValue.readOnly span[tabindex]:first-child
+span.bocEnumValue.disabled option
```
MichaelKetting, line 68: could you please check I'm not sure we actually have to anything for this. at least right now. later on, we may want to update the styles for disabled data entry elements to better match the theme colors.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,92 @@
...
 
-.bocEnumValue td
+div.bocEnumValue input:hover
 {
-  white-space: nowrap;
+  cursor: pointer;
 }
 
-.bocEnumValue td label
+div.bocEnumValue input:disabled
```
MichaelKetting, line 46: this may (or may not?) also be redudant

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,80 @@
 span.bocEnumValue,
```
MichaelKetting, line 3: hmm.. last tweak (I hope). I think we can drop the `div` and `span` selectors from .bocEnumValue in the file. this apears to be quite redundant.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,80 @@
...
 
-.bocEnumValue.disabled
+.bocEnumValue td label
 {
+  padding: var(--remotion-themed-padding);
 }
 
-span.bocEnumValue select
+div.bocEnumValue td,
+div.bocEnumValue td span
```
MichaelKetting, line 22: please split this into two blocks and merge the td blocks. the elements are nested and there's no technical reason (I can see) why these style should always match. They just are the same, right now.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,80 @@
...
-.bocEnumValue td
+div.bocEnumValue input:not([disabled]):hover
 {
-  white-space: nowrap;
+  cursor: pointer;
 }
 
-.bocEnumValue td label
+span.bocEnumValue select:not([disabled]):hover,
+span.bocEnumValue:not(.disabled) option:hover
```
MichaelKetting, line 47: I think this is now redundant? or rather, should be 
```suggestion
.bocEnumValue select:not([disabled]) option:hover
```

also, please slit this up into separate blocks.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -3,76 +3,80 @@
...
 
-.bocEnumValue td
+div.bocEnumValue input:not([disabled]):hover
 {
-  white-space: nowrap;
+  cursor: pointer;
 }
 
-.bocEnumValue td label
+span.bocEnumValue select:not([disabled]):hover,
```
MichaelKetting, line 46: the select hover stuff should be after the normal select stuff.

MichaelKetting (changes requested):
some thoughts in structuring. I would like to do a live demo when it's done.

MichaelKetting (approved):

PR #193, opened 2021-08-31
RM-7332 Setting Enabled to false in BocEnumValue with ControlType ListBox does not render a disabled attribute

FlorianLautner:


Remotion/ObjectBinding/Web/UI/Controls/Styles.cs
```diff
@@ -50,6 +50,11 @@ public enum BocTextBoxMode
     PasswordNoRender
   }
 
+  class CompatibleListBox : ListBox
+  {
+      public override bool SupportsDisabledAttribute => true;
+  }
```
MichaelKetting, line 56:
```suggestion
  private class CompatibleListBox : ListBox
  {
    public override bool SupportsDisabledAttribute => true;
  }
```

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #198, opened 2021-09-06
RM-7926: TreeView styling

iam-mholle:


Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
...
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a:hover > span
 {
+  text-decoration: underline;
 }
 
-ul.treeViewNodeChildren > li,
-ul.treeViewNodeChildrenNoLines > li
+ul.treeViewRoot span.treeViewNode:hover .treeViewNodeHead:before,
+ul.treeViewRoot span.treeViewNode .treeViewNodeHeadSelected:before,
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead:before,
```
MichaelKetting, line 46: focus underline only works for nodes that have an icon.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
...
-ul.treeViewTopLevelNodeChildren > li
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a > span,
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a:hover > span
 {
+  text-decoration: underline;
 }
 
-ul.treeViewNodeChildren > li,
-ul.treeViewNodeChildrenNoLines > li
+ul.treeViewRoot span.treeViewNode:hover .treeViewNodeHead:before,
```
MichaelKetting, line 44: Hover highlight doesn't work starting where the absolute positon is greater than the current viewport.
image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
...
 }
 
-/* Applied to the top level node's children area if the expandre is hidden for the top level nodes. */
-ul.treeViewTopLevelNodeChildren
+ul.treeViewRoot a
 {
-  margin: 0;
-  padding: 0;
-  list-style: none;
+  color: black;
```
MichaelKetting, line 14:
```suggestion
  color: var(--color-standard-text);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
...
-
 ul.treeViewRoot
 {
+  --treeview-chevron-size: 1em;
+  --treeview-chevron-gap: 0.5em;
+  --treeview-padding: calc(var(--treeview-chevron-size) + var(--treeview-chevron-gap));
+
+  box-sizing: border-box;
   margin: 0;
   padding: 0;
```
MichaelKetting, line 9: redundant (see line 23)

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
-/* Style applied to the TreeView. */
-
 ul.treeViewRoot
 {
+  --treeview-chevron-size: 1em;
+  --treeview-chevron-gap: 0.5em;
+  --treeview-padding: calc(var(--treeview-chevron-size) + var(--treeview-chevron-gap));
+
+  box-sizing: border-box;
```
MichaelKetting, line 7: shouldn't we do border-box for the other tree elements, too? Or, otherwise, drop it? (am not sure why it's needed in tree.)

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
...
+ul.treeViewRoot span.treeViewNodeHeadSelected a:focus,
+ul.treeViewRoot li[role=treeitem]
 {
   outline: none;
 }
 
-ul.treeViewRoot > li,
-ul.treeViewTopLevelNodeChildren > li
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a > span,
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a:hover > span
```
MichaelKetting, line 39: I'm not sure I get the difference here. `a:hover` is already covered by `a`, isn't it?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,570 +1,588 @@
...
 }
 
-/* We want the tree node to span the entire line to better handle clicking and styling of the selected line.
-  Since it is not possible to use a nesting-count in CSS, we need to do an explicit nesting for a reasonable depth (50 levels). 
-  Once the maximum depth is reached, the formatting and behavior will revert to nesting via parent padding. */
+/*region Padding calculation*/
 ul.treeViewRoot span.treeViewNode
 {
-  padding-left: calc(19px * 0);
+  padding-left: calc(var(--treeview-padding)  * 0);
```
MichaelKetting, line 84: you've got an extra whitespace in the calc (in all the calcs)

MichaelKetting, line 84: I just had an idea, could we render the nesting level as a css variable on the respective node:
https://stackoverflow.com/questions/20490704/combine-calc-with-attr-in-css

That way, we solve this mess once and for all. When it works, we should also update the NovaGray theme.

MichaelKetting (changes requested):
Note: I need to do another linting run once the feedback is incorporated to completely understand the new css.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
iam-mholle: Tested in current Edge & Firefox, both working. What testsite did you use? (Tested it after incorporating other changes, but they should not interfere)

MichaelKetting: http://localhost:60400/fx-develop/ObjectBinding.Web.Test/SingleTestTreeView.aspx

iam-mholle: Works for me, please test again.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,697 +1,92 @@
...
-ul.treeViewRoot ul ul ul ul ul  ul ul ul ul ul  ul span.treeViewNode
+ul.treeViewRoot .treeViewNode,
+ul.treeViewRoot .treeViewNodeHead > a,
+ul.treeViewRoot .treeViewNodeHeadSelected > a
 {
-  padding-left: calc(19px * 11);
+  display: flex;
+  align-items: center;
+  padding-top: calc(0.5 * var(--remotion-themed-padding));
+  padding-bottom: calc(0.5 * var(--remotion-themed-padding));
```
MichaelKetting, line 70:
```suggestion
  padding-top: var(--remotion-themed-spacing-half);
  padding-bottom: var(--remotion-themed-spacing-half);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,697 +1,92 @@
...
-ul.treeViewRoot li
+ul.treeViewRoot span.treeViewNodeHeadSelected a:focus,
+ul.treeViewRoot li[role=treeitem]
 {
   outline: none;
 }
 
-ul.treeViewRoot > li,
-ul.treeViewTopLevelNodeChildren > li
+ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a > span
```
MichaelKetting, line 38:
```suggestion
ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHead a > span
ul.treeViewRoot li:focus > span.treeViewNode span.treeViewNodeHeadSelected a > span
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,697 +1,92 @@
...
-{
-  padding-left: calc(19px * 10);
+  bottom: 0;
+  z-index: -1;
 }
 
-ul.treeViewRoot ul ul ul ul ul  ul ul ul ul ul  ul span.treeViewNode
+ul.treeViewRoot .treeViewNode,
+ul.treeViewRoot .treeViewNodeHead > a,
+ul.treeViewRoot .treeViewNodeHeadSelected > a
```
MichaelKetting, line 65: Please make this consistent. You're using `span.treeViewNode` in most of the rest of the file.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,697 +1,92 @@
...
-ul.treeViewRoot ul ul ul ul ul  ul ul ul ul span.treeViewNode
-{
-  padding-left: calc(19px * 9);
-}
-
-ul.treeViewRoot ul ul ul ul ul  ul ul ul ul ul span.treeViewNode
-{
-  padding-left: calc(19px * 10);
+  bottom: 0;
+  z-index: -1;
```
MichaelKetting, line 60:
```suggestion
  z-index: var(--z-index-hidden);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,697 +1,92 @@
...
 }
 
-ul.treeViewRoot ul ul ul ul ul  ul ul ul ul ul  ul span.treeViewNode
+ul.treeViewRoot .treeViewNode,
+ul.treeViewRoot .treeViewNodeHead > a,
```
MichaelKetting, line 64: This here described two separate nodes that just happen to have the same css, correct? In that case, please split it up. Otherwise, please elaborate why parent and child are guaranteed to always require the same css.

Same topic, other question: why do we need the half-spacing on the anchor in addtion to a separate half-spacing on the treeViewNode. Couldn't we just do a full spacing on the treeViewNode?

And I couldn't observe an effect with display and align-items for Anchor, but this may be accidental due to the currently rendered text.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -686,6 +687,7 @@ private void RenderNodes (HtmlTextWriter writer, WebTreeNodeCollection nodes, bo
         writer.AddStyleAttribute ("white-space", "nowrap");
```
MichaelKetting, line 687: The wrap-option doesn't work in the new CSS. The idea is, if the node is too long, it should wrap. It's okay if the nowrap becomes a css class on the treeview's root and is then applied via css file instead of via markup.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
iam-mholle: > And I couldn't observe an effect with display and align-items for Anchor, but this may be accidental due to the currently rendered text.

This makes sure that the text and the icon, and the chevron are aligned.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
iam-mholle: Please elaborate, when should it wrap?

MichaelKetting: image
You set a width on the tree, and the treeview will wrap the individual node instead of showing a scrollbar. It will show a horizontal scrollbar, when the node depth exceeds the width.

iam-mholle: Please have a look, I think I got it right.

MichaelKetting (changes requested):
Wordwrap Doesn't work consistently
image

image

iam-mholle:
Yes, if there is a scrollbar, the nodes wrap after hitting the new width. Should it rather behave like this? 
image

MichaelKetting:
> 
> 
> Yes, if there is a scrollbar, the nodes wrap after hitting the new width. Should it rather behave like this?

Ah, okay, now I get it (maybe). Thing is, the scroll bit's always been kind of weird when it comes to wrapping the text while being able to have a scrollbar when the nesting requires it. The screenshot documents the existing behavior so it's certainly the behavior that's the least suprising. If that's easy to do, I would prefer it for now and perhaps later consider better alternatives (e.g. setting the viewport-width as max-width for each invidiual node). Thoughts?

iam-mholle:
> > Yes, if there is a scrollbar, the nodes wrap after hitting the new width. Should it rather behave like this?
> 
> Ah, okay, now I get it (maybe). Thing is, the scroll bit's always been kind of weird when it comes to wrapping the text while being able to have a scrollbar when the nesting requires it. The screenshot documents the existing behavior so it's certainly the behavior that's the least suprising. If that's easy to do, I would prefer it for now and perhaps later consider better alternatives (e.g. setting the viewport-width as max-width for each invidiual node). Thoughts?

Wouldn't it be better to sort this out right away with a UX-savy person? E.g. the next designer meeting?

MichaelKetting:
> Wouldn't it be better to sort this out right away with a UX-savy person? E.g. the next designer meeting?

In theorey, yes. But there's two considerations: we can't do this in the current task due to scheduling. We need to do this based on live data downstream. So, for now the easiest route is to stick with the status quo instead of changing the existing behavior (can't go wrong there) and yes, I'll put it on the list for discussion-topics.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -1,697 +1,112 @@
...
-  padding: 1px;
-  /* Synchronized with focus/hover */
-  padding-top: 2px;
-  padding-bottom: 2px;
-
-  border: 1px solid #569CE0;
-  background-color: #C5E0FC;
+  --current-padding: calc(var(--nesting-depth) * var(--treeview-padding));
+  --chevron-size: calc(var(--treeview-chevron-size) + var(--treeview-chevron-gap));
+  width: calc(var(--treeview-width) - var(--current-padding) - var(--chevron-size));
```
MichaelKetting, line 90: Hmm...the standard usecase is to set the --treeview-width to 100% so I don't think we actually need to have a variable for this. Reasoning: 100% is evaluated here based on the parent and unless I'm mistaken, using calc(100%) here will either never work or always work. Either way, we should be able to drop the css variable for treeview-width. Is that correct?

MichaelKetting (changes requested):
See comment

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
iam-mholle: The point of the variable is to limit the width of the text node, resulting in the word wrapping behavior from before. But I see your point, if the width is set to something like 100%, or limited by the parent width without an explicit width set, this calculation would not work. It was the only solution I found to restore the old wrapping behavior, while working with the new "highlight the whole line" design.

MichaelKetting: Meh, okay. Then we skip this bit, just add the nowordwrap css class to the root and apply disable the wordwrap. What's the behavior then? It will break at the end of the longest line, correct? Please create a bug ticket with "wordwrap behavior is undefined" and we already have an open question for this anyway so we keep it simple for now.

iam-mholle: > What's the behavior then? It will break at the end of the longest line, correct?

yes. I'll add a ticket.

MichaelKetting (approved):

PR #201, opened 2021-09-08
RM-7924 Create styles for new theme for FormGrid

FlorianLautner:


Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
...
-  border-top: none;
-  border-left: none;
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  border-color: White;
+  padding: var(--remotion-themed-spacing) 0;
+  border-bottom: var(--remotion-themed-grid-border);
 }
 
 /*  Style applied to the cells in the first row after the title row  */
```
derlistigelurch, line 71: I don't think any of these comments are needed anymore.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
...
-  color: Red;
+  color: var(--color-indicator-error);
 }
 
 /*  Style applied to the validation notice that can be displayed by the ValidationStateViewer  */
 .formGridValidationNotice
 {
-  color: Red;
+  color: var(--color-indicator-error);
 }
```
derlistigelurch, line 93: I think it's best if you merge these rules.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -116,23 +106,21 @@ td.formGridTopDataRow
   transition-duration: 0;
   transition-delay: 0;
```
derlistigelurch, line 107:
```suggestion
  transition-duration: 0s;
  transition-delay: 0s;
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
...
-  border-left: none;
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  width: 17%;
+  box-sizing: border-box;
+  width: 5%;
+  padding-top: var(--remotion-themed-spacing);
+  padding-right: var(--remotion-themed-spacing);
+  padding-bottom: var(--remotion-themed-spacing);
+  padding-left: 0;
```
derlistigelurch, line 25: You could write it that way, since all properties except `padding-left` have the same value.

```suggestion
  padding: var(--remotion-themed-spacing);
  padding-left: 0;
```

derlistigelurch (changes requested):

derlistigelurch (approved):

Remotion/Web/Core/UI/Controls/FormGridManager.cs
```diff
@@ -2345,6 +2330,21 @@ protected virtual void LoadMarkersIntoCell (FormGridRow dataRow)
         else
           dataRow.MarkersCell.Controls.Add(CreateBlankMarker());
       }
+
+      //  ValidationMarker and RequiredMarker share right-hand position
```
MichaelKetting, line 2334:
```suggestion
      //  ValidationMarker and RequiredMarker share right-hand position
      //  ValidationMarker takes precedence      
```

Remotion/Web/Core/UI/Controls/FormGridManager.cs
```diff
@@ -2314,23 +2314,8 @@ protected virtual void LoadMarkersIntoCell (FormGridRow dataRow)
...
-        dataRow.MarkersCell.Controls.Add(dataRow.RequiredMarker);
-      }
-      else if (ShowValidationMarkers || ShowRequiredMarkers)
-      {
-        dataRow.MarkersCell.Controls.Add(CreateBlankMarker());
-      }
-
-      //  HelpProvider takes right-hand side in column
+      //  HelpProvider takes left-hand side in column
+      //  HelpProvider takes precedence
```
MichaelKetting, line 2318:
```suggestion
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
 /*  Style applied to the form grid table's outer structure  */
-table.formGridTable
+.formGridTable
 {
   border-collapse: collapse;
-  border-width: 0;
   width: 100%;
+  background-color: var(--color-standard-background);
```
MichaelKetting, line 6:
```suggestion
```
I don't think there should be a background color here.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
 }
 
-td.formGridTitleCell,
-td.formGridSubTitleCell
+.formGridTitleCell,
+.formGridSubTitleCell
 {
   font-weight: bold;
```
MichaelKetting, line 12: I've looked around and I think I found a good example for how the titlecell should look:
image

Also, we need a configurable padding, settable via a css variable so the formgrid can be displayed with and without a padding because we can't do the gray background otherwise. The default is probably going to be padding, but it can be overridden to zero via CSS

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
   font-size: 125%;
```
MichaelKetting, line 13: font-weight will be done via font (Roboto-bold).
font-size will be different for title and subtitle. We also want to do a remotion-themed-heading-fontsize and remotion-themed-subheading-fontsize for this. Please check with @derlistigelurch for the names. He's doing the fonts.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
 {
-  background-color: #F0F0F0;
-  text-align: right;
-  padding: 0.3em 1em 0.3em 1em;
-  border-top: none;
-  border-left: none;
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  width: 17%;
+  box-sizing: border-box;
```
MichaelKetting, line 20: Here, we should not do box-sizing. This is too much of a table to really benefit.
```suggestion
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
+  width: 5%;
+  padding-top: var(--remotion-themed-spacing);
+  padding-right: var(--remotion-themed-spacing);
+  padding-bottom: var(--remotion-themed-spacing);
+  padding-left: 0;
+  border-bottom: var(--remotion-themed-grid-border);
+  font-weight: bold;
```
MichaelKetting, line 27: This will be done via font-face

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
-  background-color: #F0F0F0;
-  text-align: right;
-  padding: 0.3em 1em 0.3em 1em;
-  border-top: none;
-  border-left: none;
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  width: 17%;
+  box-sizing: border-box;
+  width: 5%;
```
MichaelKetting, line 21: why width 5%

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
+  padding-top: var(--remotion-themed-spacing);
+  padding-right: var(--remotion-themed-spacing);
+  padding-bottom: var(--remotion-themed-spacing);
+  padding-left: 0;
```
MichaelKetting, line 25: We need this to be smarter. It's possible to have extra columns to the left of the labels columns, it's just never used. So, we shold condition this to it being the first TD in the TR, otherwise the padding is set to spacing.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,84 @@
...
-  color: Red;
-}
-
-/*  Style applied to the validation notice that can be displayed by the ValidationStateViewer  */
+.formGridValidationMessage,
+.formGridValidator,
 .formGridValidationNotice
 {
-  color: Red;
+  color: var(--color-indicator-error);
```
MichaelKetting, line 81: please see update in ticket regarding color.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -113,26 +92,24 @@ td.formGridTopDataRow
...
+  border: var(--remotion-themed-alert-border);
+  background-color: var(--color-alert-background);
 }
 
 .formGridMarkersCell a[aria-label]:not([title]):focus::after,
 .formGridMarkersCell a[aria-label]:not([title]):hover::after
 {
   visibility: visible;
-  transition-delay: 0.7s;
+  transition-delay: var(--transition-delay);
```
MichaelKetting, line 114: Better name please. maybe --transition-delay-tooltip?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -113,26 +92,24 @@ td.formGridTopDataRow
...
-  padding: 0.25em;
-  border: 1px solid Black;
-  background-color: Lightyellow;
-  box-sizing: border-box;
+  /* popup spawns at the bottom right corner of the icon's hover highlighting */
+  top: calc(var(--default-icon-size) + var(--remotion-themed-spacing-half));
+  z-index: var(--z-index-max);
+  padding: var(--remotion-themed-spacing-half);
+  border: var(--remotion-themed-alert-border);
+  background-color: var(--color-alert-background);
```
MichaelKetting, line 107: we should make those "tooltip" instead of "alert". And I just noticed that firefox and chrome on WIndows 10 use the same tooltip style. please emulate these colors (font, border, background) for tooltips.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -113,26 +92,24 @@ td.formGridTopDataRow
...
   position: absolute;
-  left: 0;
-  top: 1.2em;
-  z-index: 1000;
-  padding: 0.25em;
-  border: 1px solid Black;
-  background-color: Lightyellow;
-  box-sizing: border-box;
+  /* popup spawns at the bottom right corner of the icon's hover highlighting */
+  top: calc(var(--default-icon-size) + var(--remotion-themed-spacing-half));
```
MichaelKetting, line 103: Okay, this is because the top starts at the icon's top and then we have to push it down an exrtra half spacing?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
MichaelKetting, line 107: we also need to set the text color.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -26,6 +26,7 @@
   --remotion-themed-border-radius: 2px;
   --remotion-themed-border-width: 1px;
   --remotion-themed-border: var(--remotion-themed-border-width) solid var(--color-standard-border);
+  --remotion-themed-grid-border: var(--remotion-themed-border-width) solid var(--color-standard-separator);
```
MichaelKetting, line 29: Please move this into the formgrid.css and make a formgrid-specific name. See BocList-branch. It's a setting that's just relevant to FormGrid.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
FlorianLautner: I solved it by setting `padding-right` for each `td` that is not the `:last-child` and then added `padding-left` if it's not `:first-child` (or `:only-child`, such as `formGridTitleCell`, for example). This way, each `td` gets both `padding-left` and `padding-right` except for the first `td` (only `padding-left`) and the last `td` (only `padding-right`). I also set the value to `--remotion-themed-spacing-half` instead of `--remotion-themed-spacing` so that the `padding-left` and `padding-right` of two neighboring `td`s aggregate to a "full" padding, thus avoiding that the space between two `td`s doesn't get too wide.

MichaelKetting: sounds reasonable.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
...
 }
 
-td.formGridMarkersCell img
+.formGridSubTitleCell
 {
-  min-width: 12px;
-  border: none;
-  height: 16px;
-  width: 16px;
+  font-size: var(--remotion-themed-subheading-fontsize);
```
MichaelKetting, line 37: those get just `font` set to `--remotion-thrmed-subheading-font`

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
...
-  padding: 0.3em 1em 0.3em 0em;
-  border-top: none;
-  border-left: none;
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  width: auto;
+  width: 15%;
+  padding: var(--remotion-themed-spacing) 0;
+  border-bottom: var(--formgrid-border);
+  font-family: var(--remotion-themed-font-bold);
```
MichaelKetting, line 45:
```suggestion
  font: var(--remotion-themed-font-bold);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -113,26 +103,26 @@ td.formGridTopDataRow
...
-  border: 1px solid Black;
-  background-color: Lightyellow;
-  box-sizing: border-box;
+  /* popup spawns at the bottom right corner of the icon's hover highlighting */
+  top: calc(var(--default-icon-size) + var(--remotion-themed-spacing-half));
+  padding: var(--remotion-themed-spacing) calc(var(--remotion-themed-spacing) + var(--remotion-themed-spacing-half));
+  border: var(--remotion-themed-tooltip-border);
+  color: var(--color-tooltip-text);
+  background-color: var(--color-tooltip-background);
+  font-size: 90%;
```
MichaelKetting, line 119: --font-tooltip (or --tooltip-font etc, please chose based on the naming rules we have)

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
 /*  Style applied to the form grid table's outer structure  */
-table.formGridTable
+.formGridTable
```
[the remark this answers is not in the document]
FlorianLautner: With `:root`, `--formgrid-border` is now globally accessible, though. Not that it poses a problem in this case, I assume, but do we really want that? Or put differently, if we make it globally accessible with `:root`, why not put it into `Common.css` in the first place, then it is at least visible/obvious to everyone?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,95 @@
...
 {
   border-collapse: collapse;
-  border-width: 0;
   width: 100%;
 }
 
-td.formGridTitleCell,
-td.formGridSubTitleCell
+.formGridTable tr td:not(:last-child, :only-child)
```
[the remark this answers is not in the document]
FlorianLautner: Using the actual classes is difficult, because if there are one or more columns added (to the left of `formGridLabelsCell`, for example) I cannot possibly tell at this point which classes these `td`s will have. If these cells do have one of the given classes (`formGridLabelsCell`, `formGridMarkersCell`, `formGridControlsCell`), though, then it wouldn't be a problem, of course.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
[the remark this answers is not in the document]
MichaelKetting: @a-ctor thoughts?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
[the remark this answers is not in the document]
MichaelKetting: I'm not sure I completely get the problem. Maybe it's based on syntax: 

`td.formGridLabelsCell:first-child` does not apply the first TD with the specified css class, but rather to the first TD in the row if it also has the specified css class. So, we should be okay?

image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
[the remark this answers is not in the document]
a-ctor: I think both solutions are acceptable. Having it in the same file makes development easier as you quickly know what relevant variables are available to you. On the other hand, it is also beneficial to a have all of the "public API" declared in the same file.

I personally prefer the variables in the relevant files because if you want to style a FormGrid, for example, you go to `FormGrid.css` and look at the top lines. In the `common.css` you have to pick the relevant styles out of 100 variable declarations.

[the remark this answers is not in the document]
MichaelKetting: okay, thanks and I tend towards the same reason for placing them in the same file.

 @a-ctor, the other question is regarding the variable being available as "public" vs. just within the formgrid.
 Hmm... maybe also get @iam-mholle  opinon.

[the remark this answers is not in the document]
iam-mholle: 👍🏼 for the same file, 👍🏼 for `.formGridTable` instead of `:root` - why make it globally useable if its specific to this control?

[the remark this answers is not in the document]
a-ctor: If we do not want anyone to change a variable it might be better to put it on the control class. But all the classes that have a chance of being overriden should be on `:root` and not the form control class. Otherwise it makes overriding the variable more difficult than it needs to be.

I would prefer adding all variables to `:root` because this allows them to be overridden without much drama, but there is the problem of differentiating public from private variables. If we found a good way to fix this I would go with all vars on `:root`. Maybe commenting `public` and `private` regions is enough? Or a prefix for private variables?

```
:root
{
  /* public */
  --my-public-var: ... ;

  /* private */
  --my-private-var: ... ;
}
```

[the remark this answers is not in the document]
a-ctor: Another solution would be to declare private variables in the control class but do so in a way that keeps the original value if set:

```
.my-control
{
  --my-private-var = var(--my-private-var, default if not set);
}
```

This way, the variable is not declared as public API but can be overriden if needed as if it was declared in `:root`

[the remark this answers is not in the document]
MichaelKetting: Okay, I think I've now a good overview and we should move this discussion outside of the current scope. There's three four places where we have some variables (FormGrid, SingleView/MultiView, and BocList). A few of those variables are extensionpoints and aside from SingleView/MultiView we have them inside the css-class at this time. I would suggest using the css-class for now (@a-ctor that's RM-7923) and discuss this in a design session regarding further improvements. Otherwise, we might blow the scope.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,103 @@
...
-/*  Style applied to the validation notice that can be displayed by the ValidationStateViewer  */
+.formGridTopDataRow
+{
+  border-top: var(--formgrid-border);
+}
+
+.formGridValidationMessage,
+.formGridValidator,
 .formGridValidationNotice
 {
```
MichaelKetting, line 99:
```suggestion
{
  padding-top: var(--remotion-themed-spacing);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,103 @@
...
 
-/*  Style applied to the cells in the first row after the title row  */
-td.formGridTopDataRow
+.formGridMarkersCell img
 {
-  border-top: solid 1px #AAAAAA;
+  height: var(--default-icon-size);
+  width: var(--default-icon-size);
+  padding: var(--remotion-themed-spacing-half);
+  vertical-align: middle;
```
MichaelKetting, line 76:
```suggestion
```
please check, I'm not sure this style actually does somehting.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,103 @@
...
 
-/*  Style applied to the cells containing the controls  */
-td.formGridControlsCell
+.formGridLabelsCell:first-child
+{
+  padding-left: 0;
+}
+
+.formGridMarkersCell
+{
```
MichaelKetting, line 49:
```suggestion
{
  padding-top: calc(var(--remotion-themed-line-height) / 2 - var(--remotion-themed-spacing-half) / 2);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,103 @@
...
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  width: 1%;
+  padding: var(--remotion-themed-spacing) var(--remotion-themed-spacing-half);
+  border-bottom: var(--formgrid-border);
 }
 
-td.formGridMarkersCell img
+.formGridLabelsCell
 {
```
MichaelKetting, line 38:
```suggestion
{
  padding-top: calc(var(--remotion-themed-line-height) / 2 + var(--remotion-themed-spacing-half));
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/FormGrid.css
```diff
@@ -1,105 +1,103 @@
...
-  white-space: nowrap;
-  padding: 0.3em 0.5em 0.3em 0.5em;
-  border-top: none;
-  border-left: none;
-  border-right: none;
-  border-bottom: solid 1px #C8C8C8;
-  width: 1%;
+  padding: var(--remotion-themed-spacing) var(--remotion-themed-spacing-half);
+  border-bottom: var(--formgrid-border);
 }
```
MichaelKetting, line 35:
```suggestion
  vertical-align: top;
}
```

MichaelKetting (changes requested):
Needed to adjust the veritical alignment and before there's any questions: no, it's not nice but it gets the job done. The problem is that we do know where the first line starts when using controls based on our theme but the labels, we can only asume to be text that needs to get pushed down. So, yeah, the used formula works after testing it and I don't have a better solution at this time.

MichaelKetting (approved):

PR #202, opened 2021-09-09
Feature/rm 7928 support configuring button type web button

a-ctor:
https://re-motion.atlassian.net/browse/RM-7928

### DIFF
--- a/Remotion/Web/Contracts.DiagnosticMetadata/DiagnosticMetadataAttributes.cs
+++ b/Remotion/Web/Contracts.DiagnosticMetadata/DiagnosticMetadataAttributes.cs
@@ -33,6 +33,7 @@ namespace Remotion.Web.Contracts.DiagnosticMetadata
     public static readonly string TriggersNavigation = "data-triggers-navigation";
     public static readonly string TriggersPostBack = "data-triggers-postback";
 
+    public static readonly string ButtonType = "data-button-type";
 
     public static readonly string WebTreeViewBadgeValue = "data-webtreeview-badge-value";
     public static readonly string WebTreeViewBadgeDescription = "data-webtreeview-badge-description";
--- a/Remotion/Web/Core/UI/Controls/ButtonType.cs
+++ b/Remotion/Web/Core/UI/Controls/ButtonType.cs
@@ -1,0 +1,23 @@
+﻿namespace Remotion.Web.UI.Controls
+{
+  /// <summary>
+  /// Defines how a button is displayed on the page.
+  /// </summary>
+  public enum ButtonType
+  {
+    /// <summary>
+    /// The button is displayed using the standard style.
+    /// </summary>
+    Standard,
+
+    /// <summary>
+    /// The button is displayed using a style indicating a primary action.
+    /// </summary>
+    Primary,
+
+    /// <summary>
+    /// The button is displayed using a style indicating a supplemental action.
+    /// </summary>
+    Supplemental
+  }
+}
--- a/Remotion/Web/Core/UI/Controls/WebButton.cs
+++ b/Remotion/Web/Core/UI/Controls/WebButton.cs
@@ -62,6 +62,7 @@ namespace Remotion.Web.UI.Controls
     private bool _hasPagePreRenderCompleted;
 
     private TextWithHotkey _textWithHotkey;
+    private ButtonType _buttonType;
 
     public WebButton ()
     {
@@ -204,7 +205,13 @@ namespace Remotion.Web.UI.Controls
       var originalCssClass = (CssClass ?? "").Replace (DisabledCssClass, "").Trim();
       var isCssStyleOverridden = !string.IsNullOrEmpty (originalCssClass);
       var computedCssClass = isCssStyleOverridden ? originalCssClass : CssClassBase;
+      if (ButtonType == ButtonType.Primary)
+        computedCssClass += " " + CssClassPrimary;
+      else if (ButtonType == ButtonType.Supplemental)
+        computedCssClass += " " + CssClassSupplemental;
+
       computedCssClass += " " + CssClassThemed;
+
       ControlStyle.CssClass = computedCssClass;
 
       base.AddAttributesToRender (writer);
@@ -236,6 +243,8 @@ namespace Remotion.Web.UI.Controls
 
       var triggersPostBack = !OnClientClick.Contains (";return ");
       writer.AddAttribute (DiagnosticMetadataAttributes.TriggersPostBack, triggersPostBack.ToString().ToLower());
+
+      writer.AddAttribute (DiagnosticMetadataAttributes.ButtonType, ButtonType.ToString());
     }
 
     protected override PostBackOptions GetPostBackOptions ()
@@ -346,6 +355,21 @@ namespace Remotion.Web.UI.Controls
 
       if (Icon != null)
         Icon.LoadResources (resourceManager);
+    }
+
+    /// <summary>
+    /// Gets or sets the button type that determines how the button is displayed on the page.
+    /// </summary>
+    /// <remarks>
+    /// Depending on the button types the css classes <see cref="CssClassPrimary"/> or <see cref="CssClassSupplemental"/> will be applied to the button.
+    /// </remarks>
+    [Description ("Determines how the button is displayed on the page.")]
+    [Category ("Appearance")]
+    [DefaultValue (UI.Controls.ButtonType.Standard)]
+    public ButtonType ButtonType
+    {
+      get { return _buttonType; }
+      set { _buttonType = value; }
     }
 
     /// <summary> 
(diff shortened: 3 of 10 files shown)


Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -38,14 +38,22 @@ button.webButton.primary
...
+button.webButton.primary:not(:disabled):hover,
+button.webButton.primary:not(:disabled):focus
 {
   background-color: var(--color-primary-highlight);
 }
 
 button.webButton.secondary
 {
+  margin: var(--remotion-themed-padding);
+  padding: 0 var(--remotion-themed-padding);
```
MichaelKetting, line 50:
```suggestion
  margin: var(--remotion-themed-spacing);
  padding: 0 var(--remotion-themed-spacing);
```

Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -205,6 +206,11 @@ protected override void AddAttributesToRender (HtmlTextWriter writer)
       var isCssStyleOverridden = !string.IsNullOrEmpty (originalCssClass);
       var computedCssClass = isCssStyleOverridden ? originalCssClass : CssClassBase;
       computedCssClass += " " + CssClassThemed;
+      if (ButtonType == WebButtonType.Primary)
+        computedCssClass += " " + CssClassPrimary;
+      else if (ButtonType == WebButtonType.Secondary)
+        computedCssClass += " " + CssClassSecondary;
```
MichaelKetting, line 212:
```suggestion
      if (ButtonType == WebButtonType.Primary)
        computedCssClass += " " + CssClassPrimary;
      else if (ButtonType == WebButtonType.Secondary)
        computedCssClass += " " + CssClassSecondary;
      computedCssClass += " " + CssClassThemed;
```

Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -348,6 +354,18 @@ protected virtual void LoadResources (IResourceManager resourceManager)
         Icon.LoadResources (resourceManager);
     }
 
+    /// <summary>
+    /// Gets or sets the button type that determines how the button is displayed on the page.
+    /// </summary>
+    /// <remarks>
+    /// Depending on the button types the css classes <see cref="CssClassPrimary"/> or <see cref="CssClassSecondary"/> will be applied to the button.
+    /// </remarks>
+    public WebButtonType ButtonType
```
MichaelKetting, line 363:
```suggestion
    [Description ("Determines how the button is displayed on the page.")]
    [Category ("Appearance")]
    [DefaultValue (WebButtonType.Standard)]
    public WebButtonType ButtonType
```

Remotion/Web/Core/UI/Controls/WebButtonType.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Remotion.Web.UI.Controls
+{
+  /// <summary>
+  /// Defines how a <see cref="WebButton"/> is displayed on the page.
+  /// </summary>
+  public enum WebButtonType
```
MichaelKetting, line 6: Sorry, I just realized that we want to resue the ButtonType also in DropDownMenu, so we should keep using `ButtonType` as originally specified instead of `WebButtonType`.

Remotion/Web/Core/UI/Controls/WebButtonType.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    Standard,
+
+    /// <summary>
+    /// The <see cref="WebButton"/> is displayed using a style indicating a primary action.
+    /// </summary>
+    Primary,
+
+    /// <summary>
+    /// The <see cref="WebButton"/> is displayed using a style indicating a secondary action.
```
MichaelKetting, line 19: Given that `Standard` is used next to `Primary` on the page, maybe we should get a better name for `Secondary`. Problem is, same as during the design of this task, I don't have a good name. `Extra`, `Additonal`, or `Other`. Ideas / Thoughts?

Remotion/Web/Development.WebTesting.ControlObjects/WebButtonControlObject.cs
```diff
@@ -50,6 +53,21 @@ public string GetText ()
       return Scope.Text.Trim();
     }
 
+    /// <summary>
+    /// Returns the button type of the button.
+    /// </summary>
+    public WebButtonType GetButtonType ()
```
MichaelKetting, line 59: Please switch to data annotion for this. by using the css class we take a dependency and would actually break if someone where to indeed override the property on the WebButton.

MichaelKetting (changes requested):

Remotion/Web/Core/UI/Controls/WebButtonType.cs
a-ctor: Bootstrap calls it a `link` button since it does not have a button outline

Remotion/Web/Development.WebTesting.ControlObjects/WebButtonControlObject.cs
a-ctor: The rendering as well? Or add a data annotation in addition to the css class.

Styling is a lot more readable and easier if it is done with a css class so I would prefer not to remove it.

Remotion/Web/Core/UI/Controls/WebButtonType.cs
MichaelKetting: yeah, I can see why, but that doesn't make it better when used with the dropdownmenu. It also makes an assumption on the look&feel which we shouldn't because the no-outline-thing could change in a future theme.

Remotion/Web/Development.WebTesting.ControlObjects/WebButtonControlObject.cs
MichaelKetting: In addition. there's the css class for styling and the data-annation that's rendered when the diag-features are turned on.

Remotion/Web/Core/UI/Controls/WebButtonType.cs
a-ctor: `supplemental` it is

Remotion/Web/Contracts.DiagnosticMetadata/DiagnosticMetadataAttributes.cs
```diff
@@ -34,6 +34,8 @@ public static class DiagnosticMetadataAttributes
     public static readonly string TriggersPostBack = "data-triggers-postback";
 
 
+    public static readonly string WebButtonButtonType = "data-webbutton-button-type";
+
```
MichaelKetting, line 38:
```suggestion
```suggestion
    public static readonly string ButtonType = "data-button-type";


```
We can simplify this.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -38,14 +38,22 @@ button.webButton.primary
...
+button.webButton.primary:not(:disabled):hover,
+button.webButton.primary:not(:disabled):focus
 {
   background-color: var(--color-primary-highlight);
 }
 
-button.webButton.secondary
+button.webButton.supplemental
 {
+  margin: var(--remotion-themed-spacing);
```
MichaelKetting, line 49: why do we need margin here. can we make this to work just like the other button types, only without border?

Remotion/Web/Core/UI/Controls/ButtonType.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Remotion.Web.UI.Controls
+{
+  /// <summary>
+  /// Defines how a <see cref="WebButton"/> is displayed on the page.
```
MichaelKetting, line 4: Do you want to just add `DropDownMenu` in the next commit or update the description to not include the type name? maybe just "button face"?

Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -547,6 +571,30 @@ public virtual string CssClassThemed
...
+    /// <summary>
+    /// Gets the CSS-Class applied to the <c>div</c> when the <see cref="ButtonType"/> is set to <see cref="Controls.ButtonType"/>.<see cref="Controls.ButtonType.Primary"/>.
+    /// </summary>
+    /// <remarks> 
+    ///   <para> Class: <c>primary</c>. </para>
+    ///   <para> Applied in addition to the regular CSS-Class.</para>
+    /// </remarks>
+    public virtual string CssClassPrimary
+    {
+      get { return "primary"; }
```
MichaelKetting, line 583: Should we move those to CssClassDefinition as `ButtonTypePrimary` and `ButtonTypeSupplemental`?

Remotion/Web/Development.WebTesting.ControlObjects/ButtonType.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Remotion.Web.Development.WebTesting.ControlObjects
+{
+  /// <summary>
+  /// Defines how a <see cref="WebButtonControlObject"/> is displayed on the page.
```
MichaelKetting, line 4: Update during the DropDownMenu commit or make it more generic now?

Remotion/Web/Development.WebTesting.ControlObjects/WebButtonControlObject.cs
```diff
@@ -27,6 +28,9 @@ namespace Remotion.Web.Development.WebTesting.ControlObjects
   public class WebButtonControlObject
       : WebFormsControlObjectWithDiagnosticMetadata, IClickableControlObject, IControlObjectWithText, IStyledControlObject, ISupportsDisabledState
   {
+    private const string c_primaryButtonTypeCssClass = "primary";
+    private const string c_supplementalButtonTypeCssClass = "supplemental";
+
```
MichaelKetting, line 33:
```suggestion
```

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
a-ctor: Isn't the supplemental button supposed to be smaller than the other buttons?

MichaelKetting: No, it's just borderless. The hover highlight etc may still use the full size of the button.

Remotion/Web/Contracts.DiagnosticMetadata/DiagnosticMetadataAttributes.cs
```diff
@@ -34,6 +34,8 @@ public static class DiagnosticMetadataAttributes
     public static readonly string TriggersPostBack = "data-triggers-postback";
```
MichaelKetting, line 36:
```suggestion
```

MichaelKetting (approved):
Please fix the whitespace issue, then it looks done.

PR #205, opened 2021-09-13
RM-7992 Add new font

derlistigelurch:
https://re-motion.atlassian.net/browse/RM-7992


Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -48,6 +48,29 @@
...
+
+@font-face {
+  font-family: 'Roboto-Regular';
+  src: url('../../../Html/Roboto-Regular.ttf') format('truetype');
+}
+
+*
+{
+  font-family: var(--remotion-themed-font-standard);
 }
```
a-ctor, line 74: It is sad that we have to do it like this because now fonts are no longer inherited and need to be applied to every control. Maybe it would be better to apply it only to `body` and reset all the pre-defined browser styles that explicitly set a `font-family`, but this is a lot more work ^^

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -68,6 +68,7 @@ span.DropDownMenuSelect a.DropDownMenuButton
 span.DropDownMenuSelect:not(.disabled) a.DropDownMenuLabel
 {
   color: var(--color-standard-text);
+  font-family: var(--remotion-themed-font-bold);
```
a-ctor, line 71: Because of the `:not(...)` selector the font is not applied to disabled controls

a-ctor (changes requested):
Based on the issue description we are still missing the font declarations for `FormGrid` and `TabbedMenu`.

I guess they we left out because there are not on the new theme yet, but we should either do them here anyway or add a reminder in the ticket to make sure it does not get lost.

SecurityManager/Clients.Web.Test/Global.asax.cs
```diff
@@ -58,6 +58,9 @@ protected void Application_Start (object sender, EventArgs e)
           {
               new ResourcePathMapping ("Remotion.Web/Html/jquery-1.6.4.js", @"..\..\Remotion\Web\Core\res\Html\jquery-1.6.4.js"),
               new ResourcePathMapping ("Remotion.Web/Html/jquery.iFrameShim.js", @"..\..\Remotion\Web\Core\res\Html\jquery.iFrameShim.js"),
+              new ResourcePathMapping ("Remotion.Web/Html/Roboto-Regular.ttf", @"..\..\Web\Core\res\Html\Roboto-Regular.ttf"),
+              new ResourcePathMapping ("Remotion.Web/Html/Roboto-Medium.ttf", @"..\..\Web\Core\res\Html\Roboto-Medium.ttf"),
+              new ResourcePathMapping ("Remotion.Web/Html/Roboto-Bold.ttf", @"..\..\Web\Core\res\Html\Roboto-Bold.ttf"),
```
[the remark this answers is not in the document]
MichaelKetting: Side note: we really need to consider extending ResourcePathMapping to support merging two directories so we don't have to go on listing a ton of files: https://re-motion.atlassian.net/browse/RM-8037

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
derlistigelurch: What would be the benefit of doing it this way, because `*` already selects all elements on the site and changes the font.🤔

a-ctor: If you want to change the font you have to child elements because of the `*` selector. So if you have a `<code>` block that should have another font and you want, for example, to use an `<b>` element inside your CSS needs to explicitly handle the `<b>` tag because the `*` rule overrides the font again.

I don't actually know if this is an important thing for us or if it can stay this way (cc @MichaelKetting)

MichaelKetting: we don't want a * rule here. it may be valid, that we declare the css variables in Common.css and then apply the standard-font via style.css. This does not solve the question how we deal with the individual controls. I don't hink this can be solved via review-ping-pong?

derlistigelurch: In `style.css` there is already a rule for the `font-family`.
```css
body, input, select, textarea, button
{
  font-family: 'Arial', 'Helvetica', sans-serif;
}
```
An easy solution would be to use the same selectors in the `common.css` rule.

derlistigelurch: How was the issue with the controls handled in before?

MichaelKetting: before, we just said that by importing style.css, we globally define the fonts. this decision (importing style.css) was done by the application developer, not by the controls library author. so, what I want is that just because we add common.css, this should not globally set the font. globally setting the font may only be done via style.css.

maybe the easiest way would be if we define the three font-variables in common.css without specifiying the font, setting the value of the font-variables in style.css and using the font-variables for bold and medium at those places where we need to use medium or bold font.

You can include the weight in the generic font, see here: https://stackoverflow.com/questions/4218549/one-css-declaration-for-all-css-font-properties
`font: bold italic 15px/20px arial,sans-serif;`

That way, we can make the medium and bold font variables to just specify the weight in common.css and then override via style.css. it's not pefect, but probably a good comproimise.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -48,6 +48,9 @@
   --chromium-border-radius: 1px;
   --mozilla-box-shadow: 0 0 0 2px royalblue;
   --mozilla-border-radius: 1px;
+  --remotion-themed-font-heading: normal 700 14px Arial, 'Helvetica', sans-serif;
+  --remotion-themed-font-bold: normal 500 14px Arial, 'Helvetica', sans-serif;
+  --remotion-themed-font-standard: normal 400 14px Arial, 'Helvetica', sans-serif;
```
a-ctor, line 53: If I understood mike correctly, this should only apply the font weight and inherit the rest. `style.css` should override those with the fonts that we want.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -58,6 +58,7 @@ span.DropDownMenuSelect a.DropDownMenuLabel
   padding: var(--remotion-themed-spacing);
   border-right: var(--remotion-themed-border-width) solid var(--color-contrast-separator);
   outline: none;
+  font-family: var(--remotion-themed-font-bold);
```
a-ctor, line 61: The full specifier will not work for `font-family` so we need to use `font` here. There are also a few other places where we need to make this change.

```suggestion
  font: var(--remotion-themed-font-bold);
```

a-ctor (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
MichaelKetting: correct, if we can do this that way. if we need to name the font, maybe we can substitute `inherit` or `auto`, otherwise, go with just `sans-serif` for the font type. and the size is set to `100%` or `inherit`

a-ctor (approved):

Remotion/ObjectBinding/Web.Test/Global.asax.cs
```diff
@@ -126,7 +129,6 @@ protected void Application_Start (Object sender, EventArgs e)
                 new ResourcePathMapping ("Remotion.ObjectBinding.Web/Html", @$"..\..\ObjectBinding\Web.ClientScript\bin\{configuration}\dist"),
                 new ResourcePathMapping ("Remotion.ObjectBinding.Web/Themes", @"..\..\ObjectBinding\Web\res\Themes"),
                 new ResourcePathMapping ("Remotion.Web.Preview", @"..\..\Web\Preview\res"),
-                new ResourcePathMapping ("Remotion.ObjectBinding.Web.Preview", @"..\..\ObjectBinding\Web.Preview\res")
```
MichaelKetting, line 129: why did you remove this?

Remotion/Web/Core/Res/HTML/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
MichaelKetting, line 1: Please rename to Roboto-License.txt

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Style.css
```diff
@@ -35,14 +35,35 @@ Black
 
 /* Basic styles used to define the font, text-size, etc. */
 
+@font-face {
+  font-family: 'Roboto-Medium';
+  src: url('../../../Html/Roboto-Medium.ttf') format('truetype');
```
MichaelKetting, line 40: Okay, so we can only import it like this? "good" (as in, okay, we have to do it this way then and deal with the consequences).

So, since we cannot include the font from the code-behind but only via the CSS file, we can now move the TTF files into the themes html folder isntead of the shared one. Reasoning: we no longer have to do a global registration. Also, things get easier, no changes in global.asax, etc.

MichaelKetting (changes requested):
Please add a follow-up ticket for us to spend some more time analyzing if we can improve the loading of the ttf files:
image

Remotion/ObjectBinding/Web.Test/Global.asax.cs
derlistigelurch: ups that was a mistake

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Style.css
```diff
@@ -35,14 +35,35 @@ Black
 
 /* Basic styles used to define the font, text-size, etc. */
 
+@font-face {
+  font-family: 'Roboto-Medium';
+  src: url('../Html/Roboto-Medium.ttf') format('truetype');
```
MichaelKetting, line 40:
```suggestion
  src: url('Roboto-Medium.ttf') format('truetype');
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Style.css
```diff
@@ -35,14 +35,35 @@ Black
...
+}
+
+@font-face {
+  font-family: 'Roboto-Bold';
+  src: url('../Html/Roboto-Bold.ttf') format('truetype');
```
MichaelKetting, line 45:
```suggestion
  src: url('Roboto-Bold.ttf') format('truetype');
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Style.css
```diff
@@ -35,14 +35,35 @@ Black
...
+}
+
+@font-face {
+  font-family: 'Roboto-Regular';
+  src: url('../Html/Roboto-Regular.ttf') format('truetype');
```
MichaelKetting, line 50:
```suggestion
  src: url('Roboto-Regular.ttf') format('truetype');
```

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #208, opened 2021-09-15
RM-7929 Add primary/supplemental variants for the DropDownMenu

a-ctor:
https://re-motion.atlassian.net/browse/RM-7929


iam-mholle (changes requested):
Please add renderer tests asserting the css classes are rendered. PR is great apart from this.

iam-mholle (approved):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -24,6 +24,36 @@ span.DropDownMenuSelect
   cursor: pointer;
 }
 
+span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled)
+{
+  --color-standard-text: var(--color-primary-text);
```
MichaelKetting, line 29: please don't.
I don't think we should override the standard properties in a usecase-specific way. This feels like it could have unintended sideeffects in a theoretical future scanario.

in WebButton we just set color, doesn't this work, here, too? Otherwise, we should intoduce a variable and switch that one as needed.

Remotion/Web/Core/Res/Themes/NovaViso/Image/sprite.svg
```diff
@@ -9,6 +9,11 @@
         <polyline points="6.5 11 9.5 8 6.5 5" />
       </g>
     </g>
+    <g id="SingleChevronWhite">
+      <g stroke="#ffffff" stroke-width="1.2" fill="none" fill-rule="evenodd" stroke-linecap="square" stroke-linejoin="round">
```
MichaelKetting, line 13: Quick question to @iam-mholle: can we override the color inside the svg file, i.e. have a generic black chevron and then make it white when we use it in line 44?

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
```diff
@@ -236,11 +236,30 @@ protected string CssClassDropDownButton
...
       AddStandardAttributesToRender (renderingContext);
       if (string.IsNullOrEmpty (renderingContext.Control.CssClass) && string.IsNullOrEmpty (renderingContext.Control.Attributes["class"]))
-        renderingContext.Writer.AddAttribute (HtmlTextWriterAttribute.Class, CssClassBase);
+      {
+        var cssClass = CssClassBase;
+
+        if (renderingContext.Control.ButtonType == ButtonType.Primary)
+          cssClass += " " + CssClassDropDownButtonPrimary;
+        else if (renderingContext.Control.ButtonType == ButtonType.Supplemental)
+          cssClass += " " + CssClassDropDownButtonSupplemental;
```
MichaelKetting, line 259: random thought: would this be better readable with a switch expression?

Remotion/Web/UnitTests/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRendererTest.cs
```diff
@@ -31,6 +31,7 @@
 using Remotion.Web.UI.Controls.DropDownMenuImplementation;
 using Remotion.Web.UI.Controls.DropDownMenuImplementation.Rendering;
 using Remotion.Web.UI.Controls.Rendering;
+using ButtonType = Remotion.Web.UI.Controls.ButtonType;
```
MichaelKetting, line 34: Why do need this in the Unittests? Didn't do a checkout, so it just seems strange?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -24,6 +24,36 @@ span.DropDownMenuSelect
...
+  background-color: var(--color-primary);
+  border: none;
+}
+
```
MichaelKetting, line 33: I'm missing the update to the a.DropDownMenuLabel in primary. we should make this either white or the light-blue variant and then, during hover, the border turns primary-color. quick demo or different screenhots?

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/Image/sprite.svg
iam-mholle: It should be possible to style it via the css section in the svg file. Not 100% sure tho.

MichaelKetting: @iam-mholle thanks!
@a-ctor I think we should try this timeboxed to remove duplication. Should it make things much more complicated, we can revisit, of course. But the color-shift may also be relevant for making disabled icons, etc, so we may need this anyway in the (very near) future.

a-ctor: It is probably easiest to apply a filter to the image element and have just one black item. For example, in the this case we just need a `filter: invert()` to get to the white chevron. Other colors can be created using more complex filters.

Remotion/Web/UnitTests/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRendererTest.cs
a-ctor: Because there is a `System.Web.UI.WebControls.ButtonType`

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
a-ctor: I don't think I know what the two solutions are that you want to compare.

I have added the color styling and also introduced a border on hover. Let me know if there is something else that I should test

Remotion/Web/Core/Res/Themes/NovaViso/Image/sprite.svg
MichaelKetting: okay, then we stick with the status quo for now. Using a filter for this feels like overkill.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
MichaelKetting: That's the one:
image
image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -24,6 +24,42 @@ span.DropDownMenuSelect
...
+span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled).focus
+{
+  background-color: var(--color-primary-highlight);
+  border-color: var(--color-primary-highlight);
+}
+
+span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled) a.DropDownMenuButton img
+{
+  filter: invert();
+}
```
MichaelKetting, line 50: What I mentioned in the other chat, if filter in a css file is the only way, we stick with doing this in the svg file via duplication.

What I was trying to achive was applying color-styling (but no filters) directly inside the svg file to separate the shape from the color, since the svg-file already has a layered approach for the chevron (declaring the shape and using the shape). Since we can't colorize the shape inside the svg file at the using-place, we stick with duplication.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #209, opened 2021-09-17
RM-7918 Create custom style for checkbox and radiobutton

derlistigelurch:


Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
```diff
@@ -6,10 +6,11 @@
   align-items: center;
 }
 
-.bocCheckBox input
+.bocCheckBox input[type=checkbox]
```
iam-mholle, line 9: is the `[type=checkbox]` really needed?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
```diff
@@ -6,10 +6,11 @@
...
 {
   position: absolute;
   z-index: var(--z-index-hidden);
+  opacity: 0;
```
iam-mholle, line 13: You might want to use `visibility: hidden` here, even if it does the same thing.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -89,3 +91,49 @@
     line height remains unchanged. */
   content: "\200b";
 }
+
+.bocEnumValue input[type=radio]
+{
+  z-index: var(--z-index-hidden);
+  position: absolute;
+  opacity: 0;
```
iam-mholle, line 99: You might want to use `visibility: hidden`.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -89,3 +91,49 @@
...
+
+.bocEnumValue input[type=radio] + label::before
+{
+  background-image: url('../Image/sprite.svg#RadioButtonFalse');
+  height: var(--default-icon-size);
+  width: var(--default-icon-size);
+  display: inline-block;
+  content: "";
+  margin-bottom: -3px;
+  margin-right: 4px;
```
iam-mholle, line 115: Shouldn't the margins also be expressed in `em`? Not sure about that tho.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/Image/sprite.svg
```diff
@@ -92,4 +92,15 @@
...
   </svg>
+  <svg class="sprite-part" id="RadioButtonFalse" width="16px" height="16px" viewBox="0 0 16 16">
+    <g>
+      <ellipse ry="7.46081" rx="7.42946" id="svg_2" cy="8.03135" cx="7.96865" stroke="#0c6594" fill="#ffffff"/>
+    </g>
+  </svg>
+  <svg class="sprite-part" id="RadioButtonTrue" width="16px" height="16px" viewBox="0 0 16 16">
+    <g>
+      <ellipse fill="#ffffff" stroke="#0c6594" cx="7.96865" cy="8.03135" id="svg_2" rx="7.42946" ry="7.46081"/>
+      <ellipse fill="#0c6594" cx="7.93884" cy="8.09174" id="svg_4" rx="5.5" ry="5.5"/>
```
iam-mholle, line 103:
```suggestion
      <circle fill="#ffffff" stroke="#0c6594" cy="8" cx="8" r="7.4" />
    </g>
  </svg>
  <svg class="sprite-part" id="RadioButtonTrue" width="16px" height="16px" viewBox="0 0 16 16">
    <g>
      <circle fill="#ffffff" stroke="#0c6594" cx="8" cy="8" r="7.4" />
      <circle fill="#0c6594" cx="8" cy="8" r="5.5" />
```

iam-mholle (changes requested):
The disabled radiobuttons aren't aligned properly, but I'm not sure if it is part of this PR.
image

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
derlistigelurch: no, it's not needed

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
derlistigelurch: No unfortunately they are not doing the same thing. With `visibility: hidden` the checkbox/radiobutton is no longer tabbable.

iam-mholle: Interesting, then ignore the 2 comments about it, sorry about that.

iam-mholle (approved):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -89,3 +91,54 @@
     line height remains unchanged. */
   content: "\200b";
 }
+
+.bocEnumValue input[type=radio]
```
MichaelKetting, line 95: Please check, I think we can move this whole logic to the Common.css and apply to to all input[type=checkbox|radiobutton] that have remotion-themed applied or are nested in remotion-themed.

We expect/require that these elements are followed by a label/span.

Please don't forget to move the icons to the Remotion.Web sprite.

In the BocList, you will also find occurances of checkbox/radiobutton. Those should be automatically covered once we use the remotion-themed approach.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -89,3 +91,54 @@
...
+  margin-bottom: -0.25em;
+  margin-right: 0.25em;
+}
+
+.bocEnumValue input[type=radio]:checked + label::before
+{
+  background-image: url('../Image/sprite.svg#RadioButtonTrue');
+}
+
+.bocEnumValue input[type=radio]:disabled + label:before
```
MichaelKetting, line 123: please check for consistency, both in the new code and in the existing code regarding before/after

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -89,3 +91,54 @@
...
+}
+
+.bocEnumValue input[type=radio] + label::before
+{
+  background-image: url('../Image/sprite.svg#RadioButtonFalse');
+  height: var(--default-icon-size);
+  width: var(--default-icon-size);
+  display: inline-block;
+  content: "";
+  margin-bottom: -0.25em;
```
MichaelKetting, line 114: are the margins for adjusing the icon due to svg viewbox? we should make them into css variables so this becomes clearer. In particular the .disabled style is a bit lost otherwise.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/Image/sprite.svg
```diff
@@ -92,4 +92,15 @@
...
+    <g>
+      <circle fill="#ffffff" stroke="#0c6594" cy="8" cx="8" r="7.4" />
+    </g>
+  </svg>
+  <svg class="sprite-part" id="RadioButtonTrue" width="16px" height="16px" viewBox="0 0 16 16">
+    <g>
+      <circle fill="#ffffff" stroke="#0c6594" cx="8" cy="8" r="7.4" />
+      <circle fill="#0c6594" cx="8" cy="8" r="5.5" />
+    </g>
+  </svg>
```
MichaelKetting, line 105:
```suggestion
  <svg class="sprite-part" id="RadioButtonFalse" width="16px" height="16px" viewBox="0 0 16 16">
    <g>
      <circle fill="#ffffff" stroke="#979797" stroke-width="1.1" cy="8" cx="8" r="7.4" />
    </g>
  </svg>
  <svg class="sprite-part" id="RadioButtonTrue" width="16px" height="16px" viewBox="0 0 16 16">
    <g>
      <circle fill="#ffffff" stroke="#0c6594" stroke-width="1.1" cx="8" cy="8" r="7.4" />
      <circle fill="#0c6594" cx="8" cy="8" r="5.3" />
    </g>
  </svg>
```

Some fine tuning to make it look smoother.

Also, the non-selected variant should be gray
image

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/Image/sprite.svg
```diff
@@ -92,4 +92,15 @@
       <use href="#SingleChevron" />
     </g>
   </svg>
```
MichaelKetting, line 94: I've noticed that the checkbox uses transparent instead of white background. we need to update this.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
derlistigelurch:
```css
.remotion-themed input[type="radio"],
.remotion-themed input[type="checkbox"],
input[type="radio"] .remotion-themed,
input[type="checkbox"] .remotion-themed
```

If I put this in the `Common.css` the default `checkbox` and `radiobutton` in the `boclist` would also be affected.

MichaelKetting:
```
input[type="radio"] .remotion-themed,
input[type="checkbox"] .remotion-themed
```
Those two need to go, then there is no conflict. Same as with input[type=text], select, etc

derlistigelurch: No, in both cases the `remotion-themed` class comes before the `input` element, so I think there is no way of doing this without adding additional selectors. Maybe this should not go inside the `Common.css`.

MichaelKetting: I'm sorry, but then I don't understand the problem because if the input is inside remotion-themed, we want to have the theme. Please elaborate where this is not the intended outcome.

derlistigelurch: Alright, the problem is that the `bocList` cell contains a default `checkbox` and no `label`, therefore when adding the rule inside the `Common.css` the `checkbox` just disappears.

image

image

MichaelKetting: correct. this needs to be updated to add an empty span that covers this part.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -117,6 +117,14 @@ input[type="password"].remotion-themed,
...
 }
 
+.remotion-themed input[type="radio"],
+.remotion-themed input[type="checkbox"]
+{
+  z-index: var(--z-index-hidden);
+  position: absolute;
+  opacity: 0;
+}
+
```
MichaelKetting, line 127: quick note: all the other stuff based on label and span also needs to go here

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocSelectorColumnRenderer.cs
```diff
@@ -168,6 +175,9 @@ private void RenderDataRowSelectorControl (BocListRenderingContext renderingCont
 
       renderingContext.Writer.RenderBeginTag (HtmlTextWriterTag.Input);
       renderingContext.Writer.RenderEndTag();
+      renderingContext.Writer.AddAttribute (HtmlTextWriterAttribute.Class, "checkbox-visualizer");
```
MichaelKetting, line 178: I don't think we need to add a css-class here and before. We should be able to do this with the `remotion-themed` css class on the input element  and moving all the css and svg content into remotion.web. The only thing we need to do in objectbinding should be adding the css-class and the empry span (and click handler, apparently).

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocSelectorColumnRenderer.cs
```diff
@@ -134,12 +133,20 @@ private void RenderTitleRowSelectorControl (BocListRenderingContext renderingCon
                         + ");";
         renderingContext.Writer.AddAttribute (HtmlTextWriterAttribute.Onclick, script);
       }
-
```
MichaelKetting, line 137: revert

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocSelectorColumnRenderer.cs
```diff
@@ -134,12 +133,20 @@ private void RenderTitleRowSelectorControl (BocListRenderingContext renderingCon
...
       if (_renderingFeatures.EnableDiagnosticMetadata)
         renderingContext.Writer.AddAttribute (DiagnosticMetadataAttributesForObjectBinding.BocListWellKnownSelectAllControl, "true");
 
       renderingContext.Writer.RenderBeginTag (HtmlTextWriterTag.Input);
       renderingContext.Writer.RenderEndTag(); // Input-checkbox
+      renderingContext.Writer.AddAttribute (HtmlTextWriterAttribute.Class, "checkbox-visualizer");
+      if (renderingContext.Control.HasClientScript)
```
MichaelKetting, line 142: please check, I believe all the javascript is done in the boclist.ts file.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -377,6 +377,15 @@ class BocList
     selectedRows.OnSelectionChanged (bocList, false);
   }
 
+  public static OnCustomCheckboxClick (elementName: string): void
```
MichaelKetting, line 380: since I noticed this in the repo:

I was referring to integrating this code with the rest of the click-handling code already available in BocList.ts for setting the checkbox/radiobutton. We do not need any Javascript-Hookup in the C# code behind. We just need to wire up the relevant DOM elements. 

There is also code in BocList.ts that makes the entire cell work as the checkbox so maybe we just need to make the select element click "transparent"?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/Image/sprite.svg
```diff
@@ -92,4 +92,15 @@
...
+  <svg class="sprite-part" id="RadioButtonFalse" width="16px" height="16px" viewBox="0 0 16 16">
```
MichaelKetting, line 95: Checkbox and Radiobutton should ONLY be located in Web/spite.svg. They should NOT be in ObjectBinding.Web/sprite.svg

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -132,6 +132,57 @@ input[type="password"].remotion-themed,
...
+.remotion-themed input[type=checkbox] + span
```
MichaelKetting, line 143: You need to do these styles for both `span` and `label`

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -132,6 +132,57 @@ input[type="password"].remotion-themed,
...
+  background-image: url('../Image/sprite.svg#CheckBoxTrue');
+}
+
+.remotion-themed input[type=radio]:checked + label:before,
+.remotion-themed input[type=radio]:checked + span:before
+{
+  background-image: url('../Image/sprite.svg#RadioButtonTrue');
+}
+
+.remotion-themed input[type=radio]:disabled + label:before
```
MichaelKetting, line 159: we need this also for checkbox

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -132,6 +132,57 @@ input[type="password"].remotion-themed,
...
+{
+  opacity: 0.5;
+  filter: grayscale(100%);
+}
+
+.remotion-themed input[type=checkbox] + span:before
```
MichaelKetting, line 165: I believe checkbox and radiobutton would have the same declarations, except the actual image.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -132,6 +132,57 @@ input[type="password"].remotion-themed,
...
+  background-image: url('../Image/sprite.svg#CheckBoxFalse');
+  display: inline-block;
+  content: "";
+  margin: var(--remotion-themed-spacing);
+  vertical-align: middle;
+  height: 1.5em;
+  width: 1.5em;
+}
+
+.remotion-themed input[type=radio] + label::before,
```
MichaelKetting, line 176: please check for consistency. sometimes there's `::`, then there's `:` in you change set.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocSelectorColumnRenderer.cs
```diff
@@ -168,6 +169,9 @@ private void RenderDataRowSelectorControl (BocListRenderingContext renderingCont
 
       renderingContext.Writer.RenderBeginTag (HtmlTextWriterTag.Input);
       renderingContext.Writer.RenderEndTag();
+      renderingContext.Writer.AddAttribute (HtmlTextWriterAttribute.Class, "checkbox-visualizer");
```
MichaelKetting, line 172: Why do we still have this css class here?

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -169,7 +169,10 @@ private ListControl GetListControl (BocEnumValueRenderingContext renderingContex
       var isDropDownList = renderingContext.Control.ListControlStyle.ControlType == ListControlType.DropDownList;
 
       if (isRadioButtonList)
+      {
         listControl.Attributes.Add (HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.RadioGroup);
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, "remotion-themed");
```
MichaelKetting, line 174: There's a property / constant for this. See `WebButton.CssClassThemed` for an example

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -189,7 +189,7 @@ div.bocList div.bocListTableBlock div.bocListNavigator > span > input
   padding: var(--remotion-themed-spacing-small);
   margin: var(--remotion-themed-spacing-small) var(--remotion-themed-spacing);
   text-align: right;
-  width: calc(0.6em * var(--size));
+  width: calc(0.6em * var(--default-icon-size));
```
MichaelKetting, line 192: Why did you change this code?
```suggestion
  width: calc(0.6em * var(--size));
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -1,11 +1,18 @@
-/* Style applied to the BocEnumValue. */
+.bocEnumValue
+{
+  --label-alignment-margin : 0.125em;
+  --icon-alignment-margin : 0.25em;
+  --negative-icon-alignment-margin : -0.25em;
+  --negative-disabled-label-alignment-margin : -4px;
```
MichaelKetting, line 6: Please add a todo for this in RM-7968. I will need to tweak this for names and sizes (e.g. it may be some of those dimensions are actually dependent on other variables...)

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
```diff
@@ -6,56 +6,24 @@
...
-
 .bocCheckBox span.description
 {
   display: inline-flex;
   align-items: center;
 }
 
 .bocCheckBox img,
-.bocCheckBox input + span.checkbox-visualizer::before
+.bocCheckBox input + span.checkbox-visualizer:before
```
MichaelKetting, line 16: we no longer have this css class. Also, the alignment stuff should (probably?) come from Common.css. It really depends on which parts can be seen as generic and which parts are specific. a good hint is if the same rules are needed for BocBoolean, BocCheckbox and in BOclist. Also we should be consistent between radiobutton and checkbox

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -123,6 +123,14 @@ class BocList
       selectedRows.SelectAllSelectorControls = document.querySelectorAll<HTMLInputElement>('input[name="' + selectAllSelectorControlName + '"]')!;
       selectedRows.OnSelectionChanged = onSelectionChangedHandler;
 
+      var visualizer = document.querySelectorAll('input[name="' + selectAllSelectorControlName + '"] + span'/*.checkbox-visualizer*/)!;
```
MichaelKetting, line 126:
```suggestion
      var visualizer = document.querySelectorAll('input[name="' + selectAllSelectorControlName + '"] + span')!;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -123,6 +123,14 @@ class BocList
...
+      visualizer.forEach(checkbox => {
+          checkbox.addEventListener('click', function (evt) {
+              evt.stopPropagation();
+                  selectedRows.SelectAllSelectorControls![0]!.click();
```
MichaelKetting, line 130: is the indent incorrect? also please check the braces, we use a different layout than here in the diff.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -377,6 +385,15 @@ class BocList
     selectedRows.OnSelectionChanged (bocList, false);
   }
 
+  // public static OnCustomCheckboxClick (elementName: string): void
```
MichaelKetting, line 388: dead code

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/Image/sprite.svg
derlistigelurch: the `readonly` `bocCheckbox`, automatically creates the image path. If the `svg` is not in the `Objectbinding.Web/sprite.svg` the `readonly` image cannot be displayed.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
derlistigelurch: Because '--size' is not defined in any css file

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/Image/sprite.svg
MichaelKetting: Hmm... right. considering the complexities: (read: I want to get away from duplicating identical images)

In `BocBooleanValueResourceSetFactory`, if we change 
```
    private string GetResourceUrl (string icon)
    {
      return _resourceUrlFactory.CreateThemedResourceUrl (typeof (BocBooleanValueResourceSetFactory), ResourceType.Image, icon).GetUrl();
    }
```
to 
```
    private string GetResourceUrl (string icon)
    {
      return _resourceUrlFactory.CreateThemedResourceUrl (typeof (HtmlHeadAppenderExtensions), ResourceType.Image, icon).GetUrl();
    }
```
we solve this by getting the URLs from Remotion.Web instead of Remotion.ObjectBinding.Web. Don't forget: This will need updating in the unit tests, too.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
MichaelKetting: I know, It's only added during rendering. This is correct code. (we could always discuss a concept of "declaring" such injected variables to make Rider/Reshaprer shut up. But that's out of scope here.

MichaelKetting: Oh, and if you check, `--size` is set to the digit count, it's not in PX.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRendererTest.cs
```diff
@@ -353,12 +351,6 @@ private void CheckImage (bool value, XmlNode outerSpan, string altText)
       Html.AssertAttribute (image, "alt", altText);
     }
 
-    private void CheckCheckBoxVisualizer (XmlNode outerSpan)
```
MichaelKetting, line 356: Hmm... we can (should?) keep the check and continue to assert the span. It's just the css class that's done.

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValueImplementation/Rendering/BocEnumValueRenderer.cs
```diff
@@ -169,7 +169,10 @@ private ListControl GetListControl (BocEnumValueRenderingContext renderingContex
       var isDropDownList = renderingContext.Control.ListControlStyle.ControlType == ListControlType.DropDownList;
 
       if (isRadioButtonList)
+      {
         listControl.Attributes.Add (HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.RadioGroup);
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassThemed);
```
MichaelKetting, line 174: okay, turns out we already add the CssClassThemed to the root element or the BocEnumValue, so please recheck and then we can drop the redundant css class here.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -214,4 +276,26 @@ img.disabled,
...
+.mozilla .remotion-themed span[role=checkbox]:focus > img,
+.mozilla .remotion-themed span[role=radio]:focus > img,
+.mozilla .remotion-themed input[type=checkbox]:focus + span:before,
+.mozilla .remotion-themed input[type=checkbox]:focus + label:before,
+.mozilla .remotion-themed input[type=radio]:focus + span:before,
+.mozilla .remotion-themed input[type=radio]:focus + label:before
+{
+  box-shadow: var(--mozilla-box-shadow);
+  border-radius: var(--mozilla-border-radius);
+}
```
MichaelKetting, line 301: This code is dead once you rebase on top of the NovaViso branch since the focus-task is already implemented now.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -132,6 +132,68 @@ input[type="password"].remotion-themed,
...
+
+.remotion-themed input[type=checkbox] + label:before,
+.remotion-themed input[type=checkbox] + span:before
+{
+  background-image: url('../Image/sprite.svg#CheckBoxFalse');
+  margin: var(--remotion-themed-spacing);
+  vertical-align: middle;
+  height: 1.5em;
+  width: 1.5em;
```
MichaelKetting, line 165:
```suggestion
  height: var(--remotion-themed-img-width);
  width: (--remotion-themed-img-width);
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -132,6 +132,68 @@ input[type="password"].remotion-themed,
...
+}
+
+.remotion-themed input[type=radio] + label:before,
+.remotion-themed input[type=radio] + span:before
+{
+  background-image: url('../Image/sprite.svg#RadioButtonFalse');
+  height: var(--default-icon-size);
+  width: var(--default-icon-size);
```
MichaelKetting, line 173:
```suggestion
  height: var(--remotion-themed-img-width);
  width: var(--remotion-themed-img-width);
```
I think this should be indetical to checkbox. We can then also move the height/width to the shared declarations.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
```diff
@@ -6,56 +6,24 @@
...
-  z-index: var(--z-index-hidden);
-}
-
 .bocCheckBox span.description
 {
   display: inline-flex;
   align-items: center;
 }
 
 .bocCheckBox img,
```
MichaelKetting, line 15:
```suggestion
.bocCheckBox img
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocCheckBox.css
```diff
@@ -6,56 +6,24 @@
...
-.bocCheckBox input + span.checkbox-visualizer::before
+.bocCheckBox input + span:before
```
MichaelKetting, line 16:
```suggestion
```
This is now redundant for input+span with common.css.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
```diff
@@ -101,3 +109,24 @@
...
+
+.bocEnumValue input[type=radio]:disabled + label
+{
+  margin-bottom: var(--negative-disabled-label-alignment-margin);
+}
+
+.bocEnumValue.radioButtonList.disabled
+{
+  margin-top: var(--negative-icon-alignment-margin);
+}
```
MichaelKetting, line 132: You're using a different way to align the radiobutton items compared to the checkboxes. Is this actually needed or can we use the same logic here (actually, also then over in common.css) as we do with checkbox?

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocEnumValue.css
derlistigelurch: No, this alignment adjustments are only needed in BocEnumValue.css, therefore it can stay there.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRendererTest.cs
```diff
@@ -285,14 +285,14 @@ private void CheckRender (bool value, string spanText)
         Html.AssertAttribute (valueSpan, StubValidationErrorRenderer.ValidationErrorsAttribute, c_validationErrors);
 
         CheckImage (value, valueSpan, spanText);
-        CheckCheckBoxVisualizer (outerSpan);
+        CheckOuterSpan(outerSpan);
```
MichaelKetting, line 288:
```suggestion
        CheckOuterSpan (outerSpan);
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRendererTest.cs
```diff
@@ -285,14 +285,14 @@ private void CheckRender (bool value, string spanText)
...
 
         AssertValidationErrors (outerSpan);
       }
       else
       {
         CheckInput (value, outerSpan);
-        CheckCheckBoxVisualizer (outerSpan);
+        CheckOuterSpan(outerSpan);
```
MichaelKetting, line 295:
```suggestion
        CheckOuterSpan (outerSpan);
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRendererTest.cs
```diff
@@ -353,10 +353,9 @@ private void CheckImage (bool value, XmlNode outerSpan, string altText)
       Html.AssertAttribute (image, "alt", altText);
     }
 
-    private void CheckCheckBoxVisualizer (XmlNode outerSpan)
+    private static void CheckOuterSpan (XmlNode outerSpan)
```
MichaelKetting, line 356:
```suggestion
    private void CheckOuterSpan (XmlNode outerSpan)
```

MichaelKetting (approved):
Please update the codestyle requests and then this seems to be done.

PR #211, opened 2021-09-22
RM-7912 Create styles for new theme for BocReferenceValue, BocAutoCompleteReferenceValue

FlorianLautner:
https://re-motion.atlassian.net/browse/RM-7912

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
@@ -1,248 +1,160 @@
 /* Style applied to the BocAutoCompleteReferenceValue. */
 
-span.bocAutoCompleteReferenceValue
+.bocAutoCompleteReferenceValue
 {
-  width: 21.5em;
+  --bocautocomplete-input-spacing-offset: calc(var(--remotion-themed-spacing) - 0.25em);
+  min-height: var(--remotion-themed-singleline-min-height);
+}
+
+.bocAutoCompleteReferenceValue span.body
+{
+  display: inline-grid;
+  grid-template-areas: "icon content optionsMenu";
+  grid-template-columns: var(--default-icon-size) var(--remotion-themed-control-width) 1fr;
+  gap: var(--remotion-themed-spacing);
+  align-items: center;
+}
+
+.bocAutoCompleteReferenceValue span.icon
+{
+  grid-area: icon;
+}
+
+.bocAutoCompleteReferenceValue span.icon img
+{
   position: relative;
-  display: inline-block;
-  vertical-align: middle;
-  overflow: hidden;
-  min-height: 1.7em;
+  width: var(--default-icon-size);
+  height: var(--default-icon-size);
+  padding-left: calc(var(--remotion-themed-spacing) + var(--remotion-themed-border-width));
+  pointer-events: none;
+
+  /* Currently used for overriding img.Icon rule in Style.css. */
+  vertical-align: top;
 }
 
-span.bocAutoCompleteReferenceValue span.body
+.bocAutoCompleteReferenceValue span.content
 {
-  display: block;
+  grid-row: 1;
+  grid-column: icon / content;
+  display: inline-grid;
+  grid-template-areas: "input button";
+  grid-template-columns: 1fr var(--remotion-themed-img-width);
+  gap: var(--remotion-themed-spacing-small);
+  align-items: center;
 }
 
-span.bocAutoCompleteReferenceValue input
+.bocAutoCompleteReferenceValue span.content span.bocAutoCompleteReferenceValueInput
 {
+  grid-row: 1;
+  grid-column: input / button;
+  height: 100%;
+}
+
+.bocAutoCompleteReferenceValue span.content span.bocAutoCompleteReferenceValueInput input,
+.bocAutoCompleteReferenceValue span.content.hasIcon span.bocAutoCompleteReferenceValueInput input
+{
+  box-sizing: border-box;
+  height: 100%;
   width: 100%;
-  display: inline-block;
 }
 
-span.bocAutoCompleteReferenceValue input.error
+.bocAutoCompleteReferenceValue span.content span.bocAutoCompleteReferenceValueInput input
+{
+  padding-left: calc(var(--remotion-themed-spacing) + var(--remotion-themed-border-width));
+  padding-right: calc(var(--remotion-themed-img-width) + var(--remotion-themed-spacing-small));
+}
+
+.bocAutoCompleteReferenceValue span.content.hasIcon span.bocAutoCompleteReferenceValueInput input
+{
+  padding-left: calc(var(--default-icon-size) + var(--remotion-themed-spacing) + var(--remotion-themed-border-width) + var(--bocautocomplete-input-spacing-offset));
+}
+
+.bocAutoCompleteReferenceValue input.error
 {
   background-image: url('../Image/sprite.svg#AutoCompleteReferenceValueErrorIndicator');
-  background-size: 16px 16px;
-  background-position: right center;
+  background-size: var(--remotion-themed-line-height) var(--remotion-themed-line-height);
+  background-position: right var(--remotion-themed-img-width) center;
   background-repeat: no-repeat;
 }
 
-span.bocAutoCompleteReferenceValueButton
+.bocAutoCompleteReferenceValue span.bocAutoCompleteReferenceValueButton
 {
-  display: inline-block;
-  position: absolute;
-  margin: 0;
-  padding: 0;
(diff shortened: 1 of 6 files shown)


Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
```diff
@@ -1,248 +1,146 @@
...
 
-span.bocAutoCompleteReferenceValueButton img:active
+.bocAutoCompleteReferenceValue span.icon img
 {
-  background-color: #C5E0FC;
+  position: relative;
+  width: var(--default-icon-size);
+  height: var(--default-icon-size);
+  padding-left: var(--remotion-themed-spacing);
+  z-index: 1;
```
derlistigelurch, line 70:
```suggestion
  z-index: --var(--z-index-max);
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
```diff
@@ -1,248 +1,146 @@
...
 
-/* hasOptionsMenu */
-span.bocAutoCompleteReferenceValue span.optionsMenu span.DropDownMenuSelect,
-span.bocAutoCompleteReferenceValue.disabled span.optionsMenu span.DropDownMenuSelect,
-span.bocAutoCompleteReferenceValue.readOnly span.optionsMenu span.DropDownMenuSelect
+.bocAutoCompleteReferenceValue.readOnly span.content.hasIcon span
 {
-  height: 1.5em;
-  top: 0;
+  padding-left: calc(var(--default-icon-size) + 2 * var(--remotion-themed-spacing) + var(--remotion-themed-border-width));
```
derlistigelurch, line 102: Maybe introduce a new variable for `2 * var(--remotion-themed-spacing)`, you could name it `--remotion-themed-spacing-double`

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
```diff
@@ -1,248 +1,146 @@
...
+  background-repeat: no-repeat;
 }
 
-span.bocAutoCompleteReferenceValueInput input
+.bocAutoCompleteReferenceValue span.bocAutoCompleteReferenceValueButton
 {
-  height: 1.2em;
-  line-height: 1em;
+  position: relative;
+  float: right;
```
derlistigelurch, line 127: Is `float: right` really needed, or can this also be done with a flex property?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
```diff
@@ -1,248 +1,146 @@
...
+  display: flex;
+  align-items: center;
+  width: var(--remotion-themed-img-width);
+  height: var(--remotion-themed-singleline-min-height);
+  background-color: transparent;
```
derlistigelurch, line 132: `background-color: transparent;` is the default value, so you should be able to delete this without breaking anything

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.jquery.css
```diff
@@ -1,142 +1,121 @@
...
 
-.ac_even,
-.ac_even.ac_over.ac_disabled
-
+.ac_results li.ac_disabled > div img
 {
-  background-color: White;
-  color: Black;
+  opacity: 0.4;
+  filter: progid:DXImageTransform.Microsoft.BasicImage(opacity=0.40), progid:DXImageTransform.Microsoft.BasicImage(grayscale=1);
```
derlistigelurch, line 87: We don't do that here
```suggestion
  opacity: 0.5;
  filter: grayscale(100%);
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocReferenceValue.css
```diff
@@ -1,160 +1,104 @@
...
-/* withoutOptionsMenu */
-span.bocReferenceValue span.content.withoutOptionsMenu
+.bocReferenceValue span.icon img
 {
-  right: 0;
+  position: relative;
+  width: var(--default-icon-size);
+  height: var(--default-icon-size);
+  padding-left: var(--remotion-themed-spacing);
+  z-index: 1;
```
derlistigelurch, line 37:
```suggestion
  z-index: var(--z-index-max);
```

derlistigelurch (changes requested):
image

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
MichaelKetting: no, I just dropped the double variable. if we need 2x spacing because it's meant to be padding above+below, we should do this inline.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
FlorianLautner: In this case, `z-index` cannot be set to `--z-index-max` for the icon since it would then interfere with the drop down menu of the search results, as it would still be visible when the drop down menu is expanded.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
FlorianLautner: Since the button actually overlays the input, I think it has to be done by `float: right` since flex does not support this behavior, at least that's the way I understood it.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
MichaelKetting: disclaimer: I only got the gist from the comments

I'm very careful about the z-index. using z-index may result in unwanted sideeffects when using popup panels and may result either in the z-index hiding the elemnt when it's used in a popup with a higher z-index or peeking through another overlay. -> I will need to really understand why there is no other way to do the layout.

derlistigelurch (commented):
image
image

The distance between the image and the text is not the same, can this be ignored? @MichaelKetting

MichaelKetting:
> 
> 
> image
> image
> 
> The distance between the image and the text is not the same, can this be ignored? @MichaelKetting

I see that there's a difference and the vertical alignment isn't symetrical. My goal would be that the icon is equidistant left and right and that the vertical alignment is adjusted to the text. What happens if the icon is vertically shifted -> It should have the same alignment as the (standalone) dropdown menu with icon, I think.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
FlorianLautner: If the z-index of the icon is not higher than that of the input element, then the icon is not visible. This is because we wanted to overlay these two so that the icon is visible within the input on its left-hand side. Setting it to --z-index-max, though, would lead to other icons of input elements below the current one to be visible through the expanded drop down menu since then the drop down results and the input icon would have the same z-index. Thus, I set the input icon to 1 so that it is visible within the input element, but the drop down results still overlay the icon.

MichaelKetting: We may need to move this into a live session, but I don't undestand why the icon is "before" / "lower" than the textbox given the icon is rendered / inserted after the textbox. So, there should be no need for z-index (as far as I can tell from this discussion).

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
```diff
@@ -40,6 +61,7 @@
   height: 100%;
 }
 
+
```
derlistigelurch, line 64: You can delete this new line

derlistigelurch (approved):

MichaelKetting (approved):
review notes worked in as commit

PR #212, opened 2021-09-22
Feature/rm 7998 unified focus for data controls

a-ctor:
https://re-motion.atlassian.net/browse/RM-7998

### DIFF
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
@@ -37,20 +37,25 @@ span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled) a.Drop
   border-right-color: var(--color-primary-text);
 }
 
-span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled):hover,
-span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled):focus,
-span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled).focus
+span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled):hover
 {
   background-color: var(--color-primary-highlight);
   border-color: var(--color-primary-highlight);
 }
+
+span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled):focus,
+span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled).focus
+{
+  background-color: var(--color-primary);
+  border-color: white;
+}     
 
 span.DropDownMenuContainer.primary span.DropDownMenuSelect:not(.disabled) a.DropDownMenuButton img
 {
   background-image: url('../Image/sprite.svg#DropDownMenuButtonWhite');
 }
 
-span.DropDownMenuContainer.supplemental span.DropDownMenuSelect:not(:hover, :focus),
+span.DropDownMenuContainer.supplemental span.DropDownMenuSelect:not(:hover, :focus, .focus),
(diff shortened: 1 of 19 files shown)


Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabbedMenu.css
```diff
@@ -73,31 +73,36 @@ div.tabbedMainMenu span.tabStripTabSelected a,
 div.tabbedSubMenu span.tabStripTab a,
 div.tabbedSubMenu span.tabStripTabSelected a
 {
-  border: none;
+  border: unset;
+  background-color: unset;
```
MichaelKetting, line 77: After seeing this in action, `unset` just feels more explicit and consistent than setting `transparent`

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabbedMultiView.css
```diff
@@ -41,7 +41,8 @@ div.tabbedMultiViewActiveView > div.contentBorder
 {
   padding: var(--tabbedmultiview-view-padding);
   overflow: auto;
-  border: 1px solid transparent;
+  /* Offset for focus-border */
+  border: var(--remotion-themed-border-placeholder);
```
MichaelKetting, line 45: we should never use literal values if there is a correspnding variable. And since we keep needing the placeholder border, I introduced it.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -10,17 +10,18 @@ button.webButton
...
-button.webButton:focus-within
+button.webButton:hover
 {
   background-color: var(--remotion-themed-background-color-hover);
   border: var(--remotion-themed-border-hover);
   box-shadow: var(--remotion-themed-box-shadow-hover);
 }
 
-button.webButton:focus-within
+button.webButton:focus
```
MichaelKetting, line 20: button is focusable. there is never going to be a child element that gets the focus.

focus-within is only needed when the acutally focused target and the highlight effects aren't on the same element.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -106,18 +106,18 @@ ul.DropDownMenuOptions li a
   grid-template-columns: auto minmax(0, 1fr);
   grid-template-areas: "img span";
   gap: var(--remotion-themed-spacing);
-  padding: var(--remotion-themed-spacing);
+  padding: calc(var(--remotion-themed-spacing) - var(--remotion-themed-border-width));
```
MichaelKetting, line 109: for the full-width paddings, subtracting the border-width is helpful to better stick to the intended layout. this should still not be a global subtraction from the spacing since this is used accross the entire css. It may be helpful to introduce a `--remotion-themed-spacing-with-border`, though.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/ListMenu.css
```diff
@@ -12,18 +13,32 @@ td.listMenuRow
   gap: var(--remotion-themed-spacing-small);
 }
 
+span.listMenuItem,
+span.listMenuItemDisabled
+{
+  display: block;
+  width: 100%;
```
MichaelKetting, line 20: turns out, the items looked funny when having single-item-per-line.

MichaelKetting (changes requested):
Open:
Focus for checkbox/boolean/radiobuttons
Focus for various input controls
Focus / hover for interactive elements in BocList

MichaelKetting (approved):
It's not perfect (read: I need to do some itnegration work and fine tuning, but it's good enough for now)

PR #213, opened 2021-09-23
RM-7937 Align icons and text in various ui elements

a-ctor:
https://re-motion.atlassian.net/browse/RM-7937

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
@@ -116,6 +116,16 @@
   cursor: default;
 }
 
+.bocAutoCompleteReferenceValue.readOnly span.body > span.icon
+{
+  align-self: start;
+  padding: var(--remotion-themed-control-padding-readonly);
+  padding-right: 0;
+  padding-bottom: 0;
+  padding-left: 0;
+  border: var(--remotion-themed-border-readonly);
+}
+
 .bocAutoCompleteReferenceValue.readOnly span.content
 {
   display: inline-flex;
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocReferenceValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocReferenceValue.css
@@ -69,6 +69,16 @@
   cursor: default;
 }
 
+.bocReferenceValue.readOnly span.body > span.icon
+{
+  align-self: start;
+  padding: var(--remotion-themed-control-padding-readonly);
+  padding-right: 0;
+  padding-bottom: 0;
+  padding-left: 0;
+  border: var(--remotion-themed-border-readonly);
+}
+
 .bocReferenceValue.readOnly span[tabindex]
 {
   min-height: var(--remotion-themed-singleline-min-height);
--- a/Remotion/Web/Core/Res/Themes/NovaGray/HTML/TabStrip.css
+++ b/Remotion/Web/Core/Res/Themes/NovaGray/HTML/TabStrip.css
@@ -142,6 +142,11 @@ body.msie *.tabStripTabSelected a::before
   vertical-align: middle;
 }
 
+*.tabStripTabsPane *.anchorBody > img + span
+{
+  margin-left: 0.3em;
+}
+
 *.tabStripTabsPane *.anchorBody > span
 {
 }
--- a/Remotion/Web/Core/Res/Themes/NovaGray/HTML/WebButton.css
+++ b/Remotion/Web/Core/Res/Themes/NovaGray/HTML/WebButton.css
@@ -60,6 +60,11 @@ button.webButton span.buttonBody img
   border: none;
 }
 
+button.webButton span.buttonBody > img + span
+{
+  margin-left: 0.3em;
+}
+
 button.webButton:focus span.buttonBody,
 button.webButton:hover span.buttonBody
 {
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
@@ -84,3 +84,16 @@
   border-color: var(--color-disabled-border);
   background-color: var(--color-disabled-background);
 }
+
+.tabStripTabWrapper .tabStripTab a span.anchorBody,
+.tabStripTabWrapper .tabStripTabSelected a span.anchorBody
+{
+  display: flex;
+  gap: var(--remotion-themed-spacing);
+}
+
+.tabStripTabWrapper .tabStripTab a span.anchorBody > img,
+.tabStripTabWrapper .tabStripTabSelected a span.anchorBody > img
+{
+  align-self: start;
+}
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.NetFramework/ObjectBinding.Web.Development.WebTesting.TestSite.NetFramework.csproj
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.NetFramework/ObjectBinding.Web.Development.WebTesting.TestSite.NetFramework.csproj
@@ -116,6 +116,7 @@
     <Content Include="objects\Remotion.ObjectBinding.Sample.Person\00000000-0000-0000-0000-000000000001" />
     <Content Include="objects\Remotion.ObjectBinding.Sample.Person\00000000-0000-0000-0000-000000000009" />
     <Content Include="objects\Remotion.ObjectBinding.Sample.Person\00000000-0000-0000-0000-00000000000A" />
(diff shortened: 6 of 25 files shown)


Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
```diff
@@ -52,6 +52,18 @@
   background-color: var(--color-standard-background);
 }
 
+.tabStripTabWrapper .tabStripTab a span.anchorBody,
```
MichaelKetting, line 55: I'm not 100% sure what the most consistent approach is but my gut feeling says this should go after all the `a` rules since it's nested inside `a`

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
```diff
@@ -28,7 +28,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  height: calc(var(--remotion-themed-line-height) + var(--remotion-themed-spacing));
+  min-height: calc(var(--remotion-themed-line-height) + var(--remotion-themed-spacing));
```
MichaelKetting, line 31: tabstrip-items: when can this line-wrap?

MichaelKetting (changes requested):
The Icon should be not be pulled down to the middle of the text but stay in the same place as for single-line text, if we can manage that?

a-ctor:
@MichaelKetting so it should look like this? 
image

Do you want it like this also for the BocList? Because there it fits better in the center IMO
image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
a-ctor: It wraps when there is not enough space (parent container is limited in size), but it had a rendering problem where the `height` was fixed, which caused the text to overflow from the element. Thus, this change.

MichaelKetting:
> 
> 
> @MichaelKetting so it should look like this?
> image

Yes, that's the offiical one.


> Do you want it like this also for the BocList? Because there it fits better in the center IMO
> image

Hmm...in the BocList cell... for the moment, we should do a top-alignment since that's the default right now. It may very well be that this gets updated later on, but for now, we stick with no-special-cases (so long as it it doesn't become unwieldy)

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
MichaelKetting: hmm... we have another spot where the text escapes the border in such a case. we should leave this for now and please make a note in our common.css task with the open todos that I need to check how we deal with overflow-text for such elemnts. it may well be that the text will be cut off isntead of allowed to wrap in such cases.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
```diff
@@ -52,6 +52,12 @@
   background-color: var(--color-standard-background);
 }
 
+.tabStripTabWrapper .tabStripTab a span.anchorBody > img,
+.tabStripTabWrapper .tabStripTabSelected a span.anchorBody > img
+{
+  align-self: baseline;
+}
+
```
MichaelKetting, line 60:
```suggestion
```
moved down for correct nesting

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
```diff
@@ -84,3 +90,9 @@
   border-color: var(--color-disabled-border);
   background-color: var(--color-disabled-background);
 }
+
+.tabStripTabWrapper .tabStripTab a span.anchorBody,
+.tabStripTabWrapper .tabStripTabSelected a span.anchorBody
+{
+  display: flex;
+}
```
MichaelKetting, line 98:
```suggestion

.tabStripTabWrapper .tabStripTab a span.anchorBody,
.tabStripTabWrapper .tabStripTabSelected a span.anchorBody
{
  display: flex;
}

.tabStripTabWrapper .tabStripTab a span.anchorBody > img,
.tabStripTabWrapper .tabStripTabSelected a span.anchorBody > img
{
  align-self: baseline;
}

```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
```diff
@@ -84,3 +90,9 @@
   border-color: var(--color-disabled-border);
   background-color: var(--color-disabled-background);
 }
+
+.tabStripTabWrapper .tabStripTab a span.anchorBody,
+.tabStripTabWrapper .tabStripTabSelected a span.anchorBody
+{
+  display: flex;
```
MichaelKetting, line 97: Why do we not need `gap` here, but we do in webbutton?

MichaelKetting (changes requested):
It does look okay but I didn't try it out and would like to do a shared session before we're done.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
a-ctor: Because of the way the tab strip is rendered... It renders a NBSP and a span containing the text, which is interpreted by flexbox as two separate elements and as such the NBSP is acting as a spacer. In the WebButton there is no span being rendered. The NBSP and the text is treated as the same element and as such we need a gap to have spacing between the icon and the text.

MichaelKetting: hmm...since the spacing is probably not identical, maybe we could easily select and then set the nbsp to hidden via css and then use gap for the rest. don't know, didn't try. or, perhaps somehting for a follow-up refacotring task to clean this up.

Remotion/Web/Core/Res/Themes/NovaGray/HTML/TabStrip.css
```diff
@@ -142,6 +142,11 @@ body.msie *.tabStripTabSelected a::before
   vertical-align: middle;
 }
 
+*.tabStripTabsPane *.anchorBody > img + span
+{
+  margin-left: 3px;
```
MichaelKetting, line 147: please use `em` here. white space width increases with font-size, so we need to use `em` instead of `px`.

I'm guessing 0.3em should do the trick, but please test it. it doesn't have to be pixel-perfect but it should be a good match and I don't want to go into milli-fractions

Remotion/Web/Core/Res/Themes/NovaGray/HTML/WebButton.css
```diff
@@ -60,6 +60,11 @@ button.webButton span.buttonBody img
   border: none;
 }
 
+button.webButton span.buttonBody > img + span
+{
+  margin-left: 3px;
```
MichaelKetting, line 65: see TabStrip.css

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocReferenceValue.css
```diff
@@ -69,6 +69,13 @@
   cursor: default;
 }
 
+.bocReferenceValue.readOnly span.body > span.icon
+{
+  align-self: start;
+  padding-top: calc(var(--remotion-themed-spacing) - var(--remotion-themed-border-width));
+  border-top: 1px solid transparent;
```
MichaelKetting, line 76: this feels strange. first you subtract 1px, then you add it again. I didn't analyze the rest of the file for the underlying reason, you elaborate (screenshot or live)? And either way, we have the border-width var that should be used here, too, in case there is a good reason why we need both a padding-top and a border-top.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.css
```diff
@@ -116,6 +116,13 @@
   cursor: default;
 }
 
+.bocAutoCompleteReferenceValue.readOnly span.body > span.icon
+{
+  align-self: start;
+  padding-top: calc(var(--remotion-themed-spacing) - var(--remotion-themed-border-width));
+  border-top: 1px solid transparent;
```
MichaelKetting, line 123: See BocReferenceValue.css

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocReferenceValue.css
a-ctor: Yeah this part is sad. Because we did not include the padding in the grid layout simply using `align-self: start` will put the icon above the text (since the text has padding but the button does not). For some reason some magic pixels (~2px) are missing when I apply the same padding as the text element and it does not line up. So even if we use the same padding we are still off...

I don't know what the best calculation is that we can put here to have the correct spacing. Adapting the CSS grid to include the spacing would be ideal but this takes more time.

MichaelKetting (approved):

PR #216, opened 2021-10-01
RM-7933: Refactor BocAutoCompleReferenceValue and BocReferenceValue to use CssGrid or Flexgrid for layout

derlistigelurch:


MichaelKetting (commented):
resolved with a slightly different approach

PR #217, opened 2021-10-04
RM-8035 Remove caching from TypeExtensions.CanAscribeTo

derlistigelurch:
https://re-motion.atlassian.net/browse/RM-8035


iam-mholle:
Please add caching to usage sites where necessary and re-request review.

Remotion/Core/Core/Reflection/TypeExtensions.cs
```diff
@@ -112,53 +89,38 @@ private static bool CanAscribeInternal (Type type, Type ascribeeType)
...
-      {
-        if (ascribeeType.IsAssignableFrom (type))
-          return Type.EmptyTypes;
-        else
-          throw ArgumentUtility.CreateArgumentTypeException ("type", type, ascribeeType);
-      }
-      else if (ascribeeType.IsInterface)
-        return GetAscribedGenericInterfaceArgumentsInternal (type, ascribeeType);
-      else
-        return GetAscribedGenericClassArgumentsInternal (type, ascribeeType);
```
MichaelKetting, line 133: please use this block (and add missing braces) instead of dropping the ELSE notiations. 

Reasoning: IF/ELSE is better readable when not doing alternative code branches instead of having a single main branch and several early-exit conditions.

derlistigelurch: hmm ok, how about a `switch statement` or a `switch expression` to get rid of the `if`s and `else`s?
e.g.
```Csharp

return ascribeeType switch
{
    { IsGenericType: false } when ascribeeType.IsAssignableFrom (type) => Type.EmptyTypes,
    { IsGenericType: false } => throw ArgumentUtility.CreateArgumentTypeException ("type", type, ascribeeType),
    _ => ascribeeType.IsInterface
            ? GetAscribedGenericInterfaceArgumentsInternal (type, ascribeeType)
            : GetAscribedGenericClassArgumentsInternal (type, ascribeeType)
};
```

MichaelKetting: no, just keep the original, please (refactoring of this is out of scope for the current task).  

Also: switch (or, rather pattern matching) should be used for actual pattern cases, not just as an alternative to boolean conditions in an IF. The block structure is actually helpful in the above statement.

Remotion/Core/Core/Reflection/TypeExtensions.cs
```diff
@@ -189,25 +150,17 @@ private static IReadOnlyList<Type> GetAscribedGenericClassArgumentsInternal (Typ
...
 
       if (ascribeeType != ascribeeGenericTypeDefinition)
         return ascribeeType.IsAssignableFrom (type);
       else
-        return ascribeeType.IsAssignableFrom (type.GetGenericTypeDefinition ());
+        return ascribeeType.IsAssignableFrom (type.GetGenericTypeDefinition());
     }
   }
-}
+}
```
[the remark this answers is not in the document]
MichaelKetting: ReflectionUtility.GetObjectListTypeParameter
probably cache. Maybe a single cache with IsObjectList result since we will need the typeparameter anyway for any ObjectList<> we find.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Mapping/PersistentMixinFinder.cs
```diff
@@ -38,7 +41,12 @@ public static bool IsPersistenceRelevant (Type mixinType)
     {
       ArgumentUtility.CheckNotNull ("mixinType", mixinType);
 
-      return TypeExtensions.CanAscribeTo (mixinType, typeof (DomainObjectMixin<,>));
+      if (!typeof (IDomainObjectMixin).IsAssignableFrom (mixinType))
+        return false;
+
+      return s_canAscribeCache.GetOrAdd (
+          Tuple.Create (mixinType, typeof (DomainObjectMixin<,>)),
```
MichaelKetting, line 48: see earlier remarks

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -37,6 +38,9 @@ public static class ReflectionUtility
     private static readonly Type s_binaryPropertyValueType = typeof (byte[]);
     private static readonly Type s_typePropertyValueType = typeof (Type);
     private static readonly Type s_objectIDPropertyValueType = typeof (ObjectID);
+    private static readonly ConcurrentDictionary<Tuple<Type, Type>, bool> s_canAscribeCache = new();
+    private static readonly ConcurrentDictionary<Tuple<Type, Type>, Type> s_ascribedGenericArgumentsCache = new();
```
MichaelKetting, line 42: see earlier remarks on unify caches, no tuple key since it's always ObjectList. And naming should then be specific to the usecase

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -262,11 +266,16 @@ public static Type GetObjectListTypeParameter (Type type)
...
-        return null;
-      return typeParameter;
+      Type ValueFactory (Tuple<Type, Type> key)
+      {
+        var (type, objectListType) = key;
+        var typeParameter = type.GetAscribedGenericArguments (objectListType)[0];
+        return typeParameter.IsGenericParameter ? null : typeParameter;
+      }
+
+      s_ascribedGenericArgumentsCacheValueFactory ??= ValueFactory;
```
MichaelKetting, line 276: We should check what the compiler does. There's rules regarding what's getting automatic delegate caching code generated by the c# compiler.

Remotion/Mixins/Core/MixinTypeUtility.cs
```diff
@@ -188,7 +190,10 @@ public static bool HasMixin (Type targetOrConcreteType, Type mixinType)
       ArgumentUtility.CheckNotNull ("mixinType", mixinType);
 
       return GetMixinTypes (targetOrConcreteType)
-          .FirstOrDefault (configuredMixinType => Reflection.TypeExtensions.CanAscribeTo (configuredMixinType, mixinType));
+          .FirstOrDefault (
+              configuredMixinType => s_canAscribeCache.GetOrAdd (
+                  Tuple.Create (configuredMixinType, mixinType),
```
MichaelKetting, line 195: ValueTuple.

MichaelKetting (changes requested):
Please check the non-remarked files too. They're mostly repeats of earlier input so I short-curcuited the review a bit.

Remotion/Data/DomainObjects/DataManagement/RelationEndPoints/VirtualEndPoints/CollectionEndPoints/DomainObjectCollectionFactory.cs
```diff
@@ -30,6 +31,9 @@ namespace Remotion.Data.DomainObjects.DataManagement.RelationEndPoints.VirtualEn
   /// <seealso cref="ObjectListFactory"/>
   public class DomainObjectCollectionFactory
   {
+    private static readonly ConcurrentDictionary<Tuple<Type, Type>, bool> s_canAscribeCache = new();
+    private static readonly ConcurrentDictionary<Tuple<Type, Type>, IReadOnlyList<Type>> s_ascribedGenericArgumentsCache = new();
```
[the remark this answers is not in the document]
derlistigelurch: > I think we can unify this into a common cache structure.

I don't know if this is a good idea. The two caches store different data and should therefore remain separated. 🤔

Remotion/Data/DomainObjects.ObjectBinding.UnitTests/BindableDomainObjectImplementationTest.cs
```diff
@@ -19,13 +19,24 @@
...
 {
   [TestFixture]
   public class BindableDomainObjectImplementationTest : TestBase
   {
+    [OneTimeSetUp]
+    public void OneTimeSetup ()
+    {
+      Assertion.DebugAssert (
+          !TypeExtensions.CanAscribeTo (typeof (BindableDomainObjectImplementation), typeof (Mixin<,>)),
+          "we assume the mixin does not have a base object");
```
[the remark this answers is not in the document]
MichaelKetting: neither: please add a new unit test.

Remotion/ObjectBinding/Core/BindableObject/PropertyReflector.cs
```diff
@@ -166,6 +166,9 @@ protected virtual Type GetUnderlyingType ()
 
     protected virtual Type GetItemType ()
     {
+      if (!typeof (IEnumerable).IsAssignableFrom (_propertyInfo.PropertyType))
+        return _propertyInfo.PropertyType;
```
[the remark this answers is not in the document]
derlistigelurch: There are no tests specifically for this class, should I create a new one, or is this case already covered by existing tests?
@MichaelKetting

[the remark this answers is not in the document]
derlistigelurch: ok it's already covered by existing tests.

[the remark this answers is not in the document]
derlistigelurch: I thought I have to write a new test to check if
```csharp
if (!typeof (IEnumerable).IsAssignableFrom (_propertyInfo.PropertyType))
        return _propertyInfo.PropertyType;
```
 works correctly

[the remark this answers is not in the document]
derlistigelurch: > Unrelated to the test: I noticed that GetItemType is called more often than necessary given the performance overhead involved. Please check this an make suggestions.

We could simply store the value of `GetItemType()` in a variable instead of calling the method over and over again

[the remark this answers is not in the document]
MichaelKetting: no, a mock doesn't change things. The problem is that we explicity added a shortcut to test for an interface common to all the collection types and return if it's not a collection type. so, you can't really "see" the extra conditionin the code, which is a typical "problem" of a performance optimizaiton. Of course, if you could observe the added IF, this would likely mean that a bug was introduced.  Considering code coverage, a test for a property of type IEnumerable could be helpful. That way, we have both branches covered (the first and the last one). The first branch is hit when a non-collection type is used, the last one, when it's a sequence and in between, for actual supported collections.

For testing of protected members,  you'd just derive and make method exposed via a public helper (unless we just use `PrivateInvole`).

Remotion/ObjectBinding/UnitTests/BindableObject/BindableObjectBaseTest.cs
```diff
@@ -31,6 +34,14 @@ public class BindableObjectBaseTest
     private Mock<IBusinessObjectProperty> _propertyFake;
     private Mock<IBusinessObjectClass> _businessObjectClassFake;
 
+    [OneTimeSetUp]
+    public void OneTimeSetup ()
+    {
+      Assertion.DebugAssert (
+          !TypeExtensions.CanAscribeTo (typeof (BindableObjectMixin), typeof (Mixin<,>)),
```
MichaelKetting, line 41: see other test

Remotion/ObjectBinding/UnitTests/BindableObject/BindableObjectWithIdentityBaseImplementationTest.cs
```diff
@@ -76,5 +79,13 @@ public void DisplayName_ViaImplementation_Overridden ()
       var implementation = BindableObjectWithIdentityBaseImplementation.Create (wrapper);
       Assert.That (implementation.DisplayName, Is.EqualTo ("Overrotten!"));
     }
+
+    [Test]
+    public void CanAscribeTo_BindableObjectWithIdentityBaseImplementation_Mixin ()
+    {
+      Assert.That (
+          !TypeExtensions.CanAscribeTo (typeof (BindableObjectWithIdentityBaseImplementation), typeof (Mixin<,>)),
```
MichaelKetting, line 87: see other test

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
```diff
@@ -46,9 +49,21 @@ public IValidatedTypeResolver InnerResolver
...
       ArgumentUtility.CheckNotNull ("collectorType", collectorType);
 
-      if (TypeExtensions.CanAscribeTo (collectorType, typeof(IValidationRuleCollector<>)))
-        return TypeExtensions.GetAscribedGenericArguments (collectorType, typeof(IValidationRuleCollector<>)).Single();
-      return _validatedTypeResolver.GetValidatedType (collectorType);
+      return s_canAscribeToGenericArgumentsOfIValidationRuleCollectorCache.GetOrAdd (
+              collectorType,
+              static (t, resolver) =>
+              {
+                var canAscribeTo = t.CanAscribeTo (typeof (IValidationRuleCollector<>));
```
MichaelKetting, line 56: We can do a test for IValidationRuleCollector first.

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
```diff
@@ -46,9 +49,21 @@ public IValidatedTypeResolver InnerResolver
     {
       ArgumentUtility.CheckNotNull ("collectorType", collectorType);
 
-      if (TypeExtensions.CanAscribeTo (collectorType, typeof(IValidationRuleCollector<>)))
-        return TypeExtensions.GetAscribedGenericArguments (collectorType, typeof(IValidationRuleCollector<>)).Single();
-      return _validatedTypeResolver.GetValidatedType (collectorType);
+      return s_canAscribeToGenericArgumentsOfIValidationRuleCollectorCache.GetOrAdd (
```
MichaelKetting, line 52: I'm not sure I understand why the IF and ELSE branch got merged into a single cache. It may be sensible to prevent multiple hits on CanAscribe for the ELSE branch. Unfortunately, it will also introduce a static cache for the result of an instance operation (_validatedTypeResolver.GetValidatedType).

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
```diff
@@ -29,6 +31,7 @@ namespace Remotion.Validation.Implementation
   public class GenericTypeAwareValidatedTypeResolverDecorator : IValidatedTypeResolver
   {
     private readonly IValidatedTypeResolver _validatedTypeResolver;
+    private static readonly ConcurrentDictionary<Type, ValueTuple<bool, Type?>> s_canAscribeToGenericArgumentsOfIValidationRuleCollectorCache = new();
```
MichaelKetting, line 34: s_genericValidationRuleCollectorTypeCache

MichaelKetting, line 34: static members are before instance members

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/DataManagement/RelationEndPoints/VirtualEndPoints/CollectionEndPoints/DomainObjectCollectionFactory.cs
```diff
@@ -30,6 +32,8 @@ namespace Remotion.Data.DomainObjects.DataManagement.RelationEndPoints.VirtualEn
   /// <seealso cref="ObjectListFactory"/>
   public class DomainObjectCollectionFactory
   {
+    private static readonly ConcurrentDictionary<Type, ValueTuple<bool, Type>> s_canAscribeToGenericArgumentsOfIEnumerableCache = new();
```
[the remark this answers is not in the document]
derlistigelurch: From my understanding `ValueTuples` do not support named values.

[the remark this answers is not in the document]
MichaelKetting: Oh yes, they do. it's just a question of syntax:
`ConcurrentDictionary<Type, (bool CanAscribe, Type ItemType)>`

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
derlistigelurch: What would you suggest?

MichaelKetting: I'm thinking we should do this in two parts. first, just introduce the cache but leave the code unchanged. Next, we should consider the second optimization step.

If we need more caching:
This decorator is configured to be only initialized once per IoC container so it may well be okay to use an instance cache instead of a static cache. We could also decide to use a multilayered cache, falling back to the static cache for the .net types and then caching per-instance the complete result.

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
derlistigelurch: What do you mean with test for `IValidationRuleCollector`?
If i check it the cache would be useless, because it would invoke `CanAscribeTo` all the time🤔

MichaelKetting: `typeof (IValidationRuleCollector).IsAssignableFrom(t) &&  t.CanAscribeTo (typeof (IValidationRuleCollector<>))`

Remotion/Data/DomainObjects/Mapping/PersistentMixinFinder.cs
```diff
@@ -38,7 +41,10 @@ public static bool IsPersistenceRelevant (Type mixinType)
     {
       ArgumentUtility.CheckNotNull ("mixinType", mixinType);
 
-      return TypeExtensions.CanAscribeTo (mixinType, typeof (DomainObjectMixin<,>));
+      if (!typeof (IDomainObjectMixin).IsAssignableFrom (mixinType))
+        return false;
+
+      return s_canAscribeToDomainObjectMixinCache.GetOrAdd (mixinType, static type => type.CanAscribeTo (typeof (DomainObjectMixin<,>)));
```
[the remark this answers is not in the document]
derlistigelurch: So if the generic test is not needed the value of the dictionary is always `true`, or would `type.CanAscribeTo (typeof (DomainObjectMixin)` be sufficient?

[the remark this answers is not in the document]
MichaelKetting: always true. Interface `IDomainObjectMixin` is a sufficient test

Remotion/Data/DomainObjects/DataManagement/RelationEndPoints/VirtualEndPoints/CollectionEndPoints/DomainObjectCollectionFactory.cs
```diff
@@ -135,10 +139,19 @@ public DomainObjectCollection CreateReadOnlyCollection (Type collectionType, IEn
...
+      return s_canAscribeToGenericArgumentsOfIEnumerableCache.GetOrAdd (
+              collectionType,
+              static type =>
+              {
+                var canAscribeTo = type.CanAscribeTo (typeof (IEnumerable<>));
+                if (!canAscribeTo)
+                  return ValueTuple.Create (canAscribeTo, (Type) null);
+
+                var ascribedGenericArgument = type.GetAscribedGenericArguments (typeof (IEnumerable<>)).Single();
+                return ValueTuple.Create (canAscribeTo, ascribedGenericArgument);
```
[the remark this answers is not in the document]
derlistigelurch: I think your solution looks better, because it's more clean and tidy

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -262,10 +267,24 @@ public static Type GetObjectListTypeParameter (Type type)
...
+              t.CanAscribeTo (typeof (ObjectList<>)),
+              t.GetAscribedGenericArguments (typeof (ObjectList<>))[0]),
+          static (k, v) =>
+          {
+            if (v.Item2 is not null)
+              return v;
+
+            return ValueTuple.Create (
+                v.Item1,
+                k.GetAscribedGenericArguments (typeof (ObjectList<>))[0]);
```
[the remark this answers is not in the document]
derlistigelurch: I can't populate the cache in one run, because some tests require `GetAscribedGenericArguments` to throw an exception in this method (`GetObjectListTypeParameter`), which only happens if `canAscribe` is `false`, therefore I can't guard this with an `if (canAscribe)` like i did in `IsObjectList` on line `160` because then it will return `null` for `Item2` aka `ItemType` and the test would fail.

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
derlistigelurch: I dropped the `if` and `else` because the cache value will be created by the `valueFactory`.

MichaelKetting: plz send meeting request :)

Remotion/Data/DomainObjects/ReflectionUtility.cs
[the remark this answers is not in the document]
MichaelKetting: The exception can be thrown after we test for canascribe==false from the cache. We only need to ensure that nothing bad happens, not that the exact same exceptions happen.

Remotion/Data/DomainObjects.ObjectBinding.UnitTests/BindableDomainObjectImplementationTest.cs
```diff
@@ -94,5 +96,13 @@ public void DisplayName_ViaImplementation_Overridden ()
       var implementation = BindableDomainObjectImplementation.Create (wrapper);
       Assert.That (implementation.DisplayName, Is.EqualTo ("TheDisplayName"));
     }
+
+    [Test]
+    public void CanAscribeTo_BindableDomainObjectImplementation_Mixin ()
+    {
+      Assert.IsTrue(
+          !typeof (BindableDomainObjectImplementation).CanAscribeTo (typeof (Mixin<,>)),
+          "we assume the mixin does not have a base object");
```
MichaelKetting, line 105: Misunderstanding:
> Please is Is.True syntax in assertion.
> Please update to use extension method.

```suggestion
      Assert.IsTrue(
          !typeof (BindableDomainObjectImplementation).CanAscribeTo (typeof (Mixin<,>)),
          "we assume the mixin does not have a base object");
```
```suggestion
      Assert.That (typeof (BindableDomainObjectImplementation).CanAscribeTo (typeof (Mixin<,>)), Is.False);
```

I've also been able to update the descriptions for better context.

Remotion/ObjectBinding/UnitTests/BindableObject/BindableObjectWithIdentityBaseImplementationTest.cs
```diff
@@ -76,5 +78,13 @@ public void DisplayName_ViaImplementation_Overridden ()
       var implementation = BindableObjectWithIdentityBaseImplementation.Create (wrapper);
       Assert.That (implementation.DisplayName, Is.EqualTo ("Overrotten!"));
     }
+
+    [Test]
+    public void CanAscribeTo_BindableObjectWithIdentityBaseImplementation_Mixin ()
+    {
+      Assert.IsTrue (
+          !typeof (BindableObjectWithIdentityBaseImplementation).CanAscribeTo (typeof (Mixin<,>)),
+          "we assume the mixin does not have a base object");
```
MichaelKetting, line 87:
```suggestion
      Assert.IsTrue (
          !typeof (BindableObjectWithIdentityBaseImplementation).CanAscribeTo (typeof (Mixin<,>)),
          "we assume the mixin does not have a base object");
```
```suggestion
      Assert.That (typeof (BindableObjectWithIdentityBaseImplementation).CanAscribeTo (typeof (Mixin<,>)), Is.False);
```

Remotion/ObjectBinding/UnitTests/BindableObject/BindableObjectWithIdentityBaseImplementationTest.cs
```diff
@@ -76,5 +78,13 @@ public void DisplayName_ViaImplementation_Overridden ()
       var implementation = BindableObjectWithIdentityBaseImplementation.Create (wrapper);
       Assert.That (implementation.DisplayName, Is.EqualTo ("Overrotten!"));
     }
+
+    [Test]
+    public void CanAscribeTo_BindableObjectWithIdentityBaseImplementation_Mixin ()
```
MichaelKetting, line 83:
```suggestion
    public void BindableObjectWithIdentityMixin_ImplementsMixinWithoutBaseObjectDependency ()
```

Remotion/Data/DomainObjects.ObjectBinding.UnitTests/BindableDomainObjectImplementationTest.cs
```diff
@@ -94,5 +96,13 @@ public void DisplayName_ViaImplementation_Overridden ()
       var implementation = BindableDomainObjectImplementation.Create (wrapper);
       Assert.That (implementation.DisplayName, Is.EqualTo ("TheDisplayName"));
     }
+
+    [Test]
+    public void CanAscribeTo_BindableDomainObjectImplementation_Mixin ()
```
MichaelKetting, line 101:
```suggestion
    public void BindableDomainObjectMixin_ImplementsMixinWithoutBaseObjectDependency ()
```

Remotion/ObjectBinding/UnitTests/BindableObject/BindableObjectBaseTest.cs
```diff
@@ -31,6 +34,14 @@ public class BindableObjectBaseTest
...
     private Mock<IBusinessObjectClass> _businessObjectClassFake;
 
+    [OneTimeSetUp]
+    public void OneTimeSetup ()
+    {
+      Assertion.IsTrue (
+          !typeof (BindableObjectMixin).CanAscribeTo (typeof (Mixin<,>)),
+          "we assume the mixin does not have a base object");
+    }
+
```
MichaelKetting, line 44:
```suggestion
    [Test]
    public void BindableObjectMixin_ImplementsMixinWithoutBaseObjectDependency ()
    {
      Assertion.That (typeof (BindableObjectMixin).CanAscribeTo (typeof (Mixin<,>)), Is.False);
    }
    
```
Please also move down to the test code.

Remotion/Mixins/Core/MixinTypeUtility.cs
```diff
@@ -188,7 +190,10 @@ public static bool HasMixin (Type targetOrConcreteType, Type mixinType)
       ArgumentUtility.CheckNotNull ("mixinType", mixinType);
 
       return GetMixinTypes (targetOrConcreteType)
-          .FirstOrDefault (configuredMixinType => Reflection.TypeExtensions.CanAscribeTo (configuredMixinType, mixinType));
+          .FirstOrDefault (
+              configuredMixinType => s_mixinTypeCache.GetOrAdd (
+                  ValueTuple.Create (configuredMixinType, mixinType),
+                  static tuple => Reflection.TypeExtensions.CanAscribeTo (tuple.Item1, tuple.Item2)));
```
MichaelKetting, line 196:
```suggestion
                  static t => t.ConfiguredMixinType.CanAscribeTo (t.MixinType)));
```

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -37,6 +39,18 @@ public static class ReflectionUtility
     private static readonly Type s_binaryPropertyValueType = typeof (byte[]);
     private static readonly Type s_typePropertyValueType = typeof (Type);
     private static readonly Type s_objectIDPropertyValueType = typeof (ObjectID);
+    private static readonly ConcurrentDictionary<Type, (bool CanAscribeTo, Type ItemType)> s_canAscribeToGenericArgumentsOfObjectListCache = new();
+
+    private static readonly Func<Type, ValueTuple<bool, Type>> s_cacheValueFactory =
+        static t =>
+        {
+          var canAscribeTo = t.CanAscribeTo (typeof (IReadOnlyList<>)) && t.CanAscribeTo (typeof (ObjectList<>));
```
MichaelKetting, line 47:
```suggestion
          var canAscribeTo = typeof (IReadOnlyList<DomainObject>).IsAssignableFrom (t) && t.CanAscribeTo (typeof (ObjectList<>));
```

The idea was that we can short curcuit the expesive check for types that arent actually in the running. And while I'm at it, we should use `typeof (DomainObjectCollection)` instead of `typeof (IReadOnlyList<DomainObject>)` here, since this the better/closer match. I probably didn't think of this the first time.

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -37,6 +39,18 @@ public static class ReflectionUtility
     private static readonly Type s_binaryPropertyValueType = typeof (byte[]);
     private static readonly Type s_typePropertyValueType = typeof (Type);
     private static readonly Type s_objectIDPropertyValueType = typeof (ObjectID);
+    private static readonly ConcurrentDictionary<Type, (bool CanAscribeTo, Type ItemType)> s_canAscribeToGenericArgumentsOfObjectListCache = new();
```
MichaelKetting, line 42:
```suggestion
    private static readonly ConcurrentDictionary<Type, (bool CanAscribeTo, Type ItemType)> s_objectListTypeCache = new();
```

Remotion/Mixins/Core/MixinTypeUtility.cs
```diff
@@ -38,6 +38,8 @@ public static class MixinTypeUtility
     private static readonly ConcurrentDictionary<Type, ReadOnlyCollection<Type>> s_exactMixinTypesCache =
         new ConcurrentDictionary<Type, ReadOnlyCollection<Type>>();
 
+    private static readonly ConcurrentDictionary<ValueTuple<Type, Type>, bool> s_mixinTypeCache = new();
```
MichaelKetting, line 41: We should make this named, too.
```suggestion
    private static readonly ConcurrentDictionary<(Type ConfiguredMixinType, Type MixinType), bool> s_mixinTypeCache = new();
```

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -262,10 +276,14 @@ public static Type GetObjectListTypeParameter (Type type)
     {
       ArgumentUtility.CheckNotNull ("type", type);
 
-      var typeParameters = type.GetAscribedGenericArguments (typeof (ObjectList<>));
-      var typeParameter = typeParameters[0];
+      var typeParameter = s_objectListTypeCache.GetOrAdd (type, s_cacheValueFactory).ItemType;
+
+      if(typeParameter is null)
```
MichaelKetting, line 281:
```suggestion
      if (typeParameter is null)
```

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -37,6 +39,18 @@ public static class ReflectionUtility
     private static readonly Type s_binaryPropertyValueType = typeof (byte[]);
     private static readonly Type s_typePropertyValueType = typeof (Type);
     private static readonly Type s_objectIDPropertyValueType = typeof (ObjectID);
+    private static readonly ConcurrentDictionary<Type, (bool CanAscribeTo, Type ItemType)> s_objectListTypeCache = new();
+
+    private static readonly Func<Type, ValueTuple<bool, Type>> s_cacheValueFactory =
```
MichaelKetting, line 44:
```suggestion
    private static readonly Func<Type, ValueTuple<bool, Type>> s_objectListTypeCacheValueFactory =
```

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -262,10 +276,14 @@ public static Type GetObjectListTypeParameter (Type type)
     {
       ArgumentUtility.CheckNotNull ("type", type);
 
-      var typeParameters = type.GetAscribedGenericArguments (typeof (ObjectList<>));
-      var typeParameter = typeParameters[0];
+      var typeParameter = s_objectListTypeCache.GetOrAdd (type, s_cacheValueFactory).ItemType;
```
MichaelKetting, line 279:
```suggestion
      var typeParameter = s_objectListTypeCache.GetOrAdd (type, s_objectListTypeCacheValueFactory).ItemType;
```

Remotion/Data/DomainObjects/ReflectionUtility.cs
```diff
@@ -154,7 +168,7 @@ public static bool IsObjectList (Type type)
     {
       ArgumentUtility.CheckNotNull ("type", type);
 
-      return type.CanAscribeTo (typeof (ObjectList<>));
+      return s_objectListTypeCache.GetOrAdd (type, s_cacheValueFactory).CanAscribeTo;
```
MichaelKetting, line 171:
```suggestion
      return s_objectListTypeCache.GetOrAdd (type, s_objectListTypeCacheValueFactory).CanAscribeTo;
```

Remotion/Validation/Core/Implementation/GenericTypeAwareValidatedTypeResolverDecorator.cs
```diff
@@ -46,9 +49,20 @@ public IValidatedTypeResolver InnerResolver
...
+                var canAscribeTo = typeof (IValidationRuleCollector).IsAssignableFrom (t) && t.CanAscribeTo (typeof (IValidationRuleCollector<>));
+                return ValueTuple.Create (
+                    canAscribeTo,
+                    canAscribeTo
+                        ? t.GetAscribedGenericArguments (typeof (IValidationRuleCollector<>))[0]
+                        : null);
+              })
+          .ItemType;
+
+      return itemType ?? _validatedTypeResolver.GetValidatedType (collectorType);
```
MichaelKetting, line 65: please split
```suggestion
      if (itemType != null)
        return itemType 

      return _validatedTypeResolver.GetValidatedType (collectorType);
```

This provides better results during coverage analysis

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #219, opened 2021-10-05
RM-8047 ContextAwareTypeUtility can fail during static initialization due to recursive callchain

derlistigelurch:
https://re-motion.atlassian.net/browse/RM-8052
https://re-motion.atlassian.net/browse/RM-8047

### DIFF
--- a/Remotion/Core/Core/Reflection/ContextAwareTypeUtility.cs
+++ b/Remotion/Core/Core/Reflection/ContextAwareTypeUtility.cs
@@ -34,11 +34,13 @@ namespace Remotion.Reflection
     private class Fields
     {
       public readonly Lazy<ITypeDiscoveryService> DefaultTypeDiscoveryService =
-          new Lazy<ITypeDiscoveryService> (TypeDiscoveryConfiguration.Current.CreateTypeDiscoveryService, 
+          new Lazy<ITypeDiscoveryService> (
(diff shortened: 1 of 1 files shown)


Remotion/Core/Core/Reflection/ContextAwareTypeUtility.cs
```diff
@@ -42,7 +42,9 @@ private class Fields
               LazyThreadSafetyMode.ExecutionAndPublication);
     }
 
-    private static readonly Fields s_fields = new Fields();
+    private static readonly Lazy<Fields> s_lazyFields = new(LazyThreadSafetyMode.ExecutionAndPublication);
```
iam-mholle, line 45:
```suggestion
    private static readonly Lazy<Fields> s_lazyFields = new (LazyThreadSafetyMode.ExecutionAndPublication);
```

iam-mholle (approved):

iam-mholle:
FYI this PR targets the duplication, not the "main" issue which would be https://re-motion.atlassian.net/browse/RM-8047.

MichaelKetting (changes requested):
Please switch to 8047 for documentation purporses.

The fix itself targets the wrong "level". You need too decouple the intiailization of the members in `Fields`. 

Said another way, TypeDiscoveryConfiguration.Current and TypeResolutionConfiguration.Current must be prevented to execute in the same call. 

See the description in RM-8047

MichaelKetting (approved):

PR #224, opened 2021-10-13
RM-8066 Remove general jQuery usages from BocAutoComplete

a-ctor:
https://re-motion.atlassian.net/browse/RM-8066

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
@@ -119,9 +119,6 @@ type BocAutoCompleteReferenceValue_PositioningOptions = {
 
 interface JQueryStatic {
     Autocompleter: BocAutoCompleteReferenceValue_AutocompleterStatic;
-
-    data(el: Element, key: "ac_data"): BocAutoCompleteReferenceValue_CacheRowEntry;
-    data(el: Element, key: "ac_data", value: BocAutoCompleteReferenceValue_CacheRowEntry): void;
 }
 
 interface JQuery {
@@ -175,9 +172,9 @@ type BocAutoCompleteReferenceValue_AutocompleterStatic = {
     Cache: typeof BocAutoCompleteReferenceValue_Cache;
     Select: typeof BocAutoCompleteReferenceValue_Select;
     InformationPopUp: typeof BocAutoCompleteReferenceValue_InformationPopUp;
-    Selection(field: HTMLTextAreaElement, start: number, end: number, focusInputAfterSelection: boolean): void;
-    calculateSpaceAround(element: JQuery): BocAutoCompleteReferenceValue_SpaceAround;
-    applyPositionToPopUp(reference: JQuery, popUp: JQuery, options: BocAutoCompleteReferenceValue_PositioningOptions): void;
+    Selection(field: HTMLInputElement, start: number, end: number, focusInputAfterSelection: boolean): void;
+    calculateSpaceAround(element: HTMLElement): BocAutoCompleteReferenceValue_SpaceAround;
+    applyPositionToPopUp(reference: HTMLElement, popUp: HTMLElement, options: BocAutoCompleteReferenceValue_PositioningOptions): void;
 };
 
 type BocAutoCompleteReferenceValue_CollectedElements = {
@@ -243,7 +240,6 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
             serviceMethodSearch: string,
             serviceMethodSearchExact: string,
             initialOptions: PartialWithRequiredProperties<BocAutoCompleteReferenceValue_Options, keyof BocAutoCompleteReferenceValue_RequiredOptions>): JQuery {
-            const $input = $(this);
             // This assignment builds the final options object and also ensures that all the option types are set up correctly
             const options: BocAutoCompleteReferenceValue_Options = Object.assign({}, defaultOptions, {
                 // re-motion: instead of a single URL property, use separate service URL and service method properties. 
@@ -309,7 +305,7 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
         }
     });
 
-    $.Autocompleter = <any>function(input: HTMLTextAreaElement, options: BocAutoCompleteReferenceValue_Options) {
+    $.Autocompleter = <any>function(input: HTMLInputElement, options: BocAutoCompleteReferenceValue_Options) {
 
       const KEY = {
             SHIFT: 16,
@@ -331,7 +327,7 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
         };
 
         // Create $ object for input element
-        const $input = Remotion.jQuery (input).addClass(options.inputClass);
+        input.classList.add(options.inputClass);
         //aria-controls has been removed since it does not provide a meaningful assistance at this point in time (IE 11 / JAWS 18)
         //$input.attr ({ "aria-controls" : options.selectListID });
 
@@ -359,7 +355,7 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
             mouseDownOnSelect: false,
             // holds the last text the user entered into the input element
             previousValue: '',
-            lastKeyPressValue: null
+            lastKeyPressValue: null as Nullable<string>
         };
 
         const select = new BocAutoCompleteReferenceValue_Select(options, input, selectCurrent, state);
@@ -371,17 +367,17 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
         const informationPopUp = new BocAutoCompleteReferenceValue_InformationPopUp(options, input);
         let blockSubmit: Optional<boolean>;
 
-        $input.bind("keydown.autocomplete", function(event) {
+        $(input).bind("keydown.autocomplete", function(event) {
             // track last key pressed
             state.lastKeyPressCode = event.keyCode;
             clearTimeout(timeout);
             // re-motion: cancel an already running request
             stopLoading();
             abortRequest();
-            if (state.lastKeyPressValue !== null && state.lastKeyPressValue != $input.val()) {
+            if (state.lastKeyPressValue !== null && state.lastKeyPressValue != input.value) {
                 invalidateResult();
             }
-            state.lastKeyPressValue = $input.val();
+            state.lastKeyPressValue = input.value;
 
             switch (event.keyCode)
             {
@@ -393,7 +389,7 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
                         options.clearRequestError();
                         select.prev();
                     } else {
-                        onChange(true, $input.val());
+                        onChange(true, input.value);
                     }
                     return;
 
@@ -405,7 +401,7 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
                         options.clearRequestError();
                         select.next();
                     } else {
-                        onChange(true, $input.val());
+                        onChange(true, input.value);
                     }
                     return;
 
@@ -417,7 +413,7 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
                         options.clearRequestError();
                         select.pageUp();
                     } else {
-                        onChange(true, $input.val());
+                        onChange(true, input.value);
                     }
                     return;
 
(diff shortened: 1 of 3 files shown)


Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -396,21 +406,40 @@ class LayoutUtility
...
+    // For border-box: we need to remove padding/border
+    if (style.boxSizing !== "border-box")
+    {
+      if (extra === undefined)
+        return value;
+
+      let adjustment = 0.0;
+      for (const orientation of orientations)
+      {
+        if (extra === "padding")
```
MichaelKetting, line 423: Am I missing the border?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -371,17 +367,17 @@ function NotNullAssert(v: unknown): asserts v is NotNull {}
         const informationPopUp = new BocAutoCompleteReferenceValue_InformationPopUp(options, input);
         let blockSubmit: Optional<boolean>;
 
-        $input.bind("keydown.autocomplete", function(event) {
+        $(input).bind("keydown.autocomplete", function(event) {
```
MichaelKetting, line 370: Replacing the event wiring will be done together with replacing the jquery-hookup?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1134,78 +1132,99 @@ class BocAutoCompleteReferenceValue_Select {
...
             if (this.target(event).nodeName && this.target(event).nodeName.toUpperCase() === 'LI')
             {
-                this.active = Remotion.jQuery("li", this.list).removeClass(this.CLASSES.ACTIVE).index(this.target(event));
-                const activeItem = Remotion.jQuery(this.target (event));
-                activeItem.addClass(this.CLASSES.ACTIVE);
+                const listElements = Array.from(this.list.querySelectorAll<HTMLElement>("li"));
+                for (const listElement of listElements)
+                    listElement.classList.remove(this.CLASSES.ACTIVE);
+
+                const activeItem = this.target (event);
```
MichaelKetting, line 1201:
```suggestion
                const activeItem = this.target(event);
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -1682,88 +1733,86 @@ Remotion.jQuery.Autocompleter.calculateSpaceAround = function(element: JQuery):
...
+    let contentHeight = Math.max(0, Math.max(...(Array.from(popUpDiv.children) as HTMLElement[]).map(function (el) { return el.offsetHeight + el.offsetTop; })));
 
-    let contentWidth = popUp.data('popUpContentWidth');
+    let contentWidth = parseInt(popUp.dataset['popUpContentWidth']!, 10);
     if (!isVisibe)
     {
-        contentWidth = Math.max(0, Math.max(...popUpDiv.children().map(function (this: HTMLElement) { return this.offsetWidth + this.offsetLeft; }).get()));
-        popUp.data('popUpContentWidth', contentWidth);
+        contentWidth = Math.max(0, Math.max(...(Array.from(popUpDiv.children) as HTMLElement[]).map(function (el) { return el.offsetWidth + el.offsetLeft; })));
+        popUp.dataset['popUpContentWidth'] = "" + contentWidth;
```
MichaelKetting, line 1755:
```suggestion
        popUp.dataset['popUpContentWidth'] = '' + contentWidth;
```
for consistency

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -439,8 +468,32 @@ class LayoutUtility
...
+    // parseInt ignores the "px" suffix so we do not have to remove here
+    return element.offsetWidth + parseInt(style.marginLeft) + parseInt(style.marginRight);
+  }
+
   public static IsVisible(element: HTMLElement)
   {
     return element.offsetWidth !== 0 || element.offsetHeight !== 0 || element.getClientRects().length !== 0;
   }
+
+  public static Hide(element: HTMLElement)
```
MichaelKetting, line 485: We should use these methods in the rest of the JS as well. (separate task)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
```diff
@@ -119,9 +119,6 @@ type BocAutoCompleteReferenceValue_PositioningOptions = {
 
 interface JQueryStatic {
```
MichaelKetting, line 120: I think we should replace the jquery type with a custom type that only exposes the apis we still "support" at this time. Unless that's really a lot of work instead of just an hour or so.

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -396,21 +406,40 @@ class LayoutUtility
...
-    if (style.boxSizing !== "border-box")
-      return value;
-    
     const orientations: CssOrientation[] = dimension === "width"
       ? ["Left", "Right"]
       : ["Top", "Bottom"];
 
+    // For content-box: value is already the width/height
+    // For border-box: we need to remove padding/border
+    if (style.boxSizing !== "border-box")
```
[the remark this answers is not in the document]
a-ctor: This way it also handles a case in which an invalid/no value was specified and `content-box` is used as default

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: No because there are already methods that return the size with border and padding, here we just want the padding.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
a-ctor: Yes this will be done in the follow up task

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
a-ctor: The jquery type is removed in the follow up issue. Most likely, it is going to be replaced by a class that inherits from an HTML element. So this should resolve itself.

Remotion/Web/ClientScript/Scripts/Utilities.ts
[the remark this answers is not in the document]
MichaelKetting: okay, then we should do an IF/ELSE and please switch so have a TRUE condition in the IF instead of a FALSE condition.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.jquery.ts
MichaelKetting: My idea was, to make sure with typesafety that we catch all cleanups that we want to do here. What process did you use to ensure that we don't have a jquery.getouterwidth() or jquery.data() or some other jquery function not releated to wiring-up or events left in the code?

a-ctor: I have removed all mentions of `$` and `jQuery` from the file, going through each and evaluating if it can be replaced. The only jQuery left are the arguments passed into the autocomplete function and the jQuery construct itself. To check I also removed the jQuery types and checked each of the ocurring errors since those are related to jQuery.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -377,44 +377,131 @@ class LayoutUtility
...
+
+    htmlElement.style.position = "absolute";
+    htmlElement.style.visibility = "hidden";
+    htmlElement.style.display = "block";
+
+    const result = this.GetElementDimension(htmlElement, dimension, extra);
+
+    htmlElement.style.position = previousPosition;
+    htmlElement.style.visibility = previousVisibility;
+    htmlElement.style.display = previousDisplay;
```
MichaelKetting, line 427: Okay, you're forcing a re-draw here. I'm not sure that this doesn't result in flicker-sideeffects in the UI. There's definitely an effect to the DOM or updating the properties wouldn't be needed.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -377,44 +377,131 @@ class LayoutUtility
...
+  {
+    return this.GetElementDimensionWithElementShow(element, "height", "border");
+  }
+
+  public static GetOuterWidth(element: HTMLElement)
+  {
+    return this.GetElementDimensionWithElementShow(element, "width", "border");
+  }
+
+  private static GetElementDimensionWithElementShow(element: Element, dimension: "width" | "height", extra?: "padding" | "border")
```
MichaelKetting, line 408: Please rename for better understanding of the name. `...WithElementShow` doesn't really help me understand what the method does.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -377,44 +377,131 @@ class LayoutUtility
...
+
+        // Subtract border
+        const borderProperty = `border${orientation}Width` as `border${CssOrientation}Width`;
+        adjustment -= parseFloat(style[borderProperty]) || 0;
+      }
+
+      return offsetValue + adjustment;
+    }
+
+    const value = parseFloat(style[dimension]) || 0;
```
MichaelKetting, line 464: I would prefer to start an else block here. The IF isn't just a guard condition for an early exit but rather an entire branch of exeution logic so the two blocks should be connected.

Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -377,44 +377,131 @@ class LayoutUtility
...
-    // For content-box: value is already correct
-    // For border-box: we need to remove padding and border from the value
-    if (style.boxSizing !== "border-box")
-      return value;
-    
     const orientations: CssOrientation[] = dimension === "width"
       ? ["Left", "Right"]
       : ["Top", "Bottom"];
 
+    const offsetValue = dimension === "width" ? element.offsetWidth : element.offsetHeight;
```
MichaelKetting, line 440: I would like to discuss the logic of this code, in particular offsetValue>0 case and how this all works. I'm not sure if this means we need some code comments to explain the algorithm.

I'm also wondering if the offsetValue>0 and the style.boxSizing === "border-box" blocks should be unified or not. They are identical, except for the seed value but I don't know if this is by accident or by design.

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: That code was built to mirror the behavior of jQuery. Since we set and unset in the same JS call it should not redraw while in `this.GetElementDimension(...)`

Remotion/Web/ClientScript/Scripts/Utilities.ts
a-ctor: If an element is not visible its `offsetXXX` value will return `0`. If this is the case we need to calculate the dimension from the computed style since that should still provide the correct values.

I don;t understand what you mean with "offsetValue>0 and the style.boxSizing === "border-box" blocks should be unified or not"

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: okay, great.

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: `offsetValue>0` please make this into a named variable and add a short reasinign comment. There's also the question now, why the invisible element doesn't need to differentiate between border-box and content-box.

unify: lines 443-461 and and 471-486 look identical to me. if we unify the `offsetValue` and `value` properties inside the relavant code paths, we could merge the two sections. then we have a signle IF branch for not-visible || border-box and and ELSE for the rest.

Question is: good idea or not (and why)

MichaelKetting (approved):
Good. I can read it and I'll assume there's no bugs in in the layout calc :)

PR #235, opened 2021-10-29
RM-7595: Integrate NRT in ObjectBinding Web

iam-mholle:


Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
```diff
@@ -163,47 +171,52 @@ private void EnsureBusinessObject()
     }
     else
     {
-      IBusinessObjectReferenceProperty property = Property;
+      IBusinessObjectReferenceProperty? property = Property;
       string businessObjectID = ItemID;
       BusinessObject = ((IBusinessObjectClassWithIdentity) property.ReferenceClass).GetObject (businessObjectID);
```
MichaelKetting, line 176: Why does this compile?

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
```diff
@@ -220,14 +233,16 @@ public class BusinessObjectPropertyTreeNode: BocTreeNode
   /// <summary>
   ///   The <see cref="IBusinessObjectReferenceProperty"/> of this <see cref="BusinessObjectPropertyTreeNode"/>.
   /// </summary>
+  [AllowNull]
```
MichaelKetting, line 236: why?

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
```diff
@@ -239,17 +254,20 @@ protected override string DisplayedTypeName
...
+    BusinessObjectTreeNode? parentNode = (BusinessObjectTreeNode?) ParentNode;
     if (parentNode == null)
       throw new InvalidOperationException ("BusinessObjectPropertyTreeNode with ItemID '" + ItemID + "' has no parent node but property nodes cannot be used as root nodes.");
 
-    IBusinessObjectProperty property = parentNode.BusinessObject.BusinessObjectClass.GetPropertyDefinition (ItemID);
-    Property = (IBusinessObjectReferenceProperty) property;
+    Assertion.IsNotNull (parentNode.BusinessObject, "parentNode.BusinessObject must not be null.");
+
+    IBusinessObjectProperty? property = parentNode.BusinessObject.BusinessObjectClass.GetPropertyDefinition (ItemID);
+    _property = (IBusinessObjectReferenceProperty) Assertion.IsNotNull (property, "No property definition exists for the id '{0}'.", ItemID);
```
MichaelKetting, line 270: `"Property '{0}' does not exist on class '{1}'.", ItemID, parentNode.BusinessObject.BusinessObjectClass.Identifer);

Remotion/Web/Core/UI/Controls/Command.cs
```diff
@@ -970,6 +970,7 @@ public virtual WxeFunctionCommandInfo WxeFunctionCommand
     [DesignerSerializationVisibility (DesignerSerializationVisibility.Hidden)]
     [Browsable (false)]
     [EditorBrowsable (EditorBrowsableState.Advanced)]
+    [AllowNull]
```
MichaelKetting, line 973: Okay, (see also the other comments), now I remember AllowNull. Can we drop this and just keep the null-coalesce. It probalby makes things less ambiguous going forward.

MichaelKetting (commented):
partial review. most of the control types are left to do and a few other classes.
112 files are reviewed.

Remotion/ObjectBinding/Web/Services/BusinessObjectWithIdentityProxy.cs
```diff
@@ -46,18 +47,21 @@ public BusinessObjectWithIdentityProxy (IBusinessObjectWithIdentity obj)
...
 
-    public string DisplayName
+    [DisallowNull]
+    public string? DisplayName
     {
       get { return _displayName; }
       set { _displayName = ArgumentUtility.CheckNotNullOrEmpty ("value", value); }
     }
 
+    [AllowNull]
```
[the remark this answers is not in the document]
iam-mholle: What guard? I added the attribute because the setter handles `null` values, even if its not needed. But it would make sense to drop it, together with the guard + add an argument check.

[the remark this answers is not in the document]
MichaelKetting: Sorry, I meant "fallback" not guard, my bad. My reasoning was, that there's no need to have the AllowNull since there's no observable reason why someone needs to assign null here. That's my reasoning but it may be flawed :)

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
```diff
@@ -322,14 +324,19 @@ protected object GetValue ()
...
-    protected void SetValue (object value)
+    protected void SetValue (object? value)
     {
       _value = value;
 
       if (Property != null && _value != null)
-        _enumerationValueInfo = Property.GetValueInfoByValue (_value, GetBusinessObject());
+      {
+        var businessObject = Assertion.IsNotNull (GetBusinessObject(), "GetBusinessObject() must not return a null value.");
+        _enumerationValueInfo = Property.GetValueInfoByValue (_value, businessObject);
```
MichaelKetting, line 334: the businessObject parameter is nullable now.
```suggestion
        _enumerationValueInfo = Property.GetValueInfoByValue (_value, GetBusinessObject());
```

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
```diff
@@ -388,7 +395,7 @@ protected void EnsureValue ()
       else if (_internalValue != null && Property != null)
       {
         //  Can get a new EnumerationValueInfo
-        _enumerationValueInfo = Property.GetValueInfoByIdentifier (_internalValue, GetBusinessObject());
+        _enumerationValueInfo = Property.GetValueInfoByIdentifier (_internalValue, GetBusinessObject()!)!;
```
MichaelKetting, line 398:
```suggestion
        _enumerationValueInfo = Property.GetValueInfoByIdentifier (_internalValue, GetBusinessObject());
```

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
```diff
@@ -650,12 +657,12 @@ protected override void LoadResources (IResourceManager resourceManager, IGlobal
...
     /// </value>
     /// <remarks> Only used to simplify access to the <see cref="IEnumerationValueInfo"/>. </remarks>
-    protected IEnumerationValueInfo EnumerationValueInfo
+    protected IEnumerationValueInfo? EnumerationValueInfo
     {
       get
       {
         if (_enumerationValueInfo == null && Property != null && _value != null)
-          _enumerationValueInfo = Property.GetValueInfoByValue (_value, GetBusinessObject());
+          _enumerationValueInfo = Property.GetValueInfoByValue (_value, GetBusinessObject()!);
```
MichaelKetting, line 665:
```suggestion
          _enumerationValueInfo = Property.GetValueInfoByValue (_value, GetBusinessObject());
```

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
```diff
@@ -683,10 +690,12 @@ private IEnumerationValueInfo[] GetEnabledValues ()
     {
       if (Property == null)
         return new IEnumerationValueInfo[0];
-      return Property.GetEnabledValues (GetBusinessObject());
+
+      var businessObject = Assertion.IsNotNull (GetBusinessObject(), "GetBusinessObject() must not be null.");
+      return Property.GetEnabledValues (businessObject);
```
MichaelKetting, line 695:
```suggestion
      return Property.GetEnabledValues (GetBusinessObject());
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditModeController.cs
```diff
@@ -68,15 +68,17 @@ public EditModeController (IEditModeHost editModeHost)
...
       if (_editModeHost.Value == null || index >= _editModeHost.Value.Count)
         throw new ArgumentOutOfRangeException ("index", "The index must not point to an object past the elements in the Value collection");
 
       if (_editMode == EditMode.None)
         return null;
 
-      var editedRowID = _editModeHost.RowIDProvider.GetItemRowID (new BocListRow (index, (IBusinessObject) _editModeHost.Value[index]));
+      Assertion.IsNotNull (_editedRowIDs, "_editedRowIDs must not be null.");
+
+      var editedRowID = _editModeHost.RowIDProvider.GetItemRowID (new BocListRow (index, (IBusinessObject) _editModeHost.Value[index]!));
```
MichaelKetting, line 81: Value nullablity will need cleanup after rebase on develop

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditModeController.cs
```diff
@@ -256,14 +260,19 @@ public void EndListEditMode (bool saveChanges, BocColumnDefinition[] columns)
...
         var editedRows = _editedRowIDs.Select (
             (editedRowID, index) =>
             {
               var bocListRow = _editModeHost.RowIDProvider.GetRowFromItemRowID (_editModeHost.Value, editedRowID);
 
               if (bocListRow == null)
-                return Tuple.Create (-1, _rows[index].GetDataSource().BusinessObject);
+              {
+                var businessObject = _rows[index].GetDataSource().BusinessObject;
+                Assertion.IsNotNull (businessObject, "The business object bound to index {0} must not be null.", index);
```
MichaelKetting, line 273:
```suggestion
                Assertion.IsNotNull (businessObject, "The business object at row index {0} must not be null.", index);
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditModeController.cs
```diff
@@ -375,8 +386,11 @@ public void SynchronizeEditModeControls (BocColumnDefinition[] columns)
       if (!IsListEditModeActive)
         return new Tuple<BocListRow, EditableRow>[0];
 
+      Assertion.IsNotNull (_editedRowIDs, "_editedRowIDs must not be null.");
+      Assertion.IsNotNull (_editModeHost.Value, "_editModeHost.Value must not be null.");
+
       var availableRows = _editModeHost.Value.Cast<IBusinessObject>().Select ((o, i) => new BocListRow (i, o)).ToList();
-      var editedRows = _editedRowIDs.Select (rowID => _editModeHost.RowIDProvider.GetRowFromItemRowID (_editModeHost.Value, rowID)).ToList();
+      var editedRows = _editedRowIDs.Select (rowID => _editModeHost.RowIDProvider.GetRowFromItemRowID (_editModeHost.Value, rowID)!).ToList();
```
MichaelKetting, line 393: there's some scary nullability stuff going on in this type and in EditableRow. maybe we should add an improvement ticket?

Remotion/ObjectBinding/Web/UI/Controls/BocListView.cs
```diff
@@ -64,9 +65,9 @@ protected override void OnOwnerControlChanged()
     _columnDefinitions.OwnerControl = OwnerControl;
   }
 
-  public override string ToString()
+  public override string? ToString()
```
MichaelKetting, line 68:
```suggestion
  public override string ToString()
```

Remotion/ObjectBinding/Web/UI/Controls/BocListView.cs
```diff
@@ -25,18 +26,18 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
 [ParseChildren (true, "ColumnDefinitionCollection")]
 public class BocListView: BusinessObjectControlItem
 {
-  private string _itemID;
-  private object _title;
+  private string? _itemID;
+  private object? _title;
```
MichaelKetting, line 30: unrelated bug: I think it's a bug that title's an object, not a string.

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
```diff
@@ -929,11 +929,13 @@ bool IBocReferenceValueBase.IsIconEnabled()
...
     }
 
-    IconInfo IBocReferenceValueBase.GetIcon ()
+    IconInfo? IBocReferenceValueBase.GetIcon ()
     {
       var businessObjectClass = GetBusinessObjectClass ();
       if (businessObjectClass == null)
         return null;
+
+      Assertion.IsNotNull (Value, "Value must not be null.");
```
MichaelKetting, line 938:
```suggestion
```
GetIcon supports null

Remotion/ObjectBinding/Web/IBusinessObjectWebUIService.cs
```diff
@@ -44,7 +44,7 @@ public interface IBusinessObjectWebUIService : IBusinessObjectService
     HelpInfo GetHelpInfo (
```
MichaelKetting, line 44: GetIcon and GetToolTip support null values.

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
```diff
@@ -888,25 +888,25 @@ protected virtual void PreRenderOptionsMenu ()
     ///     The default implementation used the <see cref="IBusinessObjectWithIdentity.DisplayName"/> property to get the display name.
     ///   </para>
     /// </remarks>
-    protected virtual string GetDisplayName (IBusinessObjectWithIdentity businessObject)
+    protected virtual string? GetDisplayName (IBusinessObjectWithIdentity businessObject)
```
MichaelKetting, line 891:
```suggestion
    protected virtual string GetDisplayName (IBusinessObjectWithIdentity businessObject)
```
plz check

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
```diff
@@ -888,25 +888,25 @@ protected virtual void PreRenderOptionsMenu ()
...
     {
       ArgumentUtility.CheckNotNull ("businessObject", businessObject);
       return businessObject.GetAccessibleDisplayName();
     }
 
-    private bool IsNullValue (string newValue)
+    private bool IsNullValue (string? newValue)
```
MichaelKetting, line 897:
```suggestion
    private bool IsNullValue (string newValue)
```

MichaelKetting (commented):
Todo: BocList, + 2nd review of EditModeControl, EditableRow

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -1035,7 +1040,7 @@ protected override void OnPreRender (EventArgs e)
         if (currentRow == null)
           _newPageIndex = null;
         else
-          _newPageIndex = currentRow.SortedIndex / _pageSize.Value;
```
MichaelKetting, line 1038: can we link _pageSize != null with IsPagingEnabled = true?

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -2383,7 +2401,7 @@ private IEnumerable<BocListRow> GetSelectedRowsInternal ()
         return Enumerable.Empty<BocListRow>();
 
       return _selectorControlCheckedState
-          .Select (rowID => RowIDProvider.GetRowFromItemRowID (Value, rowID))
+          .Select (rowID => RowIDProvider.GetRowFromItemRowID (Value, rowID)!)
```
MichaelKetting, line 2404: Why the `!`?

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -2431,7 +2449,7 @@ public void SetSelectedRows (int[] selectedRows)
         }
       }
 
-      SetSelectedRows (selectedRows.Select (rowIndex => new BocListRow (rowIndex, (IBusinessObject) Value[rowIndex])).ToArray());
+      SetSelectedRows (selectedRows.Select (rowIndex => new BocListRow (rowIndex, (IBusinessObject) Value[rowIndex]!)).ToArray());
```
MichaelKetting, line 2452: The thing with `Value being IList` that will go away with the rebase

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditModeController.cs
```diff
@@ -209,6 +211,8 @@ public void EndRowEditMode (bool saveChanges, BocColumnDefinition[] columns)
       {
         var editedRow = GetEditedRow();
 
+        Assertion.DebugIsNotNull (_rows[0].GetDataSource(), "The data source of _rows[0] must not be null.");
```
MichaelKetting, line 214: What's the pattern for DebugIsNotNull between using a verbose msg vs letting Rider insert a default `_rows[0].GetDataSource() != null` message?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -234,13 +240,15 @@ protected void AddToValidators (int columnIndex, IEnumerable<BaseValidator> vali
     {
       ArgumentUtility.CheckNotNull ("validators", validators);
 
-      ControlCollection cellValidators = GetValidators (columnIndex);
+      ControlCollection? cellValidators = GetValidators (columnIndex);
```
MichaelKetting, line 243: I think we should add an assertion here and then we can drop the `!` two lines down.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -283,7 +292,7 @@ protected void CreateValidators (int columnIndex)
     {
       if (HasEditControl (columnIndex))
       {
-        IBusinessObjectBoundEditableWebControl editControl = GetEditControl (columnIndex);
+        IBusinessObjectBoundEditableWebControl editControl = GetEditControl (columnIndex)!;
```
MichaelKetting, line 295: Assertion

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -302,13 +311,15 @@ protected void PrepareValidation (int columnIndex)
...
```
MichaelKetting, line 314: Assertion

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -326,7 +337,7 @@ protected bool Validate (int columnIndex)
 
       if (HasValidators (columnIndex))
       {
-        ControlCollection cellValidators = GetValidators (columnIndex);
+        ControlCollection cellValidators = GetValidators (columnIndex)!;
```
MichaelKetting, line 340: Assertion

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -356,7 +367,7 @@ public string[] GetTrackedClientIDs()
     protected string[] GetTrackedClientIDs (int columnIndex)
     {
       if (HasEditControl (columnIndex))
-        return GetEditControl (columnIndex).GetTrackedClientIDs();
+        return GetEditControl (columnIndex)!.GetTrackedClientIDs();
```
MichaelKetting, line 370: Assertion

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -377,21 +388,23 @@ public bool IsDirty()
     protected bool IsDirty (int columnIndex)
     {
       if (HasEditControl (columnIndex))
-        return GetEditControl (columnIndex).IsDirty;
+        return GetEditControl (columnIndex)!.IsDirty;
```
MichaelKetting, line 391: Assertion

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -377,21 +388,23 @@ public bool IsDirty()
...
       else
         return false;
     }
 
     public bool IsRequired (int columnIndex)
     {
       if (HasEditControl (columnIndex))
-        return GetEditControl (columnIndex).IsRequired;
+        return GetEditControl (columnIndex)!.IsRequired;
```
MichaelKetting, line 399: Assertion

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -410,8 +423,9 @@ public IBusinessObjectBoundEditableWebControl[] GetEditControlsAsArray()
       if (! HasEditControl (columnIndex))
         return;
   
-      ControlCollection validators = GetValidators (columnIndex);
+      ControlCollection validators = GetValidators (columnIndex)!;
```
MichaelKetting, line 426: Assertion (I hope that's correct))

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
MichaelKetting: In Data, we have the pattern of appending "Mandatory" for a not-null variant of an api. Maybe it's a good idea (follow-up task) to add private methods for GetEditControl and GetValidators that return not-null values and check with an exception if it is null. That way, we can get rid of the need for inline-assertions.

-> GetEditControlMandatory(...)
-> GetValidatorsMandatory(...)

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocDateTimeValueValidatorBase.cs
```diff
@@ -54,6 +55,7 @@ protected override bool ControlPropertiesValid ()
     }
 
     /// <summary> Gets or sets the input control to validate. </summary>
+    [AllowNull]
     public new string ControlToValidate
```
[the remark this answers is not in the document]
iam-mholle: Because `?` sets setter and getter both nullable. In this case the setter accepts null, but the returned value is never null.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
```diff
@@ -91,14 +94,16 @@ public IBusinessObjectWithIdentity BusinessObject
   ///   The <see cref="IBusinessObjectReferenceProperty"/> that was used to access the 
   ///   <see cref="IBusinessObjectWithIdentity"/> of this <see cref="BusinessObjectTreeNode"/>.
   /// </summary>
+  [AllowNull]
```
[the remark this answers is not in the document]
iam-mholle: Null can be set, but not retrieved. Am I wrong?

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
iam-mholle: Compiler knows property is not null. I removed the `?`.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
iam-mholle: Null can be set but not retrieved. Am I wrong?

Remotion/Web/Core/UI/Controls/Command.cs
iam-mholle: Why would we drop it and keep dead code? The fallback value would stay there for years. If AllowNull should be removed, then the fallback value too imo.

MichaelKetting: the problem ist, that NRT isn't something that's enforcable right now for projects. It's just a usage hint as an opt-in. (same reason we can't drop the arg-null checks). I just figure, the AllowNull isn't helping so we can drop it while the fallback makes our code safe for our own usage.

=> I would prefer to mark the "here we could pass null as well" information as obsolete, but that's not possible. So, dropping the attribute accomplishes the same effect, to me. TBH, if that's not a good reason, we can keep the attribute. It's a gut feeling thing about how the api is used, that's all.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
MichaelKetting: Smae discussion: https://github.com/re-motion/Framework/pull/235#discussion_r753320779

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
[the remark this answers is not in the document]
MichaelKetting: I'm not sure. EnsureProperty is an IF/ELSEIF and I think you could construct something that has a TreeNode that's not covered by this. 

It may well be that there's going to be a null-ref exception somewhere if you do something like this, but I can't promise the property contract without a deepdive. It's probably a nullbug, but we should go with nullable for now, for Property and add a bugfix-todo for this.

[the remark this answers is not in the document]
iam-mholle: This case is currently guarded by an assertion in the last line of EnsureProperty

Remotion/Web/Core/UI/Controls/Command.cs
iam-mholle: When dropping the nullability and adding an argument check, the contract is enforced for projects - would this be the way to go?

Remotion/ObjectBinding/Web/UI/Controls/BocTreeView.cs
```diff
@@ -734,12 +749,12 @@ public override Control TargetControl
...
     [DesignerSerializationVisibility (DesignerSerializationVisibility.Hidden)]
-    public new IBusinessObjectReferenceProperty Property
+    public new IBusinessObjectReferenceProperty? Property
     {
-      get { return (IBusinessObjectReferenceProperty) base.Property; }
+      get { return (IBusinessObjectReferenceProperty?) base.Property; }
       set
       {
-        IBusinessObjectReferenceProperty property = ArgumentUtility.CheckType<IBusinessObjectReferenceProperty> ("value", value);
+        IBusinessObjectReferenceProperty property = ArgumentUtility.CheckNotNullAndType<IBusinessObjectReferenceProperty> ("value", value!);
```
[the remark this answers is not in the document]
iam-mholle: The bug is then in the next line, as `value.IsList` would throw a `NullReferenceException`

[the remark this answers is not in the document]
iam-mholle: I'll change it to `if (value?.IsList == false)`

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
iam-mholle: `GetValueInfoByIdentifier` might return `null`. How should this be handled? Assertion in the following line?

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
iam-mholle: It already is, but this doesn't help in this case

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
iam-mholle: Linq + NRT aren't working that well. The method returns `IEnumerable<BocListRow>`, the return type would be `IEnumerable<BocListRow?>` without the `!`, `.Where (... != null)` does not influence the type.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditModeController.cs
iam-mholle: I thought the message should always be verbose in this codebase - I'd use the generated message in most cases.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
```diff
@@ -163,47 +171,52 @@ private void EnsureBusinessObject()
...
       if (_property == null) // This test could be omitted if graceful recovery is wanted.
         throw new InvalidOperationException ("Could not find IBusinessObjectReferenceProperty '" + _propertyIdentifier + "'.");
     }
     else if (propertyParentNode != null)
     {
       Property = propertyParentNode.Property;
-      return;
     }
+
+    Assertion.DebugIsNotNull (_property, "_property must not be null at this point.");
```
MichaelKetting, line 206: Might trigger a false positive check. Needs a todo-ticket to check the correct/allowed flows in this method. 

EnsureProperty can currently not ensure that the property will never be null.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
```diff
@@ -163,47 +171,52 @@ private void EnsureBusinessObject()
...
     else
     {
-      IBusinessObjectReferenceProperty property = Property;
+      IBusinessObjectReferenceProperty? property = Property;
       string businessObjectID = ItemID;
       BusinessObject = ((IBusinessObjectClassWithIdentity) property.ReferenceClass).GetObject (businessObjectID);
     }
   }
 
+  [MemberNotNull (nameof (_property))]
```
MichaelKetting, line 180: Drop (see line 206 (not-null guard)

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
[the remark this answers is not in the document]
MichaelKetting: Ah, yes. I think this guard's a bug until proofen otherwise. I'll add a remark for this.

MichaelKetting (commented):
See comment

Remotion/Web/Core/UI/Controls/Command.cs
MichaelKetting: this would break the project if they do not use NRT and not notice that null is not recommended here. That's the problem with weak constraints. We can just fallback during the setter execution to a not=null value. The only reason to not officially allow null is that it's not really helpful to pass null. So, we keep the attribute and just shelve this?

iam-mholle: Isn't this what breaking changes are for?

I don't have a strong opinion on what to do here, I just ask the questions to have the full understanding for future reference.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeView.cs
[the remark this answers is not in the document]
MichaelKetting: oh, yes, I see it now. Yeah, the evlis a good idea.!

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
MichaelKetting: yeah, when the internalValue is not null, the result should be null either. => assertion.IsNotNull with a reference to the actual string value that resulted in null.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
iam-mholle: Is it really necessary here? There is no way to link `HasEditControl` and `GetEditControl` but its pretty clear here imo.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditModeController.cs
MichaelKetting: I think I picked this up in Data NRTs. For the not-null debug messages that are just there to enfore NRT semantics, the generated messages should be fine, sometimes with added context (e.g. "foo != null when bar is true"). The verbose message isn't any more expressive and right now, the generated debug message gives us more details on the error that the verbose message. Also, I think there's a bunch of Debug-assertions that don't even have messages => the generated one is good for these cases.

Remotion/Web/Core/UI/Controls/Command.cs
MichaelKetting: Yeah, excatly. I just don't want to have a breaking change for this because it's not helpful to anybody. It's mainly based on the consideration of what's helpful for the application developer and doesn't introduce bugs. So, if we make this a breaking change, it would only be discovered by reading the docs and then looking up the usages. That doesn't happen. Thus, breaking changes that don't express themselves as a compiler error need to be carefully evaluated and published very visibly. So, this particualr change just isn't worth it (nobody is going to benefit from us locking down the setter). 

The soft hint (not having the AllowNull attribute) on the other hand might give someone a compiler error after they switch, they will notice and clean it up. But if they down't nothing breaks, more satisfaction all round.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
iam-mholle: But in this case, `EnsureProperty` really ensures that `_property` is not `null`.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
MichaelKetting: With the Mandatory-remark (see further down) this becomes moot anyway. so, yeah, okay.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
MichaelKetting: Updated: given EnsureProperty cannot ensure that _property != null (at this time), the Property getter would be nullable anyway.

iam-mholle: There are 2 `EnsureProperty` methods in this file. This one ensures it - if I don't see the case where it doesn't, please enlighten me.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
```diff
@@ -82,11 +83,11 @@ public interface IBocList
 
     /// <summary> Gets or sets the text that is displayed in the index column's title row. </summary>
     /// <remarks> The value will not be HTML encoded. </remarks>
-    string IndexColumnTitle { get; }
+    string? IndexColumnTitle { get; }
```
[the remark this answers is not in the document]
iam-mholle: Ok, but they're `string?` for now right?

Remotion/ObjectBinding/Web/UI/Controls/BocTreeNode.cs
MichaelKetting: discussed.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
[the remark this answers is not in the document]
MichaelKetting: yep. just wanted to mention it for context and when we do the rebase.

Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
iam-mholle: This means I could also add `[return: NotNullIfNotNull ("identifier")]` to the method declaration, right?

iam-mholle: nevermind, the callchain beneath is too complex - I'll add the assertion

Remotion/ObjectBinding/Web/UI/Controls/BocTreeView.cs
```diff
@@ -207,9 +207,10 @@ private void RenderTreeNodeMenu (HtmlTextWriter writer, WebTreeNode node, DropDo
       {
         var businessObjectWebServiceContext = CreateBusinessObjectWebServiceContext();
         var nodeBusinessObjectProperty = ((node as BusinessObjectTreeNode)?.Property ?? (node as BusinessObjectPropertyTreeNode)?.Property)?.Identifier;
-        var nodeBusinessObject = (node as BusinessObjectTreeNode)?.BusinessObject.UniqueIdentifier;
+        var nodeBusinessObject = (node as BusinessObjectTreeNode)?.BusinessObject;
+        var nodeBusinessObjectUniqueIdentifier = nodeBusinessObject?.UniqueIdentifier;
```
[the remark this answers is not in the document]
iam-mholle: I think I forgot to re-join the statements after removing an assertion.

Remotion/ObjectBinding/Web/UI/Controls/BocTreeView.cs
```diff
@@ -384,7 +396,9 @@ private void EvaluateTreeNode (WebTreeNode node)
 
     private void CreateAndAppendBusinessObjectNodeChildren (BusinessObjectTreeNode businessObjectNode)
     {
-      IBusinessObjectWithIdentity businessObject = businessObjectNode.BusinessObject;
+      IBusinessObjectWithIdentity? businessObject = businessObjectNode.BusinessObject;
+
+      Assertion.IsNotNull (businessObject, "The business object of the node must not be null.");
```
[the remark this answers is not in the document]
iam-mholle: How would I word the message then?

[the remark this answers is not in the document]
MichaelKetting: Wtih a lot of null-fallbacks, I guess. 

=> So, yeah, we leave it. This was more of a commentary on the state of things, sorry. And in case you've got a good idea because I don't.

Remotion/ObjectBinding/Web/IBusinessObjectWebUIService.cs
```diff
@@ -25,8 +25,8 @@ namespace Remotion.ObjectBinding.Web
   /// </summary>
   public interface IBusinessObjectWebUIService : IBusinessObjectService
   {
-    IconInfo GetIcon (IBusinessObject obj);
-    string GetToolTip (IBusinessObject obj);
+    IconInfo? GetIcon (IBusinessObject? obj);
+    string GetToolTip (IBusinessObject? obj);
```
MichaelKetting, line 29:
```suggestion
    string? GetToolTip (IBusinessObject? obj);
```

Remotion/ObjectBinding/Web/IBusinessObjectWebUIService.cs
MichaelKetting, line 44:
```suggestion
    HelpInfo? GetHelpInfo (
```

Remotion/ObjectBinding/Web/ObjectBinding.Web.csproj
```diff
@@ -30,5 +29,21 @@
...
+    <None Remove="UI\Controls\BocListImplementation\UniqueIdentifierBasedRowIDProvider.cs.orig" />
+    <None Remove="UI\Controls\BocListImplementation\NullValueRowIDProvider.cs.orig" />
+    <None Remove="UI\Controls\BocListImplementation\IndexBasedRowIDProvider.cs.orig" />
+    <None Remove="UI\Controls\BocListImplementation\IRowIDProvider.cs.orig" />
+    <None Remove="UI\Controls\BocListImplementation\IBocList.cs.orig" />
+    <None Remove="UI\Controls\BocListImplementation\EditableRowSupport\IEditModeHost.cs.orig" />
+    <None Remove="UI\Controls\BocList.cs.orig" />
+    <None Remove="UI\Controls\BocList.RowMenuSupport.cs.orig" />
+    <None Remove="UI\Controls\BocList.EditModeHost.cs.orig" />
+  </ItemGroup>
```
MichaelKetting, line 47:
```suggestion
```
Drop? (including the files, if they exist)

Remotion/ObjectBinding/Web/Services/BusinessObjectWithIdentityProxy.cs
```diff
@@ -46,13 +47,15 @@ public BusinessObjectWithIdentityProxy (IBusinessObjectWithIdentity obj)
       _displayName = obj.GetAccessibleDisplayName();
     }
 
-    public string UniqueIdentifier
+    [DisallowNull]
+    public string? UniqueIdentifier
     {
       get { return _uniqueIdentifier; }
       set { _uniqueIdentifier = ArgumentUtility.CheckNotNullOrEmpty ("value", value); }
     }
```
MichaelKetting, line 55: Cool null-"bug". It's obviously written this way, and it works, it's just...stupid, I guess?

TODO-Ticket to make this always nullable (after thinking about whether this is what we want)

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/UniqueIdentifierBasedRowIDProvider.cs
```diff
@@ -53,19 +53,19 @@ public BocListRow GetRowFromItemRowID (IReadOnlyList<IBusinessObject> values, st
       int rowIndex = tuple.Item1;
       string uniqueIdentifier = tuple.Item2;
 
-      if (rowIndex < values.Count && ((IBusinessObjectWithIdentity) values[rowIndex]).UniqueIdentifier == uniqueIdentifier)
+      if (rowIndex < values.Count && ((IBusinessObjectWithIdentity) values[rowIndex]!).UniqueIdentifier == uniqueIdentifier)
```
MichaelKetting, line 56: Aren't those null-forgives obsolete, now? If so, could you please check in the PR for Value-indexer with null-forgiving operators?

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #238, opened 2021-11-03
RM-8110 Fix problem where the AsyncLocalStorageProvider did not work well with ExecutionContext.Run

a-ctor:
https://re-motion.atlassian.net/browse/RM-8110

### DIFF
--- a/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
+++ b/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
@@ -19,25 +19,191 @@ using System.Collections;
 using System.Threading;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
+using System.Runtime.CompilerServices;
+using Remotion.Logging;
 
 namespace Remotion.Context
 {
-  [ImplementationFor (typeof (ISafeContextStorageProvider), Position = 1)]
+  [ImplementationFor (typeof (ISafeContextStorageProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
   public class AsyncLocalStorageProvider : ISafeContextStorageProvider
   {
-    private static readonly AsyncLocal<Hashtable> s_data = new (AsyncLocalValueChangedHandler);
-
-    private static Hashtable Data => s_data.Value ??= new Hashtable();
-
-    private static void AsyncLocalValueChangedHandler (AsyncLocalValueChangedArgs<Hashtable> args)
+    /// <summary>
+    /// Represents the context which stores values and is itself stored in the <see cref="AsyncLocal{T}"/>.
+    /// Is immutable because of the strange immutability constraints of <see cref="ExecutionContext"/>.
+    /// </summary>
+    /// <remarks>
+    /// In addition to the data stored in a hashtabe it also contains a reference to the previous context.
+    /// This is used to detect if a context is restored and reapply the values (see <see cref="AsyncLocalStorageProvider.AsyncLocalValueChangedHandler"/> below).
+    /// </remarks>
+    private class StorageContext
     {
+      public static readonly StorageContext Empty = new StorageContext (null, null);
+
+      public static StorageContext FromPrevious (StorageContext? previous) => previous != null ? new StorageContext (null, new WeakReference<StorageContext>(previous)) : Empty;
+
+      private readonly WeakReference<StorageContext>? _previous;
+      private readonly Hashtable? _hashtable;
+
+      public StorageContext? Previous => _previous != null && _previous.TryGetTarget (out var result) ? result : null;
+
+      private StorageContext (Hashtable? hashtable, WeakReference<StorageContext>? previous)
+      {
+        _previous = previous;
+        _hashtable = hashtable;
+      }
+
+      public object? GetData (string key) => _hashtable?[key];
+
+      public StorageContext SetData (string key, object? newValue)
+      {
+        // We do not need to store null values as getting a non-existing value will return null (and there is no "Contains")
+        if (newValue == null)
+          return FreeData (key);
+
+        if (_hashtable == null)
+        {
+          var hashtable = new Hashtable();
+          hashtable[key] = newValue;
+
+          return new StorageContext (hashtable, _previous);
+        }
+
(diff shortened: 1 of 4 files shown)


Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -14,38 +14,199 @@
...
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
 
+#if TRACE_ASYNCLOCAL
+using System.Runtime.CompilerServices;
+#endif
+
 namespace Remotion.Context
 {
   [ImplementationFor (typeof (ISafeContextStorageProvider), Position = 1)]
```
MichaelKetting, line 34: We don't have instance state in either implementation of ISafeContextStorageProvider -> please verify and then add Lifetime = Singleton to the attribute. 

By doing this, we 1. don't change the semantics, 2. improve performance (side benefit), and 3. enable us to register a special instance that supports the Trace_AsyncLocal enabled via boolean flag instead of compiler.

So, next step: 
Add a private ctor that allows setting a boolean flag to enable the Trace-semantics. An application developer is then able to simply reconfigure the system to enable tracing.

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -14,38 +14,199 @@
...
+      var threadID = Thread.CurrentThread.ManagedThreadId;
+      var contextChanged = args.ThreadContextChanged ? "yes" : "no";
+      var previousContainerIdentifier = GetObjectIdentifier (previousContainer);
+      var currentContainerIdentifier = GetObjectIdentifier (currentContainer);
+      Console.WriteLine ($"[{threadID}] change(cnx change {contextChanged}, prev {previousContainerIdentifier}, curr: {currentContainerIdentifier})");
+#endif
+
       if (!args.ThreadContextChanged)
+      {
+        Debug.Assert (args.CurrentValue != null, "args.CurrentValue != null");
```
MichaelKetting, line 152:
```suggestion
        Assertion.DebugAssert (args.CurrentValue != null, "args.CurrentValue != null");
```

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -14,38 +14,199 @@
...
+    {
+      var previousContainer = args.PreviousValue;
+      var currentContainer = args.CurrentValue;
+
+#if TRACE_ASYNCLOCAL
+      var threadID = Thread.CurrentThread.ManagedThreadId;
+      var contextChanged = args.ThreadContextChanged ? "yes" : "no";
+      var previousContainerIdentifier = GetObjectIdentifier (previousContainer);
+      var currentContainerIdentifier = GetObjectIdentifier (currentContainer);
+      Console.WriteLine ($"[{threadID}] change(cnx change {contextChanged}, prev {previousContainerIdentifier}, curr: {currentContainerIdentifier})");
```
MichaelKetting, line 147: I'm really considering using a logger here instead. it just feels cleaner.

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -14,38 +14,199 @@
...
+          return this;
+
+        var newHashtable = (Hashtable) _hashtable.Clone();
+        newHashtable.Remove (key);
+
+        return new StorageContext (newHashtable, _previous);
+      }
+    }
+
+    private static readonly AsyncLocal<StorageContext> s_context = new(AsyncLocalValueChangedHandler);
```
MichaelKetting, line 104:
```suggestion
    private static readonly AsyncLocal<StorageContext> s_context = new (AsyncLocalValueChangedHandler);
```

MichaelKetting (changes requested):
Only did a big picture review as a first step. review of tests and implementation logic to follow.

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
a-ctor: How is the application developer supposed to enable tracing if the constructor is private?

MichaelKetting: Forgot: and add a factory method with a good name.

a-ctor: So public constructor that calls private constructor with `false`, Private constructor taking tracing boolean, and a factory method that calls the private constructor with tracing set to `true`?

MichaelKetting (approved):
I'll just go and speak a silent prayer to the cpu register...

MichaelKetting (changes requested):
Managed to miss that there's PR comments :o

a-ctor:
I have fixed them an rebased on top of develop

MichaelKetting (approved):

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderTest.cs
```diff
@@ -76,19 +76,5 @@ public void FreeData_WithoutValue ()
       Assert.That (_provider.GetData ("Foo"), Is.Null);
       Assert.That (_testContext.Items.Contains ("Foo"), Is.False);
     }
-
-    [Test]
-    public void FallbackToCallContext_IfNoCurrentHttpContext ()
```
MichaelKetting, line 81: Why did you drop thsi test?

MichaelKetting: ah, I see. We can use PrivateInvoke.GetPrivateField(...) for this. It's not perfect but a good compromise for now.

Historical context information: back when we started with the SafeContextProvider and the IoC infrastructure, we couldn't do decorator implementations. Today, I might do the HttpContextStorageProvider as a decorator. Might still be a good refactoring but probably better suited to a separate task.

MichaelKetting (changes requested):
re-enable test

MichaelKetting (approved):

PR #239, opened 2021-11-05
RM-8009: Update datepicker on immediate parent

iam-mholle:


derlistigelurch (approved):

Remotion/Web/ClientScript/Scripts/DatePicker.ts
```diff
@@ -210,14 +210,12 @@ class DatePickerFrame
...
   {
-    if (window.parent !== window)
-    {
-      window.parent.DatePickerFrame.Calendar_SelectionChanged(value);
-    }
-    else
-    {
-      DatePicker.UpdateValue(value);
-    }
+    window.parent.DatePickerFrame.UpdateDatePickerValue(value);
```
MichaelKetting, line 213: For testing purposes, I think we should also add a branch for `window.parent===null` (or equivalent). This happens when openingt he DatePicker.aspx stand alone (i.e outside of the iframe). While this isn't a regular usecase, it may be interesting/revalant for developing/debugging something in date picker.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #241, opened 2021-11-05
RM-8014 Opening two Functions with WxeExecutor.ExecuteFunctionExternal seems to do nothing

derlistigelurch:


Remotion/Web/Core/ExecutionEngine/Infrastructure/WxeExecutor.cs
```diff
@@ -119,8 +119,10 @@ public void ExecuteFunctionExternal (WxeFunction function, Control sender, WxeCa
           href,
           options.Target,
           options.Features ?? string.Empty);
-      _page.ClientScript.RegisterClientScriptBlock (_page, typeof (WxeExecutor), "WxeExecuteFunction", openScript);
-      _page.RegisterClientSidePageEventHandler (SmartPageEvents.OnLoaded, "WxeExecuteFunction", functionName);
+
+      var key = Guid.NewGuid().ToString();
```
iam-mholle, line 123: The code works as intended, I can't think of any bad side effects. As an improvement tho, a new guid is created a few lines above, you could store its string in a variable and reuse it here.

iam-mholle (changes requested):

iam-mholle (approved):

MichaelKetting (approved):

PR #242, opened 2021-11-08
RM-7526: Rework cache to no longer keep a strong reference to resource managers

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7526

### DIFF
--- a/Remotion/Web/Core/UI/Controls/RendererBase.cs
+++ b/Remotion/Web/Core/UI/Controls/RendererBase.cs
@@ -16,8 +16,8 @@
 // 
 using System;
 using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
 using System.Text;
-using System.Threading;
 using System.Web.UI;
 using Remotion.Globalization;
 using Remotion.Utilities;
@@ -34,8 +34,7 @@ namespace Remotion.Web.UI.Controls
   public abstract class RendererBase<TControl>
       where TControl : IStyledControl, IControlWithDiagnosticMetadata
   {
-    private readonly ConcurrentDictionary<Tuple<Type, IResourceManager>, Lazy<IResourceManager>> _resourceManagerCache =
-        new ConcurrentDictionary<Tuple<Type, IResourceManager>, Lazy<IResourceManager>>();
+    private readonly ConcurrentDictionary<Type, ConditionalWeakTable<IResourceManager, Lazy<IResourceManager>>> _resourceManagerCache = new();
 
     private readonly IResourceUrlFactory _resourceUrlFactory;
     private readonly IGlobalizationService _globalizationService;
@@ -144,13 +143,19 @@ namespace Remotion.Web.UI.Controls
       ArgumentUtility.CheckNotNull ("localResourcesType", localResourcesType);
       ArgumentUtility.CheckNotNull ("controlResourceManager", controlResourceManager);
 
-      var cachedResourceManager = _resourceManagerCache.GetOrAdd (
-          Tuple.Create (localResourcesType, controlResourceManager),
-          key => new Lazy<IResourceManager> (
-              () => ResourceManagerSet.Create (key.Item2, _globalizationService.GetResourceManager (key.Item1)),
-              LazyThreadSafetyMode.ExecutionAndPublication));
+      var table = _resourceManagerCache
+          .GetOrAdd (
+              localResourcesType,
+              _ => new ConditionalWeakTable<IResourceManager, Lazy<IResourceManager>>());
 
-      return cachedResourceManager.Value;
+      if (table.TryGetValue (controlResourceManager, out var cachedResult))
+        return cachedResult.Value;
+
+      var result = table.GetValue (
+          controlResourceManager,
+          mgr => new Lazy<IResourceManager> (() => ResourceManagerSet.Create (mgr, _globalizationService.GetResourceManager (localResourcesType))));
+
+      return result.Value;
     }
   }
 }
--- a/Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
+++ b/Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
@@ -1,0 +1,107 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using JetBrains.Annotations;
+using Moq;
+using NUnit.Framework;
+using Remotion.Globalization;
+using Remotion.Web.UI.Controls;
+using Remotion.Web.UI.Controls.Rendering;
+
+namespace Remotion.Web.UnitTests.Core.UI.Controls.Rendering
+{
+  [TestFixture]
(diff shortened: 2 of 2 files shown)


Remotion/Web/Core/UI/Controls/RendererBase.cs
```diff
@@ -144,13 +143,15 @@ protected IResourceManager GetResourceManager (Type localResourcesType, IResourc
...
-          key => new Lazy<IResourceManager> (
-              () => ResourceManagerSet.Create (key.Item2, _globalizationService.GetResourceManager (key.Item1)),
-              LazyThreadSafetyMode.ExecutionAndPublication));
+      var result = _resourceManagerCache
+          .GetOrAdd (
+              localResourcesType,
+              _ => new ConditionalWeakTable<IResourceManager, Lazy<IResourceManager>>())
+          .GetValue (
+              controlResourceManager,
+              mgr => new Lazy<IResourceManager> (() => ResourceManagerSet.Create (mgr, _globalizationService.GetResourceManager (localResourcesType))));
```
a-ctor, line 152: I was skeptical at first if we need the `Lazy<T>` here, because we always directly create the value in line 154. But it turns out doing things like this has the benefit of thread safety as two contending threads will only create two `Lazy<T>` objects one of which is thrown away by the dictionary and then when accessing the value on of the threads will block.

a-ctor (approved):

Remotion/Web/Core/UI/Controls/RendererBase.cs
```diff
@@ -144,13 +143,15 @@ protected IResourceManager GetResourceManager (Type localResourcesType, IResourc
...
-      var cachedResourceManager = _resourceManagerCache.GetOrAdd (
-          Tuple.Create (localResourcesType, controlResourceManager),
-          key => new Lazy<IResourceManager> (
-              () => ResourceManagerSet.Create (key.Item2, _globalizationService.GetResourceManager (key.Item1)),
-              LazyThreadSafetyMode.ExecutionAndPublication));
+      var result = _resourceManagerCache
+          .GetOrAdd (
+              localResourcesType,
+              _ => new ConditionalWeakTable<IResourceManager, Lazy<IResourceManager>>())
+          .GetValue (
```
MichaelKetting, line 150: Please refactor to use double check:
first TryGetValue and if it's false, we do GetValue. Otherwise, we always create the closure during value access.

MichaelKetting (changes requested):
please add a unit test for RenderBase (RenderBaseTest) using a TestableRenderBase class to show that the control resource manager isn't kept alive. It's testable and kind of an important feature of this bugfix so we should test ist.

Remotion/Web/Core/UI/Controls/RendererBase.cs
iam-mholle: Hmm there is another problem, if I'm not mistaken, delegate passed into the lazy also keeps a strong reference to the resource manager. The resource manager cannot be freed this way. I might need to change the type to `Lazy<WeakReference<IResourceManager>>` and adapt usages accordingly.

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+  [TestFixture]
+  public class RendererBaseTest : RendererTestBase
+  {
+    [Test]
+    public void GetResourceManager_DoesNotKeepResourceManagerAlive ()
+    {
+      var renderer = new TestableRendererBase (Mock.Of<IResourceUrlFactory>(), GlobalizationService, Mock.Of<IRenderingFeatures>());
+
+      var resourceManagerWeakReference = CallGetResourceManagerWithManagerAndReturnItsWeakReference(renderer);
```
MichaelKetting, line 35: Since we're going to update the whitespace anyway in the next couple days, I'll leave it up to use to fix, not fix the whitepace bugs here and in line 64

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
```diff
@@ -0,0 +1,74 @@
...
+      GC.Collect();
+      GC.WaitForPendingFinalizers();
+      GC.Collect();
+
+      Assert.That (resourceManagerWeakReference.IsAlive, Is.False);
+    }
+
+    private static WeakReference CallGetResourceManagerWithManagerAndReturnItsWeakReference (TestableRendererBase renderer)
+    {
+      var resourceManager = ResourceManagerSet.Create (NullResourceManager.Instance);
```
MichaelKetting, line 46: Just for consideration (the test would be red anyway, if this changes so there's no hidden test-bug):

There's no contract that ResourcemanagerSet.Create couldn't return the Null-instance for empty sets instead of always creating a set. For the current testcase, a fake IResourceManager would probably be a more robust choice regarding future refactorings.

Remotion/Web/Core/UI/Controls/RendererBase.cs
```diff
@@ -144,13 +143,21 @@ protected IResourceManager GetResourceManager (Type localResourcesType, IResourc
...
-              () => ResourceManagerSet.Create (key.Item2, _globalizationService.GetResourceManager (key.Item1)),
-              LazyThreadSafetyMode.ExecutionAndPublication));
+      var table = _resourceManagerCache
+          .GetOrAdd (
+              localResourcesType,
+              _ => new ConditionalWeakTable<IResourceManager, Lazy<IResourceManager>>());
 
-      return cachedResourceManager.Value;
+      if (table.TryGetValue (controlResourceManager, out var cachedResult))
+      {
```
MichaelKetting, line 152: As you like: Personally, I drop them for lines like this in favor of a more compact IF structure.

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
```diff
@@ -0,0 +1,74 @@
...
+using Remotion.Web.UI.Controls;
+using Remotion.Web.UI.Controls.Rendering;
+
+namespace Remotion.Web.UnitTests.Core.UI.Controls.Rendering
+{
+  [TestFixture]
+  public class RendererBaseTest : RendererTestBase
+  {
+    [Test]
+    public void GetResourceManager_DoesNotKeepResourceManagerAlive ()
```
MichaelKetting, line 31: Hmm... I think we should also add a test to show the same instance is returned twice for the same key while it's alive and a new instance is returned if the key's expired. Should have seen then earlier. sorry!

MichaelKetting (changes requested):

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
iam-mholle: I can't assert that a collected instance is different from a new one directly, I guess the existing test + a new one showing that the same instance is returned for the same keys is enough?

MichaelKetting: > I can't assert that a collected instance is different from a new one directly,
True. 

> I guess the existing test + a new one showing that the same instance is returned for the same keys is enough?
+1

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+      var resourceManager = new FakeResourceManager();
+      var resourceManagerReference = new WeakReference (resourceManager);
+
+      _ = renderer.GetResourceManager (typeof (string), resourceManager);
+
+      return resourceManagerReference;
+    }
+
+    private class FakeResourceManager : IResourceManager
```
MichaelKetting, line 70: Quick question: did you go with an actual type because I used "fake" instead of "stub" or because Moq keeps the mock instances alive?

If it's the first case, I probably shouldn't be so fancy with my words.

MichaelKetting (approved):

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/RendererBaseTest.cs
iam-mholle: I thought that Moq might keep the mock instance alive inside of a static repository of some sort, so I didn't event try it out. I could tho if you're interested.

MichaelKetting: as you like

PR #247, opened 2021-11-12
RM-8120: Change type of `WebButton.Text` from `string` to `WebString`

iam-mholle:
<details>
  <summary>Task List</summary>
  
- [x] WebButton.Text is defined on System.Web.Button. To change the datatype of the Text property, the property is redefined on WebButton.  
- [x] System.Web.Button.Text persists its value in the ViewState. For compatibility, this practice is continued and the raw value (WebString.GetValue()) is written into the same slot used by System.Web.Button.Text when setting WebButton.Text. In addition, the WebString.Type is also written into the ViewState. 
- [x] When getting WebButton.Text, the Remotion.WebString is reconstiuted from the two ViewState values. If Type is not available in the ViewState, WebStringType.PlainText is assumed. This makes setting the System.Web.Button.Text property safe.
- [x] Remove usage of HotkeyFormatter.FormatHotkey(...) and HotkeyParser.Parse(...) from WebButton and verify that property System.Web.Button.AccessKey works as a replacement (i.e. renders as the html attribute accesskey). Remove redunant code left over from hotkey parsing.
- [x] To correctly render the value attribute during AddAttributesToRender, System.Web.Button.Text must be temporarily set to the raw value  (WebString.GetValue()) before the base call then restored after the base call is completed. This way, System.Web.Button can add the value attribute, which is automatically encoded.
- [x] Rework behavior of WebButton.IsLegacyButtonEnabled, which results in rendering an html input element of type button. 
- [x] Instead of System.String.IsNullOrEmpty, Remotion.WebString.IsEmpty can be used.
- [x] Properties of type Remotion.WebString are declared as not-nullable.
- [x] When rendering DiagnosticMetadataAttributes.Content, the raw value (WebString.GetValue()) is used. Html tags are only stripped if the WebString.Type is set to Html. 
- [x] Update WebButtonDiagnosticMetadataTest with new cases (plaintext and encoded)
- [x] Update WebButtonControlObject to expose the AccessKey and add a web test.
- [x] When WebButton.Text is loaded via the help of ResourceManagerUtility.GetGlobalResourceKey(...) based on the resource-key-encoded property value, the WebStringType from the property is preserved. If the property does not use Remotion.WebString, WebStringType.Plaintext is assumed.
- [x] Button.Tooltip and Button.AccessKey are not changed since they are always rendered as encoded attributes by ASP.NET.
# Breaking Changes to document:
- [x] Changed of datatype for the Text property from String to WebString
- [x] Change of default encoding for System.Web.Button.Text from plain to encoded.
- Change of AccessKey support
  - [x] AccessKey must be set manually
  - [x] The access key must be highlighted manually in WebButton.Text, if required.
  - [x] & in WebButton.Text  is no longer used to indicate the access key and should be removed.
  
</details>

### DIFF
--- a/Remotion/Web/Core/UI/Controls/WebButton.cs
+++ b/Remotion/Web/Core/UI/Controls/WebButton.cs
@@ -26,8 +26,8 @@ using Remotion.Security;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
 using Remotion.Web.Contracts.DiagnosticMetadata;
+using Remotion.Web.Globalization;
 using Remotion.Web.Infrastructure;
-using Remotion.Web.UI.Controls.Hotkey;
 using Remotion.Web.UI.Controls.Rendering;
 using Remotion.Web.UI.Controls.WebButtonImplementation;
 using Remotion.Web.UI.Controls.WebButtonImplementation.Rendering;
@@ -49,6 +49,9 @@ namespace Remotion.Web.UI.Controls
   {
     private static readonly object s_clickEvent = new object();
 
+    private const string c_textViewStateKey = nameof (Text);
+    private const string c_textWebStringViewStateKey = c_textViewStateKey + "_" + nameof (WebStringType);
+
     private IconInfo _icon;
     private readonly IRenderingFeatures _renderingFeatures;
     private PostBackOptions? _options;
@@ -61,13 +64,37 @@ namespace Remotion.Web.UI.Controls
     private bool _requiresSynchronousPostBack;
     private bool _hasPagePreRenderCompleted;
 
-    private TextWithHotkey? _textWithHotkey;
     private ButtonType _buttonType;
 
     public WebButton ()
     {
       _icon = new IconInfo();
       _renderingFeatures = SafeServiceLocator.Current.GetInstance<IRenderingFeatures>();
+    }
+
+    [Category ("Appearance")]
+    [Description ("The text to be shown on the button.")]
+    [DefaultValue (typeof (WebString), "")]
+    public new WebString Text
+    {
+      get
+      {
+        var value = (string?) ViewState[c_textViewStateKey];
+        var type = (WebStringType?) ViewState[c_textWebStringViewStateKey] ?? WebStringType.PlainText;
+
+        return type switch
+        {
+            WebStringType.PlainText => WebString.CreateFromText (value),
+            WebStringType.Encoded => WebString.CreateFromHtml (value),
+            _ => throw new InvalidOperationException (
+                $"The value for key '{c_textWebStringViewStateKey}' in the ViewState contains invalid data '{type}'."),
+        };
+      }
+      set
+      {
+        ViewState[c_textViewStateKey] = value.GetValue();
+        ViewState[c_textWebStringViewStateKey] = value.Type;
+      }
     }
 
     protected override void OnInit (EventArgs e)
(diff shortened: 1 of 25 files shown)


Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -134,25 +158,15 @@ private void Page_PreRenderComplete (object? sender, EventArgs e)
...
+        writer.AddAttribute (HtmlTextWriterAttribute.Accesskey, AccessKey);
 
       if (IsLegacyButtonEnabled)
       {
-        if (!string.IsNullOrEmpty (_textWithHotkey!.Text))
-          Text = _textWithHotkey.Text;
-        else if (_icon != null && _icon.HasRenderingInformation)
-          Text = _icon.AlternateText;
+        if (Text.IsEmpty && _icon != null && _icon.HasRenderingInformation)
+          base.Text = _icon.AlternateText;
```
MichaelKetting, line 169: I think we need to move this down to the base-call of AddAttributesToRender.

Before, it worked because we just always set base.Text. Now, it's at least cleaner (and possibly incorrect otherwise) if we don't make this the else-branch for assigning the base.Text value.

Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -215,8 +229,13 @@ protected override void AddAttributesToRender (HtmlTextWriter writer)
 
       ControlStyle.CssClass = computedCssClass;
 
+      var currentText = Text;
+      base.Text = Text.GetValue();
+
```
MichaelKetting, line 234:
```suggestion
      var currentText = Text;
      if (Text.IsEmpty && _icon != null && _icon.HasRenderingInformation)
        base.Text = _icon.AltText
      else
        base.Text = Text.GetValue();
```

Please check the IF-semantics, it's just a hip-shot right now.

Remotion/Web/UnitTests/Core/UI/Controls/WebButtonTests/WebButtonDiagnosticMetadataTest.cs
```diff
@@ -45,6 +45,20 @@ public void RenderDiagnosticMetadataAttributes ()
...
+    {
+      var webButton = new TestWebButton { ID = "WebButton", Text = WebString.CreateFromHtml ("<p>My Button</p>") };
+
+      var renderedText = RenderControl (webButton);
+
+      Assert.That (renderedText, Does.Contain (DiagnosticMetadataAttributes.ControlType + "=\"WebButton\""));
+      Assert.That (renderedText, Does.Contain (DiagnosticMetadataAttributes.ItemID + "=\"" + webButton.ID + "\""));
+      Assert.That (renderedText, Does.Contain (DiagnosticMetadataAttributes.Content + "=\"My Button\""));
+      Assert.That (renderedText, Does.Not.Contains (DiagnosticMetadataAttributes.CommandName));
+      Assert.That (renderedText, Does.Contain (DiagnosticMetadataAttributes.TriggersPostBack + "=\"true\""));
```
MichaelKetting, line 59: Please drop the redundant assertions (trigger, controltype, etc) that are not part of the current test's focus.

Remotion/Web/UnitTests/Core/UI/Controls/WebButtonTests/EncodingTest.cs
```diff
@@ -0,0 +1,81 @@
...
+      _webButton.ID = "WebButton";
+    }
+
+    [Test]
+    public void Text_WithPlainText ()
+    {
+      _webButton.Text = WebString.CreateFromText ("test");
+      var result = RenderControl (_webButton);
+
+      Assert.That (result, Does.Contain (@"<span class=""buttonBody""><span>test</span></span>"));
```
MichaelKetting, line 42: We have a control test infrastructure. See BoctextValueRendererTest etc. Removes details such as insignificant whilespace and attribute ordering. Please update.

Setup: Html = new HtmlHelper();

Test:
_webButton.Render (Html.Writer);
var document = Html.GetResultDocument();
Html.AssertChildElementCount (document.DocumentElement, 1);
var span = Html.GetAssertedChildElement (document, "button", 0);
Html.AssertAttribute (span, "class", "buttonBody");

etc etc

MichaelKetting (changes requested):

Remotion/Web/Core/UI/Controls/WebButton.cs
iam-mholle: if `!Text.IsEmpty`, `Text` should be set alright.

MichaelKetting (approved):

Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -70,6 +72,31 @@ public WebButton ()
       _renderingFeatures = SafeServiceLocator.Current.GetInstance<IRenderingFeatures>();
     }
 
+    [Category ("Appearance")]
+    [Description ("Button_Text")]
```
MichaelKetting, line 76: Please resolve the actual text.

Remotion/Web/Core/UI/Controls/WebButton.cs
```diff
@@ -70,6 +72,31 @@ public WebButton ()
...
+    [DefaultValue ("")]
```
MichaelKetting, line 77: Please check if you should add typeof(WebString), too.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #248, opened 2021-11-12
RM-8131 Change `FormGridLabel.Text` type from `string` to `WebString`

iam-mholle:
<details>
  <summary>Task List</summary>

- [x] FormGridLabel.Text is defined on System.Web.Label. To change the datatype of the Text property, the property is redefined on FormGridLabel.  
- [x] System.Web.Label.Text persists its value in the ViewState. For compatibility, this practice is continued and the raw value (WebString.GetValue()) is written into the same slot used by System.Web.Label.Text when setting FormGridLabel.Text. In addition, the WebString.Type is also written into the ViewState. 
- [x] When getting FormGridLabel.Text, the Remotion.WebString is reconstiuted from the two ViewState values. If Type is not available in the ViewState, WebStringType.PlainText is assumed. This makes setting the System.Web.Label.Text property safe.
- [x] Render FormGridLabel.Text during RenderContents (see existing WebLinkButton for example). Alternatively, we could set and reset the Text value before just calling the base implementation of RenderContents but this seems less clean. Neither solution is perfect but this way we’re consistent.
- [x] Instead of System.String.IsNullOrEmpty, Remotion.WebString.IsEmpty can be used.
- [x] Properties of type Remotion.WebString are declared as not-nullable.
- [x] When FormGridLabel.Text is loaded via the help of ResourceManagerUtility.GetGlobalResourceKey(...) based on the resource-key-encoded property value, the WebStringType from the property is preserved.
# Breaking Changes to document:
- [x] Changed of datatype for the Text property from String to WebString
- [x] Change of default encoding for System.Web.Label.Text from plain to encoded.

</details>

### DIFF
--- a/Remotion/Web/Core/UI/Controls/FormGridLabel.cs
+++ b/Remotion/Web/Core/UI/Controls/FormGridLabel.cs
@@ -23,9 +23,9 @@ using System.Web.UI;
 using System.Web.UI.WebControls;
 using Remotion.Globalization;
 using Remotion.Utilities;
+using Remotion.Web.Globalization;
 using Remotion.Web.Infrastructure;
 using Remotion.Web.UI.Globalization;
-using Remotion.Web.Utilities;
 
 namespace Remotion.Web.UI.Controls
 {
@@ -37,8 +37,36 @@ namespace Remotion.Web.UI.Controls
 [ToolboxItemFilter("System.Web.UI")]
 public class FormGridLabel: Label, ISmartControl
 {
+  private const string c_textViewStateKey = nameof (Text);
+  private const string c_textWebStringViewStateKey = c_textViewStateKey + "_" + nameof (WebStringType);
+
   private bool _required = false;
   private string? _helpUrl = null;
+
+  [Category ("Appearance")]
+  [Description ("The text to be shown for the Label.")]
+  [DefaultValue (typeof (WebString), "")]
+  public new WebString Text
+  {
+    get
+    {
+      var value = (string?) ViewState[c_textViewStateKey];
+      var type = (WebStringType?) ViewState[c_textWebStringViewStateKey] ?? WebStringType.PlainText;
+
+      return type switch
+      {
+          WebStringType.PlainText => WebString.CreateFromText (value),
+          WebStringType.Encoded => WebString.CreateFromHtml (value),
+          _ => throw new InvalidOperationException (
+              $"The value for key '{c_textWebStringViewStateKey}' in the ViewState contains invalid data '{type}'."),
+      };
+    }
+    set
+    {
+      ViewState[c_textViewStateKey] = value.GetValue();
+      ViewState[c_textWebStringViewStateKey] = value.Type;
+    }
+  }
 
   [Category("Behavior")]
   [DefaultValue (false)]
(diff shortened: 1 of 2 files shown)


Remotion/Web/Core/UI/Controls/FormGridLabel.cs
```diff
@@ -37,9 +37,34 @@ namespace Remotion.Web.UI.Controls
 [ToolboxItemFilter("System.Web.UI")]
 public class FormGridLabel: Label, ISmartControl
 {
+  private const string c_textWebStringViewStateKey = nameof (Text) + "_" + nameof (WebStringType);
+
   private bool _required = false;
   private string? _helpUrl = null;
 
+  public new WebString Text
```
MichaelKetting, line 45: Attributes from base declaration

Remotion/Web/Core/UI/Controls/FormGridLabel.cs
```diff
@@ -37,9 +37,34 @@ namespace Remotion.Web.UI.Controls
...
+  {
+    get
+    {
+      var value = (string?) ViewState[nameof (Text)];
```
MichaelKetting, line 49: same question as with WebButton regarding const-keys.

Remotion/Web/Core/UI/Controls/FormGridLabel.cs
```diff
@@ -37,9 +37,34 @@ namespace Remotion.Web.UI.Controls
...
+      var type = (WebStringType?) ViewState[c_textWebStringViewStateKey];
+      type ??= WebStringType.PlainText;
```
MichaelKetting, line 51: smae question as with WebButton regarding fallback

Remotion/Web/Core/UI/Controls/ValidationStateViewer.cs
```diff
@@ -223,7 +223,7 @@ protected virtual void RenderValidationMessages (HtmlTextWriter writer)
             if (control is SmartLabel)
               text = ((SmartLabel) control).GetText();
             else if (control is FormGridLabel)
-              text = ((FormGridLabel) control).Text;
+              text = ((FormGridLabel) control).Text.GetValue();
```
MichaelKetting, line 226: please use use ToString(HtmlWithTransformedLineBreaks)

MichaelKetting (changes requested):

Remotion/Web/Core/UI/Controls/FormGridLabel.cs
```diff
@@ -37,9 +37,37 @@ namespace Remotion.Web.UI.Controls
...
 {
+  private const string c_textViewStateKey = nameof (Text);
+  private const string c_textWebStringViewStateKey = c_textViewStateKey + "_" + nameof (WebStringType);
+
   private bool _required = false;
   private string? _helpUrl = null;
 
+  [Category("Appearance")]
+  [Description("Label_Text")]
+  [DefaultValue("")]
```
MichaelKetting, line 48: Please resolve the Label_Text string
Please check if we should add typeof (WebString) next to the empty string.

MichaelKetting (changes requested):

PR #253, opened 2021-11-15
RM-8132 Change type of SmartLabel.Text from System.String to Remotion…

a-ctor:
https://re-motion.atlassian.net/browse/RM-8132

### DIFF
--- a/Remotion/Web/Core/UI/Controls/SmartLabel.cs
+++ b/Remotion/Web/Core/UI/Controls/SmartLabel.cs
@@ -20,6 +20,7 @@ using System.Web.UI;
 using System.Web.UI.WebControls;
 using Remotion.Globalization;
 using Remotion.Utilities;
+using Remotion.Web.Globalization;
 using Remotion.Web.Infrastructure;
 using Remotion.Web.UI.Globalization;
 using Remotion.Web.Utilities;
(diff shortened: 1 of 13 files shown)


Remotion/Web/Core/UI/Controls/SmartLabel.cs
```diff
@@ -119,7 +120,7 @@ public string GetText()
         text = "[Label for " + ForControl + "]";
     }
     ForControl = forControlBackUp;
-    return text;
+    return WebString.CreateFromText (text);
```
MichaelKetting, line 123: One of the next tasks will change the returntype of smartControl.DisplayName to WebString (line 118). We can either refactor now to allow easier substitution or the type or later when the next task gets done.

From what I'm reading it will either be to refactor into return statements for text or individual wrapping of the text results. Either way, line 123 will be changed again, either back to the original form or to be deleted entirely. 

You choice when we do this.

MichaelKetting (approved):
I've added an optional refactoring note.

Remotion/Web/Core/UI/Controls/SmartLabel.cs
a-ctor: I think that is better done when the time comes instead of digging up this feature branch again.

PR #255, opened 2021-11-16
RM-8134 Change type of text properties on DropDownMenu and ListMenu to WebString

a-ctor:
https://re-motion.atlassian.net/browse/RM-8134

### DIFF
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenu.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenu.cs
@@ -24,6 +24,7 @@ using Remotion.Mixins;
 using Remotion.ObjectBinding.Web.Services;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
+using Remotion.Web;
 using Remotion.Web.ExecutionEngine;
 using Remotion.Web.Services;
 using Remotion.Web.UI;
@@ -188,8 +189,7 @@ public class BocDropDownMenu : BusinessObjectBoundWebControl, IBocMenuItemContai
     if (businessObject != null)
     {
       _dropDownMenu.GetSelectionCount = "function() { return 1; }";
-      var titleText = GetTitleText (businessObject);
-      _dropDownMenu.TitleText = System.Web.HttpUtility.HtmlEncode (titleText);
+      _dropDownMenu.TitleText = GetTitleText (businessObject);
 
      if (_enableIcon)
      {
@@ -202,14 +202,15 @@ public class BocDropDownMenu : BusinessObjectBoundWebControl, IBocMenuItemContai
     }
   }
 
-  protected virtual string GetTitleText (IBusinessObject businessObject)
+  protected virtual WebString GetTitleText (IBusinessObject businessObject)
   {
     ArgumentUtility.CheckNotNull ("businessObject", businessObject);
+
+    var titleText = businessObject is IBusinessObjectWithIdentity businessObjectWithIdentity
+        ? businessObjectWithIdentity.GetAccessibleDisplayName()
+        : businessObject.ToString();
     
-    if (businessObject is IBusinessObjectWithIdentity)
-      return ((IBusinessObjectWithIdentity) businessObject).GetAccessibleDisplayName();
-    else
-      return businessObject.ToString();
+    return WebString.CreateFromText (titleText);
   }
 
   protected override void Render (HtmlTextWriter writer)
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenuColumnDefinition.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenuColumnDefinition.cs
@@ -21,6 +21,8 @@ using CommonServiceLocator;
 using Remotion.Globalization;
 using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering;
 using Remotion.Utilities;
+using Remotion.Web;
+using Remotion.Web.Globalization;
 using Remotion.Web.UI.Controls;
 using Remotion.Web.UI.Globalization;
 
@@ -30,7 +32,7 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
   public class BocDropDownMenuColumnDefinition : BocColumnDefinition
   {
     private bool _showMenuTitle = true;
-    private string _menuTitleText = string.Empty;
+    private WebString _menuTitleText;
     private IconInfo _menuTitleIcon;
 
     public BocDropDownMenuColumnDefinition ()
@@ -65,16 +67,16 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     }
 
     /// <summary> Gets or sets the text displayed as the menu title. </summary>
-    /// <value> A <see cref="string"/> displayed as the menu's title. </value>
+    /// <value> A <see cref="WebString"/> displayed as the menu's title. </value>
     [PersistenceMode (PersistenceMode.Attribute)]
     [Category ("Appearance")]
     [Description ("The menu title, can be empty.")]
-    [DefaultValue ("")]
+    [DefaultValue (typeof (WebString), "")]
     [NotifyParentProperty (true)]
-    public string MenuTitleText
+    public WebString MenuTitleText
     {
       get { return _menuTitleText; }
-      set { _menuTitleText = value ?? string.Empty; }
+      set { _menuTitleText = value; }
     }
 
     /// <summary> Gets or sets the icon displayed in the menu's title field. </summary>
@@ -99,9 +101,9 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
       
       base.LoadResources (resourceManager, globalizationService);
 
-      string key = ResourceManagerUtility.GetGlobalResourceKey (MenuTitleText);
+      string key = ResourceManagerUtility.GetGlobalResourceKey (MenuTitleText.GetValue());
       if (!string.IsNullOrEmpty (key))
-        MenuTitleText = resourceManager.GetString (key);
+        MenuTitleText = resourceManager.GetWebString (key, MenuTitleText.Type);
 
       if (MenuTitleIcon != null)
         MenuTitleIcon.LoadResources (resourceManager);
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
@@ -43,6 +43,7 @@ using Remotion.Utilities;
 using Remotion.Web;
 using Remotion.Web.Contracts.DiagnosticMetadata;
 using Remotion.Web.ExecutionEngine;
+using Remotion.Web.Globalization;
 using Remotion.Web.Infrastructure;
 using Remotion.Web.Services;
 using Remotion.Web.UI;
@@ -255,7 +256,7 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     private bool _showListMenu = true;
 
     private RowMenuDisplay _rowMenuDisplay = RowMenuDisplay.Undefined;
-    private string _optionsTitle;
+    private WebString _optionsTitle;
     private string[] _hiddenMenuItems;
     private Unit _menuBlockOffset = Unit.Empty;
     private Unit _menuBlockItemOffset = Unit.Empty;
@@ -2061,9 +2062,9 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
       if (! string.IsNullOrEmpty (key))
         EmptyListMessage = resourceManager.GetString (key);
 
-      key = ResourceManagerUtility.GetGlobalResourceKey (OptionsTitle);
+      key = ResourceManagerUtility.GetGlobalResourceKey (OptionsTitle.GetValue());
       if (! string.IsNullOrEmpty (key))
-        OptionsTitle = resourceManager.GetString (key);
+        OptionsTitle = resourceManager.GetWebString (key, OptionsTitle.Type);
 
       key = ResourceManagerUtility.GetGlobalResourceKey (AvailableViewsListTitle);
       if (! string.IsNullOrEmpty (key))
@@ -3504,8 +3505,8 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     /// <summary> Gets or sets the text that is rendered as a label for the <c>options menu</c>. </summary>
     [Category ("Menu")]
     [Description ("The text that is rendered as a label for the options menu.")]
-    [DefaultValue ("")]
-    public string OptionsTitle
+    [DefaultValue (typeof (WebString), "")]
+    public WebString OptionsTitle
     {
       get { return _optionsTitle; }
       set { _optionsTitle = value; }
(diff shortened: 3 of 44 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenuColumnDefinition.cs
```diff
@@ -65,16 +67,16 @@ public bool ShowMenuTitle
     }
 
     /// <summary> Gets or sets the text displayed as the menu title. </summary>
-    /// <value> A <see cref="string"/> displayed as the menu's title. </value>
+    /// <value> A <see cref="WebString"/> displayed as the menu's title. </value>
     [PersistenceMode (PersistenceMode.Attribute)]
     [Category ("Appearance")]
     [Description ("The menu title, can be empty.")]
     [DefaultValue ("")]
```
MichaelKetting, line 74: ASP.NET prefers we add typeof (WebString) to the DefaultValue.

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -3505,7 +3506,7 @@ public string AvailableViewsListTitle
     [Category ("Menu")]
     [Description ("The text that is rendered as a label for the options menu.")]
     [DefaultValue ("")]
```
MichaelKetting, line 3508:
```suggestion
    [DefaultValue ("")]
```
```suggestion
    [DefaultValue (typeof (WebString), "")]
```

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
```diff
@@ -313,7 +315,7 @@ public WebMenuItemCollection OptionsMenuItems
     [Category ("Menu")]
     [Description ("The text that is rendered as a label for the options menu.")]
     [DefaultValue ("")]
```
MichaelKetting, line 317:
```suggestion
    [DefaultValue (typeof (WebString), "")]
```
```suggestion
    [DefaultValue ("")]
```

SecurityManager/Clients.Web/Classes/OrganizationalStructure/BocListInlineEditingConfigurator.cs
```diff
@@ -75,14 +75,14 @@ public virtual void Configure<TBusinessObject> (BocList bocList, Func<TBusinessO
           new BocMenuItem
           {
               ItemID = "New",
-              Text = GlobalResourcesHelper.GetString (_globalizationService, GlobalResources.New),
+              Text = WebString.CreateFromText (GlobalResourcesHelper.GetString (_globalizationService, GlobalResources.New)),
```
MichaelKetting, line 78:
```suggestion
              Text = WebString.CreateFromText (GlobalResourcesHelper.GetString (_globalizationService, GlobalResources.New)),
```
```suggestion
              Text = GlobalResourcesHelper.GetText (_globalizationService, GlobalResources.New),
```
We already have a utility for this since the WebButton branch.

SecurityManager/Clients.Web/Classes/OrganizationalStructure/BocListInlineEditingConfigurator.cs
```diff
@@ -75,14 +75,14 @@ public virtual void Configure<TBusinessObject> (BocList bocList, Func<TBusinessO
...
               Command = new InlineEditingNewItemMenuItemCommand<TBusinessObject> (newObjectFactory)
           });
       bocList.ListMenuItems.Add (
           new BocMenuItem
           {
               ItemID = "Delete",
-              Text = GlobalResourcesHelper.GetString (_globalizationService, GlobalResources.Delete),
+              Text = WebString.CreateFromText (GlobalResourcesHelper.GetString (_globalizationService, GlobalResources.Delete)),
```
MichaelKetting, line 85:
```suggestion
              Text = WebString.CreateFromText (GlobalResourcesHelper.GetString (_globalizationService, GlobalResources.Delete)),
```
```suggestion
              Text = GlobalResourcesHelper.GetText (_globalizationService, GlobalResources.Delete),
```

Remotion/Web/Core/UI/Controls/WebMenuItem.cs
```diff
@@ -169,21 +177,20 @@ public string Category
 
     [PersistenceMode (PersistenceMode.Attribute)]
     [Category ("Appearance")]
-    [Description ("The text displayed in this menu item. Use '-' for a separator menu item. The value will not be HTML encoded.")]
+    [Description ("The text displayed in this menu item. Use '-' for a separator menu item.")]
     [NotifyParentProperty (true)]
     [DefaultValue ("")]
```
MichaelKetting, line 182:
```suggestion
    [DefaultValue ("")]
```
```suggestion
    [DefaultValue (typeof (WebString), "")]
```

Remotion/Web/Core/UI/Controls/WebMenuItem.cs
```diff
@@ -169,21 +177,20 @@ public string Category
 
     [PersistenceMode (PersistenceMode.Attribute)]
     [Category ("Appearance")]
-    [Description ("The text displayed in this menu item. Use '-' for a separator menu item. The value will not be HTML encoded.")]
+    [Description ("The text displayed in this menu item. Use '-' for a separator menu item.")]
```
MichaelKetting, line 180: Good catch!

Remotion/Web/Core/UI/Controls/WebMenuItem.cs
```diff
@@ -169,21 +177,20 @@ public string Category
...
-      set { _text = value ?? string.Empty; }
+      set { _text = value; }
     }
 
     [DesignerSerializationVisibility (DesignerSerializationVisibility.Hidden)]
     [Browsable (false)]
     public bool IsSeparator
     {
-      get { return _text == c_separator; }
+      get { return _text.GetValue() == c_separator; }
```
MichaelKetting, line 193: Should we make the c_separator into a WebString static instead of a string const and then we can do a compare here? Didn't check what that means for all the usages of c_separator, though.

Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenu.cs
```diff
@@ -202,14 +202,15 @@ private void PreRenderDropDownMenu()
...
   }
 
-  protected virtual string GetTitleText (IBusinessObject businessObject)
+  protected virtual WebString GetTitleText (IBusinessObject businessObject)
   {
     ArgumentUtility.CheckNotNull ("businessObject", businessObject);
+
+    var titleText = businessObject is IBusinessObjectWithIdentity
+        ? ((IBusinessObjectWithIdentity) businessObject).GetAccessibleDisplayName()
+        : businessObject.ToString();
```
MichaelKetting, line 211:
```suggestion
    var titleText = businessObject is IBusinessObjectWithIdentity
        ? ((IBusinessObjectWithIdentity) businessObject).GetAccessibleDisplayName()
        : businessObject.ToString();
```
```suggestion
    var titleText = businessObject switch
    {
        IBusinessObjectWithIdentity businessObjectWithIdentity => businessObjectWithIdentity.GetAccessibleDisplayName(),
        _ => businessObject.ToString()
    };
```

Do you think we should do a switch expression instead. The semantic of the IF is a type switch so it feels more correct to do pattern matching instead of an if/ternary. Thoughts?

Remotion/Web/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRenderer.cs
```diff
@@ -323,8 +325,13 @@ private void AppendListMenuItem (ListMenuRenderingContext renderingContext, Stri
         if (!string.IsNullOrEmpty(menuItem.ItemID))
           diagnosticMetadataDictionary.Add(DiagnosticMetadataAttributes.ItemID, menuItem.ItemID);
 
-        if (!string.IsNullOrEmpty(menuItem.Text))
-          diagnosticMetadataDictionary.Add(DiagnosticMetadataAttributes.Content, HtmlUtility.StripHtmlTags(menuItem.Text));
+        if (!menuItem.Text.IsEmpty)
+        {
+          var contentAttributeValue = menuItem.Text.Type == WebStringType.Encoded
+              ? HtmlUtility.StripHtmlTags (menuItem.Text.GetValue())
+              : menuItem.Text.GetValue();
```
MichaelKetting, line 332: What do you think, should we do an overload for StripHtmlTags that accepts WebString and wraps this semantic? It feels like a good idea given we already have this web-test-helper method anyway.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocDropDownMenu.cs
a-ctor: In this case I think doing a switch expression does not add much because we only have two branches, but I changed the ternary operator to use pattern matching, removing the need for casting

Remotion/Web/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRenderer.cs
a-ctor: I think it is a good idea, that also solves the problem for future tasks because the code now compiles without problems :D

Remotion/Web/Core/UI/Controls/WebMenuItem.cs
MichaelKetting: Ah, yes, in case someone has the idea of setting `-` with encoded. Hadn't thought of that but you probably can't prevent such nonsense. so, good choice!

Remotion/Web/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRenderer.cs
MichaelKetting: > hat also solves the problem for future tasks because the code now compiles without problems :D
Am not sure how to interpret the "compiles without problems" part. Was this about doing this change "now vs later" or something else?

a-ctor: Other code that uses the `Content` diagnostic metadata attribute already uses the `HttpUtility.StripHtmlTags` but with the `string` overload. If a future task changes the type of a property to `WebString` it will just compile without change because the `WebString` overload of `HttpUtility.StripHtmlTags` is used

MichaelKetting: Could you pelase create a ticket for this refactoring? thanks!

a-ctor: https://re-motion.atlassian.net/browse/RM-8170

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
```diff
@@ -410,15 +410,17 @@ private void AppendMenuItem (DropDownMenuRenderingContext renderingContext, Stri
 
       string icon = GetIconUrl (renderingContext, menuItem, showIcon);
       string disabledIcon = GetDisabledIconUrl (renderingContext, menuItem, showIcon);
-      string text = showText ? "'" + menuItem.Text + "'" : "null";
-      string diagnosticMetadataText = showText ? menuItem.Text : "";
+      string text = showText
+          ? "'" + ScriptUtility.EscapeClientScript (menuItem.Text.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks)) + "'"
```
MichaelKetting, line 414: Another possible usecase for an overload. please ticket :)

Remotion/Web/Core/UI/Controls/WebMenuItem.cs
```diff
@@ -433,9 +440,9 @@ public virtual void LoadResources (IResourceManager resourceManager, IGlobalizat
       if (!string.IsNullOrEmpty (key))
         Category = resourceManager.GetString (key);
 
-      key = ResourceManagerUtility.GetGlobalResourceKey (Text);
+      key = ResourceManagerUtility.GetGlobalResourceKey (Text.GetValue());
       if (!string.IsNullOrEmpty (key))
-        Text = resourceManager.GetString (key);
+        Text = resourceManager.GetWebString (key, Text.Type);
```
MichaelKetting, line 445: Do you have any good ideas how we can put those lines into a utility method?

```      key = ResourceManagerUtility.GetGlobalResourceKey (Text.GetValue());
      if (!string.IsNullOrEmpty (key))
        Text = resourceManager.GetWebString (key, Text.Type);```
I can only think of using a by-ref for the value type, which might be okay, but feels kind of strange to me.

File: Remotion/Web/Core/Utilities/HtmlUtility.cs [hunk 13]
```diff
@@ -59,6 +59,13 @@ public static string StripHtmlTags ([NotNull] string text)
       return s_stripHtmlTagsRegex.Replace (text, string.Empty);
     }
 
+    public static string StripHtmlTags (WebString text)
```
MichaelKetting, line 62: I'm missing the unit tests here. please add.

MichaelKetting (changes requested):

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
a-ctor: https://re-motion.atlassian.net/browse/RM-8172

Remotion/Web/Core/UI/Controls/WebMenuItem.cs
a-ctor: We want to convert the following lines into a utility not just the if and the assignment?

```c#
key = ResourceManagerUtility.GetGlobalResourceKey (Text.GetValue());
if (!string.IsNullOrEmpty (key))
  Text = resourceManager.GetWebString (key, Text.Type);
```

Since we have the assignment on an arbitrary property it gets a bit trick. This are the solutions I came up with:

 - Use the backing field as `ref` target and move the code into a helper method
   `resourceManager.TryGetWebString(Text, ref _text);`
 - Use `out` and a `bool` return to get simplify the code around the if: 
   `if (resourceManager.TryGetWebString(Text, out var result) Text = result;`
 - Use a delegate to set the property. Useful if we need the property setter

MichaelKetting: yeah, it's the complete pattern that's interesting. doing just the `if` isn't all that helpful. 

And that's what I was afraid of - there's not really a good solution, is there. Missed the bit about Property vs Field here, too, so....

The delegate's expensive, I think. We'd need to pass the property as an expression and then get the getter and setter delegates or do direct invokes. So, just accept it like this and maybe create an analyzer for this pattern?

MichaelKetting (approved):

PR #257, opened 2021-11-17
RM-8141 Change type of text properties on `BocCheckBox` and `BocBooleanValue` from `string` to `WebString`

iam-mholle:


derlistigelurch (approved):

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
```diff
@@ -141,23 +141,23 @@ public void RenderTrue ()
     {
       _booleanValue.Setup (mock => mock.Enabled).Returns (true);
       _booleanValue.Object.Value = true;
-      CheckRendering ("true", true.ToString(), "TrueIconUrl", _booleanValue.Object.TrueDescription);
+      CheckRendering ("true", true.ToString(), "TrueIconUrl", _booleanValue.Object.TrueDescription.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks));
```
MichaelKetting, line 144: Given the uniformity of this change, I would recommend changing the parameter type and passing webstring instead of string to improve maintainability: 
1. if we need to change/add tests, they also require the "long" form.
2. if we change this now instead of later, we only have a single place in the code where this change happens instead of changing 10 occurances now and then changing the same 10 occurances later on, reducing effort down the line and the total affectd changeset.

against this speaks the fact that value isn't passed as a nullable boolean but also a string. That was done to have more control over the passed string in case a test needs to do something special. I'm not sure I would still do it this way to today instead of passing a nullable boolean.

pro/cons for this refactoring?

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRendererTest.cs
```diff
@@ -98,14 +99,14 @@ public void SetUp ()
     public void RenderTrue ()
     {
       _checkbox.Setup (mock => mock.Enabled).Returns (true);
-      CheckRender (true, _checkbox.Object.TrueDescription);
+      CheckRender (true, _checkbox.Object.TrueDescription.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks));
```
MichaelKetting, line 102: Same question as with BocBooleanRendererTest

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRenderer.cs
```diff
@@ -269,11 +269,11 @@ private void RegisterStarupScriptIfNeeded (BocBooleanValueRenderingContext rende
       scriptBuilder.Append (", ");
       scriptBuilder.Append (requiredFlag);
       scriptBuilder.Append (", ");
-      AppendStringValueOrNullToScript (scriptBuilder, renderingContext.Control.TrueDescription);
+      AppendStringValueOrNullToScript (scriptBuilder, renderingContext.Control.TrueDescription.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks));
```
MichaelKetting, line 272: please add a ticket for adding an overload for this. (fyi: we already have a ticket for EscapeClientStript)

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRenderer.cs
```diff
@@ -297,25 +297,25 @@ private void RegisterStarupScriptIfNeeded (BocBooleanValueRenderingContext rende
       {
         checkedState = HtmlAriaCheckedAttributeValue.Mixed;
         imageUrl = resourceSet.NullIconUrl;
-        description = string.IsNullOrEmpty (renderingContext.Control.NullDescription)
-            ? resourceSet.DefaultNullDescription
-            : renderingContext.Control.NullDescription;
+        description = renderingContext.Control.NullDescription.IsEmpty
```
MichaelKetting, line 300: Please change type of description to WebString. That way, we don't do this in every line, removing the code duplication and syntactical noise.

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
```diff
@@ -207,8 +207,12 @@ private string GetScriptParameters (BocCheckBoxRenderingContext renderingContext
...
                       + checkBox + ", "
                       + label + ", "
-                      + (string.IsNullOrEmpty (renderingContext.Control.TrueDescription) ? "null" : "'" + renderingContext.Control.TrueDescription + "'") + ", "
-                      + (string.IsNullOrEmpty (renderingContext.Control.FalseDescription) ? "null" : "'" + renderingContext.Control.FalseDescription + "'") + ");";
+                      + (renderingContext.Control.TrueDescription.IsEmpty
+                          ? "null"
+                          : "'" + renderingContext.Control.TrueDescription.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks) + "'") + ", "
+                      + (renderingContext.Control.FalseDescription.IsEmpty
+                          ? "null"
+                          : "'" + renderingContext.Control.FalseDescription.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks) + "'") + ");";
```
MichaelKetting, line 215: It's a bug that we don't use EscapeClientScript here. Please update (it's mentioned also in other tasks where this oversight was more obvious during task design)

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
```diff
@@ -257,8 +261,13 @@ private string GetDescription (BocCheckBoxRenderingContext renderingContext)
         string defaultTrueDescription = renderingContext.Control.DefaultTrueDescription;
         string defaultFalseDescription = renderingContext.Control.DefaultFalseDescription;
```
MichaelKetting, line 262: Please update the type for the Default... values, too (consistency). in the end, we can return WebString instead of string.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
iam-mholle: Changing it makes sense

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
```diff
@@ -222,44 +227,50 @@ private void RegisterStartupScriptIfNeeded (BocCheckBoxRenderingContext renderin
...
+    private void PrepareImage (BocCheckBoxRenderingContext renderingContext, Image imageControl, WebString? description)
     {
       var imageUrl = ResourceUrlFactory.CreateThemedResourceUrl (
           typeof (HtmlHeadAppenderExtensions),
           ResourceType.Image,
           renderingContext.Control.Value.Value ? c_trueIcon : c_falseIcon);
 
       imageControl.ImageUrl = imageUrl.GetUrl();
-      imageControl.AlternateText = description ?? string.Empty;
+      imageControl.AlternateText = description?.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks) ?? string.Empty;
```
MichaelKetting, line 243:
```suggestion
      imageControl.AlternateText = description?.GetValue() ?? string.Empty;
```
Sorry, I missed this before. I think this is the correct way to handle this (please check, I could miss something). AlternateText is an attribute, if we do Html or HtmlWithLinebreaks, the AlternateText would contain and then show the html characters.

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
```diff
@@ -222,44 +227,50 @@ private void RegisterStartupScriptIfNeeded (BocCheckBoxRenderingContext renderin
...
       labelControl.CssClass = "description";
-      labelControl.Text = description;
+      labelControl.Text = description?.ToString (WebStringEncoding.HtmlWithTransformedLineBreaks);
       labelControl.Width = Unit.Empty;
       labelControl.Height = Unit.Empty;
       labelControl.ApplyStyle (renderingContext.Control.LabelStyle);
     }
 
-    private string GetDescription (BocCheckBoxRenderingContext renderingContext)
+    private WebString? GetDescription (BocCheckBoxRenderingContext renderingContext)
```
MichaelKetting, line 256: Please check, I think we can drop the nullability and simplify the code for the null/empty case. I didn't see any place where null is actually handled beyond making sure that the string value is empty when it is passed on to by asp.net.

MichaelKetting (changes requested):
sorry, found a bug and a code simplifcation that makes sense to clean up now that we have a good datatype for this.

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
iam-mholle: shouldn't this be `description?.ToString (WebStringEncoding.Attribute)` then?

MichaelKetting: yeah, good question. given that asp.net already uses encoding when AlternateText is rendered, we might end up with double-encode. I don't know if there's an actual example to be found where this results in a problem, (maybe quotes?) but my gut feeling says that this would lead to problems down the line.

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
```diff
@@ -222,45 +227,52 @@ private void RegisterStartupScriptIfNeeded (BocCheckBoxRenderingContext renderin
...
       labelControl.ApplyStyle (renderingContext.Control.LabelStyle);
     }
 
-    private string GetDescription (BocCheckBoxRenderingContext renderingContext)
+    private WebString GetDescription (BocCheckBoxRenderingContext renderingContext)
     {
-      string trueDescription = null;
-      string falseDescription = null;
+      WebString? trueDescription = null;
+      WebString? falseDescription = null;
```
MichaelKetting, line 259: Question about stylistic approach: why not set this to WebString.Empty? I guess either is fine but using WebString.Empty seems like it's less "noise".

MichaelKetting (approved):
I've got a stylic question but that's just commentary unless you think it's actually a relevant improvement.

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocCheckBoxRenderer.cs
iam-mholle: Makes sense, I'll change it

PR #261, opened 2021-11-18
Feature/rm 8139 change text properties web tree view boc tree view web string

a-ctor:
https://re-motion.atlassian.net/browse/RM-8139

### DIFF
--- a/Remotion/Web/Core/WebStringConverter.cs
+++ b/Remotion/Web/Core/WebStringConverter.cs
@@ -100,7 +100,7 @@ namespace Remotion.Web
 
       if (input.Type == WebStringType.PlainText)
       {
-        if (rawString.StartsWith (c_htmlPrefix))
+        if (rawString.StartsWith (c_htmlPrefix) || rawString.StartsWith (c_textPrefix))
           return c_textPrefix + rawString;
 
         return rawString;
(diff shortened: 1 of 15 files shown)


Remotion/Web/Core/WebStringConverter.cs
```diff
@@ -100,7 +100,7 @@ private object ConvertToString (in WebString input)
 
       if (input.Type == WebStringType.PlainText)
       {
-        if (rawString.StartsWith (c_htmlPrefix))
+        if (rawString.StartsWith (c_htmlPrefix) || rawString.StartsWith (c_textPrefix))
```
MichaelKetting, line 103: Why this change?

Remotion/Web/UnitTests/Core/WebStringConverterTest.cs
```diff
@@ -66,6 +66,7 @@ public void ConvertToString (WebString? input, string expectedOutput)
         new object[] { WebString.CreateFromText (string.Empty), string.Empty },
         new object[] { WebString.CreateFromText ("test"), "test" },
         new object[] { WebString.CreateFromText ("(html)test"), "(text)(html)test" },
+        new object[] { WebString.CreateFromText ("(text)test"), "(text)(text)test" },
```
MichaelKetting, line 69: See https://github.com/re-motion/Framework/pull/261/files#diff-fe47fd88d0742bdd44b6142569bf1c5762f84c8178e53b12c080c2b1d7d65aaaL103-R103

MichaelKetting (changes requested):

Remotion/Web/Core/WebStringConverter.cs
a-ctor: Without it the conversion is not correct in the special case that the text starts with `(text)`. For example:
 - we have a WebString with text `(text)asd` and a type plaintext
 - we convert it to a `string` using `ConvertTo` which as of now will return `(text)asd` instead of `(text)(text)asd`
 - we convert it back to a `WebString` using `ConvertFrom` but this will remove the `(text)` prefix leaving us with a web string with text `asd` type plaintext

MichaelKetting (approved):
I guess this means we're done?

a-ctor:
I guess so

PR #262, opened 2021-11-18
RM-8194 Add Wrapper PlainTextString

a-ctor:
https://re-motion.atlassian.net/browse/RM-8194

### DIFF
--- a/Remotion/Web/Core/PlainTextString.cs
+++ b/Remotion/Web/Core/PlainTextString.cs
@@ -1,0 +1,194 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.ComponentModel;
+using System.IO;
+using System.Web;
+using System.Web.UI;
+using JetBrains.Annotations;
+using Remotion.Utilities;
+
+namespace Remotion.Web
+{
+  /// <summary>
+  /// The <see cref="PlainTextString"/> structure represents strings that are encoded before being rendered onto a web page.
+  /// </summary>
+  [TypeConverter (typeof (PlainTextStringConverter))]
+  public readonly struct PlainTextString : IEquatable<PlainTextString>
+  {
+    private static readonly string[] s_newlineSplitSeparators = { "\r\n", "\r", "\n" };
+
+    public static readonly PlainTextString Empty = new(string.Empty);
+
+    /// <summary>
+    /// Creates a <see cref="PlainTextString"/> from the specified <paramref name="text"/>.
+    /// The value will be encoded before being rendered.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


Remotion/Web/Core/WebString.cs
```diff
@@ -176,6 +176,23 @@ public void AddAttribute ([NotNull] HtmlTextWriter writer, [NotNull] string attr
       }
     }
 
+    /// <summary>
+    /// Tries to convert this <see cref="WebString"/> into a <see cref="PlainTextString"/>.
+    /// Only works if <see cref="Type"/> is <see cref="WebStringType"/>.<see cref="WebStringType.PlainText"/>,
+    /// and will throw an <see cref="InvalidOperationException"/> otherwise.
```
MichaelKetting, line 182: There's too much redundant information in the summary (i.e. the exception info). If possible, the summary should be a single line with just the happy path semantics.
```suggestion
    /// Converts this <see cref="WebString"/> into a <see cref="PlainTextString"/> if <see cref="Type"/> is <see cref="WebStringType.PlainText"/>.
```

Remotion/Web/Core/PlainTextString.cs
```diff
@@ -0,0 +1,220 @@
...
+    [CanBeNull]
+    private readonly string? _value;
+
+    private PlainTextString ([CanBeNull] string? value)
+    {
+      _value = value;
+    }
+
+    /// <summary>
+    /// Tests whether the <see cref="PlainTextString"/>s value is <see langword="null" /> or an empty string.
```
MichaelKetting, line 103: This is a good example where using "Sytem.String" instead "string" would be overkill / noise, even though strictly speaking, more correct.

Remotion/Web/Core/PlainTextString.cs
```diff
@@ -0,0 +1,220 @@
...
+    /// </summary>
+    [NotNull]
+    public string GetValue () => _value ?? string.Empty;
+
+    /// <summary>
+    /// Writes the <see cref="PlainTextString"/>'s value to the specified <paramref name="writer"/>.
+    /// The value is escaped before being written.
+    /// </summary>
+    /// <param name="writer">The <see cref="HtmlTextWriter"/> where the value will be appended to. Must not be <see langword="null" />.</param>
+    public void Write ([NotNull] HtmlTextWriter writer)
```
MichaelKetting, line 118: I think it would be a good idea to delegate to WebString for the various rendering methods. Otherwise we do have duplication and PlainTextString is kind of a special case of WebString. Once we follow up with the idea of making these into HtmlTextWriter-extensions, the problem of where to locate them gets solved anyway.

Remotion/Web/Core/PlainTextString.cs
```diff
@@ -0,0 +1,220 @@
...
+    public override string ToString ()
+    {
+      return ToString (WebStringEncoding.Html);
+    }
+
+    /// <summary>
+    /// Returns a string that represents the current <see cref="PlainTextString"/> in the specified <paramref name="encoding"/>.
+    /// </summary>
+    /// <returns>A string that represents the current <see cref="PlainTextString"/> in the specified <paramref name="encoding"/>.</returns>
+    public string ToString (WebStringEncoding encoding)
```
MichaelKetting, line 197: Should also delegate ToString to WebString. Same reason as the rendering.

Remotion/Web/UnitTests/Core/WebStringTest.cs
```diff
@@ -255,6 +255,22 @@ public void ToString_WithPlainTextTextAndHtmlWithEncodedLineBreaksEncoding_Encod
...
+
+      Assert.That (plainTextString.GetValue(), Is.EqualTo ("test"));
+    }
+
+    [Test]
+    public void ToPlainTextString_WithEncodedText_ThrowsInvalidOperationException ()
+    {
+      Assert.That(
+          () => WebString.CreateFromHtml ("test").ToPlainTextString(),
+          Throws.TypeOf<InvalidOperationException>());
```
MichaelKetting, line 271: Please assert message too. Helps with making sure that a) the correct exception case gets hit and b) the message works (for constructed messages)

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #267, opened 2021-11-22
RM-8186: Add RendererBase.AppendStringValueOrNullToScript overload

iam-mholle:


Remotion/Web/Core/UI/Controls/RendererBase.cs
```diff
@@ -117,6 +117,14 @@ protected void AppendStringValueOrNullToScript (StringBuilder scriptBuilder, str
         scriptBuilder.Append ("'").Append (ScriptUtility.EscapeClientScript (stringValue)).Append ("'");
     }
 
+    protected void AppendStringValueOrNullToScript (StringBuilder scriptBuilder, WebString? stringValue)
+    {
+      if (stringValue.HasValue)
```
MichaelKetting, line 122: 1. should only render if not empty. See system.string overload.

2. nullable webstring or just webstring: the advantage of letting it be nullable is that we can keep the overload name as is. On the other hand, we already have NRT in Remotion.Web and the other overload isn't nullable, either. So, shoud we just make this not-nullable and update the name on both overloads or alternatively, add nullalibity for the other overload too? I'm thinking by adding nullability, we have a more robust solution for the name (less chance of this needing to change in the future), but should add the `?` to the other overload.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #268, opened 2021-11-23
Feature/rm 8124 update validation infrastructure to use plain text error messages

a-ctor:
https://re-motion.atlassian.net/browse/RM-8124

### DIFF
--- a/Remotion/Development/Web/UnitTesting/UI/Controls/Rendering/StubValidationErrorRenderer.cs
+++ b/Remotion/Development/Web/UnitTesting/UI/Controls/Rendering/StubValidationErrorRenderer.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Web.UI;
 using Remotion.Utilities;
+using Remotion.Web;
 using Remotion.Web.UI.Controls.Rendering;
 
 namespace Remotion.Development.Web.UnitTesting.UI.Controls.Rendering
@@ -33,7 +34,7 @@ namespace Remotion.Development.Web.UnitTesting.UI.Controls.Rendering
     public void SetValidationErrorsReferenceOnControl (
         IAttributeAccessor attributeAccessor,
         string validationErrorID,
-        IReadOnlyCollection<string> validationErrors)
+        IReadOnlyCollection<PlainTextString> validationErrors)
     {
       ArgumentUtility.CheckNotNull ("attributeAccessor", attributeAccessor);
       ArgumentUtility.CheckNotNullOrEmpty ("validationErrorID", validationErrorID);
@@ -46,7 +47,7 @@ namespace Remotion.Development.Web.UnitTesting.UI.Controls.Rendering
     public void AddValidationErrorsReference (
         AttributeCollection attributeCollection,
         string validationErrorID,
-        IReadOnlyCollection<string> validationErrors)
+        IReadOnlyCollection<PlainTextString> validationErrors)
     {
       ArgumentUtility.CheckNotNull ("attributeCollection", attributeCollection);
       ArgumentUtility.CheckNotNullOrEmpty ("validationErrorID", validationErrorID);
@@ -56,7 +57,10 @@ namespace Remotion.Development.Web.UnitTesting.UI.Controls.Rendering
       attributeCollection[ValidationErrorsAttribute] = string.Join (" ", validationErrors);
     }
 
-    public void RenderValidationErrors (HtmlTextWriter htmlTextWriter, string validationErrorID, IReadOnlyCollection<string> validationErrors)
+    public void RenderValidationErrors (
+        HtmlTextWriter htmlTextWriter,
+        string validationErrorID,
+        IReadOnlyCollection<PlainTextString> validationErrors)
     {
       ArgumentUtility.CheckNotNull ("htmlTextWriter", htmlTextWriter);
       ArgumentUtility.CheckNotNullOrEmpty ("validationErrorID", validationErrorID);
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
@@ -51,7 +51,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocBooleanValueImplem
     private const string c_keyValueName = "MyBooleanValue_Value";
     private const string c_displayValueName = "MyBooleanValue_DisplayValue";
     private const string c_labelID = "Label";
-    private const string c_validationErrors = "ValidationError";
+
+    private static readonly PlainTextString s_validationErrors = PlainTextString.CreateFromText ("ValidationError");
 
     private string _startupScript;
     private string _keyDownScript;
@@ -83,7 +84,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocBooleanValueImplem
       _booleanValue.Setup (mock => mock.GetValueName ()).Returns (c_keyValueName);
       _booleanValue.Setup (mock => mock.GetDisplayValueName()).Returns (c_displayValueName);
       _booleanValue.Setup (mock => mock.GetLabelIDs()).Returns (EnumerableUtility.Singleton (c_labelID));
-      _booleanValue.Setup (mock => mock.GetValidationErrors()).Returns (EnumerableUtility.Singleton (c_validationErrors));
+      _booleanValue.Setup (mock => mock.GetValidationErrors()).Returns (EnumerableUtility.Singleton (s_validationErrors));
       
       string startupScriptKey = typeof (BocBooleanValueRenderer).FullName + "_Startup_" + _resourceSet.ResourceKey;
       _startupScript = string.Format (
@@ -435,7 +436,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocBooleanValueImplem
     private void AssertValidationErrors (XmlNode node)
     {
       Html.AssertAttribute (node, StubValidationErrorRenderer.ValidationErrorsIDAttribute, c_clientID + "_ValidationErrors");
-      Html.AssertAttribute (node, StubValidationErrorRenderer.ValidationErrorsAttribute, c_validationErrors);
+      Html.AssertAttribute (node, StubValidationErrorRenderer.ValidationErrorsAttribute, s_validationErrors);
     }
 
     private void CheckHiddenField (XmlNode outerSpan, string value)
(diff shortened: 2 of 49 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
```diff
@@ -964,9 +964,13 @@ string IBocReferenceValueBase.GetLabelText ()
...
     }
 
-    IEnumerable<string> IBocReferenceValueBase.GetValidationErrors ()
+    IEnumerable<PlainTextString> IBocReferenceValueBase.GetValidationErrors ()
     {
-      return GetRegisteredValidators().Where (v => !v.IsValid).Select (v => v.ErrorMessage).Distinct();
+      return GetRegisteredValidators()
+          .Where (v => !v.IsValid)
+          .Select (v => v.ErrorMessage)
+          .Select (PlainTextString.CreateFromText)
```
MichaelKetting, line 972: Shouldn't we flip this and make the distinct first? 
(all occurances)

Also, should we take heart and add an extension method for `IEnumerable<PlaintextString> GetDistinctValidationErrors()(IEnumerable<IValidator>  this)`?

Remotion/Web/Core/UI/Controls/FormGridManager.cs
```diff
@@ -1675,13 +1675,13 @@ private void RegisterValidationErrors (FormGridRow dataRow)
...
-          if (!string.IsNullOrEmpty (validationMessage))
+          if (!validationMessage.IsEmpty)
           {
             if (toolTip.Length > 0)
               toolTip.AppendLine();
-            toolTip.Append (validationMessage);
+            toolTip.Append (validationMessage.GetValue());
           }
         }
         dataRow.ValidationMarker = CreateValidationMarker (toolTip.ToString());
```
MichaelKetting, line 1687: Update the api to pass a PlainTextString?

Remotion/Web/Core/UI/Controls/ValidationError.cs
```diff
@@ -42,7 +42,7 @@ public class ValidationError
     private ControlCollection? _labels;
 
     /// <summary> The message to be displayed to the user. </summary>
-    private string? _validationMessage;
+    private PlainTextString? _validationMessage;
```
MichaelKetting, line 45: not null, please. that's just based on NRT

Remotion/Web/Core/UI/Controls/ValidationError.cs
```diff
@@ -144,7 +142,7 @@ public HtmlGenericControl ToLabel (string cssClass)
     {
       HtmlGenericControl label = new HtmlGenericControl ("label");
 
-      label.Controls.Add (ToSpan (ValidationMessage));
+      label.Controls.Add (ToSpan (ValidationMessage.GetValue()));
```
MichaelKetting, line 145: Great, I just found a bug, here and with the other usage of ToSpan.

ToSpan gets a cssclass, not the validationmessage. => pass `null`, since we already add the css lcass to the parent element.

This should be a dedicated issue, though.

Remotion/Web/UnitTests/Core/UI/Controls/Rendering/ValidationErrorRendererTest.cs
```diff
@@ -182,7 +212,7 @@ public void RenderValidationError ()
       validationErrorRenderer.RenderValidationErrors (
           _html.Writer,
           "ValidationErrorID",
-          new[] { "ValidationError1", "ValidationError2" });
+          new[] { PlainTextString.CreateFromText ("ValidationError1"), PlainTextString.CreateFromText ("ValidationError2") });
```
MichaelKetting, line 215: We should add a line break to the individual text, too, so we can better assert the correct api usage.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
```diff
@@ -435,7 +436,7 @@ private void CheckLinkAttributes (XmlNode link, string checkedState, string desc
     private void AssertValidationErrors (XmlNode node)
     {
       Html.AssertAttribute (node, StubValidationErrorRenderer.ValidationErrorsIDAttribute, c_clientID + "_ValidationErrors");
-      Html.AssertAttribute (node, StubValidationErrorRenderer.ValidationErrorsAttribute, c_validationErrors);
+      Html.AssertAttribute (node, StubValidationErrorRenderer.ValidationErrorsAttribute, s_validationErrors.GetValue());
```
MichaelKetting, line 439: The fake-implementation renderes attributes for the validation errors. I think we should align test and fake somehow because right now it works mainly because the string doesn't contain to-be-escaped chars. Would ToString() be more correct?

Remotion/Web/Core/UI/Controls/ValidationError.cs
```diff
@@ -81,17 +81,15 @@ public ValidationError (Control validatedControl, IValidator validator)
     /// <param name="validatedControl"> The control with an invalid state. </param>
     /// <param name="validationMessage"> The message to be displayed to the user. Must not be <see langword="null"/> or empty. </param>
     /// <param name="labels">The labels containing the control's headings.</param>
-    public ValidationError (Control validatedControl, string validationMessage, ControlCollection? labels)
+    public ValidationError (Control validatedControl, PlainTextString validationMessage, ControlCollection? labels)
     {
-      ArgumentUtility.CheckNotNullOrEmpty ("validationMessage", validationMessage);
```
MichaelKetting, line 86: CheckNotEmpty is missing

MichaelKetting (changes requested):

Remotion/Development/Web/UnitTesting/UI/Controls/Rendering/StubValidationErrorRenderer.cs
```diff
@@ -56,7 +57,10 @@ public class StubValidationErrorRenderer : IValidationErrorRenderer
       attributeCollection[ValidationErrorsAttribute] = string.Join (" ", validationErrors);
```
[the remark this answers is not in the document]
a-ctor: Added to RM-8215

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
a-ctor: Shouldn't make much of a difference in this case because both boil down to a pointer but it also should not hurt. If all the subclasses do the same thing why don't we introduce a method at the level of `GetRegisteredValidators`?

Remotion/Web/Core/UI/Controls/FormGridManager.cs
a-ctor: What do you mean? Adding a `StringBuilder` method on `PlainTextString` or an extension method?

MichaelKetting: CreateValidationMarker (string) => CreateValidationMarker (PlainTextString)

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/BocReferenceValueBase.cs
MichaelKetting: There's a difference because PlainTextString.ToStirng() performs HTML Escaping of the string and then does a compare, unless I'm missing something?

Why extension methid instead of an alternative to GetRegisteredValidators()?
To me, it's easier to understand with a more modular approach. Of course, I could name the method `GetDistinctValidationErrorsForAllRegisteredValidators()` but this makes the usages of GetRegisteredValidators() (the impact in particular) more hidden and doesn't help with read-abilty. Also, there could be (in theory) other usecases for this API aside from processing the result of GetRegisteredValidators() when we put it in Remotion.Web.Utilities.ValidatorExtensions (or similar).

a-ctor: `ToString` does do encoding, but `Distinct` should use as hash set and thus the `Equals` and `GetHashCode` methods, which do not encode. `PlainTextString.CreateFromText` just wraps into the string and also does not encode. So it should (fingers crossed) change anything (except that duplicates are not converted to structures first).

Ok I am going to change the `Distinct` order and introduce an extension method

MichaelKetting: Correct, Distcint uses GetHascode and Euquals, so we don't operate on the encoded string. my bad. as you like.

Remotion/Web/Core/UI/Controls/ValidationError.cs
a-ctor: https://re-motion.atlassian.net/browse/RM-8236

Remotion/Web/Core/UI/Controls/FormGridManager.cs
```diff
@@ -1675,16 +1675,16 @@ private void RegisterValidationErrors (FormGridRow dataRow)
...
+          if (!validationMessage.IsEmpty)
           {
             if (toolTip.Length > 0)
               toolTip.AppendLine();
-            toolTip.Append (validationMessage);
+            toolTip.Append (validationMessage.GetValue());
           }
         }
-        dataRow.ValidationMarker = CreateValidationMarker (toolTip.ToString());
+        dataRow.ValidationMarker = CreateValidationMarker ((PlainTextString) toolTip.ToString());
```
MichaelKetting, line 1687: shouldn't we use PlainTextString.CreateFromText for better consistency?

Also, I apparently misremmber: we also have a cast for string to PlainTextString? Is this actually helpful or could this create more confusion since there's two ways to do things?

MichaelKetting (approved):
I found one spot with cast-vs factory method. please check and fix (unless there's a good reason to keep the cast). Otherwise, we're done here.

Remotion/Web/Core/UI/Controls/FormGridManager.cs
a-ctor: I added the factory method `CreateFromText` onto `PlainTextString` to make it more consistent with the `WebString`. Cast and factory method do the same thing so we could get rid of one of them or keep them.

I personally find it more natural to use a cast operator when you have a string and want to pass it to something that requires a `PlainTextString`. We could not do it with `WebString` since there were two options (plaintext or encoded). But I also understand if we want to use only one of the methods. Your choice.

MichaelKetting: I agree with the cast in principle but specific case with PlainTextString, the symmetry to WebString is the preference in the design. We should keep the calls consistent in re-motion, but I don't "care" about whether we have the cast or not for downstream usage, unless we want to give explicit guidance here.

PR #269, opened 2021-11-23
RM-8167: Use DiagnosticMetadataAttributes.Content instead of the Text API to fetch button text

iam-mholle:


derlistigelurch (approved):

Remotion/Web/Development.WebTesting.ControlObjects/WebButtonControlObject.cs
```diff
@@ -48,7 +48,7 @@ public IStyleInformation StyleInfo
     /// </summary>
     public string GetText ()
     {
-      return Scope.Text.Trim();
+      return Scope[DiagnosticMetadataAttributes.Content].Trim();
```
MichaelKetting, line 51: Trim() should not be necessary when we have a dedicated attribute.
```suggestion
      return Scope[DiagnosticMetadataAttributes.Content];
```

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #271, opened 2021-11-25
RM-8215 WebString Analyzer

iam-mholle:


Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,223 @@
...
+        var argumentType = GetFirstArgumentTypeSymbol (node, ctx.SemanticModel);
+        if (SymbolEqualityComparer.Default.Equals (argumentType, symbols.WebStringTypeSymbol)
+            || SymbolEqualityComparer.Default.Equals (argumentType, symbols.PlainTextStringTypeSymbol))
+        {
+          var diagnostic = Diagnostic.Create (
+              DiagnosticDescriptor,
+              ctx.Node.GetLocation(),
+              symbols.StringBuilderAppendMethodSymbol.ToString(),
+              argumentType.ToString(),
+              "Call '.ToString(WebStringEncoding.HtmlWithTransformedLineBreaks)' on the WebString instance instead.");
```
MichaelKetting, line 108: Don't we need to cover WebString and PlainTextString separately? The message only refers to WebString.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,223 @@
...
+
+    private static IMethodSymbol GetStringBuilderAppendObjectMethodSymbol (INamedTypeSymbol stringBuilderTypeSymbol, INamedTypeSymbol? objectSymbol)
+    {
+      return stringBuilderTypeSymbol
+          .GetMembers ("Append")
+          .Cast<IMethodSymbol>()
+          .Single (m => m.Parameters.Length == 1 && SymbolEqualityComparer.Default.Equals (m.Parameters.Single().Type, objectSymbol));
+    }
+
+    class SymbolContext
```
MichaelKetting, line 194:
```suggestion
    private class SymbolContext
```

MichaelKetting, line 194: Also, our style guide says that nested types are placed on the top of the class (nested types, const, static, instance fields, ctor, members)

MichaelKetting, line 194: Question: would this be a good usecase for record?

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            var objectSymbol = compilationContext.Compilation.GetTypeByMetadataName ("System.Object");
+            var stringSymbol = compilationContext.Compilation.GetTypeByMetadataName ("System.String");
+            var webStringEnumerable = compilationContext.Compilation.GetTypeByMetadataName ("System.Collections.Generic.IEnumerable`1")!
+                .Construct (webStringTypeSymbol);
+            var plainTextStringEnumerable = compilationContext.Compilation.GetTypeByMetadataName ("System.Collections.Generic.IEnumerable`1")!
+                .Construct (plainTextStringTypeSymbol);
+            var stringBuilderAppendMethodSymbol = GetStringBuilderAppendObjectMethodSymbol (stringBuilderTypeSymbol, objectSymbol);
+            var htmlTextWriterWriteMethodSymbol = GetHtmlTextWriterWriteObjectMethodSymbol (htmlTextWriterSymbol, objectSymbol);
+            var stringJoinMethodSymbols = GetStringJoinMethodSymbols (stringSymbol!);
```
MichaelKetting, line 70: Why do we have the null-forgiving here?

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,223 @@
...
+
+    private static IReadOnlyCollection<IMethodSymbol> GetStringJoinMethodSymbols (INamedTypeSymbol stringSymbol)
+    {
+      return stringSymbol
+          .GetMembers ("Join")
+          .Cast<IMethodSymbol>()
+          .ToList();
+    }
+
+    private static IMethodSymbol GetHtmlTextWriterWriteObjectMethodSymbol (INamedTypeSymbol htmlTextWriterSymbol, INamedTypeSymbol? objectSymbol)
```
MichaelKetting, line 178: Why are some of the parameters nullable and others not? Is this just derived from the usage in the comparer? The callsite doesn't seem to indicate that we need them as nullable.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,223 @@
...
+          ctx.ReportDiagnostic (diagnostic);
+        }
+      }
+    }
+
+    private static bool AnyWithResult<TSymbol> (IEnumerable<TSymbol> methods, Func<TSymbol, bool> predicate, out TSymbol match)
+        where TSymbol : ISymbol
+    {
+      match = methods.FirstOrDefault (predicate)!;
+      return match != null;
```
MichaelKetting, line 156: Stylistic question: to me, `return method.Any(m => m==predicate);` feels like it's easier to understand / more on point of the intended semantics. Good idea / different idea / something I missed?

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,223 @@
...
+          s => SymbolEqualityComparer.Default.Equals (invocationOriginalDefinitionSymbol, s),
+          out var methodSymbol))
+      {
+        var argumentTypes = GetArgumentTypeSymbols (node, ctx.SemanticModel);
+        if (AnyWithResult (
+            argumentTypes,
+            a => SymbolEqualityComparer.Default.Equals (a, symbols.WebStringTypeSymbol)
+                 || SymbolEqualityComparer.Default.Equals (a, symbols.PlainTextStringTypeSymbol)
+                 || ctx.Compilation.ClassifyConversion (a, symbols.WebStringEnumerable).IsImplicit
+                 || ctx.Compilation.ClassifyConversion (a, symbols.PlainTextStringEnumerable).IsImplicit,
```
MichaelKetting, line 138: I'm learning that it's not so easy to check "correctness" during the review when you're not fluent in Roslyn.

I think it would helpful for maintainability to add a short summary to each of the IF blocks with a desciption of the cases covered by the block. Otherwise, you have to reverse engineer the cases from the tickets that belong to the commit and by jumping to (finding) and then reading the tests just to understand what's the intent.

Another option is always to factor out methods that encapsulate each use case.

Thoughts? Pro/Con?

Comment example:
// Detects string.Join (string, (object) WebString)
// Detects string.Join (string, (object) PlainTextString)

Remotion/Web/Development.Analyzers/Web.Development.Analyzers.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <AssemblyName>Remotion.Web.Development.Analyzers</AssemblyName>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.8.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="$(OutputPath)\$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
```
MichaelKetting, line 19: This type if include means we have the assembly show up in the visual studio tree, which isn't best practice. There is the option of adding the assembly directly to the package outout, see https://github.com/NuGet/Home/issues/4704#issuecomment-315894224 or https://github.com/NuGet/Home/issues/4704#issuecomment-315915898. I haven't done a deep dive into the pro/cons of the two versions, but the first one that works should be fine.

I would recommend we consider setting up a dedicated Analyzers.targets proejct type, though. We could do this in a separate task and refactor here when we do the task. depends on how much time is sensible to invest now vs later.

Remotion/Web/Development.Analyzers/Web.Development.Analyzers.csproj
```diff
@@ -0,0 +1,27 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>9</LangVersion>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <DevelopmentDependency>true</DevelopmentDependency>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <AssemblyName>Remotion.Web.Development.Analyzers</AssemblyName>
```
MichaelKetting, line 11: I think this is mising the description to be used in the nuget package and the assembly title. please check the library projects.

Remotion/Web/Development.Analyzers.Tests/RMWEB0001_WebStringAnalyzerTests.cs
```diff
@@ -0,0 +1,254 @@
...
+      var analyzer = new RMWEB0001_WebStringAnalyzer();
+      const string input = @"
+using System.Text;
+using Remotion.Web;
+public class A
+{
+  public void Test()
+  {
+    var sb = new StringBuilder();
+    ↓sb.Append(WebString.CreateFromText(""test""));
```
MichaelKetting, line 47: Would the tests be more precise if we moved the construction of WebString outside of the method call and just used a variable? After all, we don't want to detect if someone uses WebString.CreateFromText("...") but rather if someone uses WebString.

Build/ProjectTypes/Library.props
```diff
@@ -26,4 +26,11 @@
     <PackageReference Include="JetBrains.Annotations" Version="2021.1.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="$(SolutionDir)Remotion\Web\Development.Analyzers\Web.Development.Analyzers.csproj"
```
MichaelKetting, line 30: This should be moved to WebLibary.props.

MichaelKetting (changes requested):

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
iam-mholle: Unit vs. Integration doesn't depend on the runtime but rather on the scope. If Tests doesn't satisfy you here, IntegrationTests would be the way to go imo. After all, a compilation of the input is created, with the analyzer injected, and the output of the compilation/analysis process in asserted.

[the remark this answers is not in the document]
MichaelKetting: Then it's IntegrationTests

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
```diff
@@ -0,0 +1,29 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>net48</TargetFrameworks>
+    <AssemblyTitle>Remotion Web Analyzer Tests</AssemblyTitle>
+    <AssemblyName>Remotion.Web.Development.Analyzers.Tests</AssemblyName>
+    <RootNamespace>$(AssemblyName)</RootNamespace>
+    <SignAssembly>False</SignAssembly>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Gu.Roslyn.Asserts" Version="3.3.2" />
```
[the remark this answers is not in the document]
iam-mholle: > Each dependency we introduce in the framework (non-test code) will carried into all our dev environments (for dev dependencies) and prod environments

It's a dependency of the test assembly tho.

I'm aware that the official package is on it's way (https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Analyzer.Testing/ btw), the last time I checked, it wasn't ready to be used. I'll have another look.

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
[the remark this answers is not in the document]
MichaelKetting: oh, and please cahnge the itemgroup for the the tests to the `IntegrationTestProjectFiles` in the projects.props file.

Remotion/Web/Development.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,5 @@
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|--------------------
+RMWEB0001  | Usage | Error | Wrong WebString or PlainTextString usage.
```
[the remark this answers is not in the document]
iam-mholle: Those files are required and lead to compiler errors if missing. It is unshipped because the rule currently is not released. I'm not sure how to handle this, I guess it would be too much work to keep moving entries from one file to the other on release, and I actually don't want to look into an msbuild integration (unless you really want it). I would propose to add them to the shipped file (versioned correctly) directly as they are added, leaving the unshipped file empty.

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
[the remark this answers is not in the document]
MichaelKetting: Thanks. And as I said, it's okay if there's a know out criteria against the MS package that we can quickly check first. 

As for the test assembly, the thing is, we still have this running on our build agents and we need to take a moment to consider if this has security implications. Actually, considering that we now isolate the tests with docker, that risk can probably be pushed down a bit regarding its importance. This leaves the maintenance question: if the other library is going to get retired once the MS library is up to speed, we should switch to the shiny new horse before we write code that needs migrating (provided the new horse is already shiny).

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: Whoops, forgot this occurrence after adding plaintextstring support

Remotion/Web/Development.Analyzers/AnalyzerReleases.Unshipped.md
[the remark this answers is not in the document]
MichaelKetting: looks good to me (moving it to shipped). Once a rule is done, it's done and that should be at the end of the feature branch anyway. so, for the most part, shipped will be the correct choice from the get go. If we ever have a rule that's done but kind if iffy, we can still keep it in unshipped.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: it would, it would require the IsExternalInit class to be defined though. But I guess that would be alright as its just the following:
```csharp
namespace System.Runtime.CompilerServices
{
      internal static class IsExternalInit {}
}
```

MichaelKetting: Is this because we reference netstandard2.0 isntead of net5.0?

Would switching to net5.0 be a good idea?

And I don't mind the empty internal declaration, especially if it makes the code better readable.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: `GetTypeByMetadataName` returns `INamedTypeSymbol?`, `null` if the type cannot be found. I did not add nullchecks for `object` and `string`, as I expect them to be available. That's why there's the nullforgiving operator here (the logic of `GetStringJoinMethodSymbols` requires the passed symbol to not be `null`). As for the following questions on why some parameters are nullable, `SymbolEqualityComparer.Default.Equals` accepts nullable symbols, that's why the `object` symbol passed in might aswell be `null`` - even if it won't be. It just was a decision between `?` and `!` I made. But I don't really have a strong opinion here.

MichaelKetting: okay, thanks (meaning we have good reasons for the inconsistency). (things you can't see in github)

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: I had some issues running the tests with net5.0 but I'll look into it.

Remotion/Web/Development.Analyzers/Web.Development.Analyzers.csproj
iam-mholle: > This type if include means we have the assembly show up in the visual studio tree,

That's what `Visible="false"` is for.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: I thought about adding comments but it seemed clear enough - but its not. I'll look into it.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: It would be alright, I just didn't want to iterate twice over the methods if I don't have to.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
MichaelKetting: Thanks. And it's more of a "is this actually the correct way to do this" question.

The big problem with netstandard2.0 is that microsoft has started to deprecate the concent. modern c# feautres aren't officially supported. NRT isn't officially supported (or annoated), etc. This means I try to avoid netstandard2.0 when I can use net 5 or net6 instead, hoping to make things easier moving forward.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
MichaelKetting: Am confused: why do you iterate twice with Any and only once with FirstOrDefault?

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: I agree, but iirc, analyzers should target netstandard2.0 - might be deprecated now tho.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: How would I assign the `out` parameter only using `Any`? The use-case for this method is to provide a way of checking and extracting the matching element inside of an if condition

Remotion/Web/Development.Analyzers/Web.Development.Analyzers.csproj
MichaelKetting: ah, I.C. Didn't find notice that one. Verify interesting. It may offend my puristic views but I can certainly see why this makes things easier to handle so it's probably best that I start liking :D Where did you learn about `Visible`?

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
MichaelKetting: What ever's the correct practice :)

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
MichaelKetting: Ah, there's an out parameter. Hmm...would it be cleaner to return a nullable value instead of the out parameter and then capture the result in a boolean expression in the if. 
`if (var result = Any(...) != null)`

Am not sure if that's valid C#9, though, or just wishful thinking.

Another consideration would be to find a name that works with the Try... naming convention associated with boolean + out. Or we result a value tuple of (boolean, method), but that seems strange when we could just return method. => things I do to avoid out-parameters. Thoughts?

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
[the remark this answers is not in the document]
iam-mholle: The microsoft assembly does not provide the arrow indicator support. but rather requires a span to be specified, which is better anyway imo as it also specifies the end.

A test would look like this:
```csharp
    [Test]
    public async Task StringBuilder_WithWebString_Diagnostic ()
    {
      const string input = @"
using System.Text;
using Remotion.Web;
public class A
{
  public void Test()
  {
    var sb = new StringBuilder();
    sb.Append(WebString.CreateFromText(""test""));
  }
}";

      var diagnostic = Verify.Diagnostic()
          .WithSpan(9, 5, 9, 48)
          .WithMessage ("'System.Text.StringBuilder.Append(object)' should not be used with a 'Remotion.Web.WebString' argument. Call '.ToString(WebStringEncoding.HtmlWithTransformedLineBreaks)' on the Remotion.Web.WebString instance instead.");
      await Verify.VerifyAnalyzerAsync (input, diagnostic);
    }
```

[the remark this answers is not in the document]
iam-mholle: Additionally, net5.0 works out of the box. Might be worth to switch then.

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: I think declarations are statements but assignments expressions. So it should work when the variable is already declared - which is 😒. I'll settle for some Try variant naming.

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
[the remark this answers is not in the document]
MichaelKetting: okay, then I'd say we go for it. Do it now or a separate clenaup-task done as soon as possible?

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
MichaelKetting: Okay then. 
(why the sad-face for "variable already declared"?)

And we can always brainstorm the name if you want. The Try-name is just the typical go to based on .net framework patterns.

Remotion/Web/Development.Analyzers.Tests/Web.Development.Analyzers.Tests.csproj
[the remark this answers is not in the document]
iam-mholle: I'd just add it now as it fixen the missing class for `Init` prop support/records and enables net5.0

Remotion/Web/Development.Analyzers/Web.Development.Analyzers.csproj
iam-mholle: Can't remember but it's handy

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
iam-mholle: > why the sad-face for "variable already declared"?

I don't want to declare a variable in the method only assigned and used in one branch

Remotion/Web/Development.Analyzers.Tests/RMWEB0001_WebStringAnalyzerTests.cs
iam-mholle: Knowing how the analyzer is implemented, no. but it might be a good idea to ensure future refactorings produce the expected result.

Remotion/Web/Development.Analyzers.IntegrationTests/CSharpAnalyzerVerifier.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading.Tasks;
+using System.Web.UI;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Testing;
+using Microsoft.CodeAnalysis.CSharp.Testing.NUnit;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+
+namespace Remotion.Web.Development.Analyzers.Tests
```
MichaelKetting, line 27: Namespace-update please.

Remotion/Web/Development.Analyzers.IntegrationTests/RMWEB0001_WebStringAnalyzerTests.cs
```diff
@@ -0,0 +1,246 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Verify = Remotion.Web.Development.Analyzers.Tests.CSharpAnalyzerVerifier<Web.Development.Analyzers.RMWEB0001_WebStringAnalyzer>;
```
MichaelKetting, line 20: `Verifier` not `Verify` (and I'm happy if there's a better naming option). To me, `Verify.VerifyAnalyzerAsync` looks strange and it doesn't for the regular naming convention of using nouns for types. I think only fluent APIs and Assert get this exemption from the rule.

Remotion/Web/Development.Analyzers.IntegrationTests/Web.Development.Analyzers.IntegrationTests.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <ProjectType>Test</ProjectType>
+    <UseWebForms>true</UseWebForms>
+  </PropertyGroup>
+
+  <Import Project="..\..\..\Build\Shared.build.props" />
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0-windows</TargetFrameworks>
+    <AssemblyTitle>Remotion Web Analyzer Tests</AssemblyTitle>
+    <AssemblyName>Remotion.Web.Development.Analyzers.Tests</AssemblyName>
```
MichaelKetting, line 13:
```suggestion
    <AssemblyName>Remotion.Web.Development.Analyzers.IntegrationTests</AssemblyName>
```

Remotion/Web/Development.Analyzers.IntegrationTests/Web.Development.Analyzers.IntegrationTests.csproj
```diff
@@ -0,0 +1,29 @@
...
+  <PropertyGroup>
+    <ProjectType>Test</ProjectType>
+    <UseWebForms>true</UseWebForms>
+  </PropertyGroup>
+
+  <Import Project="..\..\..\Build\Shared.build.props" />
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0-windows</TargetFrameworks>
+    <AssemblyTitle>Remotion Web Analyzer Tests</AssemblyTitle>
```
MichaelKetting, line 12:
```suggestion
    <AssemblyTitle>Remotion Web Analyzer Integration Tests</AssemblyTitle>
```

Remotion/Web/Development.Analyzers/RMWEB0001_WebStringAnalyzer.cs
```diff
@@ -0,0 +1,208 @@
...
+        return;
+
+      var node = (InvocationExpressionSyntax) ctx.Node;
+      var invocationOriginalDefinitionSymbol = invocationSymbolInfo.Symbol.OriginalDefinition;
+
+      // Check if the invoked method is StringBuilder.Append(object?)
+      if (SymbolEqualityComparer.Default.Equals (invocationOriginalDefinitionSymbol, symbols.StringBuilderAppendMethodSymbol))
+      {
+        var argumentType = GetFirstArgumentTypeSymbol (node, ctx.SemanticModel);
+        // Check if the first argument is WebString or PlainTextString
```
MichaelKetting, line 109: like the comments

Remotion/Web/Development.Analyzers/Web.Development.Analyzers.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
```
MichaelKetting, line 4: wouldn't net48;net5.0 be 'correct' here? After all, we don't have a stand-alone netstandard2.0 target anywhere. But I'm not sure how the targetframework of analyzers relates to the rest of the stack.

(see also question in WebLibrary.props)

Build/ProjectTypes/WebLibrary.props
```diff
@@ -21,4 +21,12 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="$(SolutionDir)Remotion\Web\Development.Analyzers\Web.Development.Analyzers.csproj"
+                      PrivateAssets="all"
+                      ReferenceOutputAssembly="false"
+                      SetTargetFramework="TargetFramework=netstandard2.0"
```
MichaelKetting, line 28: Is this done for the common denomitor because we have to use separate entries for ProjectReference otherwise?

Remotion/Web/Development.Analyzers.IntegrationTests/Web.Development.Analyzers.IntegrationTests.csproj
```diff
@@ -0,0 +1,29 @@
...
+
+  <PropertyGroup>
+    <ProjectType>Test</ProjectType>
+    <UseWebForms>true</UseWebForms>
+  </PropertyGroup>
+
+  <Import Project="..\..\..\Build\Shared.build.props" />
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0-windows</TargetFrameworks>
```
MichaelKetting, line 11: Just a head's up: right now, we don't run the net5.0 tests on the automated build. Build script must first learn to deal with dotnet test.

MichaelKetting (changes requested):
Some naming stuff.

Build/ProjectTypes/WebLibrary.props
iam-mholle: Yes

MichaelKetting: gut feeling tells me, this isn't perfect. We're not excercising the code during the test that we're using in "production". Since we can't test net48 variant anyway and we also want NRT, there's probably no way around this. Or can we set net5.0 here, too, and have the analyzer run for both net48 and net5.0 compilations?

iam-mholle: Using an analyzer targeting net5.0 did not work for me. An exception was thrown when creating the analyzer instance.

PR #273, opened 2021-11-29
[WIP] RM-7845 Replace Rhino.Mocks with Moq in Data

derlistigelurch:


Remotion/Data/DomainObjects.Security.UnitTests/SecurityClientTransactionExtensionTests/TestHelper.cs
```diff
@@ -144,85 +130,161 @@ public void AddExtension (IClientTransactionExtension extension)
...
+              {
+                CheckTransaction();
+                CheckTransaction();
+              })
+          .Returns(returnValue)
+          .Verifiable();
     }
 
-    public void ExpectFunctionalSecurityStrategyHasAccess (Type securableObjectType, Enum accessTypeEnum, HasStatelessAccessDelegate doDelegate)
+    public void ExpectObjectSecurityStrategyHasAccess (MockSequence sequence, SecurableObject securableObject, Enum accessTypeEnum, bool returnValue)
```
MichaelKetting, line 161: - [ ] Rework to pass Mock<SecurableObject> instead of SecurableObject
- [ ] Check if we can use a single method for both Sequence and no-Sequence.

MichaelKetting (changes requested):

PR #274, opened 2021-11-30
RM-8137: Adapt ResourceDispatcher

iam-mholle:


Remotion/Web/Core/UI/Controls/ControlItemCollection.cs
```diff
@@ -279,13 +280,13 @@ public void Dispatch (IDictionary values, Control? parent, string collectionName
...
-      foreach (DictionaryEntry entry in values)
+      foreach (var entry in values)
       {
-        string id = (string) entry.Key;
+        string id = entry.Key;
 
         IControlItem? item = Find (id);
         if (item != null)
-          ResourceDispatcher.DispatchGeneric (item, (IDictionary) entry.Value!); // TODO: not null assertion
+          ResourceDispatcher.DispatchGeneric (item, entry.Value); // TODO: not null assertion
```
MichaelKetting, line 289:
```suggestion
          ResourceDispatcher.DispatchGeneric (item, entry.Value);
```

I think NRT and the WebString value type take care of the null-assertion.

Remotion/Web/Core/UI/Controls/ControlItemCollection.cs
```diff
@@ -269,7 +270,7 @@ private bool IsSupportedType (IControlItem controlItem)
       }
     }
 
-    public void Dispatch (IDictionary values, Control? parent, string collectionName)
+    public void Dispatch (IDictionary<string, IDictionary<string, WebString>> values, Control? parent, string collectionName)
```
MichaelKetting, line 273: Disclaimer: his is just an observation based on the current changeset, since actually changing the code would result in probably more work than is justified (unless you actually want to do this change):

It would be cool if we had these typed as IReadOnlyDictionary<string, IReadOnlyDictionary<string, WebString>> since there is no write access. But we would need to rewrite (or cast) at some points in the code, which, combined with the low priority if this feature doesn't really make this more than a pet-peeve.

Also, the aggressive use of IReadOnlyCollection/IReadOnlyDictionary during specificiation of a task is something that requires more Pavlovian conditioning...

Remotion/Web/Core/UI/Globalization/ResourceDispatcher.cs
```diff
@@ -132,28 +132,38 @@ public static void Dispatch (Control control, IDictionary elements, string resou
...
+        stringProperty.SetValue (obj, propertyValue.GetValue(), Array.Empty<object>());
+      }
+      else if (webStringProperty != null)
+      {
+        webStringProperty.SetValue (obj, propertyValue, Array.Empty<object>());
+      }
+      else if (plainTextStringProperty != null)
+      {
+        plainTextStringProperty.SetValue (obj, propertyValue.ToPlainTextString(), Array.Empty<object>());
       }
```
MichaelKetting, line 159: Performance consideration: I think it would be best to either use
```
if (obj.GetType ().GetProperty (propertyName, typeof (WebString)) is { } webStringProperty)
{
}
if (obj.GetType ().GetProperty (propertyName, typeof (PlainTexttring)) is { } painTextStringProperty)
{
}
if (obj.GetType ().GetProperty (propertyName, typeof (string)) is { } stringProperty)
{
}
```
(with WebString tested first to optimize the most-likely case)

or, actually better
```
PropertyInfo? property = obj.GetType ().GetProperty (propertyName)
```
and then use IF  (or pattern matching) on the return type for string/webstring/plaintextstring. This way, we only ever check once for the property and get the best possible performance.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #277, opened 2021-12-01
RM-8216: JS error in DropDownMenu.ts EndOpenPopUp

iam-mholle:


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -407,7 +407,12 @@ class DropDownMenu
     ul.addEventListener ('mouseover', function (event)
     {
       var eventTarget = DropDownMenu.GetTarget (event, "LI");
+
       ul.querySelectorAll ("li").forEach (b => b.classList.remove (DropDownMenu._itemSelectedClassName));
+
+      if (eventTarget == null)
```
MichaelKetting, line 413: This is interesting. Doesn't this mean we have to make DropDownMenu.GetTarget return a nullable string, too?

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -407,7 +407,12 @@ class DropDownMenu
...
+        return;
+
       if (eventTarget.firstChild != null && eventTarget.firstChild.nodeName.toLowerCase() === 'a')
```
MichaelKetting, line 416: Also, we remove the highlighting, but the focus-entry remains. please check, I think right now the keyboard will continue at the point where the last hover has been even if the item is no longer highlighted since we don't reset the focus.  I didn't try it with the updated Typescript, the "intended" behavior is be that "no selection == cursor down will select the first item and cursur up the last item") -> please test

iam-mholle: This actually was both a css and js issue. Have a look at [`d071873` (#277)](https://github.com/re-motion/Framework/pull/277/commits/d071873c0c886388c87d3ec3de883ccaad952a69) for the fix.

MichaelKetting (approved):

PR #278, opened 2021-12-01
RM-8238 Restore the old CallContextStorageProvider

a-ctor:
https://re-motion.atlassian.net/browse/RM-8238

### DIFF
--- a/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
+++ b/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
@@ -24,7 +24,10 @@ using Remotion.Logging;
 
 namespace Remotion.Context
 {
+#if !NETFRAMEWORK
   [ImplementationFor(typeof(ISafeContextStorageProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
+#endif
+  [Obsolete("This type is experimental and might have unintended side effects. (Version 3.0.0-alpha.22)")]
   public class AsyncLocalStorageProvider : ISafeContextStorageProvider
   {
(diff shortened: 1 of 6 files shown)


MichaelKetting (changes requested):

Remotion/Core/Core/Context/CallContextStorageProvider.obsolete.cs
```diff
@@ -15,28 +15,52 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
```
MichaelKetting, line 15: Forgot: please update the filename (drop `.obsolete`). I only use the `.obsolete` suffix if the file contains only dead APIs without logic, not code that's still in use, even if it's obsolete. It's an indicator that the file can eventually be deleted.

- [ ] Rename

Remotion/Core/Core/Context/AsyncLocalStorageProvider.obsolete.cs
```diff
@@ -24,7 +24,10 @@
 
 namespace Remotion.Context
 {
+#if !NETFRAMEWORK
   [ImplementationFor(typeof(ISafeContextStorageProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
+#endif
+  [Obsolete("This type is experimental and might have unintended side effects. (Version 3.0.0-alpha.22)")]
```
MichaelKetting, line 30: - [ ] Undo rename

MichaelKetting (approved):
Please fix the renames (missed that one) and then we're done.

PR #286, opened 2021-12-09
RM-8279: Extend RMWEB0001 to detect String.Format(...) usage errors

iam-mholle:


SecurityManager/Clients.Web/Classes/SecurableClassDefinitionTreeView.cs
```diff
@@ -62,11 +62,11 @@ protected override WebString GetText (IBusinessObjectWithIdentity businessObject
 
       var resourceManager = GetResourceManager(typeof(ResourceIdentifier));
       if (aclCount == 0)
-        return WebString.CreateFromText(string.Format(resourceManager.GetString(ResourceIdentifier.NoAclsText), text));
+        return WebString.CreateFromText(string.Format(resourceManager.GetString(ResourceIdentifier.NoAclsText), text.ToString()));
```
MichaelKetting, line 65: for re-motion code, I try to stick with the overload and pass html or htmlwithlinebreaks to make things explicit without needing to know the details of the parameterless overload.

Although, here, in particular, the intended way to solve this is actually to get the unencoded value because the callsite can handle the linebreak descision and we're just re-wrapping a string:
 WebString.CreateFromText(string.Format(resourceManager.GetString(ResourceIdentifier.NoAclsText), text.GetValue()))
```suggestion
        return WebString.CreateFromText(string.Format(resourceManager.GetString(ResourceIdentifier.NoAclsText), text.GetValue()));;
```

Side note: in the future, this will be changed to using a badge. I'll make a ticket for this.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #287, opened 2021-12-10
RM-8248: Introduce a result object to IBocAutoCompleteReferenceValueWebService.Search

iam-mholle:


Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
```diff
@@ -392,6 +391,24 @@ private IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResponse ([NotN
       }
     }
 
+    private static IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResultVariants (IDictionary<string, object> data) =>
```
MichaelKetting, line 394: Since there's an update anyway: IReadOnlyDictionary would be great instead of the IDictionary casts.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
```diff
@@ -373,11 +376,7 @@ private IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResponse ([NotN
       switch (state)
       {
         case AutoCompleteSearchService.Success:
-          var successData = (IReadOnlyCollection<object>)data;
-          return successData.Cast<IDictionary<string, object>>()
-              .Where(d => d != null) // empty JSON object (= no result) is converted to null
-              .Select(d => new SearchServiceResultItem((string)d["UniqueIdentifier"], (string)d["DisplayName"], (string)d["IconUrl"]))
-              .ToList();
+          return successParser((IReadOnlyCollection<object>)data);
```
MichaelKetting, line 379: Wouldn't this always be successParser(((IReadOnlyCollection<IDictionary<string, object>>)data).Single()) for the forseeable future? 

Is there an expected/known error scenario where the result is 200 but empty so .Single() would fail or can we expect this to never actually happen. Reasoning: Single() is a very unhelpful API in case the expectation is violated. You can (hopefully) figure things out with the stacktrace but a meaningful msg is usually preferred.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
```diff
@@ -373,11 +376,7 @@ private IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResponse ([NotN
       switch (state)
       {
         case AutoCompleteSearchService.Success:
-          var successData = (IReadOnlyCollection<object>)data;
-          return successData.Cast<IDictionary<string, object>>()
-              .Where(d => d != null) // empty JSON object (= no result) is converted to null
```
MichaelKetting, line 378: Null-items get filtered out in the result. The new code doesn't do that because we can expect this to never happen?

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
```diff
@@ -57,17 +58,17 @@ private class SelectAutoCompleteAction : WebTestAction
 
       private string _actionName = "BocAutoCompleteReferenceValueControlObject_SelectAutoCompleteAction";
 
-      [NotNull]
+      [JetBrains.Annotations.NotNull]
```
MichaelKetting, line 61: This feels noisy. would a 
`using NotNullAttribute = Jetbrains.Annotations.NutNullAttribute`
be better in terms of readability / noise?

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
iam-mholle: They do get filtered out in `ParseSearchServiceResultItems`

MichaelKetting: I may be blind, so...how? (we can also check  this live)

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
MichaelKetting: Gut feeling says it would be a good idea. (especially since have some other details anyway in this PR).

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
iam-mholle: `OfType<>` filters `null` items.

MichaelKetting: Ah, you switched from Cast to OfType.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
```diff
@@ -392,6 +410,24 @@ private IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResponse ([NotN
...
 
+    private static IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResultVariants (IReadOnlyDictionary<string, object> data) =>
+        (data["Type"] as string) switch
+        {
+            "ValueList" => ParseSearchServiceResultItems(((IReadOnlyCollection<object?>)data["Values"]).Cast<IReadOnlyDictionary<string, object>>()),
+            var unknown => throw new InvalidOperationException($"The result variant '{unknown}' is not supported."),
+        };
+
+    private static IReadOnlyList<SearchServiceResultItem> ParseSearchServiceResultItems (IEnumerable<IReadOnlyDictionary<string, object>?> data) =>
+        data.Select(ParseSearchServiceResultItem).OfType<SearchServiceResultItem>().ToArray();
```
MichaelKetting, line 421: I don't think I want to change this because the current code is much shorter than making the nullcheck + typing to a not-nullable type. Even so, it does feel strange to use a type check as a null filter. My reasoning is that as a reader I don't know from reading this particular line that we're not missing type errors via the soft-cast. E.g. if the `Parse...Item` method is re-typed to retunr a base type and then introduces a different type next to the `SearchServiceResultItem`, we'd not find this until you notice that items are missing. 

So, why not change this? basically, the naming makes things reasonable noticable and it's unlikley that this particular change would actually pop up in the future. So, right now, right here, that's probably a good thing. In the future, I think a dedidacted filter-to-not-null and convert-to-not-null extension method would probably be a good idea. thoughts?

Remotion/ObjectBinding/Web/Services/BocAutoCompleteReferenceValueSearchResult.cs
```diff
@@ -0,0 +1,39 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+namespace Remotion.ObjectBinding.Web.Services
+{
+  /// <summary>
+  /// The base search result of a <see cref="IBocAutoCompleteReferenceValueWebService.Search"/> call.
```
MichaelKetting, line 20:
```suggestion
  /// The search result of a <see cref="IBocAutoCompleteReferenceValueWebService"/>.<see cref="IBocAutoCompleteReferenceValueWebService.Search"/> call.
```

XmlDoc is funny that way. Unless you are fine with just the method name in the docs, you need to go the extra mile. (usually needed for referencing members on other types to make things readable).

And since it gets touched anyway, the 'base' qualifier isn't needed here.

Remotion/ObjectBinding/Web/Services/BocAutoCompleteReferenceValueSearchResult.cs
```diff
@@ -0,0 +1,39 @@
...
+  /// </summary>
+  public abstract class BocAutoCompleteReferenceValueSearchResult
+  {
+    /// <summary>
+    /// Creates a new <see cref="BocAutoCompleteReferenceValueSearchResultWithValueList"/> instance, containing the given values.
```
MichaelKetting, line 25:
```suggestion
    /// Creates a new <see cref="BocAutoCompleteReferenceValueSearchResult"/> based on the <paramref name="values"/>.
```

Since we're hiding the specific return type anyway, there's no need to put it into the summary. Using the paramref covers doccing the parameter for "simple" doc scenarios and adds a nice touch.

Also, after seeing this in the docs, the paramname feels better as `values` instead of `proxies`.

Remotion/ObjectBinding/Web/Services/BocAutoCompleteReferenceValueSearchResult.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    public static BocAutoCompleteReferenceValueSearchResult CreateForValueList (BusinessObjectWithIdentityProxy[] proxies) =>
+        new BocAutoCompleteReferenceValueSearchResultWithValueList(proxies);
```
MichaelKetting, line 28:
```suggestion
    public static BocAutoCompleteReferenceValueSearchResult CreateForValueList (BusinessObjectWithIdentityProxy[] values) =>
        new BocAutoCompleteReferenceValueSearchResultWithValueList(values);
```

Remotion/ObjectBinding/Web/Services/BocAutoCompleteReferenceValueSearchResult.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    /// <summary>
+    /// The discriminator for different <see cref="BocAutoCompleteReferenceValueSearchResult"/> variants.
```
MichaelKetting, line 31:
```suggestion
    /// The discriminator for different <see cref="BocAutoCompleteReferenceValueSearchResult"/> variants, used during Javascript deserialization.
```

MichaelKetting (approved):
Some doc comment improvents. Please do a compile check before merging.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/BocAutoCompleteReferenceValueControlObject.cs
iam-mholle: I guess just using `.Where(i => i != null).Cast<>()` would be enough if more clarity is wanted. Still, `OfType` is a known method with known behavior imo. I'm not sure more clarity is needed here.

MichaelKetting: Strictly speaking, `TypeOf<>` isn't a null-filter from the APIs primairy usecase. Instead, filtering null is a sideeffct next to the type filter. In particular, it tells us that we will ignore items with a not-matching type which in turn can mask bugs intrudced by changes to the type hierarchy. 

For future stuff: would you prefer `.Where(i => i != null).Cast<>()` or `.WhereNotNull()`?

iam-mholle: The same applies for the `is` operator then, which means `s is string` should be expanded to `s != null && s is string` right?

> For future stuff: would you prefer .Where(i => i != null).Cast<>() or .WhereNotNull()?

I would prefer Linq methods only in this case.

MichaelKetting: > The same applies for the `is` operator then, which means `s is string` should be expanded to `s != null && s is string` right?

Nope. There's two sceanarios:
1) ```object s;
if (s is string)```
and
1) ```object s;
if (s as string x) { return x.Length; }```

is sufficient because we need a typetest and it's okay to unify nul land the type test, even more so than before given nullable reference types.

2) ```string? s;
if (string != null) { return s.Length; }```
here, we know the type and just make this not-null.

The TypeOf scenario is only "needed" (or looking good) because there is no short way of making a null-filter and and re-typing the item from nullable to not-nullable. So, I just don't like to use a soft-cast as a null-test only instead of also a type-filter.

iam-mholle: Still, how is `if (s is string x)` different here, it also does not show the null check explicitly, thus should be expanded to `if (s is string x && x != null)`. That's just a matter of principle - should language features & standard API's with all their implication be required to be known or not.

MichaelKetting: The difference is that in `if (s is string x)` the primary intention is not to do a null check, that's just an implied sideeffect. If I want to do a nullcheck here, I write a nullcheck and if I want a soft cast, I do a soft cast and it's okay and readable that the softcast also does a null check. However, if I don't need the cast but just the nullcheck, I would not write a softcast because it changes the implied semantics of the code. 

TL;DR: if I want a cast (and optional null check), I should write a cast. If I want a nullcheck, I should write a nullcheck.

PR #304, opened 2022-01-10
RM-3166 Add History.Back handling using session state

iam-mholle:
- [x] Update resx, see https://github.com/re-motion/Framework/pull/304#issuecomment-1012102734

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/SmartPage.ts
+++ b/Remotion/Web/ClientScript/Scripts/SmartPage.ts
@@ -51,6 +51,12 @@ type SmartPage_EventMap = {
 };
 type SmartPage_Event = keyof SmartPage_EventMap;
 
+enum SmartPage_CacheDetection
+{
+  Undefined,
+  HasSubmitted,
+  HasLoaded,
+}
 
 // The context contains all information required by the smart page.
 class SmartPage_Context
@@ -101,15 +107,14 @@ class SmartPage_Context
   private _smartScrollingFieldID: Nullable<string> = null;
   // The hidden field containing the smart focusing data.
   private _smartFocusFieldID: Nullable<string> = null;
+  // The hidden field containing the page token guid.
+  private _smartPageTokenFieldID: string;
 
   private _activeElement = null;
   // The hashtable of eventhandlers: Hashtable < event-key, Array < event-handler > >
   private _eventHandlers: string[][] = [];
   private _trackedIDs: string[] = [];
   private _synchronousPostBackCommands: string[] = [];
-
-  private _cacheStateHasSubmitted = 'hasSubmitted';
-  private _cacheStateHasLoaded = 'hasLoaded';
 
   private _loadHandler = function () { SmartPage_Context.Instance!.OnLoad(); };
   private _beforeUnloadHandler = function () { return SmartPage_Context.Instance!.OnBeforeUnload(); };
@@ -139,7 +144,7 @@ class SmartPage_Context
       isDirtyStateTrackingEnabled: boolean,
       abortMessage: Nullable<string>, statusIsSubmittingMessage: Nullable<string>,
       smartScrollingFieldID: Nullable<string>, smartFocusFieldID: Nullable<string>,
-      checkFormStateFunctionName: Nullable<string>)
+      smartPageTokenFieldID: string, checkFormStateFunctionName: Nullable<string>)
   {
     ArgumentUtility.CheckNotNullAndTypeIsString('theFormID', theFormID);
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('isDirtyStateTrackingEnabled', isDirtyStateTrackingEnabled);
@@ -147,6 +152,7 @@ class SmartPage_Context
     ArgumentUtility.CheckTypeIsString('statusIsSubmittingMessage', statusIsSubmittingMessage);
     ArgumentUtility.CheckTypeIsString('smartScrollingFieldID', smartScrollingFieldID);
     ArgumentUtility.CheckTypeIsString('smartFocusFieldID', smartFocusFieldID);
+    ArgumentUtility.CheckNotNullAndTypeIsString('smartPageTokenFieldID', smartPageTokenFieldID);
     ArgumentUtility.CheckTypeIsString('checkFormStateFunctionName', checkFormStateFunctionName);
 
     this._isDirtyStateTrackingEnabled = isDirtyStateTrackingEnabled;
@@ -172,6 +178,7 @@ class SmartPage_Context
 
     this._smartScrollingFieldID = smartScrollingFieldID;
     this._smartFocusFieldID = smartFocusFieldID;
+    this._smartPageTokenFieldID = smartPageTokenFieldID
 
     this.AttachPageLevelEventHandlers();
   }
@@ -461,16 +468,24 @@ class SmartPage_Context
     }
   };
 
+  private GetPageToken(): string
+  {
+    return this.GetFormElement(this._smartPageTokenFieldID).value;
+  }
+
   // Determines whether the page was loaded from cache.
   public CheckIfCached(): void
   {
-    var field = this._theForm.SmartPage_CacheDetectionField;
-    if (field.value == this._cacheStateHasSubmitted)
+    const pageToken = this.GetPageToken();
+    const pageTokenValue = window.sessionStorage.getItem(pageToken);
+    const cacheDetection = SmartPage_CacheDetection[pageTokenValue as keyof typeof SmartPage_CacheDetection];
+
+    if (cacheDetection === SmartPage_CacheDetection.HasSubmitted)
     {
       this._hasSubmitted = true;
       this._isCached = true;
     }
-    else if (field.value == this._cacheStateHasLoaded)
+    else if (cacheDetection === SmartPage_CacheDetection.HasLoaded)
     {
       this._isCached = true;
     }
@@ -483,15 +498,13 @@ class SmartPage_Context
   // Marks the page as loaded.
   public SetCacheDetectionFieldLoaded(): void
   {
-    var field = this._theForm.SmartPage_CacheDetectionField;
-    field.value = this._cacheStateHasLoaded;
+    window.sessionStorage.setItem(this.GetPageToken(), SmartPage_CacheDetection[SmartPage_CacheDetection.HasLoaded]);
   };
 
   // Marks the page as submitted.
   public SetCacheDetectionFieldSubmitted(): void
   {
-    var field = this._theForm.SmartPage_CacheDetectionField;
-    field.value = this._cacheStateHasSubmitted;
+    window.sessionStorage.setItem(this.GetPageToken(), SmartPage_CacheDetection[SmartPage_CacheDetection.HasSubmitted]);
   };
 
   // Event handler for window.OnBeforeUnload.
--- a/Remotion/Web/Core/Globalization/WxePageInfo.fr.resx
+++ b/Remotion/Web/Core/Globalization/WxePageInfo.fr.resx
@@ -121,6 +121,6 @@
     <value>&lt;p&gt;Vous êtes sur le point de quitter la page.&lt;/p&gt;&lt;p&gt;Vous ne pouvez plus envoyer aucune saisie au serveur.&lt;/p&gt;</value>
   </data>
   <data name="Remotion.Web.ExecutionEngine.WxePageInfo.StatusIsCachedMessage" xml:space="preserve">
-    <value>Impossible de naviguer avec les boutons Suivant et Précédent.</value>
+    <value>Ne dupliquez pas ou ne rechargez pas les onglets et n'utilisez pas les boutons «Suivant» et «Précédent».</value>
   </data>
 </root>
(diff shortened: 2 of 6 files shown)


Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -525,6 +525,11 @@ protected override void OnInit (EventArgs e)
   {
     base.OnInit(e);
     RegisterRequiresControlState(this);
+
```
MichaelKetting, line 528: This implementation belongs in SmartPageInfo.  SmartPage doesn't actually have any logic itself. Originally, the idea was that anyone can implement the ISmartPage interface on a custom base-Page by delegating to the SmartPageInfo type.

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -525,6 +525,11 @@ protected override void OnInit (EventArgs e)
...
+    if (!IsPostBack)
+      ClientScript.RegisterHiddenField(this, "SmartPage_InstanceToken", Guid.NewGuid().ToString());
+    else
+      ClientScript.RegisterHiddenField(this, "SmartPage_InstanceToken", Request.Params["SmartPage_InstanceToken"]);
```
MichaelKetting, line 532: For consistent implementation pattern:
See how SmartPageInfo handles reading the c_smartScrollingID during prerender.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -391,6 +391,12 @@ class SmartPage_Context
 
     var isAsynchronous = false;
     this.PageLoaded(isAsynchronous);
+
+    let data = {
+      ...history.state,
+      pageToken: document.getElementById('SmartPage_InstanceToken')!.getAttribute('value') ?? '',
```
MichaelKetting, line 397: see SmartPage.ts, declaration/setting of `_smartFocusFieldID`. These values aren't hardcoded to prevent code (well, value) duplication.

MichaelKetting, line 397: see SmartPage.ts, there's a helper API `this.GetFormElement(id).value` we should use.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -464,13 +470,15 @@ class SmartPage_Context
   // Determines whether the page was loaded from cache.
   public CheckIfCached(): void
   {
-    var field = this._theForm.SmartPage_CacheDetectionField;
-    if (field.value == this._cacheStateHasSubmitted)
+    const pageToken = document.getElementById('SmartPage_InstanceToken')!.getAttribute('value');
+    const hasDifferentPageToken = history.state?.pageToken && history.state.pageToken !== pageToken;
+
+    if (history.state?.cacheDetection === this._cacheStateHasSubmitted)
```
MichaelKetting, line 476: Since I just spent some time re-familiarzing myself with the sematnics of `_cacheStateHasSubmitted` and `_cacheStateHasLoaded` (before figuring out those are just string constants): should this be updated to use a typed approach with string values defined via discrimininated union / enum? Don't know about the readability impact of this, I'm open for suggestions :)

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -464,13 +470,15 @@ class SmartPage_Context
   // Determines whether the page was loaded from cache.
   public CheckIfCached(): void
   {
-    var field = this._theForm.SmartPage_CacheDetectionField;
-    if (field.value == this._cacheStateHasSubmitted)
+    const pageToken = document.getElementById('SmartPage_InstanceToken')!.getAttribute('value');
+    const hasDifferentPageToken = history.state?.pageToken && history.state.pageToken !== pageToken;
```
MichaelKetting, line 474: should we add a comment for the need/semantics of the pagetoken? I think this is a case where the need is only discoverable via trial and error / reverse engineering via debugger. thoughts?

MichaelKetting (changes requested):
Some tweaks requested.

Remotion/Web/Core/UI/SmartPageImplementation/SmartPageInfo.cs
```diff
@@ -54,6 +54,7 @@ public enum ResourceIdentifier
     }
 
     public static readonly string CacheDetectionID = "SmartPage_CacheDetectionField";
+    private const string c_smartPageTokenID = "SmartPage_InstanceToken";
```
MichaelKetting, line 57:
```suggestion
    private const string c_smartPageTokenID = "SmartPage_TokenField";
```

`...Field` for consistency reason. 
`Instance` or not, that's just a preference thing and since we don't use `Instance` in the names, it feels cleaner this way (no need to think about why the variable name and the value differ) but either one is fine.

MichaelKetting (changes requested):
A few consistency bits.

Remotion/Web/Core/UI/SmartPageImplementation/SmartPageInfo.cs
```diff
@@ -302,6 +303,14 @@ private void Page_PreRenderComplete (object? sender, EventArgs eventArgs)
...
     private void PreRenderSmartPage ()
     {
+      var postBackCollection = _page.GetPostBackCollection();
+      var smartScrollingValue = postBackCollection switch
+      {
+          { } c => c[c_smartPageTokenID],
+          null => Guid.NewGuid().ToString(),
+      };
+
+      _page.ClientScript.RegisterHiddenField(_page, c_smartPageTokenID, smartScrollingValue);
```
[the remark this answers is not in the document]
iam-mholle: Whoops

MichaelKetting (changes requested):
Okay, tested this in the browser and found a problem:

Opening the same page again via a get-request triggers the "different pagetoken" scenario in Google Chrome and this lands the page in the cached=true case even though a get request was performed and the page is newly loaded. Maybe the condition in line 498 needs a hasSamePageToken-flag instead of hasDifferentPageToken-flag?

There's also the problem in Firefox when a page that wasn't previously submitted is loaded from the history: this page isn't recognized as cached and the WXE page can't properly react to this. 

These two issues might be different sides of the same coin (i.e. Chrome and Firefox behaving differently for the same underlying usecase). Either way, we need to improve this. 

I also realized that the message needs updating. I'll get us new strings.

iam-mholle:
> Okay, tested this in the browser and found a problem:
> 
> Opening the same page again via a get-request triggers the "different pagetoken" scenario in Google Chrome and this lands the page in the cached=true case even though a get request was performed and the page is newly loaded. Maybe the condition in line 498 needs a hasSamePageToken-flag instead of hasDifferentPageToken-flag?
> 
> There's also the problem in Firefox when a page that wasn't previously submitted is loaded from the history: this page isn't recognized as cached and the WXE page can't properly react to this.
> 
> These two issues might be different sides of the same coin (i.e. Chrome and Firefox behaving differently for the same underlying usecase). Either way, we need to improve this.
> 
> I also realized that the message needs updating. I'll get us new strings.

I introduced this bug while refactoring the code, I'll look into it.

MichaelKetting:
New strings for Remotion.Web.ExecutionEngine.WxePageInfo.StatusIsCachedMessage

* en: Do not duplicate or refresh tabs, or use the "Forward" and "Back" buttons.
* de: Duplizieren oder laden Sie Tabs nicht neu und verwenden Sie nicht die Schaltflächen "Vorwärts" und "Zurück".
* fr: Ne dupliquez pas ou ne rechargez pas les onglets et n'utilisez pas les boutons «Suivant» et «Précédent».
* it: Non duplicare o ricaricare le schede e non usare i pulsanti "Avanti" e "Indietro".

Remotion/Web/Core/Globalization/WxePageInfo.resx
```diff
@@ -121,6 +121,6 @@
     <value>&lt;p&gt;You are about to leave this page.&lt;/p&gt;&lt;p&gt;You will not be able to send any further input to the server.&lt;/p&gt;</value>
   </data>
   <data name="Remotion.Web.ExecutionEngine.WxePageInfo.StatusIsCachedMessage" xml:space="preserve">
-    <value>You cannot use the Forward and Back buttons.</value>
+      <value>Do not duplicate or refresh tabs, or use the "Forward" and "Back" buttons.</value>
```
MichaelKetting, line 124:
```suggestion
    <value>Do not duplicate or refresh tabs, or use the "Forward" and "Back" buttons.</value>
```

Remotion/Web/Core/Globalization/WxePageInfo.it.resx
```diff
@@ -121,6 +121,6 @@
     <value>&lt;p&gt;Si sta per lasciare questa pagina.&lt;/p&gt;&lt;p&gt;Non è possibile inviare altri input al server.&lt;/p&gt;</value>
   </data>
   <data name="Remotion.Web.ExecutionEngine.WxePageInfo.StatusIsCachedMessage" xml:space="preserve">
-    <value>Non è supportata la navigazione con i pulsanti Avanti e Indietro.</value>
+      <value>Non duplicare o ricaricare le schede e non usare i pulsanti "Avanti" e "Indietro".</value>
```
MichaelKetting, line 124:
```suggestion
    <value>Non duplicare o ricaricare le schede e non usare i pulsanti "Avanti" e "Indietro".</value>
```

Remotion/Web/Core/Globalization/WxePageInfo.fr.resx
```diff
@@ -121,6 +121,6 @@
     <value>&lt;p&gt;Vous êtes sur le point de quitter la page.&lt;/p&gt;&lt;p&gt;Vous ne pouvez plus envoyer aucune saisie au serveur.&lt;/p&gt;</value>
   </data>
   <data name="Remotion.Web.ExecutionEngine.WxePageInfo.StatusIsCachedMessage" xml:space="preserve">
-    <value>Impossible de naviguer avec les boutons Suivant et Précédent.</value>
+      <value>Ne dupliquez pas ou ne rechargez pas les onglets et n'utilisez pas les boutons «Suivant» et «Précédent».</value>
```
MichaelKetting, line 124:
```suggestion
    <value>Ne dupliquez pas ou ne rechargez pas les onglets et n'utilisez pas les boutons «Suivant» et «Précédent».</value>
```

Remotion/Web/Development.WebTesting.IntegrationTests/AnchorControlObjectTest.cs
```diff
@@ -139,10 +139,6 @@ public void TestClick ()
       var htmlAnchor = home.Anchors().GetByLocalID("MyHtmlAnchor");
       home = htmlAnchor.Click().Expect<WxePageObject>();
       Assert.That(home.Scope.FindId("TestOutputLabel").Text, Is.EqualTo("MyHtmlAnchor"));
-
-      var htmlAnchorWithJavaScriptLink = home.Anchors().GetByLocalID("MyHtmlAnchorWithJavaScriptLink");
```
MichaelKetting, line 143: Please try, I think we can solve this by appending `return false;` to the onclick in the APSX file. That was missing from the get-go and only now popped up as a problem.

` <a id="MyHtmlAnchorWithJavaScriptLink" href="#" onclick="javascript:myClickHandler('MyHtmlAnchorWithJavaScriptLink'); return false;" runat="server">MyHtmlAnchorWithJavaScriptLink</a>`

MichaelKetting (commented):
partial review: need to try/test/review the TS file before I'm done.

MichaelKetting (changes requested):
Sorry, found at least one problem:

opening the same page twice in Chrome and going back in history, keeps the "current" target in the url and does not detect the outdated page.

MichaelKetting:
Also, Firefox does not detect duplicate-tab. (works in Chrome)

Remotion/Web/Development.WebTesting.IntegrationTests/AnchorControlObjectTest.cs
MichaelKetting: Please re-check: right now, the a-tag has both a href and an onclick-handler. since the onclick-handler does not end with a `return false;` statement, clicking the anchor will first execute the onlick handler and then the href. As far as I understand the problem, executing the href with `#` will break the test. By appending `return false;` this behavior should change (unless I'm mistaken).

MichaelKetting:
Considering the problem we have with the history API, I believe we should try to use the browser's session storage (https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) instead. That way, we don't have to rely on the location target or history manipulation to achieve our requirements. Session storage will be evicted upon tab close and is cloned between tabs. Thus, we should be able to do everything we need, using the page session token as the storage key. According to the docs, the user should not be able to affect (or be notifed) when using session storage. We need to test this to make sure. If there's no new problems and sessionstorage solves all our needs, we should go with that route instead of the history api.

MichaelKetting (changes requested):
I think there's something missing here. The pageToken isn't used consistently as the key in sessionstorage. Am I missing something?

MichaelKetting (changes requested):
I think the webtests may have become significantly slower from this change. When looking at the executing tests, the history-back overlay is popping up. This alone seems to be strange since we don't do forward/back navigation and may indicate a bug. Why the test are slower but still do not fail, is not yet clear (or they may show the failure once the build is complete).

Question 1: are the tests still green?
Question 2: why is the navigation overlay appearing and how do we fix it?
Question 2a: does the navigation overlay cause the delay in the web tests?

MichaelKetting (approved):
Okay, it's working.

Please create tickets for 
* optimizing the key and removing old entries when there's too many
* and for pruning the forward history (I think this can be done?) because there's some edge cases with boing back and then forward.

PR #307, opened 2022-01-18
RM-8329 Simplify the implementation of AsyncLocalStorageProvider

a-ctor:
https://re-motion.atlassian.net/browse/RM-8329

### DIFF
--- a/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
+++ b/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
@@ -24,15 +24,28 @@ using Remotion.Utilities;
 
 namespace Remotion.Context
 {
+  /// <summary>
+  /// Implements a storage mechanism for <see cref="SafeContext"/> that uses <see cref="AsyncLocal{T}"/> to store the values.
+  /// Values flow by default and <see cref="OpenSafeContextBoundary"/> is necessary to prevent flow. 
+  /// </summary>
 #if !NETFRAMEWORK
   [ImplementationFor(typeof(ISafeContextStorageProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
 #endif
-  public class AsyncLocalStorageProvider : ISafeContextStorageProvider
+  public class AsyncLocalStorageProvider : ISafeContextStorageProvider, ISafeContextBoundaryStrategy
   {
-    private readonly AsyncLocal<Hashtable> _context = new();
+    private readonly AsyncLocal<Hashtable?> _context = new();
 
     public AsyncLocalStorageProvider ()
     {
+    }
+
+    /// <inheritdoc />
+    public SafeContextBoundary OpenSafeContextBoundary ()
+    {
+      var hashtable = _context.Value;
+      _context.Value = null;
+
+      return new SafeContextBoundary(this, hashtable);
     }
 
     /// <inheritdoc />
@@ -69,6 +82,12 @@ namespace Remotion.Context
       ArgumentUtility.CheckNotNull(nameof(key), key);
 
       _context.Value?.Remove(key);
+    }
+
+    /// <inheritdoc />
+    void ISafeContextBoundaryStrategy.RestorePreviousSafeContext (object? state)
+    {
+      _context.Value = (Hashtable?)state;
     }
   }
 }
--- a/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
+++ b/Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
@@ -17,234 +17,58 @@
 using System;
 using System.Collections;
 using System.Threading;
+#if !NETFRAMEWORK
 using Remotion.ServiceLocation;
+#endif
 using Remotion.Utilities;
-using System.Runtime.CompilerServices;
-using Remotion.Logging;
 
 namespace Remotion.Context
 {
 #if !NETFRAMEWORK
   [ImplementationFor(typeof(ISafeContextStorageProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
 #endif
-  [Obsolete("This type is experimental and might have unintended side effects. (Version 3.0.0-alpha.22)")]
   public class AsyncLocalStorageProvider : ISafeContextStorageProvider
   {
-    /// <summary>
-    /// Represents the context which stores values and is itself stored in the <see cref="AsyncLocal{T}"/>.
-    /// Is immutable because of the strange immutability constraints of <see cref="ExecutionContext"/>.
-    /// </summary>
-    /// <remarks>
-    /// In addition to the data stored in a hashtabe it also contains a reference to the previous context.
-    /// This is used to detect if a context is restored and reapply the values (see <see cref="AsyncLocalStorageProvider.AsyncLocalValueChangedHandler"/> below).
-    /// </remarks>
-    private class StorageContext
-    {
-      public static readonly StorageContext Empty = new StorageContext(null, null);
-
-      public static StorageContext FromPrevious (StorageContext? previous) => previous != null ? new StorageContext(null, new WeakReference<StorageContext>(previous)) : Empty;
-
-      private readonly WeakReference<StorageContext>? _previous;
-      private readonly Hashtable? _hashtable;
-
-      public StorageContext? Previous => _previous != null && _previous.TryGetTarget(out var result) ? result : null;
-
-      private StorageContext (Hashtable? hashtable, WeakReference<StorageContext>? previous)
-      {
-        _previous = previous;
-        _hashtable = hashtable;
-      }
-
-      public object? GetData (string key) => _hashtable?[key];
-
-      public StorageContext SetData (string key, object? newValue)
-      {
-        // We do not need to store null values as getting a non-existing value will return null (and there is no "Contains")
-        if (newValue == null)
-          return FreeData(key);
-
-        if (_hashtable == null)
-        {
-          var hashtable = new Hashtable();
-          hashtable[key] = newValue;
-
-          return new StorageContext(hashtable, _previous);
-        }
-
-        // Make sure that this set will change anything to avoid creating another container for nothing
-        var oldValue = _hashtable[key];
-        if (oldValue == newValue)
-          return this;
-
-        var newHashTable = (Hashtable)_hashtable.Clone();
-        newHashTable[key] = newValue;
-
-        return new StorageContext(newHashTable, _previous);
-      }
-
-      public StorageContext FreeData (string key)
-      {
-        if (_hashtable == null)
-          return this;
-
-        if (!_hashtable.ContainsKey(key))
-          return this;
-
-        var newHashtable = (Hashtable)_hashtable.Clone();
-        newHashtable.Remove(key);
-
-        return new StorageContext(newHashtable, _previous);
-      }
-    }
-
-    private class ObjectIdentifier
-    {
-      public int Value;
-
-      public ObjectIdentifier ()
-      {
-        Value = -1;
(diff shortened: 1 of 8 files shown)


Remotion/Web/Core/Context/HttpContextStorageProvider.cs
```diff
@@ -34,6 +34,11 @@ public class HttpContextStorageProvider : ISafeContextStorageProvider
 #pragma warning restore 618
 #endif
 
+    public SafeContextBoundary OpenSafeContextBoundary ()
+    {
+      return default;
```
MichaelKetting, line 39: I think we need the new semantic with the cleanup also for HttpContextStorageProvider in .NET 5. 

And since the semantics shouldn't differ between .net 5 and .net Framework, we also need this for Callcontext-Storageprovider.

MichaelKetting (commented):
Partial review due to the notes for SafeContextBoundary

Remotion/Web/Core/Context/HttpContextStorageProvider.cs
a-ctor: `HttpContextStorageProvider` writes the values directly into `HttpContext.Items`. How should a reset behave in this case?

MichaelKetting: hmm....removing and re-adding while the scope is open. this will "guard" scenarios such as parallel.foreach. For task etc, httpcontext should not propagate to the new thread anyway.

a-ctor: So we need to keep track of all values written to be able to remove those values at the boundary and restore the old values afterwards? But what about values from the inside scope leaking out?

We would need to save the complete `Items` collection and set it back the way it was. Which is kind of awkward since we cannot set the `Items` directly, but have to make a diff and do `Add`/`Remove` calls

Remotion/Core/Core/Context/ISafeContextBoundaryStrategy.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Remotion.Context
+{
+  /// <summary>
+  /// Implements a strategy for <see cref="SafeContextBoundary"/> to restore a previous safe context.
```
MichaelKetting, line 4:
```suggestion
  /// Implements a strategy for <see cref="SafeContextBoundary"/> to restore a previous context.
```
It's a consistency thing. `SafeContext` as a standalone type isn't an exact match, but reading `safe context` as two words seems odd.

Remotion/Core/Core/Context/ISafeContextBoundaryStrategy.cs
```diff
@@ -0,0 +1,15 @@
...
+  /// </summary>
+  /// <seealso cref="SafeContextBoundary"/>
+  /// <seealso cref="ISafeContextStorageProvider.OpenSafeContextBoundary"/>
+  public interface ISafeContextBoundaryStrategy
+  {
+    /// <summary>
+    /// Restores a previous state of the safe context from <paramref name="state"/>.
```
MichaelKetting, line 11:
```suggestion
    /// Restores a previous state of the context from <paramref name="state"/>.
```

Remotion/Core/Core/Context/ISafeContextStorageProvider.cs
```diff
@@ -24,6 +24,13 @@ namespace Remotion.Context
   // By default, use HttpContextStorageProvider. Only if not available, use CallContextStorageProvider.
   public interface ISafeContextStorageProvider
   {
+    /// <summary>
+    /// Opens a <see cref="SafeContextBoundary"/> into which values will not flow.
+    /// The previous context is restored when the <see cref="SafeContextBoundary"/> is disposed.
```
MichaelKetting, line 29:
```suggestion
    /// Opens a <see cref="SafeContextBoundary"/>, preventing the values in the current context from flowing to the new context.
    /// The previous context is restored when the <see cref="SafeContextBoundary"/> is disposed.
```

Remotion/Core/Core/Context/ISafeContextStorageProvider.cs
```diff
@@ -24,6 +24,13 @@ namespace Remotion.Context
...
+    /// </summary>
+    /// <returns>A <see cref="SafeContextBoundary"/> object with which the previous context can be restored.</returns>
+    SafeContextBoundary OpenSafeContextBoundary ();
```
MichaelKetting, line 32: The docs also imply that the boundary is unconditional, i.e. also works with CallContext and HttpContext.

Remotion/Core/Core/Context/SafeContextBoundary.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Remotion.Context
+{
+  /// <summary>
+  /// Represents a <see cref="SafeContext"/> boundary, across which values will not flow.
+  /// The previous context is restored when disposed.
+  /// </summary>
+  /// <seealso cref="ISafeContextStorageProvider.OpenSafeContextBoundary"/>
+  public readonly struct SafeContextBoundary : IDisposable
+  {
+    private readonly ISafeContextBoundaryStrategy _boundaryStrategy;
```
MichaelKetting, line 12: Why isn't NRT complaining that the `?` is missing?

Remotion/Core/UnitTests/Context/SafeContextBoundaryTests.cs
```diff
@@ -0,0 +1,30 @@
...
+      var state = new object();
+
+      var safeContextBoundary = new SafeContextBoundary(boundaryStrategyMock.Object, state);
+      safeContextBoundary.Dispose();
+
+      boundaryStrategyMock.Verify(e => e.RestorePreviousSafeContext(state));
+    }
+
+    [Test]
+    public void Dispose_WithNullBoundaryStrategy_ThrowsNothing ()
```
MichaelKetting, line 23:
```suggestion
    public void Dispose_WithDefaultInitialization_DoesNotThrow ()
```

Remotion/Core/UnitTests/Context/SafeContextBoundaryTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+      var safeContextBoundary = new SafeContextBoundary();
```
MichaelKetting, line 25: I wonder if we should use `default` here instead. C# 10 allows default ctors for structs with an implementation so this code might mask an issue in the future.

MichaelKetting (commented):
partial review, AsyncLocal-impl is left to review

Remotion/Web/Core/Context/HttpContextStorageProvider.cs
MichaelKetting: Actually: place a dictionary into the Items and switch the dictionary. Same as with AsyncLocal

MichaelKetting: (I know, it's different than before and should be documented, but over all, that solves things.)

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -106,19 +74,71 @@ public void SetData_FromOtherThread_NotAccessible ()
...
+    }
+
+    [Test]
+    public void SetData_FromThreadWithSafeContextBoundary_DoesNotOverrideValueOnMainThread ()
+    {
+      _provider.SetData(c_testKey, "1");
+
+      var t = new Thread(() =>
+      {
+        using (_provider.OpenSafeContextBoundary())
```
MichaelKetting, line 96: We should also add a test that shows that disposing the boundary is optional

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -17,234 +17,72 @@
...
-using Remotion.Logging;
 
 namespace Remotion.Context
 {
 #if !NETFRAMEWORK
   [ImplementationFor(typeof(ISafeContextStorageProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
 #endif
-  [Obsolete("This type is experimental and might have unintended side effects. (Version 3.0.0-alpha.22)")]
-  public class AsyncLocalStorageProvider : ISafeContextStorageProvider
+  public class AsyncLocalStorageProvider : ISafeContextStorageProvider, ISafeContextBoundaryStrategy
```
MichaelKetting, line 30: XML Doc

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -17,234 +17,72 @@
...
-    }
-
-    private AsyncLocalStorageProvider (bool tracingEnabled)
     {
-      _tracingEnabled = tracingEnabled;
-      _context = new AsyncLocal<StorageContext>(AsyncLocalValueChangedHandler);
     }
 
-    private void AsyncLocalValueChangedHandler (AsyncLocalValueChangedArgs<StorageContext> args)
+    public SafeContextBoundary OpenSafeContextBoundary ()
```
MichaelKetting, line 38: inherit doc?

Remotion/Core/Core/Context/AsyncLocalStorageProvider.cs
```diff
@@ -17,234 +17,72 @@
...
+      _context.Value?.Remove(key);
+    }
 
-      var newValue = oldValue.FreeData(key);
-      if (oldValue != newValue)
-        _context.Value = newValue;
+    /// <inheritdoc />
+    void ISafeContextBoundaryStrategy.RestorePreviousSafeContext (object? state)
+    {
+      _context.Value = state as Hashtable;
```
MichaelKetting, line 85: Hard cast.

Soft cast is only used when a type test is explicitly intended and a mismatched type does not indicate a bug. Here, an invalid state type will mean something's gone horribly wrong.

Remotion/Web/Core/Context/HttpContextStorageProvider.cs
MichaelKetting: Performance question: 
1. the lookup speed will depend on the key-length, the key should be just a few characters (see SafeContextKeys).
2. how does the performance of a weakreference-valueget compare to the lookup in the HttpContext.Items? If the weakreference is significantly better, we could discuss a performance optimization by caching the last dictonary in a threadlocal weakreference (there's a bit more involved...)

MichaelKetting (changes requested):

Remotion/Core/Core/Context/SafeContextBoundary.cs
a-ctor: Struct values could always be `null`, but that would not make sense with NRT. So, it is possible to define values that are non-nullable and the null checks guard only the assignment part in the constructor.

Remotion/Core/UnitTests/Context/AsyncLocalStorageProviderTest.cs
```diff
@@ -274,97 +455,226 @@ public void SuppressFlowWorksCorrectly ()
...
+    public void SetData_InParallelForeachWithUnclosedSafeContextBoundary_DoesNotSetValueForOuterScope ()
+    {
+      _provider.SetData(c_testKey, "initial");
+
+      Parallel.For(
+          0,
+          100,
+          v =>
+          {
+            _provider.OpenSafeContextBoundary();
```
MichaelKetting, line 670: this works, too? great!

MichaelKetting (approved):
please don't forget the httpcontext followup ticket

PR #309, opened 2022-01-24
RM-8340 Add support for serializing types in FlattenedSerializationInfo

a-ctor:
https://re-motion.atlassian.net/browse/RM-8340

### DIFF
--- a/Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedDeserializationInfo.cs
+++ b/Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedDeserializationInfo.cs
@@ -73,7 +73,17 @@ namespace Remotion.Data.DomainObjects.Infrastructure.Serialization
     [return: MaybeNull]
     private T GetValue<T> (bool allowNull)
     {
-      if (typeof(IFlattenedSerializable).IsAssignableFrom(typeof(T)))
+      if (typeof(T) == typeof(Type))
+      {
+        var typeName = allowNull
+            ? GetNullableValueForHandle<string>()
+            : GetValueForHandle<string>();
+
+        return typeName != null
+            ? (T)(object)GetTypeFromTypeNameOrThrow(typeName)
+            : default;
+      }
+      else if (typeof(IFlattenedSerializable).IsAssignableFrom(typeof(T)))
       {
         return GetFlattenedSerializable<T>(allowNull: allowNull);
       }
@@ -114,7 +124,7 @@ namespace Remotion.Data.DomainObjects.Infrastructure.Serialization
 
     private Func<FlattenedDeserializationInfo, object> GetInstanceFactory (string typeName)
     {
-      var type = Type.GetType(typeName, throwOnError: true, ignoreCase: false)!;
+      var type = GetTypeFromTypeNameOrThrow(typeName);
       var ctorInfo = type.GetConstructor(
           BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance,
           null,
@@ -223,6 +233,11 @@ namespace Remotion.Data.DomainObjects.Infrastructure.Serialization
 
       if (DeserializationFinished != null)
         DeserializationFinished(this, EventArgs.Empty);
+    }
+
+    private static Type GetTypeFromTypeNameOrThrow (string typeName)
+    {
+      return Type.GetType(typeName, throwOnError: true, ignoreCase: false)!;
     }
   }
 }
--- a/Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedSerializationInfo.cs
+++ b/Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedSerializationInfo.cs
@@ -16,6 +16,7 @@
 // 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using Remotion.Reflection;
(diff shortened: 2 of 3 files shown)


Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedDeserializationInfo.cs
```diff
@@ -73,7 +73,14 @@ public T GetValue<T> ()
     [return: MaybeNull]
     private T GetValue<T> (bool allowNull)
     {
-      if (typeof(IFlattenedSerializable).IsAssignableFrom(typeof(T)))
+      if (typeof(T) == typeof(Type))
+      {
+        var typeName = GetValue<string>(allowNull);
+        return typeName != null
+            ? (T)(object)Type.GetType(typeName, throwOnError: true, ignoreCase: false)!
```
MichaelKetting, line 80: We should refactor to extract the typename conversion into a helper method to remove the code duplication.

Unrelated: is .net smart enough to allow the use of `(Type)` instead of `(T)(object)`?

Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedSerializationInfo.cs
```diff
@@ -54,7 +54,9 @@ public void AddBoolValue (bool value)
     /// <remarks>Note: This will use T to decide whether to recurse into an IFlattenedSerializable or not.</remarks>
     public void AddValue<T> (T? value)
     {
-      if (typeof(IFlattenedSerializable).IsAssignableFrom(typeof(T)))
+      if (typeof(T) == typeof(Type))
+        _objectWriter.AddSimpleValue(((Type?)(object?)value)?.AssemblyQualifiedName);
```
MichaelKetting, line 58: Something I realized while checking the code duplication for type / string conversion:

While we don't need the handle here, it would be a good idea to use a handle anyway to store the type string. )The flattenedserializable works the same way.) There's just a handful of type strings in the appdomain but potentially hundreds or thousands of occurances. by putting the type string into a handle, the serialization set is compressed.

Please update.

It should be enough to simply change AddSimpleValue to AddHandle  / AddNullableHandle.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedDeserializationInfo.cs
a-ctor: Sadly, .NET is not smart enough so we need to do the cast to `object` first, thus we have a bit of code clutter.

Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedDeserializationInfo.cs
```diff
@@ -73,7 +73,17 @@ public T GetValue<T> ()
...
     {
-      if (typeof(IFlattenedSerializable).IsAssignableFrom(typeof(T)))
+      if (typeof(T) == typeof(Type))
+      {
+        var typeName = allowNull
+            ? GetNullableValueForHandle<string>()
+            : GetValueForHandle<string>();
+
+        return typeName != null
+            ? (T)(object)GetTypeFromTypeNameOrThrow(typeName)!
```
MichaelKetting, line 83: Do we need the null forgiving here?

Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedSerializationInfo.cs
```diff
@@ -54,7 +54,9 @@ public void AddBoolValue (bool value)
     /// <remarks>Note: This will use T to decide whether to recurse into an IFlattenedSerializable or not.</remarks>
     public void AddValue<T> (T? value)
     {
-      if (typeof(IFlattenedSerializable).IsAssignableFrom(typeof(T)))
+      if (typeof(T) == typeof(Type))
+        AddHandle((((Type?)(object?)value)?.AssemblyQualifiedName));
```
MichaelKetting, line 58: Same as with the serializer: need to factor out the type-string conversion. Yes, it's just one property, but there's different properties and which property we use is part of the business logic that should not be duplicated.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedSerializationInfo.cs
```diff
@@ -64,7 +67,8 @@ private void AddFlattenedSerializable (IFlattenedSerializable? serializable)
     {
       if (serializable != null)
       {
-        var typeName = serializable.GetType().AssemblyQualifiedName;
+        var type = serializable.GetType()!;
```
MichaelKetting, line 70: null-forgiving needed for c# to discover that the resulting typename is never null?

MichaelKetting (approved):
Got one NRT question, otherwise, we're done.

Remotion/Data/DomainObjects/Infrastructure/Serialization/FlattenedSerializationInfo.cs
a-ctor: I though it was because it was a result of the refactoring. But it seems like it is not needed, and that the nullability refactorings are error-prone and need re-checking....

PR #311, opened 2022-01-25
RM-7872: Assert that Thread.CurrentPrincipal.Identity property return …

iam-mholle:
…values are not null

### DIFF
--- a/Remotion/Security/Core/ThreadPrincipalProvider.cs
+++ b/Remotion/Security/Core/ThreadPrincipalProvider.cs
@@ -33,9 +33,9 @@ namespace Remotion.Security
 
     public ISecurityPrincipal GetPrincipal ()
     {
-      // TODO RM-7872: Add notnull assertions
-      IIdentity identity = Thread.CurrentPrincipal!.Identity!;
-      if (!identity.IsAuthenticated)
+      var identity = Thread.CurrentPrincipal?.Identity;
+
+      if (identity is null or { IsAuthenticated: false })
         return s_nullSecurityPrincipal;
 
       // TODO RM-7873: Add notnull assertion
--- a/Remotion/Security/UnitTests/ThreadPrincipalProviderTest.cs
+++ b/Remotion/Security/UnitTests/ThreadPrincipalProviderTest.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Security.Principal;
 using System.Threading;
+using Moq;
 using NUnit.Framework;
 
 namespace Remotion.Security.UnitTests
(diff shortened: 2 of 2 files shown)


MichaelKetting (changes requested):
Sorry, I didn't realize how much outdated detail there was in the ticket, fixed now. .NET 5.0 specifies that the CurrentPrincipal and the Identity may be null and this has already been observed (hence the upped prio). We need to do a null-principal fallback, same as with IsAuthenticated == false in line 40.

Also, please add tests now that it's acutal logic in there.

MichaelKetting (approved):

Remotion/Security/Core/ThreadPrincipalProvider.cs
```diff
@@ -33,9 +33,9 @@ public ThreadPrincipalProvider ()
 
     public ISecurityPrincipal GetPrincipal ()
     {
-      // TODO RM-7872: Add notnull assertions
-      IIdentity identity = Thread.CurrentPrincipal!.Identity!;
-      if (!identity.IsAuthenticated)
+      var identity = Thread.CurrentPrincipal?.Identity;
+
+      if (identity is null or { IsAuthenticated: false })
```
MichaelKetting, line 38: Exit-Condition code style preference: https://re-motion.atlassian.net/browse/RMTEAM-453

Remotion/Security/UnitTests/ThreadPrincipalProviderTest.cs
```diff
@@ -51,6 +52,29 @@ public void GetUser_NotAuthenticated ()
...
+    [Test]
+    public void GetPrincipal_WithCurrentPrincipalNull_ReturnsNullObject ()
+    {
+      Thread.CurrentPrincipal = null;
+#if NETFRAMEWORK
+      Assert.That(Thread.CurrentPrincipal, Is.Not.Null);
+#else
+      Assert.That(Thread.CurrentPrincipal, Is.Null);
+#endif
+      Assert.That(_principalProvider.GetPrincipal().IsNull, Is.True);
```
MichaelKetting, line 64: Interesting, .NET Framework is actually forcing a non-null value into the Thread.CurrentPrincipal when you try to set it to null for tests reasons (unless you also override the AppDomain's PrincipalPolicy).

MichaelKetting (approved):

Remotion/Security/Core/ThreadPrincipalProvider.cs
iam-mholle: @MichaelKetting  should I change it now or do we change it later?

MichaelKetting: As you like. There's certainly other places in the code base and and I'm not too concerned with the odd null case since the compiler also helps.  It just served as a reminder that there's an analyzer ticket to create :) If you think it's a quick win, we can schedule the ticket in one of the next sprints.

PR #314, opened 2022-01-26
RM-3974 Separate TypeDefinition from ClassDefinition

a-ctor:
https://re-motion.atlassian.net/browse/RM-3974

Review @MichaelKetting 
- [ ] Check ClassDefinition, TypeDefinition


Remotion/Data/DomainObjects.UnitTests/StandardMappingTest.cs
```diff
@@ -181,9 +181,14 @@ protected string GetPropertyIdentifier (Type declaringType, string shortProperty
       return declaringType.FullName + "." + shortPropertyName;
     }
 
-    protected ClassDefinition GetTypeDefinition (Type classType)
+    protected ClassDefinition GetClassDefinition (Type classType)
```
MichaelKetting, line 184: We need to introduce a wrapper for ClassDefinitionObjectMother that uses TypeDefinition as the public API. The goal is to only use GetClassDefinition or ClassDefinitionObjectMother where we actually want ClassDefinition instead of TypeDefinition.

MichaelKetting (commented):
unt tests reviewed

Remotion/Data/DomainObjects.UnitTests/Mapping/FakeMappingConfiguration.cs
```diff
@@ -58,7 +58,7 @@ public static void Reset ()
 
     private readonly StorageProviderDefinition _defaultStorageProviderDefinition;
     private readonly StorageProviderDefinition _nonPersistentProviderDefinition;
-    private readonly ReadOnlyDictionary<Type, ClassDefinition> _typeDefinitions;
+    private readonly ReadOnlyDictionary<Type, ClassDefinition> _classDefinitions;
```
MichaelKetting, line 61: why keep class and interface separate?

Remotion/Data/DomainObjects.UnitTests/Mapping/FakeMappingLoader.cs
```diff
@@ -30,15 +30,15 @@ public FakeMappingLoader ()
...
+    public TypeDefinition[] GetTypeDefinitions ()
     {
       var classDefinition = ClassDefinitionObjectMother.CreateClassDefinition("Fake", typeof(Company));
       classDefinition.SetPropertyDefinitions(new PropertyDefinitionCollection());
       classDefinition.SetDerivedClasses(new ClassDefinition[0]);
       return new[] { classDefinition };
     }
 
-    public RelationDefinition[] GetRelationDefinitions (IDictionary<Type, ClassDefinition> classDefinitions)
+    public RelationDefinition[] GetRelationDefinitions (IDictionary<Type, TypeDefinition> classDefinitions)
```
MichaelKetting, line 41: parameter name

Remotion/Data/DomainObjects.Web.IntegrationTests/SetUpFixture.cs
```diff
@@ -64,7 +64,7 @@ public void OneTimeSetUp ()
             pd => pd.Factory.CreateSchemaScriptBuilder(pd),
             new RdbmsStorageEntityDefinitionProvider(),
             new ScriptToStringConverter());
-        var scripts = scriptGenerator.GetScripts(MappingConfiguration.Current.GetTypeDefinitions()).Single();
+        var scripts = scriptGenerator.GetScripts(MappingConfiguration.Current.GetClassDefinitions()).Single();
```
MichaelKetting, line 67: revert

Remotion/Data/DomainObjects/Linq/MappingResolver.cs
```diff
@@ -156,10 +156,10 @@ public Expression ResolveTypeCheck (Expression checkedExpression, Type desiredTy
           throw new UnmappedItemException(message);
         }
 
-        var classDefinition = GetClassDefinition(desiredType);
+        var typeDefinition = GetTypeDefinition(desiredType);
         var idExpression = Expression.MakeMemberAccess(checkedExpression, s_idProperty.PropertyInfo);
         var classIDExpression = Expression.MakeMemberAccess(idExpression, s_classIDProperty.PropertyInfo);
-        var allClassDefinitions = EnumerableUtility.Singleton(classDefinition).Concat(classDefinition.GetAllDerivedClasses().Select(cd => cd));
+        var allClassDefinitions = typeDefinition.GetTypeHierarchy().OfType<ClassDefinition>();
```
MichaelKetting, line 162: When we get the interfaces, we will need to add a distinct. Or change this to do get DerivedClasses + ImplementingClasses separately.

Remotion/Data/DomainObjects/Mapping/ClassDefinitionCollectionFactory.cs
```diff
@@ -53,17 +53,17 @@ public ClassDefinition[] CreateClassDefinitionCollection (IEnumerable<Type> type
       return classDefinitions.Values.ToArray();
     }
```
MichaelKetting, line 55: I think we should expand this to TypeDefCollectionFactory and accept that the logic handles interfaces as well.

Remotion/Data/DomainObjects/Mapping/IMappingConfiguration.cs
```diff
@@ -24,11 +24,15 @@ namespace Remotion.Data.DomainObjects.Mapping
...
-    ClassDefinition GetTypeDefinition (Type classType);
-    ClassDefinition GetTypeDefinition (Type classType, Func<Type, Exception> missingTypeDefinitionExceptionFactory);
+    TypeDefinition[] GetTypeDefinitions ();
+    ClassDefinition[] GetClassDefinitions ();
+    bool ContainsTypeDefinition (Type type);
+    TypeDefinition GetTypeDefinition (Type type);
+    TypeDefinition GetTypeDefinition (Type type, Func<Type, Exception> missingTypeDefinitionExceptionFactory);
+    bool ContainsClassDefinition (Type classType);
     bool ContainsClassDefinition (string classID);
+    ClassDefinition GetClassDefinition (Type type);
```
MichaelKetting, line 34: The Type-based overloads shoudl be changed to extension methods.

GetClassDefinitions() is a big question mark. Do we need this method? And if so, where?

MichaelKetting (commented):
200 / 277

Remotion/Data/DomainObjects/Mapping/RelationDefinition.cs
```diff
@@ -85,10 +86,10 @@ public IRelationEndPointDefinition[] EndPointDefinitions
     {
```
MichaelKetting, line 86: This api (+ dependencies) seems to be only used in unit tests so we should be able to remove it completely.

Remotion/Data/DomainObjects/Mapping/SortExpressions/SortExpressionParser.cs
```diff
@@ -135,10 +135,13 @@ private PropertyDefinition ParsePropertyName (string fullOrShortPropertyName)
...
-    private IEnumerable<ClassDefinition> GetAllDerivedClassDefinitions (ClassDefinition baseClassDefinition)
+    private IEnumerable<TypeDefinition> GetAllDerivedTypeDefinitions (TypeDefinition baseTypeDefinition)
     {
-      var directlyDerived = baseClassDefinition.DerivedClasses.Cast<ClassDefinition>();
-      return directlyDerived.Concat(directlyDerived.SelectMany(cd => GetAllDerivedClassDefinitions(cd)));
+      if (baseTypeDefinition is not ClassDefinition baseClassDefinition)
+        return Enumerable.Empty<TypeDefinition>();
+
+      var directlyDerived = baseClassDefinition.DerivedClasses.Cast<TypeDefinition>();
+      return directlyDerived.Concat(directlyDerived.SelectMany(cd => GetAllDerivedTypeDefinitions(cd))); // TODO R2I Mapping: decision needed if we want to support this for interfaces as well
```
MichaelKetting, line 144: I think this has been ClassDefinition.GetAllDerivedTypes(), just manually implemented. Please check and if so, we want to change this to GetTypeHierarchy. The intention is to support properties available in the type hierarchy.

Update c# comments in line 105 and 117

Remotion/Data/DomainObjects/Mapping/Validation/Logical/CheckForClassIDIsValidValidationRule.cs
```diff
@@ -32,14 +32,14 @@ public CheckForClassIDIsValidValidationRule ()
...
     }
 
-    public MappingValidationResult Validate (ClassDefinition classDefinition)
+    public MappingValidationResult Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (!s_validClassID.IsMatch(classDefinition.ID))
+      if (typeDefinition is ClassDefinition classDefinition && !s_validClassID.IsMatch(classDefinition.ID))
```
MichaelKetting, line 39: please use separate exit conditions. 
Please add test for not-classdefinition

Remotion/Data/DomainObjects/Mapping/Validation/Logical/SortExpressionIsSupportedForCardinalityOfRelationPropertyValidationRule.cs
```diff
@@ -54,7 +54,7 @@ private MappingValidationResult Validate (IRelationEndPointDefinition relationEn
             virtualObjectRelationEndPointDefinition.PropertyInfo,
             "Property '{0}' of class '{1}' must not specify a SortExpression, because cardinality is equal to 'one'.",
```
MichaelKetting, line 55:
```suggestion
            "Property '{0}' of type '{1}' must not specify a SortExpression, because cardinality is equal to 'one'.",
```

MichaelKetting: actually, please check all validators, there's other "class" references in the messages, too, even though the validation rule isn't specfic to classdefinition

Remotion/Data/DomainObjects/Mapping/Validation/Logical/StorageClassIsSupportedValidationRule.cs
```diff
@@ -32,11 +32,11 @@ public StorageClassIsSupportedValidationRule ()
...
-    public IEnumerable<MappingValidationResult> Validate (ClassDefinition classDefinition)
+    public IEnumerable<MappingValidationResult> Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      return from PropertyDefinition propertyDefinition in classDefinition.MyPropertyDefinitions
+      return from PropertyDefinition propertyDefinition in typeDefinition.MyPropertyDefinitions
              select Validate(propertyDefinition.PropertyInfo);
     }
```
MichaelKetting, line 42: message "class" / "type"

Remotion/Data/DomainObjects/Mapping/Validation/Logical/StorageGroupTypesAreSameWithinInheritanceTreeRule.cs
```diff
@@ -21,20 +21,20 @@ namespace Remotion.Data.DomainObjects.Mapping.Validation.Logical
...
   /// Validates that the class has the same storage group type as its base class.
   /// </summary>
-  public class StorageGroupTypesAreSameWithinInheritanceTreeRule : IClassDefinitionValidationRule
+  public class StorageGroupTypesAreSameWithinInheritanceTreeRule : ITypeDefinitionValidationRule
   {
-    public MappingValidationResult Validate (ClassDefinition classDefinition)
+    public MappingValidationResult Validate (TypeDefinition typeDefinition)
     {
-      if (classDefinition.BaseClass != null)
+      if (typeDefinition is ClassDefinition classDefinition && classDefinition.BaseClass != null)
```
MichaelKetting, line 28: this needs an interface-path, too. It tests that the baseclass and all implemented interfaces satisfy this rule

Remotion/Data/DomainObjects/Mapping/Validation/Logical/StorageGroupTypesAreSameWithinInheritanceTreeRule.cs
```diff
@@ -21,20 +21,20 @@ namespace Remotion.Data.DomainObjects.Mapping.Validation.Logical
...
       {
         if (classDefinition.StorageGroupType != classDefinition.BaseClass.StorageGroupType)
         {
           var message = "Class '{0}' must have the same storage group type as its base class '{1}'.";
```
MichaelKetting, line 32: message "class" / "type"

Remotion/Data/DomainObjects/Mapping/Validation/Reflection/ClassDefinitionTypeIsSubclassOfDomainObjectValidationRule.cs
```diff
@@ -22,18 +22,18 @@ namespace Remotion.Data.DomainObjects.Mapping.Validation.Reflection
   /// <summary>
   /// Validates that a class definition type is a sub-class of domain object. 
   /// </summary>
-  public class ClassDefinitionTypeIsSubclassOfDomainObjectValidationRule : IClassDefinitionValidationRule
+  public class ClassDefinitionTypeIsSubclassOfDomainObjectValidationRule : ITypeDefinitionValidationRule
```
MichaelKetting, line 25: TODO R2I buddy: add rule for itnerface is IDomainObject

Remotion/Data/DomainObjects/Mapping/Validation/Reflection/ClassDefinitionTypeIsSubclassOfDomainObjectValidationRule.cs
```diff
@@ -22,18 +22,18 @@ namespace Remotion.Data.DomainObjects.Mapping.Validation.Reflection
...
   {
-    public MappingValidationResult Validate (ClassDefinition classDefinition)
+    public MappingValidationResult Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (!ReflectionUtility.IsDomainObject(classDefinition.ClassType))
+      if (typeDefinition is ClassDefinition classDefinition && !ReflectionUtility.IsDomainObject(classDefinition.Type))
```
MichaelKetting, line 31: split.

Remotion/Data/DomainObjects/Mapping/Validation/Reflection/InheritanceHierarchyFollowsClassHierarchyValidationRule.cs
```diff
@@ -22,23 +22,23 @@ namespace Remotion.Data.DomainObjects.Mapping.Validation.Reflection
...
     public InheritanceHierarchyFollowsClassHierarchyValidationRule ()
     {
 
     }
 
-    public MappingValidationResult Validate (ClassDefinition classDefinition)
+    public MappingValidationResult Validate (TypeDefinition typeDefinition)
     {
-      if (classDefinition.BaseClass !=null && !classDefinition.ClassType.IsSubclassOf(classDefinition.BaseClass.ClassType))
+      if (typeDefinition is ClassDefinition classDefinition && classDefinition.BaseClass !=null && !classDefinition.Type.IsSubclassOf(classDefinition.BaseClass.Type))
```
MichaelKetting, line 34: needs either an interface buddy or implementation for itnerface, too. it's okay to split this into separate cases given the complexity. if so, split to have exit condtion separate separate + test

MichaelKetting (commented):
235/277

Remotion/Data/DomainObjects/Mapping/Validation/Reflection/StorageGroupAttributeIsOnlyDefinedOncePerInheritanceHierarchyValidationRule.cs
```diff
@@ -22,17 +22,17 @@ namespace Remotion.Data.DomainObjects.Mapping.Validation.Reflection
...
     }
 
-    public MappingValidationResult Validate (ClassDefinition classDefinition)
+    public MappingValidationResult Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      return Validate(classDefinition.ClassType);
+      return Validate(typeDefinition.Type);
```
MichaelKetting, line 35: TODO R2I

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/IStorageNameProvider.cs
```diff
@@ -32,7 +32,7 @@ public interface IStorageNameProvider
     string GetClassIDColumnName ();
     string GetTimestampColumnName ();
 
-    EntityNameDefinition? GetTableName (ClassDefinition classDefinition);
+    EntityNameDefinition? GetTableName (ClassDefinition classDefinition); // TODO R2I Persistence: check usages for required changes
     EntityNameDefinition GetViewName (ClassDefinition classDefinition);
```
MichaelKetting, line 36: Same here, or switch to TypeDefintiion

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/RelationStoragePropertyDefinitionFactory.cs
```diff
@@ -83,7 +83,7 @@ public IRdbmsStoragePropertyDefinition CreateStoragePropertyDefinition (Relation
       var oppositeEndPointDefinition = relationEndPointDefinition.GetOppositeEndPointDefinition();
       var relationColumnName = _storageNameProvider.GetRelationColumnName(relationEndPointDefinition);
       var relationClassIDColumnName = _storageNameProvider.GetRelationClassIDColumnName(relationEndPointDefinition);
-      return CreateStoragePropertyDefinition(oppositeEndPointDefinition.ClassDefinition, relationColumnName, relationClassIDColumnName);
+      return CreateStoragePropertyDefinition((ClassDefinition)oppositeEndPointDefinition.TypeDefinition, relationColumnName, relationClassIDColumnName); // TODO R2I Persistence: Support TypeDefinition
```
MichaelKetting, line 86: wouldn't this be cleaner with a dedicated exception, like in the other files?

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Validation/ClassAboveTableIsAbstractValidationRule.cs
```diff
@@ -32,11 +32,15 @@ public ClassAboveTableIsAbstractValidationRule ()
...
     }
 
-    public IEnumerable<MappingValidationResult> Validate (ClassDefinition classDefinition)
+    public IEnumerable<MappingValidationResult> Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (!classDefinition.IsClassTypeResolved)
+      if (typeDefinition is not ClassDefinition classDefinition)
```
MichaelKetting, line 39: test case

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Validation/OnlyOneTablePerHierarchyValidationRule.cs
```diff
@@ -32,11 +32,15 @@ public OnlyOneTablePerHierarchyValidationRule ()
...
-    public IEnumerable<MappingValidationResult> Validate (ClassDefinition classDefinition)
+    public IEnumerable<MappingValidationResult> Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (classDefinition.BaseClass != null && classDefinition.HasStorageEntityDefinitionBeenSet && classDefinition.StorageEntityDefinition is TableDefinition)
+      if (typeDefinition is not ClassDefinition classDefinition)
+      {
+        yield return MappingValidationResult.CreateValidResult();
```
MichaelKetting, line 41: testcase.

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Validation/TableNamesAreDistinctWithinConcreteTableInheritanceHierarchyValidationRule.cs
```diff
@@ -33,22 +33,22 @@ public TableNamesAreDistinctWithinConcreteTableInheritanceHierarchyValidationRul
...
     }
 
-    public IEnumerable<MappingValidationResult> Validate (ClassDefinition classDefinition)
+    public IEnumerable<MappingValidationResult> Validate (TypeDefinition typeDefinition)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (classDefinition.BaseClass == null) //if class definition is inheritance root class
+      if (typeDefinition is ClassDefinition classDefinition && classDefinition.BaseClass == null) //if class definition is inheritance root class
```
MichaelKetting, line 40: separate condition + test

Remotion/Data/DomainObjects/Persistence/Rdbms/RdbmsProvider.cs
```diff
@@ -448,14 +449,14 @@ private ObjectID CheckStorageProvider (ObjectID id, string argumentName)
...
 
-    private void CheckClassDefinition (ClassDefinition classDefinition, string argumentName)
+    private void CheckTypeDefinition (TypeDefinition typeDefinition, string argumentName)
     {
-      if (classDefinition.StorageEntityDefinition.StorageProviderDefinition != StorageProviderDefinition)
+      if (typeDefinition.StorageEntityDefinition.StorageProviderDefinition != StorageProviderDefinition)
       {
         throw CreateArgumentException(
             argumentName,
             "The StorageProviderID '{0}' of the provided ClassDefinition does not match with this StorageProvider's ID '{1}'.",
```
MichaelKetting, line 458:
```suggestion
            "The StorageProviderID '{0}' of the provided TypeDefinition does not match with this StorageProvider's ID '{1}'.",
```

Remotion/Data/DomainObjects/Persistence/StorageEntityBasedStorageProviderDefinitionFinder.cs
```diff
@@ -22,19 +22,19 @@
...
   {
-    public StorageProviderDefinition GetStorageProviderDefinition (ClassDefinition classDefinition, string? errorMessageContext)
+    public StorageProviderDefinition GetStorageProviderDefinition (TypeDefinition typeDefinition, string? errorMessageContext)
     {
-      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+      ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (!classDefinition.HasStorageEntityDefinitionBeenSet)
+      if (!typeDefinition.HasStorageEntityDefinitionBeenSet)
         throw new InvalidOperationException("Cannot obtain storage provider for ClassDefinitions without storage entities. " + errorMessageContext);
```
MichaelKetting, line 35:
```suggestion
        throw new InvalidOperationException("Cannot obtain storage provider for TypeDefinitions without storage entities. " + errorMessageContext);
```

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,224 @@
+﻿#nullable enable
```
MichaelKetting, line 1:
```suggestion
```

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,224 @@
...
+    public abstract TypeDefinition GetInheritanceRoot ();
+
+    /// <summary>
+    /// Returns all types that are part of this type's hierarchy, including itself, using depth-first.
+    /// </summary>
+    public abstract TypeDefinition[] GetTypeHierarchy ();
+
+    protected abstract void CheckPropertyDefinitions (IEnumerable<PropertyDefinition> propertyDefinitions);
+
+    protected abstract void CheckRelationEndPointDefinitions (IEnumerable<IRelationEndPointDefinition> relationEndPoints);
```
MichaelKetting, line 222: Abstract members directly after ctor, please.

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,224 @@
...
+    {
+      IsReadOnly = true;
+    }
+
+    public override string ToString ()
+    {
+      return $"{GetType().GetFullNameChecked()}: {Type.GetFullNameSafe()}";
+    }
+
+    public PropertyDefinition? this [string propertyName]
```
MichaelKetting, line 204: This one should be up tehre with the other property-stuff

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,224 @@
...
+    private readonly Lazy<RelationEndPointDefinitionCollection> _cachedRelationEndPointDefinitions;
+    private readonly Lazy<PropertyDefinitionCollection> _cachedPropertyDefinitions;
+
+    private PropertyDefinitionCollection? _propertyDefinitions;
+    private RelationEndPointDefinitionCollection? _relationEndPoints;
+    private IStorageEntityDefinition? _storageEntityDefinition;
+
+    public Type Type { get; }
+
+    public IStorageEntityDefinition StorageEntityDefinition
```
MichaelKetting, line 23: non-field stuff after the ctor

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,224 @@
...
+    {
+      ArgumentUtility.CheckNotNull(nameof(propertyInformation), propertyInformation);
+
+      var propertyAccessorData = PropertyAccessorDataCache.ResolvePropertyAccessorData(propertyInformation);
+      return propertyAccessorData?.RelationEndPointDefinition;
+    }
+
+    public bool IsAssignableFrom (TypeDefinition other)
+    {
+      return Type.IsAssignableFrom(other.Type);
```
MichaelKetting, line 191: argcheck

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -120,66 +90,52 @@ public bool IsSameOrBaseClassOf (ClassDefinition classDefinition)
...
+        return BaseClass.GetInheritanceRoot();
 
       return this;
     }
 
-    [Obsolete("Contains (...) method was ambiguous between GetPropertyDefinitions() and MyPropertyDefinitions. Use GetPropertyDefinitions().Contains (...) instead. (Version: 3.0.0)", true)]
+    [Obsolete(
+        "Contains (...) method was ambiguous between GetPropertyDefinitions() and MyPropertyDefinitions. Use GetPropertyDefinitions().Contains (...) instead. (Version: 3.0.0)",
+        true)]
     public bool Contains (PropertyDefinition propertyDefinition)
```
MichaelKetting, line 124: should get moved to base type. the obsolete-method's just there to inform the developer what happend to the feature, so when we now swtich from class to type, the developer will be looking for thsi on TypeDef.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -120,66 +90,52 @@ public bool IsSameOrBaseClassOf (ClassDefinition classDefinition)
...
-      if (relationEndPointDefinition == null)
-        throw CreateMappingException("No relation found for class '{0}' and property '{1}'.", ID, propertyName);
-
-      return relationEndPointDefinition;
-    }
-
     public bool IsMyRelationEndPoint (IRelationEndPointDefinition relationEndPointDefinition)
     {
-      ArgumentUtility.CheckNotNull("relationEndPointDefinition", relationEndPointDefinition);
+      ArgumentUtility.CheckNotNull(nameof(relationEndPointDefinition), relationEndPointDefinition);
```
MichaelKetting, line 131: I would prefer not to do this sort of cleanup in between. it's better to do one global cleanup for this.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -446,61 +223,58 @@ private void FillAllDerivedClasses (List<ClassDefinition> allDerivedClasses)
...
       {
-        if (!ReferenceEquals(propertyDefinition.ClassDefinition, this))
+        if (!ReferenceEquals(propertyDefinition.TypeDefinition, this))
         {
-          throw CreateMappingException(
-              "Property '{0}' cannot be added to class '{1}', because it was initialized for class '{2}'.",
-              propertyDefinition.PropertyName,
-              _id,
-              propertyDefinition.ClassDefinition.ID);
+          throw new MappingException(
```
MichaelKetting, line 237: why did you change this away from the factory method? Is this just because you want to use the interpolated strings? in that case, interpolated strings should wait until we have interpolated overloads and can keep the factory methods.

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,224 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Threading;
+using Remotion.Data.DomainObjects.Persistence.Model;
+using Remotion.Reflection;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Mapping
+{
+  public abstract class TypeDefinition
```
MichaelKetting, line 12: Please add a testfixture for TypeDefinition with at leaast one empty method and an ignore with the appropriate todo-comment. That way, the missing tests popup.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/Mapping/ClassDefinitionTest.cs
```diff
@@ -545,48 +546,53 @@ public void GetAllPropertyDefinitions_ReadOnly ()
     [Test]
     public void AddPropertyToOtherClass ()
     {
-      var companyClass = ClassDefinitionObjectMother.CreateClassDefinition("Company");
-      var orderClass = ClassDefinitionObjectMother.CreateClassDefinition("Order");
+      var companyClass = ClassDefinitionObjectMother.CreateClassDefinition("Company", typeof(Company));
```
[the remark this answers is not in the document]
a-ctor: I don't think that it is worth the effort trying to simplify the usages here even if the class ID is not used. The class ID can also be useful when debugging as it is simpler to identify than the type which starts of the same for each object.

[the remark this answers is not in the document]
MichaelKetting: Maybe there's a misunderstanding, I'm suggesting we remove the ID when we now explicitly supply the Type. The information available during debugging remains the same but it keeps the test complexity equal to before the change. Now, when we specify both, we either have to keep doing it (out of habit or because we think it's the right way to do things) or we will not supply the ID for any new code, making the old code appear to be more complex due to the extra parameter. That's why dropping the redudant parameter is relevant to me.

Remotion/Data/DomainObjects/Infrastructure/TypePipe/ClassDefinitionProvider.cs
```diff
@@ -22,22 +22,22 @@
...
+  [ImplementationFor(typeof(IClassDefinitionProvider), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Single)]
+  public class ClassDefinitionProvider : IClassDefinitionProvider
   {
-    public ClassDefinition? GetTypeDefinition (Type domainObjectType)
+    public ClassDefinition? GetClassDefinition (Type domainObjectType)
     {
       ArgumentUtility.CheckNotNull("domainObjectType", domainObjectType);
 
       var mappingConfiguration = MappingConfiguration.Current;
       if (!mappingConfiguration.ContainsTypeDefinition(domainObjectType))
```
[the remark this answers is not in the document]
a-ctor: It also return the class definition so I would keep the current name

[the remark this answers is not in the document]
MichaelKetting: No, I was referring to using the ClassDefinition-only implementation `ContainsClassDefinition` instead of the any-TypeDefinition `ContainsTypeDefinition` method. By testing for any TypeDefinition instead of just ClassDefinition, the implementation is at least strange if not buggy. Don't see the entire method in the current github quote so I don't know the full damage right now.

Remotion/Data/DomainObjects/Mapping/IMappingConfiguration.cs
a-ctor: The usages of `GetClassDefinitions` are now gone and I removed the method.

Remotion/Data/DomainObjects/Mapping/ClassDefinitionCollectionFactory.cs
a-ctor: I would do this in a future task that requires it instead of pushing this change into this task.

Remotion/Data/DomainObjects/Mapping/RelationDefinition.cs
a-ctor: Is the context at the wrong position? The `EndPointDefinitions` properties does not have anything to do with the introduction of `TypeDefinition`.

Remotion/Data/DomainObjects/Mapping/ClassDefinitionCollectionFactory.cs
MichaelKetting: pelase add a R2I todo then

Remotion/Data/DomainObjects/Mapping/RelationDefinition.cs
MichaelKetting: Nope, it's this Method and related ones that also accept the `classID` as a string.
```
public IRelationEndPointDefinition? GetEndPointDefinition (string classID, string propertyName)
```
There's no production usage left, as far as I could tell.

a-ctor: Are these the methods you are talking about?
image

The `GetOppositeEndPointDefinition` still has dependencies downstream and it depends on the `GetEndPointDefinition`. I don't think it makes sense to remove only parts of the API

MichaelKetting: public IRelationEndPointDefinition? GetEndPointDefinition (string classID, string propertyName)
public IRelationEndPointDefinition? GetOppositeEndPointDefinition (string classID, string propertyName)
public bool IsEndPoint (string classID, string propertyName)

-> those 3 methods with a `string classID` parameter. The one with IRelationEndPointDefinition from the screenshot remains.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
a-ctor: What exactly is the benefit of having a factory method in this case? It just hides the type of exception that from the reader. If the method provided any conditional logic or values that must not provided I would argue for the factory method. But as it stands I don't see the value for it.

MichaelKetting: It's a change that's completely unrelated to the current task. There's several factory methods for MappingException (and possibly some other exception types in Data.DomainObjects) and the descision to remove them should be consistent accross the (related) code base. Originally, the reason was likely to remove the string.Format calls from the call sites and perhaps also to remove the code duplication of choosing the exception type. Those descisions go back to 2004 so not everything's still valid in today's world. Either way, though, the change should be done separately and consistently on the code base, unless there's a benefit to the current task.

Remotion/Data/DomainObjects/Mapping/RelationDefinition.cs
a-ctor: Shouldn't we convert those into `Type, string` overloads instead? Or is there no use case for something like that? In our tests we use one of the methods but we could inline the logic there.

MichaelKetting: Hmm... Good point. Yes, that would work and keeps the noise down.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
a-ctor: I understand the need for the factory methods if you have complex stuff that you do not want to repeat everywhere. In this case the inlined code is probably shorter than the current method invocation while also making it clear what is happening without needing to jump to the helper method.

I don't quite get the argument of we should do it everywhere or nowhere at all. I am not for removing the helper methods just for the fun but in this case the helper is just not necessary and, in my opinion, makes the code less readable. Since all the helper methods are redeclared on each class why do we need to change all at once?

MichaelKetting: My reasoning is that there is no need to change this now since the code doesn't depend on this - we don't even have to change the API etc. To me, this makes it into a completely unrelated cleanup we can do any time, but we should do it in a dedicated commit (and task). And when we do that, we can cover the other 5 or so CreateMappingExceptions, too, since it's just 5 minutes more work. But, by doing it in a separate task, we isolate the change and make it easily tracable.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Model/Validation/OnlyOneTablePerHierarchyValidationRuleTest.cs
```diff
@@ -119,5 +120,14 @@ public void HasBaseClassesAndHasTableDefinition_BaseOfBaseClassHasTableDefinitio
           + "Declaring type: Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Validation.DerivedValidationDomainObjectClass";
       AssertMappingValidationResult(validationResult, false, expectedMessage);
     }
+
+    [Test]
+    public void HasBaseClassesAndHasTableDefinition_WithNonClassDefinition_ReturnsValidResult ()
```
MichaelKetting, line 125: See other validation tests

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Model/Validation/TableNamesAreDistinctWithinConcreteTableInheritanceHierarchyValidationRuleTest.cs
```diff
@@ -139,5 +140,14 @@ public void InheritanceRoot_SameTableNameInDerivedDerivedClass ()
           + "Declaring type: Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Validation.BaseOfBaseValidationDomainObjectClass";
       AssertMappingValidationResult(validationResult, false, expectedMessage);
     }
+
+    [Test]
+    public void InheritanceRoot_NonClassDefinition_ReturnsEmptyEnumerable ()
```
MichaelKetting, line 145: See other validation tests

Remotion/Data/DomainObjects/Mapping/IMappingConfiguration.cs
```diff
@@ -24,12 +24,11 @@ namespace Remotion.Data.DomainObjects.Mapping
...
   public interface IMappingConfiguration
   {
-    ClassDefinition[] GetTypeDefinitions ();
-    bool ContainsTypeDefinition (Type classType);
-    ClassDefinition GetTypeDefinition (Type classType);
-    ClassDefinition GetTypeDefinition (Type classType, Func<Type, Exception> missingTypeDefinitionExceptionFactory);
+    TypeDefinition[] GetTypeDefinitions ();
+    bool ContainsTypeDefinition (Type type);
+    TypeDefinition GetTypeDefinition (Type type, Func<Type, Exception> missingTypeDefinitionExceptionFactory);
+    bool ContainsClassDefinition (Type classType);
```
MichaelKetting, line 30: Where's the usage for ContainsClassDefinition? (didn't do a checkout) It's assymetric to typedef so that's a question mark.

Remotion/Data/DomainObjects/Persistence/Rdbms/SchemaGeneration/RdbmsStorageEntityDefinitionProvider.cs
```diff
@@ -24,16 +24,16 @@
...
   {
-    public IEnumerable<IRdbmsStorageEntityDefinition> GetEntityDefinitions (IEnumerable<ClassDefinition> classDefinitions)
+    public IEnumerable<IRdbmsStorageEntityDefinition> GetEntityDefinitions (IEnumerable<TypeDefinition> typeDefinitions)
     {
-      ArgumentUtility.CheckNotNull("classDefinitions", classDefinitions);
+      ArgumentUtility.CheckNotNull("typeDefinitions", typeDefinitions);
 
-      return classDefinitions
+      return typeDefinitions
           .Select(cd => cd.StorageEntityDefinition)
```
MichaelKetting, line 37: cd / td

Remotion/Data/DomainObjects/Persistence/Rdbms/SchemaGeneration/ScriptGenerator.cs
```diff
@@ -47,12 +47,12 @@ public class ScriptGenerator
...
 
-    public IEnumerable<Script> GetScripts (IEnumerable<ClassDefinition> classDefinitions)
+    public IEnumerable<Script> GetScripts (IEnumerable<TypeDefinition> typeDefinitions)
     {
-      ArgumentUtility.CheckNotNull("classDefinitions", classDefinitions);
+      ArgumentUtility.CheckNotNull("typeDefinitions", typeDefinitions);
 
       var classDefinitionsByStorageProvider =
-          from cd in classDefinitions
+          from cd in typeDefinitions
```
MichaelKetting, line 55: cd / td

MichaelKetting (commented):
partial review. 6 files are left

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,248 @@
...
+
+      if (IsReadOnly)
+        throw new NotSupportedException($"Type '{Type.GetFullNameSafe()}' is read-only.");
+
+      _storageEntityDefinition = storageEntityDefinition;
+    }
+
+    public PropertyDefinitionCollection GetPropertyDefinitions ()
+    {
+      return _cachedPropertyDefinitions.Value;
```
MichaelKetting, line 133: I think we need to guard this (and relations) with IsReadOnly and either throw or return a new instance when we're not readonly. otherwise, we could accidently cache incomplete data. It's not there in the original code either, but reading it now, it feels icky

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -120,61 +90,39 @@ public bool IsSameOrBaseClassOf (ClassDefinition classDefinition)
...
-    }
 
-    [Obsolete("Contains (...) method was ambiguous between GetPropertyDefinitions() and MyPropertyDefinitions. Use GetPropertyDefinitions().Contains (...) instead. (Version: 3.0.0)", true)]
-    public bool Contains (PropertyDefinition propertyDefinition)
-    {
-      throw new NotSupportedException("Use GetPropertyDefinitions().Contains (...) instead. (Version: 3.0.0)");
-    }
+      var allDerivedTypes = new TypeDefinition[allDerivedClasses.Count];
+      for (var i = 0; i < allDerivedTypes.Length; i++)
+        allDerivedTypes[i] = allDerivedClasses[i];
```
MichaelKetting, line 101: Is this an Array.Copy manually implemented? -> switch to Array.Copy ?

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -190,217 +138,49 @@ public bool IsRelationEndPoint (IRelationEndPointDefinition relationEndPointDefi
...
-    public bool IsPartOfInheritanceHierarchy
-    {
-      get { return (BaseClass != null || DerivedClasses.Count > 0); }
-    }
-
-    public IPersistentMixinFinder PersistentMixinFinder
-    {
-      get { return _persistentMixinFinder; }
-    }
+    public override bool IsPartOfInheritanceHierarchy => (BaseClass != null || DerivedClasses.Count > 0);
```
MichaelKetting, line 181: drop the parens?

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -446,61 +216,63 @@ private void FillAllDerivedClasses (List<ClassDefinition> allDerivedClasses)
       }
     }
 
-    private void CheckPropertyDefinitions (IEnumerable<PropertyDefinition> propertyDefinitions)
+    public override string ToString ()
+    {
+      return $"{GetType().GetFullNameChecked()}: {ID}";
+    }
```
MichaelKetting, line 222: Since you're doing cleanup: switch to expression bodied member?
also noticed: someone thought that the fullname instead of the name would be a good idea here. I don't think that's true anymore in 2022. You want to keep it or change it?

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    public void SetReadOnly ()
+    {
+      IsReadOnly = true;
+    }
+
+    public override string ToString ()
+    {
+      return $"{GetType().GetFullNameChecked()}: {Type.GetFullNameSafe()}";
```
MichaelKetting, line 245: Depending on what is done in ClassDefinition, we'd use the Name instead of FullName here, too. And do you want to do Expression Bodied members for the one liners?

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -531,20 +300,24 @@ private void CheckDerivedClass (IEnumerable<ClassDefinition> derivedClasses)
...
       {
-        var handleType = typeof(DomainObjectHandle<>).MakeGenericType(ClassType);
+        var handleType = typeof(DomainObjectHandle<>).MakeGenericType(Type);
 
         var constructorInfo = handleType.GetConstructor(
-            BindingFlags.Public | BindingFlags.Instance, null, new[] { typeof(ObjectID) }, null);
+            BindingFlags.Public | BindingFlags.Instance,
+            null,
+            new[] { typeof(ObjectID) },
+            null);
```
MichaelKetting, line 311: Reformat into single line?

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -446,61 +216,63 @@ private void FillAllDerivedClasses (List<ClassDefinition> allDerivedClasses)
...
       foreach (IRelationEndPointDefinition endPointDefinition in relationEndPoints)
       {
-        var relationEndPointClassDefinition = endPointDefinition.ClassDefinition;
+        var relationEndPointClassDefinition = endPointDefinition.TypeDefinition;
         var relationEndPointPropertyName = endPointDefinition.PropertyName;
         Assertion.DebugAssert(endPointDefinition.IsAnonymous == false, "endPointDefinition.IsAnonymous == false");
-        Assertion.DebugIsNotNull(relationEndPointPropertyName, "endPointDefinition.PropertyName != null when endPointDefinition.IsAnonymous == false");
+        Assertion.DebugIsNotNull(
+            relationEndPointPropertyName,
+            "endPointDefinition.PropertyName != null when endPointDefinition.IsAnonymous == false");
```
MichaelKetting, line 260: Why wrap this? linelength is 180, this should fit, shouldn't it?

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -446,61 +216,63 @@ private void FillAllDerivedClasses (List<ClassDefinition> allDerivedClasses)
...
-              _id,
-              propertyDefinition.ClassDefinition.ID);
+          throw CreateMappingException();
+
+          MappingException CreateMappingException ()
+          {
+            return new MappingException(
+                $"Property '{propertyDefinition.PropertyName}' cannot be added to class '{ID}',"
+                + $" because it was initialized for type '{propertyDefinition.TypeDefinition.Type.GetFullNameSafe()}'.");
+          }
```
MichaelKetting, line 237: why a local method?

Remotion/Data/DomainObjects/Mapping/MappingConfigurationExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Mapping
+{
+  public static class MappingConfigurationExtensions
```
MichaelKetting, line 22: xmldoc: "extension methods for mappingconfig" (nicely formatted / grammarfixed)

Remotion/Data/DomainObjects/Mapping/MappingConfigurationExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+      return mappingConfiguration.GetTypeDefinition(type, static type => CreateMappingException(type.ToString()));
+    }
+
+    public static ClassDefinition GetClassDefinition (this IMappingConfiguration mappingConfiguration, Type type)
+    {
+      ArgumentUtility.CheckNotNull(nameof(mappingConfiguration), mappingConfiguration);
+      ArgumentUtility.CheckNotNull(nameof(type), type);
+      if (!type.IsClass)
+        throw new ArgumentException("The specified type must be a class.", nameof(type));
```
MichaelKetting, line 37: please add the type.GetSafeFullName() to the message (here and below)
"Type '{0}' must be a class."

Remotion/Data/DomainObjects/Mapping/MappingConfigurationExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Remotion.Data.DomainObjects.Mapping
+{
+  public static class MappingConfigurationExtensions
+  {
+    public static TypeDefinition GetTypeDefinition (this IMappingConfiguration mappingConfiguration, Type type)
+    {
+      ArgumentUtility.CheckNotNull(nameof(mappingConfiguration), mappingConfiguration);
+      ArgumentUtility.CheckNotNull(nameof(type), type);
+
+      return mappingConfiguration.GetTypeDefinition(type, static type => CreateMappingException(type.ToString()));
```
MichaelKetting, line 29: prbably need to inline the CreateMappingEx since there's different msgs for type and class needed.

Remotion/Data/DomainObjects/Mapping/MappingConfigurationExtensions.cs
MichaelKetting, line 22: I don't remember: do we have `MappingConfigurationExtensionsTests`?

Remotion/Data/DomainObjects.UnitTests/Mapping/Validation/TypeDefinitionValidatorTest.cs
```diff
@@ -0,0 +1,136 @@
...
+
+      var mappingValidationResults = validator.Validate(new[] { _typeDefinition1, _typeDefinition2, _typeDefinition3 }).ToArray();
+
+      _validationRuleMock1.Verify();
+      Assert.That(validator.ValidationRules.Count, Is.EqualTo(3));
+      Assert.That(mappingValidationResults.Length, Is.EqualTo(0));
+    }
+
+    [Test]
+    public void ValidateWithSeveralRulesAndClassDefinitions_InvalidResult ()
```
MichaelKetting, line 106: Those two tests don't say 'why' they're once invalid and once value. Is this the same with other validator tests or just here? if it's just here, please fix. otherwise, please point me towards them :)

Remotion/Data/DomainObjects.UnitTests/Mapping/Validation/TypeDefinitionValidatorTest.cs
```diff
@@ -0,0 +1,136 @@
...
+
+      var mappingValidationResults = validator.Validate(new[] { _typeDefinition1 }).ToArray();
+
+      _validationRuleMock1.Verify();
+      Assert.That(validator.ValidationRules.Count, Is.EqualTo(1));
+      Assert.That(mappingValidationResults.Length, Is.EqualTo(0));
+    }
+
+    [Test]
+    public void ValidateWithOneRuleAndClassDefinition_InvalidResult ()
```
MichaelKetting, line 69: Same review comment as with SeveralRules test name

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -20,67 +20,36 @@
...
-          actualDefinition.InstanceCreator,
-          "Instance creator of class definition '{0}' is not the expected object.",
-          actualDefinition.ID);
-
-      if (expectedDefinition.BaseClass == null)
+      if (expectedDefinition is ClassDefinition expectedClassDefinition)
       {
-        Assert.IsNull(actualDefinition.BaseClass, "actualDefinition.BaseClass of class definition '{0}' is not null.", expectedDefinition.ID);
+        if (actualDefinition is not ClassDefinition actualClassDefinition)
+          throw new InvalidOperationException("The two type definitions must have a matching type.");
```
MichaelKetting, line 40: Isn't that an Assert?

Also, sicne you're touching most (all?) of the lines with assert anyway, do you want to switch them over to Assert.That syntax or should we create a ticket for this? They got left behind during the big mirgration  a few years ago.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -20,67 +20,36 @@
...
       else
       {
-        Assert.IsNotNull(actualDefinition.BaseClass, "actualDefinition.BaseClass of class definition '{0}' is null.", expectedDefinition.ID);
-
-        Assert.AreEqual(
-            expectedDefinition.BaseClass.ID,
-            actualDefinition.BaseClass.ID,
-            "BaseClass of class definition '{0}' does not match.",
-            expectedDefinition.ID);
+        throw new InvalidOperationException("Unsupported type definition cannot be checked.");
```
MichaelKetting, line 46: Assert.Fail

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -92,50 +61,60 @@ public void Check (ClassDefinition expectedDefinition, ClassDefinition actualDef
...
       foreach (var expectedDefinition in expectedDefinitions)
       {
-        var actualDefinition = actualDefinitions[expectedDefinition.ClassType];
-        Check(expectedDefinition, actualDefinition);
-        CheckDerivedClasses(expectedDefinition, actualDefinition);
+        var actualDefinition = actualDefinitions[expectedDefinition.Type];
+        if (expectedDefinition is ClassDefinition expectedClassDefinition)
+        {
+          if (actualDefinition is not ClassDefinition actualClassDefinition)
+            throw new InvalidOperationException("The two type definitions must have a matching type.");
```
MichaelKetting, line 68: Assert.that

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -92,50 +61,60 @@ public void Check (ClassDefinition expectedDefinition, ClassDefinition actualDef
...
+
+          CheckClassDefinition(expectedClassDefinition, actualClassDefinition);
+          CheckDerivedClasses(expectedClassDefinition, actualClassDefinition);
+        }
+        else
+        {
+          throw new InvalidOperationException("Unsupported type definition cannot be checked.");
```
MichaelKetting, line 75: assert.fail

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Mapping/IMappingConfiguration.cs
a-ctor: The usages are in `IClassDefinitionProvider` (TypePipe infra) and in the `ObjectIDSerializer`. The methods are asymmetric as searching for a type definition using a class ID does not make much sense.

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
a-ctor: This was done in an upcoming commit together with the change on how read-only behaves.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
a-ctor: That won't work because one of them is a `List<T>`. This could be replaced with a LINQ expression but that feels so unnecessary.

Remotion/Data/DomainObjects/Mapping/IMappingConfiguration.cs
MichaelKetting: Ah, I was only referring to `ContainsClassDefinition (Type classType)`. As far as I can tell, ObjectIDSerializer uses the classID-overload instead of the type.

Okay, In IClassDefinitionProvider we're struck because there's no good way to test for classdefinition via extension methods. Which means the interface has an asymetric api. should we approximate this via an extension method?
```
ContainsClassDefinition(type)
{
  if (!type.IsClass)
    return false;
  if (mapping.ContainsTypeDefinition(type))
    return mapping.GetTypeDefinition(type) is ClassDefinition);
  return false;
}
```

it's not perfect but it does make the api consistent. In theory, we could approximate that a class-type will always be part of classdef and we simplify the check a bit, but that's so...so...

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
MichaelKetting: `allDerivedTypes = allDerivedClasses.ToArray()` then?

a-ctor: It would need to be `allDerivedClasses.Cast<TypeDefinition>().ToArray()` as the type of the list is `ClassDefinition`. And I did not like that which is why I manually wrote the copy loop.

Remotion/Data/DomainObjects.UnitTests/Mapping/Validation/TypeDefinitionValidatorTest.cs
a-ctor: I just renamed the file no idea what the test were supposed to do :p

Looking at the other validator tests, for example `SortExpressionValidatorTest`, they have the same structure.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
a-ctor: I don't think that changing the assert pattern here is worth the effort. If you really want to do it I would suggest changing this to a collection of errors that are collected and then reported. This way, one would not have to re-run a test to see what all of the differences are. It could also be used for the upcoming mapping test infrastructure to report all failures at once, which instead of only the first failure/first object.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
MichaelKetting: hmm... I see.
How about we just change the returntype from arry to IReadOnlyCollection (or IReadOnlyList)? That way, we don't need to copy and don't need to type-change?

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /// <summary>
+    /// Returns all types that are part of this type's hierarchy, including itself, using depth-first.
+    /// </summary>
+    public abstract TypeDefinition[] GetTypeHierarchy ();
+
+    protected abstract void CheckPropertyDefinitions (IEnumerable<PropertyDefinition> propertyDefinitions);
+
+    protected abstract void CheckRelationEndPointDefinitions (IEnumerable<IRelationEndPointDefinition> relationEndPoints);
```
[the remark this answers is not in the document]
a-ctor: I personally like to retain the grouping of properties/methods and have the abstract members at the top of their respective group. But I'll move the methods up to fit our style.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
MichaelKetting: The assert-syntax would just be a concsistency thing. but I don't need to touch it now. So, we just stick with changing the invalidopex to assert.

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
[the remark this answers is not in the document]
MichaelKetting: I've got the same "problem" because I, too, like the grouping. So, we keep the styleguide conform stuff for now but discuss changing the rule (inlcuding pro/con) down the road?

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
a-ctor: I don't think that returning the `List<T>` is better than reallocating the exact amount of space needed and copying. So I would either change it to the LINQ version or keep the copy as is.

MichaelKetting: Am confused. What's the problem with returning an IReadOnlyList<T> (or IReadOnlyCollection<T> if no indexer is needed). We already build a new list during the execution and then it's a readonly api that gets returned. This would actually be a perfect fit from my POV. What am I missing?

Remotion/Data/DomainObjects.UnitTests/Mapping/ClassDefinitionTest.cs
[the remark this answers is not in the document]
a-ctor: I did revert the places where I think that the class ID was superfluous. But it is difficult the find all places where it is not needed as there are some test that required the class ID to work correctly.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
a-ctor: If I understood correctly, you are proposing we change the return type to `IReadOnlyCollection<TypeDefinition>` and simply return the `List<ClassDefinition>`. I dislike returning `List<T>`s as they might occupy a lot more space than necessary. As such, I feel like creating a new array is always preferred.

I don't know what problem we are actually trying to solve with this thread. Originally the question was about `Array.Copy` not being used, which was not used as it could not be used in that scenario. I assume you want to get rid of the `for` loop? If so why?

Remotion/Data/DomainObjects.UnitTests/Mapping/ClassDefinitionTest.cs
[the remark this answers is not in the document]
MichaelKetting: Hmm... I don't remember the details from this review, it's likely just the current file, so we could add the classid fallback as type.Name when classid == null and if that doesn't introduce any new errors, just clean up the changed lines in the current file? Otherwise, we know that we can't simply add a fallback and thus have to keep things as they are now.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
MichaelKetting: Yeah, the for-loop triggered me because it's a lot of code that should have a benefit. (I'm perfectly happy to optimize code in tight loop scnearios but here, it just feels not worth it to write code that uses indexers and counters instead of something declarative)

The change of the return value is in the end is simply the final evolution that resulted from learning the less invloved approaches don't work. However, now that I see the final picture, changing the return value and returning a List<T> as an IReadOnlyCollection is actually the cleanest solution from my perspective because it removes complexity from the code. 

Regarding the memory requirement of List<T>: yes, the list grows with a power of two (thanks for reminding me, I do tend to ignore the excess space). In our scenarios: Typically, there's perhaps 500 class types in the entire domain, so we're not in a problematic territory by far. The list must be allocated anyway, which means we only save memory during the time the result object (as an array instead of a list) is being used at the call site. From a pratical POV, the result will be iterated/processed and then discared. Which means, we most likely generate less memory pressure by not copying the list and we definitely safe some cpu cycles by not copying the array. We could discuss a different approach once we decide to lazy cache the result (which may be a good idea eventually).

MichaelKetting: Historically speaking: the array-result was there because way back when, that was the only way to return a strong typed collection without implementing a collection type youself. So, the switch to IReadOnlyCollection is actually api cleanup to get rid of the nasty arrays.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+    public void GetClassDefinition_WithClassID ()
+    {
+      var classDefinition = ClassDefinitionObjectMother.CreateClassDefinition();
+      var mappingConfigurationMock = new Mock<IMappingConfiguration>(MockBehavior.Strict);
+      mappingConfigurationMock.Setup(e => e.GetClassDefinition("asdNonExistent", It.IsAny<Func<string, Exception>>())).Returns(classDefinition);
+
+      var result = mappingConfigurationMock.Object.GetClassDefinition("asdNonExistent");
+      Assert.That(result, Is.SameAs(classDefinition));
+
+      mappingConfigurationMock.VerifyAll();
```
MichaelKetting, line 158: see above

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -92,50 +61,63 @@ public void Check (ClassDefinition expectedDefinition, ClassDefinition actualDef
...
-        var actualDefinition = actualDefinitions[expectedDefinition.ClassType];
-        Check(expectedDefinition, actualDefinition);
-        CheckDerivedClasses(expectedDefinition, actualDefinition);
+        var actualDefinition = actualDefinitions[expectedDefinition.Type];
+        if (expectedDefinition is ClassDefinition expectedClassDefinition)
+        {
+          if (actualDefinition is not ClassDefinition actualClassDefinition)
+          {
+            Assert.Fail("The two type definitions must have a matching type.");
+            return;
```
MichaelKetting, line 70:
```suggestion
```
Since Assert.Fail throws, a return isn't needed.

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
```diff
@@ -120,61 +90,47 @@ public bool IsSameOrBaseClassOf (ClassDefinition classDefinition)
...
-      FillAllDerivedClasses(allDerivedClasses);
-      return allDerivedClasses.ToArray();
-    }
-
-    public ClassDefinition GetInheritanceRootClass ()
+    public override bool IsAssignableFrom (TypeDefinition other)
     {
-      if (BaseClass != null)
-        return BaseClass.GetInheritanceRootClass();
+      if (ReferenceEquals(this, other))
```
MichaelKetting, line 95: SHould we do this as a basecall instead? the base implementation also does referencequals and we ignore the value implementation here completely. Typically, there should be a reason the base call is omitted and I don't see it here. Thoughts?

MichaelKetting (changes requested):
delete test.diff.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+          () => MappingConfiguration.Current.GetTypeDefinition(typeof(IDummyInterface)),
+          Throws.TypeOf<MappingException>()
+              .With.Message.EqualTo("Mapping does not contain type 'Remotion.Data.DomainObjects.UnitTests.Mapping.MappingConfigurationExtensionsTest+IDummyInterface'."));
+    }
+
+    [Test]
+    public void ContainsClassDefinition ()
+    {
+      var mappingConfigurationMock = new Mock<IMappingConfiguration>(MockBehavior.Strict);
+      mappingConfigurationMock.Setup(e => e.ContainsTypeDefinition(typeof(DummyClass))).Returns(true);
```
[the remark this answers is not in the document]
a-ctor: `.VerifyAll()` does not require `.Verifiable()`

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+              .With.ArgumentExceptionMessageEqualTo(
+                  "The specified type 'Remotion.Data.DomainObjects.UnitTests.Mapping.MappingConfigurationExtensionsTest+IDummyInterface' must be a class.",
+                  "type"));
+    }
+
+    [Test]
+    public void GetClassDefinition_WithCustomExceptionFactory ()
+    {
+      var typeDefinition = TypeDefinitionObjectMother.CreateClassDefinition();
+      var exceptionFactoryMock = new Mock<Func<Type, Exception>>(MockBehavior.Strict);
```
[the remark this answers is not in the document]
a-ctor: In this case I think that the mock makes sense as we want to ensure that nobody is calling the func. Otherwise we would still to return a value in our dummy func thus complicating the test with unnecessary logic

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
```diff
@@ -0,0 +1,170 @@
...
+                  "The specified type 'Remotion.Data.DomainObjects.UnitTests.Mapping.MappingConfigurationExtensionsTest+IDummyInterface' must be a class.",
+                  "type"));
+    }
+
+    [Test]
+    public void GetClassDefinition_WithCustomerExceptionFactoryAndNonExistingClass_ThrowsExceptionFromFactory ()
+    {
+      var exception = new Exception("Test");
+      var exceptionFactoryMock = new Mock<Func<Type, Exception>>(MockBehavior.Strict);
+      exceptionFactoryMock.Setup(e => e(typeof(DummyClass))).Returns(exception);
```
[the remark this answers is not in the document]
a-ctor: But if we throw the exception from our own delegate the test won't fail if the SUT calls the delegate but does not throw the exception.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
a-ctor: It is needed here because `actualClassDefinition` is not defined in the else part

Remotion/Data/DomainObjects/Mapping/ClassDefinition.cs
a-ctor: I think you mixed up `TypeDefinition` and `TypeDefinitionForUnresolvedRelationPropertyType` as the method is abstract and thus not implemented in `TypeDefinition`. I also would not make it virtual as this logic should always be implemented by a subclass.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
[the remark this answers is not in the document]
MichaelKetting: Oh, that's right. Durign the Migration, we chose Verify() isntead of VerifyAll() because we wanted to separate stub vs mock calls. Do you think we should start to mix those two sematics in the codebase, chosing which ever one is the local optimum? 

If we chose to use Verify instead of VerifyAll, we probably should add a roslyn analyzer to prevent missuse. 

If we chose to mix Verify and VerifyAll, we need to be aware of the semantic difference when changing existing tests.

Thoughts?

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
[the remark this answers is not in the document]
MichaelKetting: Hmm... I guess I usually just added an Assert.Fail or a throw InvalidOpEx into the func and that was that.

Just from a code style perspective, I tend to avoid the mock these days if the alernative is equally safe but requires less mocks (and often, also less code).

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingConfigurationExtensionsTest.cs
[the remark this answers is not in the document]
MichaelKetting: > But if we throw the exception from our own delegate the test won't fail if the SUT calls the delegate but does not throw the exception.
That I don't get. We expect that the exception is thrown with Assert.That(...Throws.SameAs(expcetion). So, we have to call the delegate and use the result for the test to work. What am I missing?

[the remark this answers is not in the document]
MichaelKetting: I dug into the styleguide and the topic of stub vs mock did already get a treaty last December and Verify vs VerifyAll did factor into how to migrated from RhinoMocks to Moq. We should incorporate this existing knowledge into the discussion to prevent going in circles.

MichaelKetting (approved):
pleas edon't forget to delete test.diff

PR #319, opened 2022-02-03
Feature/rm 8830 create helper constructs for common safe context no flow scenarios

a-ctor:
https://re-motion.atlassian.net/browse/RM-8330

- [ ] re-create PR with the correct branch name once the changes have been approved


Remotion/Core/Core/Context/SafeContext.ExecutionContext.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Threading;
```
MichaelKetting, line 1: license, check all files

Remotion/Core/Core/Context/SafeContext.Parallel.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// Provides methods for executing <see cref="System.Threading.Tasks.Parallel"/> actions which are <see cref="SafeContext"/> aware.
+    /// <see cref="SafeContext"/> values do not flow into the executed Parallel action.
+    /// </summary>
+    public static class Parallel
+    {
+      /// <summary>
+      /// Opens a <see cref="SafeContextBoundary"/> in which <see cref="System.Threading.Tasks.Parallel"/> actions can be taken without
+      /// flowing <see cref="SafeContext"/> values into the callbacks.
+      /// </summary>
+      /// <example>
```
MichaelKetting, line 18: please check, I think `example` and `code` must be inside `remarks`.

Remotion/Core/Core/Context/SafeContext.Task.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+// We must use an alias as the class name is the same
+using Task_ = System.Threading.Tasks.Task;
```
MichaelKetting, line 5: I think `SystemTask` would be better when reading the code without needing to check where the `_` is coming from. Goes for the other ones, too.

Remotion/Core/UnitTests/Context/SafeContextTest.ExecutionContext.cs
```diff
@@ -0,0 +1,65 @@
...
+      {
+        ResetSafeContextStorageProvider();
+      }
+
+      [Test]
+      public void Run_WithImplicitSafeContextStorageProvider_OpensSafeContextBoundary ()
+      {
+        var safeContextStorageProvider = _safeContextProviderMock.Object;
+        using (SetupImplicitSafeContextStorageProvider(safeContextStorageProvider))
+          RunThreadStartTest((callback, state) => SafeContext.ExecutionContext.Run(ExecutionContext.Capture()!, callback, state));
```
MichaelKetting, line 38: usings already get braces. also created a ticket https://re-motion.atlassian.net/browse/RMTEAM-454

SharedSource/Development/UnitTesting/ThreadRunner.cs
```diff
@@ -56,26 +57,29 @@ public void Run ()
...
+          Thread.ResetAbort();
+          // Do not report exception in lastException, since aborting is expected behavior.
+        }
+        catch (Exception e)
+        {
+          lastException = e;
+        }
+      });
 
+      // TODO RM-8354: Remove ThreadRunner from SharedSource to allow reference to Core.Core
```
MichaelKetting, line 78: Actually, the idea was to manually add a call to the boundary in the usage sites and then update those sites once RM-8354 is done.

Remotion/Core/UnitTests/Context/SafeContextTest.Thread.cs
```diff
@@ -0,0 +1,149 @@
...
+using Moq;
+using NUnit.Framework;
+using Remotion.Context;
+
+namespace Remotion.UnitTests.Context
+{
+  public partial class SafeContextTest
+  {
+    [TestFixture]
+    public class ThreadTest
```
MichaelKetting, line 13: different design for the tests -> not nested types.

most conform choice would be a folder SafeContextTests with individual partial files as separate text fixtures. Please update

MichaelKetting (changes requested):
found mainly strucutral stuff. did not check every single method this time around

SharedSource/Development/UnitTesting/ThreadRunner.cs
a-ctor: I know, I chose to do it this way because it was easier. Some references of `ThreadRunner` have no references to `Core` so they would not be changed, while other need to be changed. Because we already plan to change how this class works in the future this seemed like the easiest choice. Also it is just unit test code

Remotion/Core/UnitTests/Context/SafeContextTest.Thread.cs
a-ctor: What is the problem with nested types in this scenario? We use nested types in the implementation and the tests should mirror that structure in my opinion.

MichaelKetting: It's an example where we want Testcase Class per Feature (http://xunitpatterns.com/Testcase%20Class%20per%20Feature.html) because each nested type of the SUT represents a feature. We never use nested test fixtures and it would introduce a new concept and thus new complexity and thus reduce maintainability of the tests. 

The idea when structuring the tests is to make them a) findable and b) allow shared setups, test data, etc. By using a nested test fixture, in particular for a case where there's tests on both the outer type and the nested type, makes the tests much harder to read and understand their interdependencies and possible sideeffects. That's why separate test fixtures in a shared folder/namespace is the pattern used for this case.

SharedSource/Development/UnitTesting/ThreadRunner.cs
MichaelKetting: Am sorry, am not convinced. 

"just unit test code" isn't a good argument. While it's true that there's some tweaks and tricks you need to do when writing a unit test in an imperfect world (e.g. doing a private call), those are done because the API of the SUT takes precedence over the unit test and not everything can easily be refactored to be both sensibly constrained and have perfect testability.

The other thing is that this hides the dependency and the effect in a really horrible way.

=> I've looked at the other repositories and it looks like we don't have any external dependencies. If you do want to keep it here, please move ThreadRunner now (i.e. in this branch or - better - a separate branch that's happening before this one)

MichaelKetting:
Oh, I noticed the branch name is wrong. this should be RM-8330

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Thread.cs
```diff
@@ -0,0 +1,144 @@
...
+  public class SafeContextThreadTest : SafeContextTestBase
+  {
+    private const int c_maxStackSize = 4096;
+
+    [Test]
+    public void New_WithCallbackAndImplicitSafeContextStorageProvider_OpensSafeContextBoundary ()
+    {
+      var safeContextStorageProvider = _safeContextProviderMock.Object;
+      using (SetupImplicitSafeContextStorageProvider(safeContextStorageProvider))
+        RunThreadStartTest(SafeContext.Thread.New);
```
MichaelKetting, line 35: braces

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Parallel.cs
```diff
@@ -0,0 +1,45 @@
...
+  public class SafeContextParallelTest : SafeContextTestBase
+  {
+    [Test]
+    public void Run_WithImplicitSafeContextStorageProvider_OpensSafeContextBoundary ()
+    {
+      var safeContextStorageProvider = _safeContextProviderMock.Object;
+      using (SetupImplicitSafeContextStorageProvider(safeContextStorageProvider))
+      {
+        using (SafeContext.Parallel.OpenSafeContextBoundary())
+          _safeContextProviderMock.Verify(e => e.OpenSafeContextBoundary());
```
MichaelKetting, line 33: braces

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.ExecutionContext.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+  [TestFixture]
+  public class SafeContextExecutionContextTest : SafeContextTestBase
+  {
+    [Test]
+    public void Run_WithImplicitSafeContextStorageProvider_OpensSafeContextBoundary ()
+    {
+      var safeContextStorageProvider = _safeContextProviderMock.Object;
+      using (SetupImplicitSafeContextStorageProvider(safeContextStorageProvider))
+        RunThreadStartTest((callback, state) => SafeContext.ExecutionContext.Run(ExecutionContext.Capture()!, callback, state));
```
MichaelKetting, line 32: braces

MichaelKetting, line 32: unit tests don't have NRT enabled so why the null-forgiving?

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.ExecutionContext.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    [Test]
+    public void Run_WithExplicitSafeContextStorageProvider_OpensSafeContextBoundary ()
+    {
+      var safeContextStorageProvider = _safeContextProviderMock.Object;
+      RunThreadStartTest((callback, state) => SafeContext.ExecutionContext.Run(safeContextStorageProvider, ExecutionContext.Capture()!, callback, state));
+    }
+
+    private void RunThreadStartTest (Action<ContextCallback, object> executionContextRun)
```
MichaelKetting, line 42: method name incorrect?

Remotion/Core/Core/Context/SafeContext.ExecutionContext.cs
```diff
@@ -0,0 +1,49 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System.Threading;
+// We must use an alias as the class name is the same
+using ExecutionContext_= System.Threading.ExecutionContext;
```
MichaelKetting, line 19: Whitespace missing

Remotion/Core/Core/Context/SafeContext.Task.cs
```diff
@@ -0,0 +1,126 @@
...
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+// We must use an alias as the class name is the same
+using Task_ = System.Threading.Tasks.Task;
```
MichaelKetting, line 21: I gather you went with the comment? My reasoning was that while reading the code, I first look at `Task_` and am confused what it means. => only after I look around and find the using that things become clearer. The main reasoning it that `Task_` on it's own doesn't give me much context about what's going on plus the trailing underscore isn't something that's part of our style conventions so there's the extra hurdle to figure out the syntax and meaning. 

That's why I want to give Task and ExecutionContext a meaningful prefix (e.g. `SystemTask` or `System_Task`), making the lines much more self explainatiory.

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.cs
```diff
@@ -0,0 +1,332 @@
...
+      // ReSharper disable once MethodSupportsCancellation
+      RunTestAction((action, _) => SafeContext.Task.Run(safeContextStorageProvider, action), false);
+    }
+
+    [Test]
+    public void Run_ActionWithCancellationAndImplicitSafeContextStorageProvider_OpensSafeContextBoundary ()
+    {
+      var safeContextStorageProvider = _safeContextProviderMock.Object;
+      using (SetupImplicitSafeContextStorageProvider(safeContextStorageProvider))
+        RunTestAction(SafeContext.Task.Run, true);
```
MichaelKetting, line 89: braces.

MichaelKetting (changes requested):

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.ExecutionContext.cs
a-ctor: ReSharper/Rider still emits a warning for a possible null assignment. Do you want to remove it anyways?

Remotion/Core/Core/Context/SafeContext.Task.cs
a-ctor: Some of my changes did not make it into the review commits because I did change that but well here we go again.

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.ExecutionContext.cs
MichaelKetting: hmm... open a discussion on this in MSTeams?

PR #325, opened 2022-02-07
RM-8347: Clear old page token session storage entries

iam-mholle:


Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -499,14 +503,27 @@ class SmartPage_Context
...
   public SetCacheDetectionFieldSubmitted(): void
   {
     window.sessionStorage.setItem(this.GetPageToken(), SmartPage_CacheDetection[SmartPage_CacheDetection.HasSubmitted]);
+    this.ClearOldPageTokenSessionStorageEntries();
   };
 
+  private ClearOldPageTokenSessionStorageEntries(): void
+  {
+    const pageTokenEntries = Object.keys(window.sessionStorage).map((k: string) => +k).filter(t => !isNaN(t));
+    const pageTokenEntryCount = pageTokenEntries.length
```
MichaelKetting, line 519: semi-colon. I'm surprised TS allows us this inconsistency.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -499,14 +503,27 @@ class SmartPage_Context
...
+    if (pageTokenEntryCount > this._pageTokenDeletionThreshold)
+    {
+      const pageTokensToDelete = pageTokenEntries.sort().splice(0, pageTokenEntryCount - this._pageTokenDeletionKeepCount);
+      setTimeout(() => pageTokensToDelete.forEach(t => window.sessionStorage.removeItem(t.toString())), this._pageTokenDeletionDelay);
```
MichaelKetting, line 523: is the timeout of > 0ms meant to actually delay the cleanup or simply to make it asynchronously. In the second case, it's enough to hardcode '0'.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -499,14 +503,27 @@ class SmartPage_Context
...
   // Marks the page as submitted.
   public SetCacheDetectionFieldSubmitted(): void
   {
     window.sessionStorage.setItem(this.GetPageToken(), SmartPage_CacheDetection[SmartPage_CacheDetection.HasSubmitted]);
+    this.ClearOldPageTokenSessionStorageEntries();
   };
 
+  private ClearOldPageTokenSessionStorageEntries(): void
+  {
+    const pageTokenEntries = Object.keys(window.sessionStorage).map((k: string) => +k).filter(t => !isNaN(t));
```
MichaelKetting, line 518: what's the `+k` syntax here?

And I think we should add a prefix, not just use a number, now that we a) limit the count and b) perform active cleanup. `pt_` or `PageToken_`

also, for an extra performance boost: we could check if the total number exceeds the keepcount before checking if that number is acutally pagetokens. until such a time as the page uses many own values in the session store, the first check will be very efficient.

As an alternative, we could (should?) also persist the actual count of pagetokens in the sessionstore. that way, we can do a quick lookup before parsing all the keys. Should there ever be another heavy user, this check would prove beneficial to the performance impact.

MichaelKetting (changes requested):
some performance and conflict-maintainability requests

MichaelKetting (changes requested):
Can we create web tests for this? If so, please add a follow-up task in the backlog.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -498,15 +504,60 @@ class SmartPage_Context
...
 
+  private SetCacheDetectionField(cacheDetectionValue: SmartPage_CacheDetection): void
+  {
+    const tokenKey = this._pageTokenSessionStoragePrefix + this.GetPageToken();
+
+    if (window.sessionStorage.getItem(tokenKey) === null)
+      this.IncreasePageTokenSessionStorageEntryCount();
+
+    window.sessionStorage.setItem(tokenKey, SmartPage_CacheDetection[cacheDetectionValue]);
+    this.ClearOldPageTokenSessionStorageEntries();
```
MichaelKetting, line 524:
```suggestion
    if (window.sessionStorage.getItem(tokenKey) === null)
    {
      this.IncreasePageTokenSessionStorageEntryCount();
      this.ClearOldPageTokenSessionStorageEntries();
    }

    window.sessionStorage.setItem(tokenKey, SmartPage_CacheDetection[cacheDetectionValue]);
```

Just for consideration - if we don't add something, we don't need to perform any cleanup. I know the count already optimizes things, but it may also make things more expressive this way. Thoughts?

MichaelKetting (approved):
Look okay now to me but I didn't try the updated cleanup logic. 

Added a code optimization suggstion, in case you agree.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
iam-mholle: I thought about that but somehow it felt wrong to increase the counter and potentially trigger a cleanup without the item actually being added at this point, that's why I chose the existing path (and order of execution). What do you think about this?

MichaelKetting: also considered the same reasoning. Landed on the other side because a) it's just 1 item in a list of 100, so the off by one "error" is not relevant for what we're trying to achieve. In the end, for me it came down to reading this method and thinking, that if I just do an update, not an insert, doing "always" a cleanup (without considereing the also implemented performance improvement) felt less optimized in this method.

But it really is a matter of preference while trying to find the local code optimum until the one extra lookup of the counter becomes a performance bottleneck (so, never?)

PR #327, opened 2022-02-08
RM-8358 Introduce keyboard-based scrolling for DropDownMenu

a-ctor:
https://re-motion.atlassian.net/browse/RM-8358

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/typings/Utilities.d.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/typings/Utilities.d.ts
@@ -75,3 +75,11 @@ declare class LayoutUtility {
     static Show(element: HTMLElement): void;
     static FormatPixelProperty(value: string | number): string
 }
+/**
+ * Contains all the CSS class definitions needed throughout UI control rendering.
+ * Changes should also be reflected in CssClassDefinition.cs.
+ */
+declare class CssClassDefinition {
+    static Scrollable: string;
+    static Themed: string;
+}
--- a/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
@@ -102,6 +102,8 @@ class DropDownMenu
   private static _statusPopupRepositionTimer: Nullable<number> = null;
   private static _blurTimer: Nullable<number> = null;
   private static _updateFocus: boolean = true;
+  private static _ignoreHoverMouseEvents: boolean = false;
+  private static _lastPositionInformation: Nullable<string> = null;
 
   public static AddMenuInfo(menuInfo: DropDownMenu_MenuInfo): void
   {
@@ -341,7 +343,7 @@ class DropDownMenu
     DropDownMenu._currentStatusPopup = statusPopup;
     document.getElementById (menuID)!.closest ('div, td, th, body')!.append (statusPopup);
 
-    DropDownMenu.ApplyPosition (statusPopup, evt, titleDivFunc());
+    DropDownMenu.ApplyPosition (statusPopup, evt, titleDivFunc(), false);
 
     if (DropDownMenu._statusPopupRepositionTimer)
       clearTimeout(DropDownMenu._statusPopupRepositionTimer);
@@ -352,7 +354,7 @@ class DropDownMenu
 
       if (DropDownMenu._currentStatusPopup && DropDownMenu._currentStatusPopup === statusPopup)
       {
-        DropDownMenu.ApplyPosition (statusPopup, null, titleDivFunc());
+        DropDownMenu.ApplyPosition (statusPopup, null, titleDivFunc(), true);
         DropDownMenu._statusPopupRepositionTimer = setTimeout(repositionHandler, DropDownMenu._repositionInterval);
       }
     };
@@ -388,7 +390,8 @@ class DropDownMenu
     menuButtonAnchor.setAttribute ('aria-expanded', 'true');
 
     var div = document.createElement('div');
-    div.className = 'DropDownMenuOptions';
+    div.classList.add('DropDownMenuOptions');
+    div.classList.add(CssClassDefinition.Themed);
     div.id = menuOptionsID;
     div.setAttribute ('role', 'menu');
     div.setAttribute ('tabindex', '-1');
@@ -396,6 +399,9 @@ class DropDownMenu
       div.setAttribute ('aria-labelledby', menuButtonAnchor.getAttribute ('aria-labelledby')!);
     else
       div.setAttribute ('aria-labelledby', menuButtonAnchor.id);
+    div.addEventListener("focus", () => this.FocusHandler());
+    div.addEventListener("blur", () => this.BlurHandler());
+
     DropDownMenu._currentPopup = div;
 
     var ul = document.createElement('ul');
@@ -412,6 +418,9 @@ class DropDownMenu
     });
     ul.addEventListener ('mouseover', function (event)
     {
+      if (DropDownMenu._ignoreHoverMouseEvents)
+        return;
+
       var eventTarget = DropDownMenu.GetTarget (event, "LI");
 
       ul.querySelectorAll ("li").forEach (b => b.classList.remove (DropDownMenu._itemSelectedClassName));
@@ -439,7 +448,7 @@ class DropDownMenu
     }
 
     var titleDivFunc = DropDownMenu.CreateTitleDivGetter (context);
-    DropDownMenu.ApplyPosition (div, evt, titleDivFunc());
+    DropDownMenu.ApplyPosition (div, evt, titleDivFunc(), false);
 
     if (DropDownMenu._repositionTimer) 
       clearTimeout(DropDownMenu._repositionTimer);
@@ -450,7 +459,7 @@ class DropDownMenu
 
       if (DropDownMenu._currentPopup && DropDownMenu._currentPopup == div && LayoutUtility.IsVisible (div))
       {
-        DropDownMenu.ApplyPosition (div, null, titleDivFunc());
+        DropDownMenu.ApplyPosition (div, null, titleDivFunc(), true);
         DropDownMenu._repositionTimer = setTimeout (repositionHandler, DropDownMenu._repositionInterval);
       }
     };
@@ -489,17 +498,28 @@ class DropDownMenu
     }
   }
 
-  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement): void
+  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement, onlyUpdateIfNecessary: boolean): void
   {
     var space_top = Math.round (LayoutUtility.GetOffset (referenceElement).top - window.pageYOffset);
     var space_bottom = Math.round (document.documentElement.clientHeight - LayoutUtility.GetOffset (referenceElement).top - LayoutUtility.GetHeight (referenceElement) + window.pageYOffset);
     var space_left = LayoutUtility.GetOffset (referenceElement).left;
     var space_right = document.documentElement.clientWidth - LayoutUtility.GetOffset (referenceElement).left - LayoutUtility.GetWidth (referenceElement);
+    
+    var positionString = `${space_top};${space_right};${space_bottom};${space_left};`;
+    if (onlyUpdateIfNecessary && positionString === this._lastPositionInformation)
+      return;
+
+    this._lastPositionInformation = positionString;
 
     // position drop-down list
     var top = clickEvent ? clickEvent.clientY : Math.max (0, space_top + referenceElement.offsetHeight);
     var left = clickEvent ? clickEvent.clientX + 'px' : 'auto';
     var right = clickEvent ? 'auto' : Math.max (0, document.documentElement.clientWidth - LayoutUtility.GetOffset (referenceElement).left - referenceElement.offsetWidth) + 'px';
+
+    const verticalSpacing = parseInt(getComputedStyle(popUpDiv).fontSize) * 2;
+
+    // Save and restore the scrollTop value to retain the scrolling after resizing
+    var scrollTop = popUpDiv.scrollTop;
 
     popUpDiv.style.top = top + 'px';
     popUpDiv.style.bottom = 'auto';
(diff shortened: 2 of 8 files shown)


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -102,6 +102,8 @@ class DropDownMenu
   private static _statusPopupRepositionTimer: Nullable<number> = null;
   private static _blurTimer: Nullable<number> = null;
   private static _updateFocus: boolean = true;
+  private static _ignoreMouseEvents: boolean = false;
```
iam-mholle, line 105: `_ignoreMouseEvents` might not be the best name, what about `_ignoreMouseHoverEvents` or something of the sort?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -489,18 +515,35 @@ class DropDownMenu
...
-  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement): void
+  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement, isPeriodic: boolean): void
   {
     var space_top = Math.round (LayoutUtility.GetOffset (referenceElement).top - window.pageYOffset);
     var space_bottom = Math.round (document.documentElement.clientHeight - LayoutUtility.GetOffset (referenceElement).top - LayoutUtility.GetHeight (referenceElement) + window.pageYOffset);
     var space_left = LayoutUtility.GetOffset (referenceElement).left;
     var space_right = document.documentElement.clientWidth - LayoutUtility.GetOffset (referenceElement).left - LayoutUtility.GetWidth (referenceElement);
 
+    // Periodic update: only update our position if it has change to allow the usage of the scrollbars
+    // Other update: it might take a periodic call to get the correct position so we force it by deleting the last position information
```
iam-mholle, line 526: Might be better off in a JSDoc comment to get IDE support

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -865,6 +927,11 @@ class DropDownMenu
           anchor.focus();
         }
       }
+
+      // A selection change might trigger hover events that, in turn, would trigger selection events
+      // So we ignore mouse events for a small amount of time to deal with cascading events triggered by the selection change
+      DropDownMenu._ignoreMouseEvents = true;
+      setTimeout(() => DropDownMenu._ignoreMouseEvents = false, 250);
```
iam-mholle, line 934: This is working great

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -489,18 +515,35 @@ class DropDownMenu
...
+    // Periodic update: only update our position if it has change to allow the usage of the scrollbars
+    // Other update: it might take a periodic call to get the correct position so we force it by deleting the last position information
+    if (isPeriodic) {
+      var positionString = `${space_top};${space_right};${space_bottom};${space_left};`;
+      if (positionString === this._lastPositionInformation)
+        return;
+
+      this._lastPositionInformation = positionString;
+    } else {
+      this._lastPositionInformation = null
```
iam-mholle, line 534: Semicolon missing

iam-mholle (approved):

MichaelKetting:
Scrolling in NovaGray doesn't behave correctly in certain edge cases in Google Chrome (window height and popup height are almost identical):
1. scrollbar is displayed but no actuall scrolling happens when the height is sufficient
image
2. popup shrinks when scrolling the page via mouse wheel.
image

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: The same property is used in the handlers for `mouseover`, `mouseout`, and `mouseleave` so adding a `Hover` seems wrong to me.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Its a private method and this information is only meant to explain the reasoning behind the implementation. But it might be useful to explain the periodic parameter in a JSDoc comment

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -489,18 +515,35 @@ class DropDownMenu
...
+  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement, isPeriodic: boolean): void
   {
     var space_top = Math.round (LayoutUtility.GetOffset (referenceElement).top - window.pageYOffset);
     var space_bottom = Math.round (document.documentElement.clientHeight - LayoutUtility.GetOffset (referenceElement).top - LayoutUtility.GetHeight (referenceElement) + window.pageYOffset);
     var space_left = LayoutUtility.GetOffset (referenceElement).left;
     var space_right = document.documentElement.clientWidth - LayoutUtility.GetOffset (referenceElement).left - LayoutUtility.GetWidth (referenceElement);
 
+    // Periodic update: only update our position if it has change to allow the usage of the scrollbars
+    // Other update: it might take a periodic call to get the correct position so we force it by deleting the last position information
+    if (isPeriodic) {
```
MichaelKetting, line 527: consistency: braces in new lines

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -489,18 +515,35 @@ class DropDownMenu
...
+    } else {
+      this._lastPositionInformation = null
+    }
+
     // position drop-down list
     var top = clickEvent ? clickEvent.clientY : Math.max (0, space_top + referenceElement.offsetHeight);
     var left = clickEvent ? clickEvent.clientX + 'px' : 'auto';
     var right = clickEvent ? 'auto' : Math.max (0, document.documentElement.clientWidth - LayoutUtility.GetOffset (referenceElement).left - referenceElement.offsetWidth) + 'px';
 
+    const verticalSpacing = parseInt(getComputedStyle(popUpDiv).fontSize) * 2;
```
MichaelKetting, line 542: I would prefer to use the height of the first menu item (no factor), not simply the font size. if we also want to support line-wrapped items, we could take the minimum value for all menu items. That way, we have a distance that's more closely aligned with the layout.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -517,25 +560,44 @@ class DropDownMenu
...
+        popUpDiv.classList.add("scrollable");
+
+        // Ensure that the popup retains the minimum size, reducing the spacing if it gets to small;
+        // This prevents the popup from disappearing completely
+        const minimumPopUpHeight = verticalSpacing * 6;
+        const targetPopUpHeight = Math.max(minimumPopUpHeight, maximumPopUpHeight);
+        const remainingSpacing = Math.floor(document.documentElement.clientHeight - verticalSpacing - targetPopUpHeight);
+        const spacing = Math.max(0, remainingSpacing);
+
+        popUpDiv.style.top = spacing + 'px';
```
MichaelKetting, line 579: naming: here be basically have two types of vertical spacing. One that's the baseline for the calculation (`verticalSpacing`) and one that's getting applied (`spacing`). We need to find better names for these, probably starting with renaming `verticalSpacing` to give it more semantic

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting, line 934: can we shorten this to 50-100ms? I.e. do we just need to deblock the cpu or do we need to deblock the human's jittery hand?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -352,7 +369,7 @@ class DropDownMenu
 
       if (DropDownMenu._currentStatusPopup && DropDownMenu._currentStatusPopup === statusPopup)
       {
-        DropDownMenu.ApplyPosition (statusPopup, null, titleDivFunc());
+        DropDownMenu.ApplyPosition (statusPopup, null, titleDivFunc(), false);
```
MichaelKetting, line 372: why is this one not periodic?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -489,18 +515,35 @@ class DropDownMenu
     }
   }
 
-  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement): void
+  private static ApplyPosition (popUpDiv: HTMLDivElement, clickEvent: Nullable<MouseEvent>, referenceElement: HTMLElement, isPeriodic: boolean): void
```
MichaelKetting, line 518: I think we should consider a different name here. From the perspective of `ApplyPosition`, it's not relevant if the call happens from a periodic operation or not. Instead, the relevant information is which kind of behavior we want - 'forced update' or 'conditional update' (we need a good name for the behavior/concept).

Additionally: We could also introduce two methods that are used in the code base and then delegate to this method with the (appropriately named) boolean parameter to make the callsites more expressive. Of course, that's not feasible for more complex overload-scenarios, but here, right now, it would work.

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Aren't those "the hover events"? Anyway, e.g. `click` is not disabled, which seemed weird to me.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: It is private, its still useful to see param docs tho when editing a 1.3k lines file 😅

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
iam-mholle: Might be a good idea to use ESLint, what do you think?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -151,8 +153,23 @@ class DropDownMenu
...
-      allButMenuButton.forEach (b => b.addEventListener ("mouseover", function () { openTarget.querySelectorAll ('.' + DropDownMenu._buttonClassName).forEach (b => b.classList.add (DropDownMenu._nestedHoverClassName)); }));
-      allButMenuButton.forEach (b => b.addEventListener ("mouseout", function () { openTarget.querySelectorAll ('.' + DropDownMenu._buttonClassName).forEach (b => b.classList.remove (DropDownMenu._nestedHoverClassName)); }));
+      const mouseOverHandler = function () { 
+        if (DropDownMenu._ignoreMouseEvents)
+          return;
+
+          openTarget.querySelectorAll ('.' + DropDownMenu._buttonClassName)
+            .forEach (b => b.classList.add (DropDownMenu._nestedHoverClassName));
+      };
+      allButMenuButton.forEach (b => b.addEventListener ("mouseover", mouseOverHandler));
```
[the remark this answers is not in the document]
a-ctor: I added them to be safe, as they do no harm and also I could not reproduce why this code is even there. `menuButton` is the span element that represents the button, and it is queried for children which are stored in `allButMenuItems`. It has no items so the list is always empty....

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: I don't think that that is a good idea as the size of the first element might change depending on the layout of popup which might cause flickering between states. I don't think that it is necessary here to have a spacing that is most closely aligned to the rest of the layout.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Firefox takes its time to report the relevant events that we want to block. A "high" timeout is not a problem here as we are only blocking hover effects, which means that the only bad thing would be that a user who does a keyboard selection cannot select something with the mouse in the next 250ms.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Its just 990 loc 😌

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor is that workable with our script files? After all, the files are much less conssitent (for various reasons) than in c#.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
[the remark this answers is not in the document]
MichaelKetting: /squints/ Please check the query. This code applies to the drop down menu button, not the menu items in the popup.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: hmm...yes, preventing this flicker on edge cases is important and would make things more complicated. Okay, we keep fontsize, but please add a comment about why we don't use itemheight to the source code.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: please make this a cosntnat and add a comment regarding Firefox. And yes, users that might immeditately do a keyboard press and fall into the e.g. < 250ms is a real scenario in QA. (we have had some fun bug reports form this)

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: yes, but in particular the braces: we have different styles in different files and different styles within some of the files. So, I'm a bit worried that enforcing brace-layout and whitespaces will mean we need to do a ton manual cleanup? Am not against enforcing these things, am just saying, it may be more involved than it was in c#

iam-mholle: ESLint can be used to automatically fix found issues too, so there would not be any manual cleanup needed - just a review I guess.

MichaelKetting (changes requested):
Sorry, found a bug while testing:

Open with Scrollbar, click on scrollbar, press cursor up/down and the entire popup becomes the focus target isntead of the menu item. Then, you can't get into the popup to menu items via keyboard. 

The expected behavior is that a click on the scrollbar followed by a cursor press will move the menu item selection.

Remotion/ObjectBinding/Web.ClientScript/Scripts/typings/Utilities.d.ts
```diff
@@ -75,3 +75,10 @@ declare class LayoutUtility {
     static Show(element: HTMLElement): void;
     static FormatPixelProperty(value: string | number): string
 }
+/**
+ * Contains all the CSS class definitions needed throughout UI control rendering.
+ * Changes should also be reflected in CssClassDefinition.cs.
+ */
+declare class CssClassDefinition {
```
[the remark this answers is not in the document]
a-ctor: In this case the comment is formatted this way to provide JSdoc information similarly to how `///` works in C#. So I think we should keep it this way for comments that act as doc on methods and classes

a-ctor:
Scrollability in the context menu of the tree view is a separate issue as the layout problems are not caused by this changeset: https://re-motion.atlassian.net/browse/RM-8376

Remotion/ObjectBinding/Web.ClientScript/Scripts/typings/Utilities.d.ts
```diff
@@ -75,3 +75,11 @@ declare class LayoutUtility {
     static Show(element: HTMLElement): void;
     static FormatPixelProperty(value: string | number): string
 }
+/**
```
MichaelKetting, line 78:
```suggestion
}

/**
```

Remotion/Web/Core/UI/Controls/CssClassDefinition.cs
```diff
@@ -18,6 +18,7 @@ namespace Remotion.Web.UI.Controls
 {
   /// <summary>
   /// Contains all the CSS class definitions needed throughout UI control rendering.
+  /// Changes should also be reflected in Utilities.ts class CssClassDefinition.
```
MichaelKetting, line 21:
```suggestion
```
Not in summary

Remotion/Web/Core/UI/Controls/CssClassDefinition.cs
```diff
@@ -41,6 +42,18 @@ public static string Disabled
...
+    /// <summary>
+    /// Gets the CSS-Class applied to an UI control when it should become scrollable.
+    /// </summary>
+    /// <remarks> 
+    ///   <para> Class: <c>scrollable</c>. </para>
+    ///   <para> Applied in addition to the regular CSS-Class.</para>
+    /// </remarks>
+    public static string Scrollable
+    {
+      get { return /* also change in Utilities.ts class CssClassDefinition */ "scrollable"; }
```
MichaelKetting, line 54:
```suggestion
      get { return "scrollable"; } // also change in Utilities.ts class CssClassDefinition
```

Remotion/Web/Core/UI/Controls/CssClassDefinition.cs
```diff
@@ -51,7 +64,7 @@ public static string Disabled
     /// </remarks>
     public static string Themed
     {
-      get { return "remotion-themed"; }
+      get { return /* also change in Utilities.ts class CssClassDefinition */ "remotion-themed"; }
```
MichaelKetting, line 67:
```suggestion
      get { return "remotion-themed"; } // also change in Utilities.ts class CssClassDefinition
```

MichaelKetting (approved):
comment-changes. Please check formatting etc and then we're done.

Remotion/ObjectBinding/Web.ClientScript/Scripts/typings/Utilities.d.ts
a-ctor: This file is auto-generated so I would not adjust formatting. This file is also just a "temporary" fix until we fixed the typescript project references between web and objectbinding.

PR #330, opened 2022-02-09
RM-8286: Add BocAutoComplete reset API via JS

iam-mholle:


iam-mholle:
@MichaelKetting I'm not sure if the methods should be called `reset` or `clear`, what do you think?

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocAutoCompleteReferenceValueControlObjectTest.cs
```diff
@@ -557,6 +557,21 @@ public void TestDropDownItemDisabledWithRequiredSelection ()
...
+      var id = bocAutoComplete.GetHtmlID();
+
+      JavaScriptExecutor.GetJavaScriptExecutor(bocAutoComplete).ExecuteScript($"document.getElementById('{id}_TextValue').reset()");
+
+      var text = bocAutoComplete.GetText();
+      var hiddenValue = home.Scope.FindCss("#body_DataEditControl_PartnerField_Normal_KeyValue").Value;
+      Assert.That(text, Is.Empty);
+      Assert.That(hiddenValue, Is.EqualTo("==null=="));
+    }
+
```
MichaelKetting, line 574: please add separate tests for autopostback and no-autopostback configurations. For the asssertions on autopostback, I think it would be best if we had the same semantics as with TestFillWith. Not sure about the no-autopostback scenario.

MichaelKetting (changes requested):

PR #332, opened 2022-02-11
RM-8349: Add web tests to cover history back handling

iam-mholle:


Remotion/Web/IntegrationTests/SmartPage/HistoryBackHandlingTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+      home.Driver.GoBack(home.Scope);
+
+      _ = home.Scope.OuterHTML; // Ensure up to date html
+      var overlayScope = home.Scope.FindCss("#WxeStatusIsCachedMessage");
+      Assert.That(overlayScope.Exists(), Is.True);
+    }
+
+    [Test]
+    public void NavigateForward_WithOverlayShown_IsNotPossible ()
```
MichaelKetting, line 128:
```suggestion
    public void NavigateForward_IsNotPossible ()
```

Unless I'm mistaken/blind, there's not test for the overlay or special operation, it's just load + back + forward, so we should use a more simple test name.

Remotion/Web/IntegrationTests/SmartPage/HistoryBackHandlingTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+      Assert.That(GetSessionStorageEntryCount(jsExecutor), Is.EqualTo(151));
+      Assert.That(GetSessionStorageEntry(jsExecutor, "pageTokenEntryCount"), Is.EqualTo(150.ToString()));
+
+      Start<WxePageObject>("WebTabStripTest.wxe");
+
+      var sessionStorageEntryCount = GetSessionStorageEntryCount(jsExecutor);
+      var sessionStorageEntryKeyValue = GetSessionStorageEntry(jsExecutor, "pageTokenEntryCount");
+      Assert.That(sessionStorageEntryCount, Is.EqualTo(101));
+      Assert.That(sessionStorageEntryKeyValue, Is.EqualTo((sessionStorageEntryCount - 1).ToString()));
```
MichaelKetting, line 97: For a complete test, we should always include a check that the added value in the sessionstate is actually correct. For the Threshold-test, test that the most recent entry is still there, plus probably the 3-from-last or so, to document the correct sorting order.

The other tests are probably already covered via the status-message tests, but the Threshold-test isn't. It would be okay to just check for expected page token values (read from the hidden field after loading the respective page load and then used for the assertion.

MichaelKetting (changes requested):

Remotion/Web/IntegrationTests/SmartPage/HistoryBackHandlingTests.cs
iam-mholle: Please have a look at [`4400fc9` (#332)](https://github.com/re-motion/Framework/pull/332/commits/4400fc9ca662f30c0a132d8f2e9ab53e29ad423a).

MichaelKetting (approved):

PR #335, opened 2022-02-15
feature/RM-8036-Remove-InternetExplorer-Artifacts-From-WebTests

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-8036


Build/Customizations/Server/Extensions.Server.targets
```diff
@@ -70,7 +70,6 @@
   <Target Name="Server_CIBuild_Chrome" DependsOnTargets="TestBuild_ChromeConfiguration;CheckProperties;SetVersionCiAndNightlyBuild;TestBuild" />
```
iam-mholle, line 70: I found the following code elements that should be removed too:
- `Remotion.Web.ExecutionEngine.WxePageInfo.c_styleFileUrlForIE`
- `Remotion.Web.ExecutionEngine.WxePageInfo.s_styleFileKeyForIE`

Remotion/Web/Development.WebTesting/Utilities/DriverExtensions.cs
```diff
@@ -48,7 +48,6 @@ public static string GetBrowserName ([NotNull] this IDriver driver)
           ChromeDriver _ => "Chrome",
           EdgeDriver _ => "Edge",
           FirefoxDriver _ => "Firefox",
-          InternetExplorerDriver _ => "InternetExplorer",
```
iam-mholle, line 51: Using directive still present

Build/Customizations/Server/Extensions.Server.targets
iam-mholle, line 70: `Remotion.Web.Development.WebTesting.CoypuElementScopeExtensions.IsTransparent` still has an IE workaround, TBD if it really only affects IE.

Remotion/Web/Development.WebTesting/WebTestConfigurationFactory.cs
```diff
@@ -122,7 +122,7 @@ public IHostingConfiguration CreateHostingConfiguration ()
     }
 
     /// <summary>
-    /// Responsible for handling Browsers other than Chrome and Internet Explorer.
+    /// Responsible for handling Browsers other than Chrome, Edge, and Firefox.
```
iam-mholle, line 125:
```suggestion
    /// Responsible for handling Browsers other than Chrome, Edge and Firefox.
```

iam-mholle (changes requested):

Build/Customizations/Server/Extensions.Server.targets
MichaelKetting: Those are out of scope for the current task (webtests only)

Remotion/Web/Development.WebTesting/WebTestConfigurationFactory.cs
MichaelKetting: Nope, starting with three options, English requires a comma before the `and`

Build/Customizations/Server/Extensions.Server.targets
MichaelKetting: Good point. they would be equivalent. I'll update the docs but leave it in.

Remotion/Web/Development.WebTesting/WebTestConfigurationFactory.cs
iam-mholle: I didn't know the oxford comma is now mandatory in the english language. imo it just looks weird. Anyway, if it's required 🤷🏼‍♂️

MichaelKetting: Interesting, didn't know it was only for them fancy talkers required. Anyhow, I think for me, it dates back to the prep lessons for my own masters thesis and I don't want to dig up whether it's actually required or just good practice for formal text. Either way, it's official enough to keep it when it's there, but may be okay to not mention it when it's missing during a review.

iam-mholle (approved):

PR #344, opened 2022-02-23
RM-8362: Screenshot support for scrollable dropdownmenus

iam-mholle:


Remotion/Web/Development.WebTesting/ScreenshotCreation/Fluent/FluentScreenshotElement.cs
```diff
@@ -55,6 +58,19 @@ public class FluentScreenshotElement<T> : IFluentScreenshotElement<T>, IFluentSc
       _minimumElementVisibility = minimumElementVisibility;
     }
 
+    public FluentScreenshotElement<T> ScrollIntoView ()
```
MichaelKetting, line 61: Hmm... Does it work if we do this as an extension method on `IFluentScreenshotElement<T> where ElementScope`?
Or, if that doesn't work, an exntesion method on `IFluentScreenshotElementWithCovariance<ElementScope>`?

That way, we don't need the guard condition, which, I believe, would make things cleaner.

Remotion/Web/Development.WebTesting/ScreenshotCreation/Fluent/FluentScreenshotElement.cs
```diff
@@ -55,6 +58,19 @@ public class FluentScreenshotElement<T> : IFluentScreenshotElement<T>, IFluentSc
...
+    {
+      if (_target is not ElementScope elementScope)
+        throw new InvalidOperationException($"'{nameof(ScrollIntoView)}' can only be used with Coypu.ElementScope instances.");
+
+      JavaScriptExecutor.GetJavaScriptExecutor(elementScope).ExecuteScript("arguments[0].scrollIntoView(true);", elementScope.Native);
+
+      // The scrolling sometimes takes some time - 200ms should be enough.
+      Thread.Sleep(250);
```
MichaelKetting, line 69: please fix comment to match used timing.

Remotion/Web/Development.WebTesting/ScreenshotCreation/Fluent/IFluentScreenshotElement`.cs
```diff
@@ -45,5 +45,11 @@ public interface IFluentScreenshotElement<T> : IFluentScreenshotElement
     /// </summary>
     [NotNull]
     ScreenshotTransformationCollection<T> Transformations { get; set; }
+
+    /// <summary>
+    /// Scrolls the target element into view. This method must only be used with <see cref="Coypu.ElementScope"/> targets.
+    /// </summary>
+    /// <exception cref="InvalidOperationException">The target is not an <see cref="Coypu.ElementScope"/> instance.</exception>
+    FluentScreenshotElement<T> ScrollIntoView ();
```
MichaelKetting, line 53: The new API is big enough and unrelated enough to warrent a separate ticket + branch. Do you agree that extracting it makes sense?

Remotion/Web/Development.WebTesting.IntegrationTests/DropDownMenuControlObjectTest.cs
```diff
@@ -299,6 +305,77 @@ public void TestHasSupplementalButtonType ()
...
+        builder.Crop(target.GetMenu(), new WebPadding(1));
+      };
+
+      var home = Start();
+      try
+      {
+        ResizeBrowserWindowTo(new Size(600, 400));
+        var fluentDropDownMenu = home.DropDownMenus().GetByLocalID("MyDropDownMenu_ManyMenuItems").ForControlObjectScreenshot();
+        fluentDropDownMenu.OpenMenu();
+        Thread.Sleep(250);
```
MichaelKetting, line 361: ScrollIntoView already does a `Thread.Sleep`. If we need a second one here, this feels like something that invalidates the API: Do we need this because `OpenMenu` takes a long time to build the dom and update the ui due to the long list of items? If so, this means everyone using this API with a long menu could run into the same issue.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/ScreenshotCreation/Fluent/FluentScreenshotElement.cs
iam-mholle: This works. That way, the API change with a separate task + ticket would not be necessary.

Remotion/Web/Development.WebTesting.IntegrationTests/DropDownMenuControlObjectTest.cs
iam-mholle: This sleep was part of the test `Remotion.Web.Development.WebTesting.IntegrationTests.ScreenshotTest.DropDownMenu` I used as reference, which was added during the original implementation of the fluent screenshotting feature - it should not be necessary anymore with the scrolling API I think, I'll test it. If it still is an issue, some way of waiting for the menu to be opened should rather be added.

Remotion/Web/Development.WebTesting/ScreenshotCreation/Fluent/FluentScreenshotElement.cs
MichaelKetting: okay, so we reduce the change to just adding an extension method, which is small enough to fit into the current PR. Hmm... I can see the logic in this.

Remotion/Web/Development.WebTesting.IntegrationTests/DropDownMenuControlObjectTest.cs
MichaelKetting: Great. So, code cleanup if possible. 

And given this delay also needed for non-screenshot tests, we may need a separate ticket to handle the OpenMenu/delay-requirement?

iam-mholle: I actually didn't find the delay in other tests, just screenshot tests.

MichaelKetting (approved):

PR #347, opened 2022-02-25
RM-8339: Implement SafeContextBoundary in HttpContextStorageProvider

iam-mholle:


Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private const string c_dataDictionaryKey = "Rm.DD";
+
+    // Ignore the obsolete warnings for AsyncLocalStorageProvider
+#pragma warning disable 618
+    private readonly AsyncLocalStorageProvider _fallbackProvider = new();
+#pragma warning restore 618
+
+    private readonly ThreadLocal<WeakReference<WrappedHashtable?>> _dataTable = new(() => new WeakReference<WrappedHashtable?>(null));
```
a-ctor, line 51: All the interactions with the cached value and the weak reference are getting difficult to follow. I think we should consider moving all of the logic into a struct which encapsulates the details of accessing the hashtable. Something like:

```c#
struct Wrapper
{
  private class WrappedHashtable { ... }

  private readonly ThreadLocal<WeakReference<WrappedHashtable?>> _dataTable;

  public bool TryGetHashtable(HttpContext context, out Hashtable result) { ... }

  public Hashtable GetOrCreateHashtable(HttpContext context) { ... }

  public Hashtable RemoveHashtable(HttpContext context) { ... }
}
```

This way all the caching logic is at one place and the actual safe context methods are streamlined. By providing the `HttpContext` as parameter we can make `WrappedHashtable` an implementation detail and simplify the outer code. Naming, maybe `CachedHashtableAccessor` or something?

What do you think?

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+      var httpContext = HttpContext.Current;
+
+      if (state is WrappedHashtable hashtable)
+      {
+        SetHashtable(hashtable);
+      }
+      else if (TryGetCachedHashtable(httpContext, out var cachedHashtable))
+      {
+        httpContext.Items.Remove(cachedHashtable);
```
a-ctor, line 141: The `Remove` method takes the key not the value so this does not work correctly. We should also write a test to catch that problem.

a-ctor (changes requested):

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,208 @@
...
+
+      public Hashtable SetHashtable (HttpContext context, Hashtable hashtable)
+      {
+        var wrappedHashtable = new WrappedHashtable(context, hashtable);
+        context.Items[c_dataDictionaryKey] = wrappedHashtable;
+        _dataTable.Value!.SetTarget(wrappedHashtable);
+        return hashtable;
+      }
+
+      private bool TryGetCachedWrappedHashtable (HttpContext httpContext, [MaybeNullWhen(false)] out WrappedHashtable hashtable)
```
a-ctor, line 96: If think we can simplify this method. We return a `WrappedHashtable`, but the usages always use only the `Data`. But if we change the out parameter to `Hashtable` it is effectively our public method above. Couldn't we just move the code into the public method and use that in the `RemoveHashtable` method? What do you think?

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        var wrappedHashtable = (WrappedHashtable?)httpContext.Items[c_dataDictionaryKey];
+        hashtable = wrappedHashtable;
+
+        if (wrappedHashtable != null)
+        {
+          _dataTable.Value!.SetTarget(wrappedHashtable);
+#pragma warning disable CS8762 // Parameter must have a non-null value when exiting in some condition.
+          return true;
+#pragma warning restore CS8762 // Parameter must have a non-null value when exiting in some condition.
```
a-ctor, line 112: Shouldn't it be possible to avoid those by using `hashtable != null` instead of `wrappedHashtable`?

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,208 @@
...
+
+    private readonly HashTableAccessor _hashTableAccessor = new(true);
+
+    public SafeContextBoundary OpenSafeContextBoundary ()
+    {
+      var httpContext = HttpContext.Current;
+      if (httpContext == null)
+      {
+        return _fallbackProvider.OpenSafeContextBoundary();
+      }
```
a-ctor, line 132: Style remark: We are mixing the way we handle the storage fallback quite a lot. Here we check for the fallback first, in the other methods we do it second. We also have `else` statements below while an exit `if` here. Maybe we should unify this to have the same pattern everywhere?

My personal preference would be to do the fallback right after the argument check like so:

```c#
// arg checks
if (httpContext == null)
  return _fallbackProvider.XXXX();
// rest
```

a-ctor (changes requested):

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    void ISafeContextBoundaryStrategy.RestorePreviousSafeContext (object? state)
+    {
+      var httpContext = HttpContext.Current;
+
+      if (state is Hashtable hashtable)
+        _hashTableAccessor.SetHashtable(httpContext, hashtable);
+      else if (_hashTableAccessor.TryGetCachedHashtable(httpContext, out _))
+        _hashTableAccessor.RemoveHashtable(httpContext);
```
a-ctor, line 178: `HashtableAccessor.RemoveHashtable` already does the check for us we can simplify this to an `else` if I am not mistaken.

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,182 @@
...
+    private readonly AsyncLocalStorageProvider _fallbackProvider = new();
+#pragma warning restore 618
+
+    private readonly HashTableAccessor _hashTableAccessor = new(true);
+
+    public SafeContextBoundary OpenSafeContextBoundary ()
+    {
+      var httpContext = HttpContext.Current;
+
+      if (httpContext == null)
```
a-ctor, line 116: style: I personally dislike the empty line in this case and would write it without. Otherwise the `httpContext` feels so left alone :(

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,182 @@
...
+        var wrappedHashtable = (WrappedHashtable?)httpContext.Items[c_dataDictionaryKey];
+        hashtable = wrappedHashtable?.Data;
+
+        if (wrappedHashtable != null)
+        {
+          _dataTable.Value!.SetTarget(wrappedHashtable);
+          return hashtable != null;
+        }
+
+        return false;
```
a-ctor, line 72: I think we can assume that `WrappedHashtable.Data` is always non-null? If that is the case I would rewrite this to make it a bit clearer what is happening. As it stands, we switch between `hashtable` and `wrappedHashtable` for null checks when their nullability should always be the same.

```suggestion
        var wrappedHashtable = (WrappedHashtable?)httpContext.Items[c_dataDictionaryKey];
        if (wrappedHashtable != null)
        {
          hashtable = wrappedHashtable?.Data;
          _dataTable.Value!.SetTarget(wrappedHashtable);
          return true;
        }

        hashtable = null;
        return false;
```

a-ctor (approved):

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Web;
+using Remotion.Context;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Context
+{
+#if !NETFRAMEWORK
+  [ImplementationFor(typeof(ISafeContextStorageProvider), Position = 0)]
+  public class HttpContextStorageProvider : ISafeContextStorageProvider, ISafeContextBoundaryStrategy
```
MichaelKetting, line 30:
```suggestion
#if !NETFRAMEWORK
  [ImplementationFor(typeof(ISafeContextStorageProvider), Position = 0)]
#endif
  public class HttpContextStorageProvider : ISafeContextStorageProvider, ISafeContextBoundaryStrategy
```

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
```diff
@@ -0,0 +1,182 @@
...
+    {
+      var httpContext = HttpContext.Current;
+
+      if (state is Hashtable hashtable)
+        _hashTableAccessor.SetHashtable(httpContext, hashtable);
+      else if (_hashTableAccessor.TryGetCachedHashtable(httpContext, out _))
+        _hashTableAccessor.RemoveHashtable(httpContext);
+    }
+  }
+#endif
```
MichaelKetting, line 177:
```suggestion
```

Remotion/Web/Core/Context/HttpContextStorageProviderNetCore.cs
MichaelKetting, line 30: Rename to  HttpContextStorageProviderWithAsyncLocalBasedFallback

Remotion/Web/Core/Context/HttpContextStorageProvider.cs
```diff
@@ -22,17 +22,11 @@
 
 namespace Remotion.Web.Context
 {
+#if NETFRAMEWORK
   [ImplementationFor(typeof(ISafeContextStorageProvider), Position = 0)]
   public class HttpContextStorageProvider : ISafeContextStorageProvider
```
MichaelKetting, line 27: Rename  HttpContextStorageProviderWithCallContextBasedFallback

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderTest.cs
```diff
@@ -49,7 +49,9 @@ public void SetData ()
     {
       _provider.SetData("Foo", 45);
       Assert.That(_provider.GetData("Foo"), Is.EqualTo(45));
```
MichaelKetting, line 51: Rename testsuite to match SUT and make entire testsuite NETFRAMEWORK-only

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderNetCoreTest.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Web;
+using NUnit.Framework;
+using Remotion.Development.Web.UnitTesting.AspNetFramework;
+using Remotion.Web.Context;
+
+namespace Remotion.Web.UnitTests.Core.Context
+{
+#if !NETFRAMEWORK
+  [TestFixture]
+  public class HttpContextStorageProviderNetCoreTest
```
MichaelKetting, line 28: Rename, please. 

Does this testsuite work "correctly" when run in full framework? If so, we should keep it always enabled. if not, which parts don't work?

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderNetCoreTest.cs
```diff
@@ -0,0 +1,112 @@
...
+      Assert.That(result, Is.Null);
+    }
+
+    [Test]
+    public void SafeContextBoundaryCreationClearsData ()
+    {
+      _provider.SetData("test", 123);
+      _provider.OpenSafeContextBoundary();
+
+      Assert.That(HttpContext.Current.Items["Rm.DD"], Is.Null);
```
MichaelKetting, line 75: should this use "test" as key?

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderNetCoreTest.cs
```diff
@@ -0,0 +1,112 @@
...
+      var context = HttpContext.Current;
+      int? threadResult = null;
+
+      var t = new Thread(
+          ctx =>
+          {
+            HttpContext.Current = (HttpContext)ctx;
+            threadResult = (int)_provider.GetData("test");
+          });
+      t.Start(context);
```
MichaelKetting, line 103: readability question: would it be easier to read to just capture `context` via a closure instead of passing it as an argument? For production code, this pattern is certainly the best way to go so closures are avoided, but in tests, it makes it harder to trace where things are coming from.

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderNetCoreTest.cs
```diff
@@ -0,0 +1,112 @@
...
+      t.Join();
+
+      var result = _provider.GetData("test");
+      Assert.That(result, Is.EqualTo(123));
+      Assert.That(threadResult, Is.EqualTo(123));
+    }
+  }
```
MichaelKetting, line 110: we need to add a test that shows that the values are not copied if we do not manually set the httpcontext?

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderNetCoreTest.cs
```diff
@@ -0,0 +1,112 @@
...
+    private HttpContext _testContext;
+
+    [SetUp]
+    public void SetUp ()
+    {
+      _testContext = HttpContextHelper.CreateHttpContext("x", "y", "z");
+      HttpContext.Current = _testContext;
+      _provider = new HttpContextStorageProvider();
+    }
+
```
MichaelKetting, line 40: We need to document the behavior that the values are not directly accessible from httpcontext.Items. Otherwise, this might pop up as a bug.

MichaelKetting (commented):
Names und Tests reviewed

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithAsyncLocalBasedFallbackTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+      WeakReference GetHashtableWeakReference ()
+      {
+        var accessorFieldInfo = typeof(HttpContextStorageProviderWithAsyncLocalBasedFallback).GetField("_hashtableAccessor", BindingFlags.Instance | BindingFlags.NonPublic);
+        var getOrCreateHashtableMethodInfo = accessorFieldInfo!.FieldType.GetMethod("GetOrCreateHashtable");
+        var accessorField = accessorFieldInfo.GetValue(_provider);
+
+        var hashtableFromField = (Hashtable)getOrCreateHashtableMethodInfo!.Invoke(accessorField, new[] { HttpContext.Current });
+
+        return new(hashtableFromField);
```
MichaelKetting, line 157: I would suggest we simplify / abstract this a bit:

Return a new WeakReference(new object)), place the WeakReference.Value (=the object) as value into the provider.SetData(...) and then just Assert that this object has been collected during GC. 

The intention of the test would remain the same (= we test that the values aren't kept alive on the thread) but we decouple ourselfs from the implementation detail.

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithAsyncLocalBasedFallbackTest.cs
```diff
@@ -0,0 +1,161 @@
...
+      t.Start();
+      t.Join();
+
+      var result = _provider.GetData("test");
+      Assert.That(result, Is.EqualTo(123));
+      Assert.That(threadResult, Is.EqualTo(123));
+    }
+
+    [Test]
+    public void NewThreadCantAccessExistingData ()
```
MichaelKetting, line 113:
```suggestion
    public void NewThreadCannotAccessExistingData ()
```

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithAsyncLocalBasedFallbackTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+      var result = _provider.GetData("test");
+      Assert.That(result, Is.Null);
+    }
+
+    [Test]
+    public void SafeContextBoundaryCreationClearsData ()
+    {
+      _provider.SetData("test", 123);
+      _provider.OpenSafeContextBoundary();
```
MichaelKetting, line 74: I think we should add a using here. It feels strange to leak the boundary during a test.

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithAsyncLocalBasedFallbackTest.cs
```diff
@@ -0,0 +1,161 @@
...
+    private HttpContext _testContext;
+
+    [SetUp]
+    public void SetUp ()
+    {
+      _testContext = HttpContextHelper.CreateHttpContext("x", "y", "z");
+      HttpContext.Current = _testContext;
+      _provider = new HttpContextStorageProviderWithAsyncLocalBasedFallback();
+    }
+
```
MichaelKetting, line 41: please add a teardown with `HttpContext.Current = nul);`. Most of the tests do it, perhaps all tests but the other HttpContextStorageProviderTest. Which would be a bug we could fix right now in the current branch, if you want to.

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithAsyncLocalBasedFallbackTest.cs
```diff
@@ -0,0 +1,161 @@
...
+        var accessorFieldInfo = typeof(HttpContextStorageProviderWithAsyncLocalBasedFallback).GetField("_hashtableAccessor", BindingFlags.Instance | BindingFlags.NonPublic);
+        var getOrCreateHashtableMethodInfo = accessorFieldInfo!.FieldType.GetMethod("GetOrCreateHashtable");
+        var accessorField = accessorFieldInfo.GetValue(_provider);
+
+        var hashtableFromField = (Hashtable)getOrCreateHashtableMethodInfo!.Invoke(accessorField, new[] { HttpContext.Current });
+
+        return new(hashtableFromField);
+      }
+    }
+  }
```
MichaelKetting, line 160: I'm missing `FallbackToAsyncLocal_IfNoCurrentHttpContext`

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithCallContextBasedFallbackTest.cs
```diff
@@ -83,14 +84,7 @@ public void FallbackToAsyncLocal_IfNoCurrentHttpContext ()
     {
```
MichaelKetting, line 84: please rename the test to `FallbackToCallContext_IfNoCurrentHttpContext`

Remotion/Web/UnitTests/Core/Context/HttpContextStorageProviderWithCallContextBasedFallbackTest.cs
```diff
@@ -24,18 +24,19 @@
...
     private HttpContext _testContext;
 
     [SetUp]
     public void SetUp ()
     {
       _testContext = HttpContextHelper.CreateHttpContext("x", "y", "z");
       HttpContext.Current = _testContext;
-      _provider = new HttpContextStorageProvider();
+      _provider = new HttpContextStorageProviderWithCallContextBasedFallback();
     }
```
MichaelKetting, line 41: please add a tear down that sets HttpContext.Current=null. That's missing in the test and since we're touching the code and global state is such a nasty thing, we shouldn't let this go unanswered.

MichaelKetting (changes requested):
A couple of details.

MichaelKetting (approved):

PR #349, opened 2022-02-28
RM-7833 Enable matching an ObjectValidationFailure to bound WebControls by adding properties

a-ctor:
https://re-motion.atlassian.net/browse/RM-7833

### DIFF
--- a/Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
+++ b/Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
@@ -1,4 +1,20 @@
-﻿using System;
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
 using System.Linq;
 using NUnit.Framework;
 using Remotion.ObjectBinding.BindableObject;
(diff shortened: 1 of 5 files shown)


Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -50,23 +50,62 @@ private static BusinessObjectValidationFailure CreateBusinessObjectValidationFai
...
+            {
+              yield return BusinessObjectValidationFailure.CreateForBusinessObject(
+                  validatedBusinessObject,
+                  objectValidationFailure.LocalizedValidationMessage);
+            }
+
+            anyFailureReported = true;
+          }
+
+          if (!anyFailureReported)
```
MichaelKetting, line 88: Not sure about the best design for the empty option when doing this with a sequence, perhaps the cleanest option is to go with ApplySideEffect(()=> hasFailuresReported = true)

=> `.properties.Where (item=>item.RelatedObject is IBusienssObject).ApplySideEffect(_=>any=true).Select(item=>GetResult(item)`

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
```diff
@@ -46,6 +46,67 @@ public void SetUp ()
...
+      var objectValidationFailure = new ObjectValidationFailure(
+          person,
+          "Error",
+          "Localized error");
+      var validationResult = new ValidationResult(new[] { objectValidationFailure });
+      var businessObjectValidationResult = BusinessObjectValidationResult.Create(validationResult);
+
+      var failures = businessObjectValidationResult.GetUnhandledValidationFailures(person).ToArray();
+      Assert.That(failures.Count, Is.EqualTo(1));
+      Assert.That(failures[0].ErrorMessage, Is.EqualTo("Localized error"));
```
MichaelKetting, line 62: Why did you skip validating the `ValidatedObject`?

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -104,31 +145,40 @@ private BusinessObjectValidationResult (IEnumerable<(BusinessObjectValidationFai
...
+    public IReadOnlyCollection<UnhandledBusinessObjectValidationFailure> GetUnhandledValidationFailures (IBusinessObject businessObject, bool includePartiallyHandledFailures = false)
     {
       ArgumentUtility.CheckNotNull("businessObject", businessObject);
 
       return _unhandledValidationFailures
           .Where(f => Equals(f.BusinessObjectValidationFailure.ValidatedObject, businessObject))
+          .Where(f => includePartiallyHandledFailures || !_handledValidationFailures.Contains(f.ValidationFailure))
           .Select(f => f.BusinessObjectValidationFailure)
           .Select(f => new UnhandledBusinessObjectValidationFailure(f.ErrorMessage, f.ValidatedProperty))
+          .Distinct()
```
MichaelKetting, line 171: I think the Distinct needs to get moved up a bit. UnhandledBusinessObjectValidationFailure doesn't have value equality, does it? and shouldn't there be a red test for this?

MichaelKetting, line 171: Hmm... thinking about this, maybe here the distinct isn't actually helpful on this overload because there shouldn't be any duplicates created by us anyway (between message and property).

MichaelKetting (changes requested):

Remotion/Validation/Core/Results/ObjectValidationFailure.cs
```diff
@@ -15,18 +15,36 @@
...
 using JetBrains.Annotations;
+using Remotion.Reflection;
+using Remotion.Utilities;
 
 namespace Remotion.Validation.Results
 {
   public class ObjectValidationFailure : ValidationFailure
   {
+    [NotNull]
+    public IReadOnlyCollection<(object RelatedObject, IPropertyInformation? Property)> AffectedProperties { get; }
```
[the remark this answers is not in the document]
a-ctor: How about `ValidatedObjectProperty`?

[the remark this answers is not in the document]
MichaelKetting: possibly. revisit once we have everything named to make sure things fit together?

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
a-ctor: Because `UnhandledBusinessObjectValidationFailure` does not have a `ValidatedObject`

Remotion/Validation/Core/Results/ValidatedObjectProperty.cs
```diff
@@ -0,0 +1,36 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using Remotion.Reflection;
+using Remotion.Utilities;
+
+namespace Remotion.Validation.Results
+{
+  public class ValidatedObjectProperty
```
MichaelKetting, line 22: docs, please

Remotion/Validation/Core/Results/ObjectValidationFailure.cs
```diff
@@ -15,18 +15,36 @@
...
 using JetBrains.Annotations;
+using Remotion.Reflection;
+using Remotion.Utilities;
 
 namespace Remotion.Validation.Results
 {
   public class ObjectValidationFailure : ValidationFailure
   {
+    [NotNull]
+    public IReadOnlyCollection<ValidatedObjectProperty> ValidatedProperties { get; }
```
MichaelKetting, line 28: Hmm... Given the naming and usage, is it a good idea to have a set of `ValidatedProperties` where the `Property` can be null? I suggest we accept that property is also mandatory on the DTO and rename the DTO to `ValidatedProperty` which makes the `Object` just the instance that holds the property value. 

Good idea? Bad idea?

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -34,11 +34,14 @@ public static BusinessObjectValidationResult Create ([NotNull] ValidationResult
...
-      var validationFailures = validationResult.Errors.Select(e => (CreateBusinessObjectValidationFailure(e), e));
+      var validationFailures = new List<(BusinessObjectValidationFailure, ValidationFailure)>();
+      foreach (var validationFailure in validationResult.Errors)
+        CreateBusinessObjectValidationFailures(validationFailure, f => validationFailures.Add((f, validationFailure)));
+
       return new BusinessObjectValidationResult(validationFailures);
     }
 
-    private static BusinessObjectValidationFailure CreateBusinessObjectValidationFailure (ValidationFailure validationFailure)
+    private static void CreateBusinessObjectValidationFailures (ValidationFailure validationFailure, Action<BusinessObjectValidationFailure> addFailure)
```
MichaelKetting, line 44: I would like to discuss the pro/cons of this approach vs the linq-based approach. it still feels clunky to me.

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
```diff
@@ -46,6 +62,67 @@ public void SetUp ()
...
+
+    [Test]
+    public void Create_WithObjectValidationFailureAndAffectedProperties_ReturnsBusinessObjectValidationFailureForEachAffectedProperty ()
+    {
+      var person = new Person();
+      var objectValidationFailure = new ObjectValidationFailure(
+          person,
+          new ValidatedObjectProperty[]
+          {
+                  new(person, _firstNameProperty),
```
MichaelKetting, line 90: I think the indent is wrong here?
(and in line 115)

MichaelKetting (changes requested):

Remotion/Validation/Core/Results/ObjectValidationFailure.cs
a-ctor: decision: Use `ValidatedProperty` and make `Property` mandatory

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
a-ctor: decision: Revert to `IEnumerable<T>` approach, but extract creation code into own methods.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -48,29 +54,78 @@ private static BusinessObjectValidationFailure CreateBusinessObjectValidationFai
...
+      {
+        return BusinessObjectValidationFailure.CreateForBusinessObjectProperty(
+            localizedValidationMessage,
+            validatedBusinessObject,
+            validatedBusinessObjectProperty);
+      }
+
+      return BusinessObjectValidationFailure.CreateForBusinessObject(
+          validatedBusinessObject,
+          localizedValidationMessage);
```
MichaelKetting, line 126: Question: are those two equivalent (=symmetric) cases? if so, an IF-ELSE instead of an IF + separate "default" behavior would be better to provide code symmetry in the code.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -48,29 +54,78 @@ private static BusinessObjectValidationFailure CreateBusinessObjectValidationFai
...
         }
       }
     }
 
+    private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObjectValidationFailuresFromValidatedProperties (
+        IEnumerable<ValidatedProperty> validatedProperties,
+        string localizedValidationMessage)
+    {
+      var iterated = false;
+      foreach (var validatedProperty in validatedProperties)
```
MichaelKetting, line 95: Do we have a written test where failure is null? The semantic, that we only set the iterated-flag when there's an actual failure created should be explicitly covered.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -48,29 +54,78 @@ private static BusinessObjectValidationFailure CreateBusinessObjectValidationFai
...
+          return EnumerableUtility.Singleton(BusinessObjectValidationFailure.Create(validationFailure.LocalizedValidationMessage));
         }
       }
     }
 
+    private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObjectValidationFailuresFromValidatedProperties (
+        IEnumerable<ValidatedProperty> validatedProperties,
+        string localizedValidationMessage)
+    {
+      var iterated = false;
```
MichaelKetting, line 94: naming: should we rename this to hasResult or hasEmptyResult? Something along those lines. 

`iterated` alone is misleading because it could also mean that we have a not-empty list of validatedProperties and not that we've found one or more items to yield return from the method.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #357, opened 2022-03-11
RM-8404 Switch test domain type discovery to whitelist

a-ctor:
https://re-motion.atlassian.net/browse/RM-8404

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
@@ -44,17 +44,24 @@ namespace Remotion.Data.DomainObjects.UnitTests.Factories
       var assemblyFinder = new CachingAssemblyFinderDecorator(new AssemblyFinder(rootAssemblyFinder, assemblyLoader));
       ITypeDiscoveryService typeDiscoveryService = new AssemblyFinderTypeDiscoveryService(assemblyFinder);
 
-      return FilteringTypeDiscoveryService.CreateFromNamespaceBlacklist(
+      var whitelistedNamespaces = new[]
+                                  {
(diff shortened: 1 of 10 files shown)


Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -44,17 +44,27 @@ public static ITypeDiscoveryService GetTypeDiscoveryService (params Assembly[] r
...
-          "Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.IntegrationTests"
-          );
+          type =>
+          {
+            var @namespace = type.Namespace ?? string.Empty;
+            for (var i = 0; i < whitelistedNamespaces.Length; i++)
+            {
+              if (@namespace.StartsWith(whitelistedNamespaces[i]))
+                return true;
+            }
```
MichaelKetting, line 64:
```suggestion
            if (whitelistedNamespaces.Any(@namespace.StartsWith))
              return true;
```
Whould this be more natural to read?

MichaelKetting (approved):
A Linq / for question, other than that, we're done.

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
a-ctor: Yes, but I wanted to keep the `for` for performance considerations as it prevents the allocation of the `IEnumerator<T>`. This might be optimized away by the JIT but in this case I would take the readability hit.

MichaelKetting: Yeah, but given this is just a onetime hit during the unit tests, I would suggest we go with readability over GC.

In general: all the one-time stuff in the Framework always goes with readability over GC impact (unless something is really problematic) and only calls that happens many times during a request (e.g. for each property access), need to be GC optimized. And in between, it depends.

PR #359, opened 2022-03-21
RM-8292: Change WebButton default border color to primary blue

iam-mholle:


iam-mholle:
@MichaelKetting I used the opportunity to remove the `input.webButton` target - no other rules apply to this selector, am I right to assume that it is obsolete? If not, I'll just change it back.

MichaelKetting:
> @MichaelKetting I used the opportunity to remove the `input.webButton` target - no other rules apply to this selector, am I right to assume that it is obsolete? If not, I'll just change it back.

@iam-mholle Please revert that one. If you set the `WebButton.UseLegacyButton` flag to true, you get the `input` element + the css class.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/WebButton.css
```diff
@@ -6,13 +6,13 @@
...
 
-button.webButton,
-input.webButton
+button.webButton
 {
   min-height: var(--webbutton-min-height);
   padding: 0;
   box-sizing: border-box;
   display: inline-block;
+  border-color: var(--color-primary);
```
MichaelKetting, line 15: I think we need to do this in Common.css. The border look&feel is unified for all button elements via Common.css, so the special color applied to buttons needs to be implemented there as well.

MichaelKetting (changes requested):

iam-mholle:
> > @MichaelKetting I used the opportunity to remove the `input.webButton` target - no other rules apply to this selector, am I right to assume that it is obsolete? If not, I'll just change it back.
> 
> @iam-mholle Please revert that one. If you set the `WebButton.UseLegacyButton` flag to true, you get the `input` element + the css class.

That means that `primary`/`supplemental` style is not available on the legacy button?

MichaelKetting:
> That means that primary/supplemental style is not available on the legacy button?

Hmm... yeah, I think we didn't notice that one. Could you please create a ticket for this?

MichaelKetting (changes requested):

PR #369, opened 2022-03-29
Feature/rm 8350 introduce interface definition

a-ctor:
https://re-motion.atlassian.net/browse/RM-8350

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Mapping/ClassDefinitionTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Mapping/ClassDefinitionTest.cs
@@ -23,14 +23,12 @@ using NUnit.Framework;
 using Remotion.Data.DomainObjects.Infrastructure;
 using Remotion.Data.DomainObjects.Infrastructure.TypePipe;
 using Remotion.Data.DomainObjects.Mapping;
-using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
 using Remotion.Data.DomainObjects.UnitTests.Mapping.MixinTestDomain;
 using Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration;
 using Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.MixedMapping;
 using Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.Model;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain.TableInheritance;
 using Remotion.Development.UnitTesting.Enumerables;
-using Remotion.Development.UnitTesting.NUnit;
 using Remotion.Mixins;
 using Remotion.Reflection;
 
@@ -83,6 +81,7 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
               typeof(Order),
               false,
               null,
+              Enumerable.Empty<InterfaceDefinition>(),
               null,
               DefaultStorageClass.Transaction,
               persistentMixinFinder.Object,
@@ -93,9 +92,11 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
       Assert.That(actual.HasStorageEntityDefinitionBeenSet, Is.False);
       Assert.That(
           () => actual.StorageEntityDefinition,
-          Throws.InvalidOperationException.With.Message.EqualTo("StorageEntityDefinition has not been set for type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order'."));
+          Throws.InvalidOperationException
+              .With.Message.EqualTo("StorageEntityDefinition has not been set for type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order'."));
       Assert.That(actual.Type, Is.SameAs(typeof(Order)));
       Assert.That(actual.BaseClass, Is.Null);
+      Assert.That(actual.ImplementedInterfaces, Is.Empty);
       Assert.That(actual.DefaultStorageClass, Is.EqualTo(DefaultStorageClass.Transaction));
       //Assert.That (actual.DerivedClasses.AreResolvedTypesRequired, Is.True);
       Assert.That(actual.IsReadOnly, Is.False);
@@ -114,6 +115,7 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
               typeof(Order),
               false,
               null,
+              Enumerable.Empty<InterfaceDefinition>(),
               null,
               DefaultStorageClass.Persistent,
               persistentMixinFinder.Object,
@@ -140,6 +142,7 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
               typeof(string),
               false,
               null,
+              Enumerable.Empty<InterfaceDefinition>(),
               null,
               DefaultStorageClass.Persistent,
               persistentMixinFinder.Object,
@@ -148,15 +151,8 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
       Assert.That(classDefinition.HandleCreator, Is.Not.Null);
       Assert.That(
           () => classDefinition.HandleCreator(DomainObjectIDs.Order1),
-          Throws.InvalidOperationException.With.Message.EqualTo("Handles cannot be created when the Type does not derive from DomainObject."));
-    }
-
-    [Test]
-    public void InitializeWithNullStorageGroupType ()
-    {
-      ClassDefinition classDefinition = ClassDefinitionObjectMother.CreateClassDefinition(classType: typeof(TIDomainBase), storageGroupType: null);
-
-      Assert.That(classDefinition.StorageGroupType, Is.Null);
+          Throws.InvalidOperationException
+              .With.Message.EqualTo("Handles cannot be created when the Type does not derive from DomainObject."));
     }
 
     [Test]
@@ -164,81 +160,7 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
     {
       ClassDefinition classDefinition = ClassDefinitionObjectMother.CreateClassDefinition(classType: typeof(TIDomainBase), storageGroupType: typeof(DBStorageGroupAttribute));
 
-      Assert.That(classDefinition.StorageGroupType, Is.Not.Null);
       Assert.That(classDefinition.StorageGroupType, Is.SameAs(typeof(DBStorageGroupAttribute)));
-    }
-
-    [Test]
-    public void SetStorageEntityDefinition ()
-    {
-      var tableDefinition = TableDefinitionObjectMother.Create(_storageProviderDefinition, new EntityNameDefinition(null, "Tablename"));
-      Assert.That(_domainBaseClass.HasStorageEntityDefinitionBeenSet, Is.False);
-
-      _domainBaseClass.SetStorageEntity(tableDefinition);
-
-      Assert.That(_domainBaseClass.StorageEntityDefinition, Is.SameAs(tableDefinition));
-      Assert.That(_domainBaseClass.HasStorageEntityDefinitionBeenSet, Is.True);
-    }
-
-    [Test]
-    public void SetStorageEntityDefinition_ClassIsReadOnly ()
-    {
-      var tableDefinition = TableDefinitionObjectMother.Create(_storageProviderDefinition, new EntityNameDefinition(null, "Tablename"));
-      _domainBaseClass.SetReadOnly();
-
-      Assert.That(
-          () => _domainBaseClass.SetStorageEntity(tableDefinition),
-          Throws.InstanceOf<NotSupportedException>()
-              .With.Message.EqualTo("Type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order' is read-only."));
-    }
-
-    [Test]
-    public void SetPropertyDefinitions ()
-    {
-      var propertyDefinition = PropertyDefinitionObjectMother.CreateForFakePropertyInfo(_domainBaseClass);
-
-      _domainBaseClass.SetPropertyDefinitions(new PropertyDefinitionCollection(new[] { propertyDefinition }, false));
-
-      Assert.That(_domainBaseClass.MyPropertyDefinitions.Count, Is.EqualTo(1));
-      Assert.That(_domainBaseClass.MyPropertyDefinitions[0], Is.SameAs(propertyDefinition));
-      Assert.That(_domainBaseClass.MyPropertyDefinitions.IsReadOnly, Is.True);
-    }
-
-    [Test]
-    public void SetPropertyDefinitions_Twice_ThrowsException ()
-    {
-      var propertyDefinition = PropertyDefinitionObjectMother.CreateForFakePropertyInfo(_domainBaseClass);
-
-      _domainBaseClass.SetPropertyDefinitions(new PropertyDefinitionCollection(new[] { propertyDefinition }, false));
-      Assert.That(
-          () => _domainBaseClass.SetPropertyDefinitions(new PropertyDefinitionCollection(new[] { propertyDefinition }, false)),
-          Throws.InvalidOperationException
-              .With.Message.EqualTo(
-                  "The property-definitions for type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order' have already been set."
-));
-    }
-
-    [Test]
-    public void SetPropertyDefinitions_ClassIsReadOnly ()
-    {
-      _domainBaseClass.SetReadOnly();
-      Assert.That(
-          () => _domainBaseClass.SetPropertyDefinitions(new PropertyDefinitionCollection(new PropertyDefinition[0], true)),
-          Throws.InstanceOf<NotSupportedException>()
-              .With.Message.EqualTo("Type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order' is read-only."));
-    }
-
-    [Test]
-    public void SetRelationEndPointDefinitions ()
-    {
-      var endPointDefinition = new VirtualObjectRelationEndPointDefinition(
-          _domainBaseClass, "Test", false, new Mock<IPropertyInformation>().Object);
-
-      _domainBaseClass.SetRelationEndPointDefinitions(new RelationEndPointDefinitionCollection(new[] { endPointDefinition }, false));
-
-      Assert.That(_domainBaseClass.MyRelationEndPointDefinitions.Count, Is.EqualTo(1));
-      Assert.That(_domainBaseClass.MyRelationEndPointDefinitions[0], Is.SameAs(endPointDefinition));
-      Assert.That(_domainBaseClass.MyRelationEndPointDefinitions.IsReadOnly, Is.True);
     }
 
     [Test]
@@ -272,29 +194,6 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
     }
 
     [Test]
-    public void SetRelationEndPointDefinitions_Twice_ThrowsException ()
-    {
-      var endPointDefinition = new VirtualObjectRelationEndPointDefinition(
-          _domainBaseClass, "Test", false, new Mock<IPropertyInformation>().Object);
-
-      _domainBaseClass.SetRelationEndPointDefinitions(new RelationEndPointDefinitionCollection(new[] { endPointDefinition }, false));
-      Assert.That(
-          () => _domainBaseClass.SetRelationEndPointDefinitions(new RelationEndPointDefinitionCollection(new[] { endPointDefinition }, false)),
-          Throws.InvalidOperationException
-              .With.Message.EqualTo("The relation end point definitions for type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order' have already been set."));
-    }
-
-    [Test]
-    public void SetRelationEndPointDefinitions_ClassIsReadonly ()
-    {
-      _domainBaseClass.SetReadOnly();
-      Assert.That(
-          () => _domainBaseClass.SetRelationEndPointDefinitions(new RelationEndPointDefinitionCollection(new IRelationEndPointDefinition[0], true)),
-          Throws.InstanceOf<NotSupportedException>()
-              .With.Message.EqualTo("Type 'Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Order' is read-only."));
-    }
-
-    [Test]
     public void SetDerivedClasses ()
     {
       _personClass.SetDerivedClasses(new[] { _customerClass });
@@ -304,22 +203,12 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
     }
 
     [Test]
-    public void SetDerivedClasses_Twice_ThrowsException ()
-    {
-      _personClass.SetDerivedClasses(new[] { _customerClass });
-      Assert.That(
-          () => _personClass.SetDerivedClasses(new[] { _customerClass }),
-          Throws.InvalidOperationException
-              .With.Message.EqualTo(
-                  "The derived-classes for class 'Person' have already been set."));
-    }
-
-    [Test]
     public void SetDerivedClasses_ClasssIsReadOnly ()
     {
-      _personClass.SetReadOnly();
+      var classDefinition = ClassDefinitionObjectMother.CreateClassDefinitionWithDefaultProperties("Person");
+      classDefinition.SetReadOnly();
       Assert.That(
-          () => _personClass.SetDerivedClasses(new[] { _orderClass }),
+          () => classDefinition.SetDerivedClasses(new[] { _orderClass }),
           Throws.InstanceOf<NotSupportedException>()
               .With.Message.EqualTo("Class 'Person' is read-only."));
     }
@@ -341,19 +230,6 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
           Throws.InstanceOf<MappingException>()
               .With.Message.EqualTo(
                   "Derived class 'Person' cannot be added to class 'Customer', because it has class 'DomainBase' as its base class definition defined."));
-    }
-
-    [Test]
-    public void SetReadOnly ()
-    {
-      ClassDefinition actual = ClassDefinitionObjectMother.CreateClassDefinition();
-      actual.SetDerivedClasses(new ClassDefinition[0]);
-      Assert.That(actual.IsReadOnly, Is.False);
-
-      actual.SetReadOnly();
-
-      Assert.That(actual.IsReadOnly, Is.True);
-      Assert.That(((ICollection<ClassDefinition>)actual.DerivedClasses).IsReadOnly, Is.True);
     }
 
     [Test]
@@ -391,16 +267,6 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping
           _distributorClass.GetRelationEndPointDefinition(
               "Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration.Partner.ContactPerson"),
           Is.Not.Null);
-    }
-
-    [Test]
-    public void GetRelationEndPointDefinitionFromEmptyPropertyName ()
-    {
-      Assert.That(
-          () => _orderClass.GetRelationEndPointDefinition(string.Empty),
-          Throws.ArgumentException
-              .With.ArgumentExceptionMessageEqualTo(
-                  "Parameter 'propertyName' cannot be empty.", "propertyName"));
     }
 
     [Test]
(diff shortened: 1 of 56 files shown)


Remotion/Data/DomainObjects/Queries/EagerFetching/FetchedRelationDataRegistrationAgentBase.cs
```diff
@@ -70,10 +70,10 @@ private void CheckClassDefinitionOfOriginatingObject (IRelationEndPointDefinitio
       Assertion.DebugAssert(originatingObject.IsNull == false, "originatingObject.IsNull == false");
       Assertion.DebugIsNotNull(originatingObject.ObjectID, "originatingObject.ObjectID != null");
 
-      var relationEndPointDefinitionInheritanceRoot = relationEndPointDefinition.TypeDefinition.GetInheritanceRoot();
-      var originatingObjectInheritanceRoot = originatingObject.ObjectID.ClassDefinition.GetInheritanceRoot();
+      var relationEndPointDefinitionInheritanceRoots = TypeDefinitionHierarchy.GetHierarchyRoots(relationEndPointDefinition.TypeDefinition);
```
MichaelKetting, line 73: Just something I remembered from a previous review (only related to this code here because of the question "static or extension"):

When I write tests that exercise extension methods, I try to write the extension method as a static method instead of using the extension syntax. That way, it doesn't look like I'm exercising a method on a mock object:

Extension syntax:
`mock.Object.DoExtensionStuff()`

Static syntax:
`Extensions.DoExtensionStuff(mock.Object)`

Thoughts? I'm thinking this could (should?) be an analyzer warning.

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionVisitor.cs
```diff
@@ -0,0 +1,134 @@
...
+        TypeDefinition typeDefinition,
+        Func<ITypeDefinitionVisitor<T>, ClassDefinition, T> classDefinitionHandler,
+        Func<ITypeDefinitionVisitor<T>, InterfaceDefinition, T> interfaceDefinitionHandler)
+    {
+      ArgumentUtility.CheckNotNull(nameof(typeDefinition), typeDefinition);
+      ArgumentUtility.CheckNotNull(nameof(classDefinitionHandler), classDefinitionHandler);
+      ArgumentUtility.CheckNotNull(nameof(interfaceDefinitionHandler), interfaceDefinitionHandler);
+
+      var visitor = new TypeDefinitionVisitor<T>(classDefinitionHandler, interfaceDefinitionHandler);
+      return visitor.Visit(typeDefinition);
```
MichaelKetting, line 118: The problem with not having `Visit (TypeDefinition)` on the visitor (see other comment) is solved by doing `typeDefinition.Accept(visitor)`

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionWalker.cs
```diff
@@ -0,0 +1,176 @@
...
+        _classDefinitionHandler = classDefinitionHandler;
+        _interfaceDefinitionHandler = interfaceDefinitionHandler;
+      }
+
+      /// <inheritdoc />
+      public override void VisitClassDefinition (ClassDefinition classDefinition)
+      {
+        ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+
+        _classDefinitionHandler(classDefinition);
```
MichaelKetting, line 50: I'm questioning if handler should be before or after the base call. Possibly, it depends on the direction of the walk.

This might be extremely usecase dependent. For instance, if we build a cache of all the properties defined by the hierarchy, we want to calculate the base-type first, then our own type.

on the other hand, if we want to return a value once we find it, we typically can stop once the first type returns the value, including our own.

Remotion/Data/DomainObjects/Linq/DomainObjectQueryGenerator.cs
```diff
@@ -274,9 +274,10 @@ private IRelationEndPointDefinition GetEagerFetchRelationEndPointDefinition (Fet
       }
 
       var propertyInfoAdapter = PropertyInfoAdapter.Create(propertyInfo);
-      var endPoint = typeDefinition.GetTypeHierarchy()
-                         .Select(cd => cd.ResolveRelationEndPoint(propertyInfoAdapter))
-                         .FirstOrDefault(ep => ep != null);
+      var endPoint = InlineTypeDefinitionWalker.WalkDescendants(
```
MichaelKetting, line 277: Here's a good example where the api no longer communicates that the walk stops once the endpoint is found.

MichaelKetting (commented):
Partial review

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/RdbmsPersistenceModelLoader.cs
```diff
@@ -90,18 +90,15 @@ public void ApplyPersistenceModelToHierarchy (TypeDefinition typeDefinition)
...
-      // ReSharper restore PossibleMultipleEnumeration
-    }
-
-    private void EnsureAllStorageEntitiesCreated (IEnumerable<TypeDefinition> typeDefinitions)
-    {
-      foreach (var typeDefinition in typeDefinitions)
-        EnsureStorageEntitiesCreated(typeDefinition);
+      InlineTypeDefinitionWalker.WalkDescendants(
+          typeDefinition,
+          EnsureStoragePropertiesCreated,
```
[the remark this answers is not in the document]
a-ctor: The double walk comes from the previous implementation that also did the check in two separate steps. So I retained the existing behavior.

Remotion/Data/DomainObjects/Mapping/ITypeDefinitionVisitor1.cs
```diff
@@ -0,0 +1,27 @@
...
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+namespace Remotion.Data.DomainObjects.Mapping
+{
+  public interface ITypeDefinitionVisitor<out T>
+  {
+    T? Visit (TypeDefinition typeDefinition);
```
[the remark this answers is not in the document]
a-ctor: As discussed in person, I am going to remove the `Visit` method in favor of using `.Accept(visitor)`

Remotion/Data/DomainObjects/Mapping/TypeDefinitionWalker.cs
```diff
@@ -0,0 +1,81 @@
...
+
+      typeDefinition.Accept(this);
+    }
+
+    /// <inheritdoc />
+    public virtual void VisitClassDefinition (ClassDefinition classDefinition)
+    {
+      ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
+
+      if (_direction == TypeDefinitionWalkerDirection.Descendants)
```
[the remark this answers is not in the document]
a-ctor: I don't see the benefit of introducing a template method here. Many extensions abort the walk in certain cases, thus not calling `base.VisitXXX()`. With a template method it would need to support that as well, by, for example, returning a boolean. This seems like it overcomplicates the implementation for no real benefit?

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionWalker.cs
a-ctor: It would be possible to add an enum to indicate when the handler is called if that behavior is needed. But I don't think we need it. All of the use cases right now do not require this behavior. In cases where things need to be built, like the property cache, you mentioned, a manual approach is probably better anyway.

Remotion/Data/DomainObjects/Mapping/TypeDefinitionWalker1.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Linq;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Mapping
+{
+  /// <summary>
+  /// Implements a <see cref="ITypeDefinitionVisitor{T}"/> that visits the type definitions in a depth-first manner
+  /// returning the first value of <typeparamref name="T"/> that is not <see langword="null"/>.
+  /// </summary>
+  public abstract class TypeDefinitionWalker<T> : ITypeDefinitionVisitor<T>
```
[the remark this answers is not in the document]
a-ctor: How about we leave the name but add a parameter for the filter? This way implementation like the inline one get another parameter with the filter condition, thus making it clearer what the behavior is.

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/RdbmsPersistenceModelLoader.cs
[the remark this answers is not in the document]
MichaelKetting: Ah, hmm... yes. Read this wrong in the diff. 

I checked the code, and doing this during the same loop would work, if this resuls in cleaner code.

Remotion/Data/DomainObjects/Mapping/InterfaceDefinition.cs
```diff
@@ -0,0 +1,209 @@
...
+        throw new InvalidOperationException("Cannot set the type definition read-only as the implementing classes are not set.");
+      if (_extendingInterfaces == null)
+        throw new InvalidOperationException("Cannot set the type definition read-only as the extending interfaces are not set.");
+
+      base.SetReadOnly();
+    }
+
+    public override bool IsAssignableFrom (TypeDefinition other)
+    {
+      if (ReferenceEquals(this, other))
```
MichaelKetting, line 123: argcheck

Remotion/Data/DomainObjects/Mapping/InterfaceDefinition.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+          throw new MappingException(
+              $"Relation end point for property '{relationEndPointPropertyName}' cannot be added to interface '{Type.GetFullNameSafe()}',"
+              + $" because it was initialized for interface '{relationEndPointTypeDefinition.Type.GetFullNameSafe()}'.");
+        }
+
+        foreach (var extendedInterface in ExtendedInterfaces)
+        {
+          var baseEndPointDefinition = extendedInterface.GetRelationEndPointDefinition(relationEndPointPropertyName);
+          if (extendedInterface != null && baseEndPointDefinition != null)
```
MichaelKetting, line 175:
```suggestion
          if (baseEndPointDefinition != null)
```

Remotion/Data/DomainObjects/Mapping/PropertyDefinitionCollection.cs
```diff
@@ -30,11 +30,13 @@ public static PropertyDefinitionCollection CreateForAllProperties (TypeDefinitio
     {
       ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (typeDefinition is not ClassDefinition classDefinition) // TODO R2I Mapping: Maybe move this into TypeDefinition instead
-        throw new NotSupportedException("Only class definitions are supported");
+      var properties = new List<PropertyDefinition>();
+      InlineTypeDefinitionWalker.WalkAncestors(
+          typeDefinition,
+          classDefinition => properties.AddRange(classDefinition.MyPropertyDefinitions),
+          interfaceDefinition => properties.AddRange(interfaceDefinition.MyPropertyDefinitions));
```
MichaelKetting, line 37: unit test ist missing for itnerface

Remotion/Data/DomainObjects/Mapping/RelationEndPointDefinitionCollection.cs
```diff
@@ -30,12 +30,13 @@ public static RelationEndPointDefinitionCollection CreateForAllRelationEndPoints
     {
       ArgumentUtility.CheckNotNull("typeDefinition", typeDefinition);
 
-      if (typeDefinition is not ClassDefinition classDefinition) // TODO R2I Mapping: Maybe move this into TypeDefinition instead
-        return new RelationEndPointDefinitionCollection(typeDefinition.MyRelationEndPointDefinitions, makeCollectionReadOnly);
+      var endPoints = new List<IRelationEndPointDefinition>();
+      InlineTypeDefinitionWalker.WalkAncestors(
+          typeDefinition,
+          classDefinition => endPoints.AddRange(classDefinition.MyRelationEndPointDefinitions),
+          interfaceDefinition => endPoints.AddRange(interfaceDefinition.MyRelationEndPointDefinitions));
```
MichaelKetting, line 37: unit test ist missing for itnerface

Remotion/Data/DomainObjects/Mapping/TypeDefinition.cs
```diff
@@ -180,14 +186,16 @@ public void SetPropertyDefinitions (PropertyDefinitionCollection propertyDefinit
...
+        ? _cachedRelationEndPointDefinitions.Value
+        : CreateRelationEndPointDefinitionCollection();
     }
 
     public IRelationEndPointDefinition? GetRelationEndPointDefinition (string propertyName)
     {
       ArgumentUtility.CheckNotNullOrEmpty(nameof(propertyName), propertyName);
 
-      return _cachedRelationEndPointDefinitions.Value[propertyName];
+      return GetRelationEndPointDefinitions()[propertyName];
```
MichaelKetting, line 198: Here and GetPropertyDefinition: since you're touching those lines and changing it to a method call: do we have a method avaiable instead of the indexer? It reads better.

Remotion/Data/DomainObjects/Mapping/TypeDefinitionHierarchy.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    /// <summary>
+    /// Returns the hierarchy (self and descendants) of the specified <paramref name="typeDefinition"/>, using a depth-first approach.
+    /// </summary>
+    /// <remarks>
+    /// The hierarchy includes the specified <paramref name="typeDefinition"/> as the first element.
+    /// </remarks>
+    public static IReadOnlyList<TypeDefinition> GetDescendantsAndSelf (TypeDefinition typeDefinition)
+    {
+      var typeHierarchyCollector = new TypeHierarchyCollector();
```
MichaelKetting, line 56: argcheck

Remotion/Data/DomainObjects/Mapping/TypeDefinitionHierarchy.cs
```diff
@@ -0,0 +1,110 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System.Collections.Generic;
+
+namespace Remotion.Data.DomainObjects.Mapping
+{
+  /// <summary>
+  /// Provides methods for getting the hierarchy of a <see cref="TypeDefinition"/>.
+  /// </summary>
+  public static class TypeDefinitionHierarchy
```
MichaelKetting, line 24: why do we skip interfaces?

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/RdbmsPersistenceModelLoader.cs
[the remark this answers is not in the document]
a-ctor: I am ignoring this issue for now since this part of the code is going to be replaced when the persistence logic is implemented

Remotion/Data/DomainObjects.UnitTests/Mapping/ClassDefinitionTest.cs
```diff
@@ -492,55 +338,37 @@ public void GetInheritedPropertyDefinition ()
           Is.Not.Null);
     }
 
+
```
MichaelKetting, line 341: is this extra blank line intentional?

Remotion/Data/DomainObjects.UnitTests/Mapping/InterfaceDefinitionTest.cs
```diff
@@ -0,0 +1,329 @@
...
+          () => actual.ExtendingInterfaces,
+          Throws.InvalidOperationException
+              .With.Message.EqualTo(
+                  "No extending interfaces have been set for interface 'Remotion.Data.DomainObjects.UnitTests.Mapping.InterfaceDefinitionTest+IDomainBase'."));
+
+      Assert.That(actual.IsReadOnly, Is.False);
+    }
+
+    [Test]
+    public void IsPartOfInheritanceHierarchy ()
```
MichaelKetting, line 103: Am realizing that a property IsPartOfInheritanceHierarchy is very strange. How is this property used?

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionTest.cs
```diff
@@ -14,13 +14,491 @@
...
+          );
+
+      _domainBaseTypeDefinition.SetPropertyDefinitions(new PropertyDefinitionCollection(new[] { propertyDefinition }, false));
+
+      Assert.That(_domainBaseTypeDefinition.MyPropertyDefinitions.Count, Is.EqualTo(1));
+      Assert.That(_domainBaseTypeDefinition.MyPropertyDefinitions[0], Is.SameAs(propertyDefinition));
+      Assert.That(_domainBaseTypeDefinition.MyPropertyDefinitions.IsReadOnly, Is.True);
+
+      _checkPropertyDefinitionsMock.Verify(_ => _(new[] { propertyDefinition }));
+    }
```
MichaelKetting, line 292: Based on symmetry with other tests, we'd use Verifiable() + Verify() instead of AAA syntax. Is there a particular benefit to using AAA?

Remotion/Data/DomainObjects/Linq/MappingResolver.cs
```diff
@@ -159,10 +161,14 @@ public Expression ResolveTypeCheck (Expression checkedExpression, Type desiredTy
...
-        var allClassIDExpressions = allClassDefinitions.Select(cd => new SqlLiteralExpression(cd.ID));
 
-        return new SqlInExpression(classIDExpression, new SqlCollectionExpression(typeof(string[]), allClassIDExpressions.Cast<Expression>()));
+        var expressions = new List<Expression>();
+        InlineTypeDefinitionWalker.WalkDescendants(
+            typeDefinition,
+            classDefinition => expressions.Add(new SqlLiteralExpression(classDefinition.ID)),
+            interfaceDefinition => throw new NotImplementedException("Interfaces are not supported")); // TODO R2I Mapping: Add support for interfaces
+
+        return new SqlInExpression(classIDExpression, new SqlCollectionExpression(typeof(string[]), expressions));
```
MichaelKetting, line 171:
```suggestion
        var allClassIDExpressions = new List<Expression>();
        InlineTypeDefinitionWalker.WalkDescendants(
            typeDefinition,
            classDefinition => expressions.Add(new SqlLiteralExpression(classDefinition.ID)),
            interfaceDefinition => throw new NotImplementedException("Interfaces are not supported")); // TODO R2I Mapping: Add support for interfaces

        return new SqlInExpression(classIDExpression, new SqlCollectionExpression(typeof(string[]), allClassIDExpressions));
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionHierarchyTest.cs
```diff
@@ -0,0 +1,105 @@
...
+    [Test]
+    public void ArePartOfSameHierarchy_WithSameTypeDefinition_ReturnsTrue ()
+    {
+      var typeDefinition = TypeDefinitionObjectMother.CreateInterfaceDefinition();
+
+      Assert.That(TypeDefinitionHierarchy.ArePartOfSameHierarchy(typeDefinition, typeDefinition), Is.True);
+    }
+
+    [Test]
+    public void GetHierarchyRoots ()
```
MichaelKetting, line 60: Dropping GetHierarchyRoots for TypeDefinitions in favor of only doing ClassDefinitions is in a later feature branch, isn't it?

Remotion/Data/DomainObjects.UnitTests/Mapping/InlineTypeDefinitionVisitorTest.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    [Test]
+    public void Visit_WithoutResult_InterfaceDefinition ()
+    {
+      var classDefinition = InterfaceDefinitionObjectMother.CreateInterfaceDefinition();
+      _voidReceiverMock.Setup(_ => _.HandleInterfaceDefinition(It.IsNotNull<ITypeDefinitionVisitor>(), classDefinition));
+
+      ExecuteVoidVisit(classDefinition);
+
+      _voidReceiverMock.VerifyAll();
```
MichaelKetting, line 92: Switch to Verify

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionWalkerWithResultTest.cs
```diff
@@ -0,0 +1,157 @@
...
+      var (domainBase, person, customer, organizationalUnit) = TypeDefinitionWalkerTest.CreateClassTestDomain();
+
+      var walker = new TestableTypeDefinitionWalker<string>(TypeDefinitionWalkerDirection.Descendants, value => value != null);
+      Assert.That(walker.VisitClassDefinition(domainBase), Is.Null);
+
+      Assert.That(walker.VisitedTypeDefinitions, Is.EqualTo(new[] { domainBase, person, customer, organizationalUnit }));
+    }
+
+    [Test]
+    public void VisitClassDefinition_WithChildrenDirectionAndFilter_AbortsTraversalOnNonNullValue ()
```
MichaelKetting, line 77: What does `AndFilter` / `AndNoFilter` stand for?

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionWalkerWithResultTest.cs
```diff
@@ -0,0 +1,157 @@
...
+          return result;
+
+        return base.VisitInterfaceDefinition(interfaceDefinition);
+      }
+    }
+
+    [Test]
+    public void VisitClassDefinition_WithChildrenDirectionAndNoFilter_IsCalledInCorrectOrder ()
+    {
+      var (domainBase, person, customer, organizationalUnit) = TypeDefinitionWalkerTest.CreateClassTestDomain();
```
MichaelKetting, line 68: this kind of sharing without a base class is a smell. Should we just move this to a base class?

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionVisitor.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+      private readonly Action<ITypeDefinitionVisitor, ClassDefinition> _classDefinitionHandler;
+      private readonly Action<ITypeDefinitionVisitor, InterfaceDefinition> _interfaceDefinitionHandler;
+
+      public TypeDefinitionVisitor (
+          Action<ITypeDefinitionVisitor, ClassDefinition> classDefinitionHandler,
+          Action<ITypeDefinitionVisitor, InterfaceDefinition> interfaceDefinitionHandler)
+      {
+        ArgumentUtility.CheckNotNull(nameof(classDefinitionHandler), classDefinitionHandler);
+        ArgumentUtility.CheckNotNull(nameof(interfaceDefinitionHandler), interfaceDefinitionHandler);
```
MichaelKetting, line 69: Just FYI: for private types, we don't need argument checks (same as with any other private method), except...

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionVisitor.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        _classDefinitionHandler = classDefinitionHandler;
+        _interfaceDefinitionHandler = interfaceDefinitionHandler;
+      }
+
+      /// <inheritdoc />
+      public void VisitClassDefinition (ClassDefinition classDefinition)
+      {
+        ArgumentUtility.CheckNotNull("classDefinition", classDefinition);
```
MichaelKetting, line 78: ...expect this argument check here is sensible since it ensures that _handler isn't invoked with a null value.

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionVisitor.cs
```diff
@@ -0,0 +1,118 @@
...
+      ArgumentUtility.CheckNotNull(nameof(classDefinitionHandler), classDefinitionHandler);
+      ArgumentUtility.CheckNotNull(nameof(interfaceDefinitionHandler), interfaceDefinitionHandler);
+
+      var visitor = new TypeDefinitionVisitor<T>(classDefinitionHandler, interfaceDefinitionHandler);
+      return typeDefinition.Accept(visitor);
+    }
+
+    public static void Visit (
+        TypeDefinition typeDefinition,
+        Action<ITypeDefinitionVisitor, ClassDefinition> classDefinitionHandler,
```
MichaelKetting, line 107: Why do we pass the visitor into the handler? The other inline-API doesn't do this and it feels kind of strange?

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionWalker.cs
```diff
@@ -0,0 +1,186 @@
...
+      private readonly Action<InterfaceDefinition> _interfaceDefinitionHandler;
+
+      public InlineTypeDefinitionWalkerImplementation (
+          TypeDefinitionWalkerDirection direction,
+          Action<ClassDefinition> classDefinitionHandler,
+          Action<InterfaceDefinition> interfaceDefinitionHandler)
+          : base(direction)
+      {
+        ArgumentUtility.CheckNotNull(nameof(classDefinitionHandler), classDefinitionHandler);
+        ArgumentUtility.CheckNotNull(nameof(interfaceDefinitionHandler), interfaceDefinitionHandler);
```
MichaelKetting, line 39: See InlineVisitor on notes about those checks

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionWalker.cs
```diff
@@ -0,0 +1,186 @@
...
+      ArgumentUtility.CheckNotNull(nameof(typeDefinition), typeDefinition);
+      ArgumentUtility.CheckNotNull(nameof(classDefinitionHandler), classDefinitionHandler);
+      ArgumentUtility.CheckNotNull(nameof(interfaceDefinitionHandler), interfaceDefinitionHandler);
+
+      var inlineTypeDefinitionWalker = new InlineTypeDefinitionWalkerImplementation(TypeDefinitionWalkerDirection.Descendants, classDefinitionHandler, interfaceDefinitionHandler);
+      typeDefinition.Accept(inlineTypeDefinitionWalker);
+    }
+
+    /// <summary>
+    /// Walks over all descendants of the specified <paramref name="typeDefinition"/> (including itself) executing the specified handlers.
```
MichaelKetting, line 164:
```suggestion
    /// Walks over the specified <paramref name="typeDefinition"/> and its descendants and executes the specified handlers until <paramref name="completionPredicate" /> returns <see langword="true" />.
```

See other docs, too

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Mapping/TypeDefinitionWalker1.cs
```diff
@@ -0,0 +1,86 @@
...
+  /// <summary>
+  /// Implements a <see cref="ITypeDefinitionVisitor{T}"/> that visits the type definitions in a depth-first manner
+  /// aborting the walk and returning the first value of <typeparamref name="T"/> that fulfills a specified predicate.
+  /// </summary>
+  public abstract class TypeDefinitionWalker<T> : ITypeDefinitionVisitor<T>
+  {
+    protected readonly TypeDefinitionWalkerDirection Direction;
+    protected readonly Func<T?, bool> CompletionPredicate;
+
+    protected TypeDefinitionWalker (TypeDefinitionWalkerDirection direction, Func<T?, bool> completionPredicate)
```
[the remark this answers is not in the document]
MichaelKetting: ah, oh, yeah. I thought we do have a transparent conversion. and the last time I used this, I was using linq syntax isntead of extension methods and the c# compiler invokes the delegate instead of passing it on to avoid this conflict. okay.

Remotion/Data/DomainObjects.UnitTests/Mapping/InterfaceDefinitionTest.cs
a-ctor: It is used to determine if a class ID column if necessary for a relation

MichaelKetting: Hmm... Move to TypeDefinitionHierarchy because it has a utility feel to it instead of a first-class citizien on the type?

MichaelKetting: or just a private logic in the persistence layer when doing the classID column?

Remotion/Data/DomainObjects/Mapping/InlineTypeDefinitionVisitor.cs
a-ctor: The inline walker already calls itself to walk the hierarchy, as such, it is not necessary to manually call the visitor. If something like that is necessary one should create a class derived from the base walker.

In the case of the inline visitor it is a much more common use case to continue calling the visitor. This is not possible without the argument as the visitor is never available from the outside.

I had a use case for this, but I fear that this was refactored as I cannot find it anymore. While I think that this solution has better usability, I have no problem removing the parameter.

MichaelKetting: Then I opt for dropping this to keep the API more straight forward until the need actually arises.

MichaelKetting (commented):
Some api questions we should satisfy before merging.

Remotion/Data/DomainObjects/Mapping/TypeDefinitionWalker1.cs
```diff
@@ -0,0 +1,87 @@
...
+  {
+    protected readonly TypeDefinitionWalkerDirection Direction;
+    protected readonly Func<T?, bool> CompletionPredicate;
+
+    protected TypeDefinitionWalker (TypeDefinitionWalkerDirection direction, Predicate<T?> match)
+    {
+      ArgumentUtility.CheckNotNull("match", match);
+
+      Direction = direction;
+      CompletionPredicate = new Func<T?, bool>(match);
```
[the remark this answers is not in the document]
a-ctor: I think that this solution provides the best of the two worlds. It is clearer for the consumer if the signaure contains a `Predicate<T>` while also not incurring that much of an overhead when creating the walker.

MichaelKetting (approved):

PR #373, opened 2022-04-01
bugfix/RM-8421-Tooltip-For-BocBooleanValue-Is-Double-Encoded

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-8421

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/Utilities.ts
+++ b/Remotion/Web/ClientScript/Scripts/Utilities.ts
@@ -88,6 +88,12 @@ class StringUtility
       ArgumentUtility.CheckTypeIsString('value', value);
       return TypeUtility.IsNull(value) || value.length == 0;
   };
+
+  public static GetPlainTextFromHtml (value: string): string
+  {
+    let domParser = new DOMParser();
+    return domParser.parseFromString(value, "text/html").documentElement.innerText;
+  };
 }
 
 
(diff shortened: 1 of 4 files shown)


Remotion/Web/ClientScript/Scripts/Utilities.ts
```diff
@@ -88,6 +88,12 @@ class StringUtility
       ArgumentUtility.CheckTypeIsString('value', value);
       return TypeUtility.IsNull(value) || value.length == 0;
   };
+
+  public static GetPlainTextFromHtml (value: string): string
+  {
+    let domParser = new DOMParser();
```
iam-mholle, line 94:
```suggestion
    const domParser = new DOMParser();
```

iam-mholle: Also, would it make sense to instantiate a private static field for the parser, instead of creating the Parser instance anew every time the function is called? Not sure if this could even have a noticeable impact.

iam-mholle (approved):

Remotion/Web/ClientScript/Scripts/Utilities.ts
MichaelKetting: Thanks, sorry about the merge beforehand. The notifications got caught up in an Outlook rule.

let/const: yeah, would have been a good idea. will get cleaned up sometime in the future :)

static field: hmm... Given that it's "Javascript" I would keep it without the static field until there's an issue. It keeps things simpler.

PR #387, opened 2022-04-08
RM-8433 Set max length of ClassWithAllDataTypes.NullableBinaryProperty from 1000000 to 1000

a-ctor:
https://re-motion.atlassian.net/browse/RM-8433


MichaelKetting (changes requested):
Please update the SetupDB.sql script to mirror this change in the datebase: `varbinary(1000)` instead of `varbinary(max)`.

MichaelKetting (changes requested):
Considering the updated semantics, should the image use "<1kB" instead of "<1 MB" to clearly communicate intent? Plz re-request if that's not a good idea.

Am surprised we don't need to change Image2 as well?

a-ctor:
whoops metric is hard...

MichaelKetting (approved):
Just for completeness' sake: it's kB with a lower-case "k".

PR #391, opened 2022-04-14
RM-8327: Fix DropDownMenuOptions width when aligned left

iam-mholle:


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -533,14 +533,11 @@ class DropDownMenu
...
-      if (LayoutUtility.GetOffset (popUpDiv).left < 0)
-      {
-        left = Math.max (0, LayoutUtility.GetOffset (referenceElement).left) + 'px';
-        popUpDiv.style.left = left;
-        popUpDiv.style.right = 'auto';
-      }
+      left = Math.max (0, LayoutUtility.GetOffset (referenceElement).left) + 'px';
+      popUpDiv.style.left = left;
+      popUpDiv.style.right = 'auto';
     }
```
MichaelKetting, line 541: It seems that not everything is solved:
image
There's also no input regarding the use of an elispis with overflow and if using a left/right-margin instead of sticking to the window border is feasible.

MichaelKetting (changes requested):

iam-mholle:
@MichaelKetting introducing a margin should now be feasible relatively easy. Should I add it with this PR?

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
```diff
@@ -225,6 +225,8 @@ ul.DropDownMenuOptions li a > span:last-of-type
   grid-area: span;
   align-self: center;
   padding-left: var(--remotion-themed-spacing);
+  overflow: hidden;
+  text-overflow: ellipsis;
```
MichaelKetting, line 229: How does the effect look in NovaGray?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -528,19 +528,35 @@ class DropDownMenu
 
     popUpDiv.style.top = top + 'px';
```
MichaelKetting, line 529: Fix for context menu behavior, right now, it does not reposition to the left of the mouse, but rather goes into overflow.

For context menu, it may be best that the menu is stuck to the right border first and then grows past the mouse cursor towards the left border (= inverse behavior of dropdown menu popup). Reasoning: By default context menu opens to the right of the mouse cursor, dropdown menu popup opens to the left of the drop down button.

MichaelKetting (commented):
reviewed what's already done. Cotnextmenu und verification of NovaGray is ToDo.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/DropDownMenu.css
iam-mholle: ATM the ellipsis is not added for NovaGray, only the alignment - the popup overflows the viewport when too large. Now I'll have to change it anyway when adding the margin, as it would look weird otherwise

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -521,39 +521,88 @@ class DropDownMenu
...
+      popUpDiv.style.left = 'auto';
+      popUpDiv.style.right = popUpMargin + 'px';
+    };
+    const clampToViewPort = () =>
+    {
+      popUpDiv.style.left = popUpMargin + 'px';
+      popUpDiv.style.right = popUpMargin + 'px';
+    };
+
+    if (isContextMenu && (requiredPopUpWidth <= maxLeftAlignedPopupWidth))
```
MichaelKetting, line 570: I'm afraid the context menu positoning isn't working yet:
image

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -521,39 +521,88 @@ class DropDownMenu
...
+    popUpDiv.style.right = 'auto';
+
+    const isContextMenu = details.align === 'left';
+    const requiredPopUpWidth = LayoutUtility.GetWidth(popUpDiv);
+    const dropDownMenuButtonWidth = LayoutUtility.GetWidth(referenceElement);
+    const maxLeftAlignedPopupWidth = (space_right - popUpMargin) + dropDownMenuButtonWidth;
+    const maxRightAlignedPopupWidth = (space_left - popUpMargin) + dropDownMenuButtonWidth;
+    const isPopUpSmallerThanViewPort = requiredPopUpWidth <= (document.documentElement.clientWidth - 2 * popUpMargin);
+
+    const alignToLeftOfReferenceElement = () =>
```
MichaelKetting, line 544: @a-ctor I find the solution @iam-mholle has taken query interesting. Given that we now have two complete different approaches for the code used with horizontal and vertical alignment, though, is this requiered due to different technical constraints or should we use the same code style?

My gut feeling says we should align the styles as soon as possible, i.e. in this task or a follow-up task done immediately. Thoughts?

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: Previously, the horizontal positioning was simply a lot simpler than the vertical one. Now both of them are more complex.  The two positioning are separate calculations that do not affect each other.

I don't think it is worth spending the time to align the styles. As they both do their own calculation they do not need to share the same style. If they depended on each other I would also make sure to have a unified code style. That said, we can probably move the two calculations into their own methods. This would make the fact more apparent and improve the readability of the positioning method

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocDropDownMenuUserControl.ascx
```diff
@@ -29,7 +29,8 @@
   <td colSpan=4><remotion:boctextvalue id=FirstNameField runat="server" ReadOnly="True" datasourcecontrol="CurrentObject" PropertyIdentifier="FirstName"></remotion:boctextvalue>&nbsp;<remotion:boctextvalue id=LastNameField runat="server" ReadOnly="True" datasourcecontrol="CurrentObject" PropertyIdentifier="LastName"></remotion:boctextvalue></td></tr>
 <tr>
   <td></td>
-  <td><remotion:BocDropDownMenu id=PartnerField runat="server" datasourcecontrol="CurrentObject" PropertyIdentifier="Partner" Width="20em" ControlServicePath="BocDropDownMenuWebService.asmx" ControlServiceArguments="ControlServiceArgs" /></td>
+  <%-- style="right:0px;position:absolute;width:100px" --%>
+  <td><remotion:BocDropDownMenu  id=PartnerField runat="server" datasourcecontrol="CurrentObject" PropertyIdentifier="Partner" Width="20em" ControlServicePath="BocDropDownMenuWebService.asmx" ControlServiceArguments="ControlServiceArgs" /></td>
```
a-ctor, line 33: Are those changes left-over test code?

a-ctor (approved):

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocDropDownMenuUserControl.ascx
iam-mholle: Yes, part of the [!d testing](https://github.com/re-motion/Framework/pull/391/commits/fe628c034ecb57ccbaba94f33128fa5d408c64d7) commit.

a-ctor: Thats what I get for looking at the whole diff :p

MichaelKetting, line 33: drop demo

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocDropDownMenuUserControl.cs
```diff
@@ -53,7 +54,7 @@ private void InitializeMenuItems ()
 
       menuItem = new BocMenuItem();
       menuItem.ItemID = "Open";
-      menuItem.Text = WebString.CreateFromText("Open");
+      menuItem.Text = WebString.CreateFromText(string.Join(" ", Enumerable.Repeat("Open", 100)));
```
MichaelKetting, line 57: drop demo

MichaelKetting (approved):

PR #396, opened 2022-04-20
RM-8434 Generate SetupDB script from the loaded type definitions

a-ctor:
https://re-motion.atlassian.net/browse/RM-8434


Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -0,0 +1,100 @@
...
+
+      var typeDefinitions = MappingConfiguration.Current.GetTypeDefinitions()
+          .Where(e => e.StorageEntityDefinition.StorageProviderID is c_testDomainProviderID or TableInheritanceMappingTest.TableInheritanceTestDomainProviderID)
+          .Where(e => !Attribute.IsDefined(e.ClassType, typeof(ExcludeFromTestDomainDBAttribute)));
+
+      return scriptGenerator.GetScripts(typeDefinitions)
+          .Select(e => new Script(e.StorageProviderDefinition, FixGeneratedScript(e.SetUpScript), FixGeneratedScript(e.TearDownScript)));
+    }
+
+    private void SaveStoredScript (string name, string newContent) // todo dev suffix
```
MichaelKetting, line 79: What's this TODO here?

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+      var filePath = Path.GetFullPath(Path.Combine(TestContext.CurrentContext.TestDirectory, "../../../Database", name));
+      Assert.True(File.Exists(filePath), "Could not find stored SQL script '{0}'.", filePath);
+
+      File.WriteAllText(filePath, newContent);
+    }
+
+    private string FixGeneratedScript (string script)
```
MichaelKetting, line 87:
```suggestion
    private string PatchGeneratedScript (string script)
```

Fix has too much of 'repair'

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -0,0 +1,100 @@
...
+          pd => pd.Factory.CreateSchemaScriptBuilder(pd),
+          new RdbmsStorageEntityDefinitionProvider(),
+          new ScriptToStringConverter());
+
+      var typeDefinitions = MappingConfiguration.Current.GetTypeDefinitions()
+          .Where(e => e.StorageEntityDefinition.StorageProviderID is c_testDomainProviderID or TableInheritanceMappingTest.TableInheritanceTestDomainProviderID)
+          .Where(e => !Attribute.IsDefined(e.ClassType, typeof(ExcludeFromTestDomainDBAttribute)));
+
+      return scriptGenerator.GetScripts(typeDefinitions)
+          .Select(e => new Script(e.StorageProviderDefinition, FixGeneratedScript(e.SetUpScript), FixGeneratedScript(e.TearDownScript)));
```
MichaelKetting, line 76:
```suggestion
          .Select(e => new Script(e.StorageProviderDefinition, PatchGeneratedScript(e.SetUpScript), PatchGeneratedScript(e.TearDownScript)));
```

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    private string FixGeneratedScript (string script)
+    {
+      return c_scriptPrefix + script.Replace("USE TestDomain", "USE DBPrefix_TestDomain");
+    }
+
+    private string GetStoredScript (string name)
+    {
+      var filePath = Path.GetFullPath(Path.Combine(TestContext.CurrentContext.TestDirectory, $"Database/{name}"));
+      Assert.True(File.Exists(filePath), "Could not find stored SQL script '{0}'.", filePath);
```
MichaelKetting, line 95:
```suggestion
      Assert.That(filePath, Does.Exist, "Could not find stored SQL script.");
```

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -0,0 +1,100 @@
...
+          .Where(e => !Attribute.IsDefined(e.ClassType, typeof(ExcludeFromTestDomainDBAttribute)));
+
+      return scriptGenerator.GetScripts(typeDefinitions)
+          .Select(e => new Script(e.StorageProviderDefinition, FixGeneratedScript(e.SetUpScript), FixGeneratedScript(e.TearDownScript)));
+    }
+
+    private void SaveStoredScript (string name, string newContent) // todo dev suffix
+    {
+      var filePath = Path.GetFullPath(Path.Combine(TestContext.CurrentContext.TestDirectory, "../../../Database", name));
+      Assert.True(File.Exists(filePath), "Could not find stored SQL script '{0}'.", filePath);
```
MichaelKetting, line 82:
```suggestion
      Assert.That(filePath, Does.Exist, "Could not find stored SQL script.");
```

MichaelKetting, line 82: Am not sure this guard is a good idea. It does help preventing us writing the file into the wrong location. however, if we ever add a new file, this would fail. Don't know which is more important.

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_Patches.sql
```diff
@@ -0,0 +1,533 @@
...
+  DROP PROCEDURE rpf_testSPQueryWithParameter;
+GO
+
+CREATE PROCEDURE rpf_testSPQueryWithParameter
+  @customerID uniqueidentifier
+AS
+  SELECT * FROM [Order] WHERE [CustomerID] = @customerID ORDER BY [OrderNo] ASC
+GO
+
+-- ==============================================
```
MichaelKetting, line 194: What's the deal with these types? Why do we have to do those manually?

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
a-ctor: TIL that there is a `Does.Exist`

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
a-ctor: You just have to add an empty file to start with. I prefer keeping the check to ensure that stuff breaks if the path logic is faulty or if the files are moved.

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_Patches.sql
a-ctor: Those are located in the default storage provider, which we did not include in the automatic schema generation

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
MichaelKetting: :) with NUnit, we may assume that there's an api to deal with collection-stuff, number-compare, String-contents, and FileSystem stuff. I usually try to assume someone already did this when I need a check and try to disprove it with google / docs :)

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_Patches.sql
MichaelKetting: I think that's not by design. we should fix this instead of doing a patch since they're perfectly fine c# types that result in tables/views

a-ctor: They should be moved into the correct storage provider? Or should we include the default storage provider?

MichaelKetting: Don't know if adding the default provider or adding a new provider group mapped to the same storage provider or simply moving those types to the testdomain provider group is the best choice. I probably would go with the last choice first and see where it leads.

MichaelKetting (approved):

PR #397, opened 2022-04-25
RM-7060: Fix exception being thrown when deselecting a tab preceding a…

iam-mholle:
…n invisible tab

### DIFF
--- a/Remotion/Web/Core/UI/Controls/WebTabCollection.cs
+++ b/Remotion/Web/Core/UI/Controls/WebTabCollection.cs
@@ -18,10 +18,10 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Web.UI;
 using Remotion.Reflection;
 using Remotion.Utilities;
-using Remotion.Web.Utilities;
 
 namespace Remotion.Web.UI.Controls
 {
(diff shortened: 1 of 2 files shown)


Remotion/Web/UnitTests/Core/UI/Controls/WebTabCollectionTest.cs
```diff
@@ -401,6 +401,34 @@ public void SelectInvisibleTab ()
...
+    _tabStrip.Tabs.Add(_tab1);
+    _tabStrip.Tabs.Add(_tab2);
+    _tab0.IsSelected = true;
+    _tab1.IsVisible = false;
+    Assert.That(_tabStrip.SelectedTab, Is.SameAs(_tab0));
+    _tab0.IsVisible = false;
+
+    Assert.That(_tabStrip.SelectedTab, Is.SameAs(_tab2));
+  }
+
```
MichaelKetting, line 431: Something from the tests to verify:
basically, the semantic is, to only select the first visible previous tab, if there is no visible next tab.

I think we need a test that shows that 
tab0.visbile=true
tab1.visibible=true
tab2.visible=false
tab3.visible=true
tab1.selected=true;
tab1.visible=false;
assert tab3.selected==true

MichaelKetting (changes requested):

iam-mholle:
The old behaviour should be restored, test cases should fit your requirements now.

Remotion/Web/Core/UI/Controls/WebTabCollection.cs
```diff
@@ -148,23 +148,13 @@ protected void DeselectTab (WebTab tab, int index)
...
-          else
-            _tabStrip.SetSelectedTabInternal(null);
-        }
-        else
-        {
-          WebTab nextTab = (WebTab)InnerList[index + 1]!;
-          _tabStrip.SetSelectedTabInternal(nextTab);
-        }
+        var webTabs = InnerList.Cast<WebTab>().ToList();
+        var nextTab = webTabs.Skip(index + 1).FirstOrDefault(IsSelectable) ?? webTabs.Take(index).LastOrDefault(IsSelectable);
```
MichaelKetting, line 152: readability and allocation prevention (the ToList isn't needed, we can just to a cast twice):
```suggestion
        var tabsBeforeSelectedTab = InnerList.Cast<WebTab>().Take(index);
        var tabsAfterSelectedTab = InnerList.Cast<WebTab>().Skip(index + 1);
        var nextTab = tabsAfterSelectedTab.FirstOrDefault(IsSelectable) ?? tabsBeforeSelectedTab .LastOrDefault(IsSelectable);
```

If we didn't incur the allocation cost for the array, we could also do C# Range syntax.

MichaelKetting (approved):
Readability optimization

PR #410, opened 2022-05-02
RM-8463 - bugfix/RM-8463-BocAutoCompleteReferenceValue-Tab-Key-Is-Broken

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-8463

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
@@ -379,7 +379,6 @@ namespace Remotion.BocAutoCompleteReferenceValue
                 case KEY.TAB:
                 case KEY.ESC:
                     // re-motion: block event bubbling
-                    event.preventDefault();
                     event.stopPropagation();
                     const wasVisible = select.visible();
                     state.mouseDownOnSelect = false;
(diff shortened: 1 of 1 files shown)


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -400,9 +398,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
-                            // stop default to prevent a form submit, Opera needs special handling
-                            blockSubmit = true;
+                        if (options.isAutoPostBackEnabled) {
+                            // stop default on auto-postback since the change-event already takes care of this.
+                            event.preventDefault();
+                            return false;
+                        } else if (wasVisible) {
+                            // stop default for visible dropdown options since RETURN should simply select the current option without triggering a postback.
+                            event.preventDefault();
                             return false;
```
MichaelKetting, line 408: Consider line 415 (ESC) also `event.preventDefault()` for `return false;`

a-ctor (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
MichaelKetting: @a-ctor this discussion about preventDefault wasn't considered during the review. The latest code now uses the updated preventDefault semantsics.

a-ctor (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: I don't know if we need `.preventDefault` here but it does not hurt. I don't think that the browser has a default behavior for ESC

MichaelKetting: This is true, but given the original code defined it as `false`, the `preventDefault()` preserves the original semantics regardless of it being needed.

PR #419, opened 2022-05-03
RM-8351 Extend the reflection loader to also load interfaces metad

a-ctor:
https://re-motion.atlassian.net/browse/RM-8351

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Mapping/Builder/ClassDefinitionBuilderNodeTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Mapping/Builder/ClassDefinitionBuilderNodeTest.cs
@@ -14,60 +14,113 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using Moq;
 using NUnit.Framework;
 using Remotion.Data.DomainObjects.Mapping;
 using Remotion.Data.DomainObjects.Mapping.Builder;
 using Remotion.Data.DomainObjects.UnitTests.Mapping.TestDomain.Integration;
-using Remotion.Development.UnitTesting;
+using Remotion.Development.UnitTesting.Enumerables;
 using Remotion.Development.UnitTesting.NUnit;
+using Remotion.FunctionalProgramming;
 
 namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
 {
   [TestFixture]
   public class ClassDefinitionBuilderNodeTest
   {
-    private abstract class Base : DomainObject
+    private interface ITop : IDomainObject
     {
     }
 
-    private abstract class Sub : Base
+    private interface IBase : IDomainObject
+    {
+    }
+
+    private interface ISub : IBase
+    {
+    }
+
+    private class Base : DomainObject
+    {
+    }
+
+    private class Sub : Base
     {
     }
 
     [TestDomain]
-    private abstract class SubWithStorageGroup : Base
+    private class SubWithStorageGroup : Base
     {
     }
 
     [Test]
     public void Initialize ()
     {
-      var classDefinitionBuilderNode = new ClassDefinitionBuilderNode(typeof(string), null);
+      var classDefinitionBuilderNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
 
-      Assert.That(classDefinitionBuilderNode.Type, Is.EqualTo(typeof(string)));
-      Assert.That(classDefinitionBuilderNode.BaseClass, Is.Null);
-      Assert.That(classDefinitionBuilderNode.DerivedClasses, Is.Empty);
+      Assert.That(classDefinitionBuilderNode.Type, Is.EqualTo(typeof(Base)));
+      Assert.That(classDefinitionBuilderNode.BaseClassNode, Is.Null);
+      Assert.That(classDefinitionBuilderNode.ImplementedInterfaceNodes, Is.Empty);
+      Assert.That(classDefinitionBuilderNode.DerivedClassNodes, Is.Empty);
       Assert.That(classDefinitionBuilderNode.ClassDefinition, Is.Null);
       Assert.That(classDefinitionBuilderNode.IsLeafNode, Is.True);
+      Assert.That(classDefinitionBuilderNode.IsInheritanceRoot, Is.True);
     }
 
     [Test]
     public void Initialize_WithBaseClass_AddsInstanceToBaseClass ()
     {
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
 
-      Assert.That(subNode.BaseClass, Is.SameAs(baseNode));
+      Assert.That(subNode.BaseClassNode, Is.SameAs(baseNode));
+      Assert.That(subNode.IsInheritanceRoot, Is.False);
       Assert.That(baseNode.IsLeafNode, Is.False);
-      Assert.That(baseNode.DerivedClasses, Is.EqualTo(new[] { subNode }));
+      Assert.That(baseNode.DerivedClassNodes, Is.EqualTo(new[] { subNode }));
+      Assert.That(baseNode.IsInheritanceRoot, Is.True);
+    }
+
+    [Test]
+    public void Initialize_WithImplementedInterfaces_AddsInstanceToImplementedInterfaces ()
+    {
+      var baseInterfaceNode = new InterfaceDefinitionBuilderNode(typeof(IBase), Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Sub), null, EnumerableUtility.Singleton(baseInterfaceNode));
+
+      Assert.That(baseNode.ImplementedInterfaceNodes, Is.EqualTo(new[] { baseInterfaceNode }));
+      Assert.That(baseInterfaceNode.IsLeafNode, Is.False);
+      Assert.That(baseInterfaceNode.ImplementingClassNodes, Is.EqualTo(new[] { baseNode }));
+    }
+
+    [Test]
+    public void Initialize_WithAlreadyImplementedInterfaces_IsNotAddedToImplementedInterfaces ()
+    {
+      var baseInterfaceNode = new InterfaceDefinitionBuilderNode(typeof(IBase), Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, EnumerableUtility.Singleton(baseInterfaceNode));
+      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, EnumerableUtility.Singleton(baseInterfaceNode));
+
+      Assert.That(baseNode.ImplementedInterfaceNodes, Is.EqualTo(new[] { baseInterfaceNode }));
+      Assert.That(subNode.ImplementedInterfaceNodes, Is.Empty);
+    }
+
+    [Test]
+    public void Initialize_WithAlreadyImplementedInterfacesButInheritanceRoot_IsAddedToImplementedInterfaces ()
+    {
+      var baseInterfaceNode = new InterfaceDefinitionBuilderNode(typeof(IBase), Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, EnumerableUtility.Singleton(baseInterfaceNode));
+      var subNode = new ClassDefinitionBuilderNode(typeof(SubWithStorageGroup), baseNode, EnumerableUtility.Singleton(baseInterfaceNode));
+
+      Assert.That(baseNode.ImplementedInterfaceNodes, Is.EqualTo(new[] { baseInterfaceNode }));
+      Assert.That(subNode.ImplementedInterfaceNodes, Is.EqualTo(new[] { baseInterfaceNode }));
     }
 
     [Test]
     public void Initialize_WithNullType_Throws ()
     {
       Assert.That(
-          () => new ClassDefinitionBuilderNode(null, null),
+          () => new ClassDefinitionBuilderNode(null, null, Enumerable.Empty<InterfaceDefinitionBuilderNode>()),
           Throws.ArgumentNullException.With.ArgumentExceptionMessageWithParameterNameEqualTo("type"));
     }
 
@@ -75,9 +128,27 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
     public void Initialize_WithNonClass_Throws ()
     {
       Assert.That(
-          () => new ClassDefinitionBuilderNode(typeof(int), null),
+          () => new ClassDefinitionBuilderNode(typeof(int), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>()),
           Throws.ArgumentException
               .With.ArgumentExceptionMessageEqualTo("The specified type must be a class.", "type"));
+    }
+
+    [Test]
+    public void Initialize_WithDomainObject_Throws ()
+    {
+      Assert.That(
+          () => new ClassDefinitionBuilderNode(typeof(DomainObject), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>()),
+          Throws.ArgumentException
+              .With.ArgumentExceptionMessageEqualTo("The specified type must not be the 'DomainObject' class.", "type"));
+    }
+
+    [Test]
+    public void Initialize_WithNonDomainObject_Throws ()
+    {
+      Assert.That(
+          () => new ClassDefinitionBuilderNode(typeof(string), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>()),
+          Throws.ArgumentException
+              .With.ArgumentExceptionMessageEqualTo("The specified type must be derived from 'DomainObject'.", "type"));
     }
 
     [Test]
@@ -87,13 +158,17 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
       var subClassDefinition = ClassDefinitionObjectMother.CreateClassDefinition(baseClass: baseClassDefinition);
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
 
-      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Base), null)).Returns(baseClassDefinition);
-      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition)).Returns(subClassDefinition);
+      mappingObjectFactoryMock
+          .Setup(_ => _.CreateClassDefinition(typeof(Base), null, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
+          .Returns(baseClassDefinition);
+      mappingObjectFactoryMock
+          .Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
+          .Returns(subClassDefinition);
 
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
       // The constructor registers the parent -> child relationship
-      _ = new ClassDefinitionBuilderNode(typeof(Sub), baseNode);
+      _ = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
       subNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
       subNode.EndBuildTypeDefinition();
 
@@ -107,8 +182,8 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
     [Test]
     public void EndBuildTypeDefinition_OnParentNode_AlsoCallEndsBuildTypeDefinitionOnChildNode ()
     {
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
 
       baseNode.EndBuildTypeDefinition();
 
@@ -119,8 +194,8 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
     [Test]
     public void EndBuildTypeDefinition_OnChildNode_AlsoCallsEndBuildTypeDefinitionOnParentNode ()
     {
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
 
       subNode.EndBuildTypeDefinition();
 
@@ -136,14 +211,14 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
 
       mappingObjectFactoryMock
-          .Setup(_ => _.CreateClassDefinition(typeof(Base), null))
+          .Setup(_ => _.CreateClassDefinition(typeof(Base), null, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
           .Returns(baseClassDefinition);
       mappingObjectFactoryMock
-          .Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition))
+          .Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
           .Returns(subClassDefinition);
 
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(Sub), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
       subNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
 
       Assert.That(baseNode.ClassDefinition, Is.SameAs(baseClassDefinition));
@@ -156,23 +231,25 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
       var baseClassDefinition = ClassDefinitionObjectMother.CreateClassDefinition();
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
 
-      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Base), null)).Returns(baseClassDefinition);
+      mappingObjectFactoryMock
+          .Setup(_ => _.CreateClassDefinition(typeof(Base), null, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
+          .Returns(baseClassDefinition);
 
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
       baseNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
       baseNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
 
       Assert.That(baseNode.ClassDefinition, Is.SameAs(baseClassDefinition));
 
-      mappingObjectFactoryMock.Verify(_ => _.CreateClassDefinition(typeof(Base), null), Times.Once);
+      mappingObjectFactoryMock.Verify(_ => _.CreateClassDefinition(typeof(Base), null, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())), Times.Once);
     }
 
     [Test]
-    public void BeginBuildTypeDefinition_ReadOnly_Throws ()
+    public void BeginBuildTypeDefinition_AlreadyConstructedNode_Throws ()
     {
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
 
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
       baseNode.EndBuildTypeDefinition();
 
       Assert.That(
@@ -188,12 +265,12 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
 
       mappingObjectFactoryMock
-          .Setup(_ => _.CreateClassDefinition(typeof(SubWithStorageGroup), null))
+          .Setup(_ => _.CreateClassDefinition(typeof(SubWithStorageGroup), null, It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
           .Returns(subClassDefinition)
           .Verifiable();
 
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(SubWithStorageGroup), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(SubWithStorageGroup), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
       subNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
 
       Assert.That(baseNode.ClassDefinition, Is.Null);
@@ -203,12 +280,33 @@ namespace Remotion.Data.DomainObjects.UnitTests.Mapping.Builder
     }
 
     [Test]
+    public void BeginBuildTypeDefinition_WithImplementedInterfaces_BuildsAndLinksInterfaces ()
+    {
+      var implementedInterfaceDefinition = InterfaceDefinitionObjectMother.CreateInterfaceDefinition(typeof(IBase));
+      var classDefinition = ClassDefinitionObjectMother.CreateClassDefinition(classType: typeof(Base), implementedInterfaces: new[] { implementedInterfaceDefinition });
+      var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
+
+      mappingObjectFactoryMock
+          .Setup(_ => _.CreateInterfaceDefinition(typeof(IBase), It.Is<IEnumerable<InterfaceDefinition>>(e => e.IsEmpty())))
+          .Returns(implementedInterfaceDefinition);
+      mappingObjectFactoryMock
+          .Setup(_ => _.CreateClassDefinition(typeof(Base), null, It.Is<IEnumerable<InterfaceDefinition>>(e => e.SequenceEqual(new[] { implementedInterfaceDefinition }))))
+          .Returns(classDefinition);
+
+      var implementedInterfaceNode = new InterfaceDefinitionBuilderNode(typeof(IBase), Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var classNode = new ClassDefinitionBuilderNode(typeof(Base), null, new[] { implementedInterfaceNode });
+      classNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
+
+      Assert.That(classNode.ClassDefinition, Is.EqualTo(classDefinition));
+    }
+
+    [Test]
     public void BeginBuildTypeDefinition_WithNonLeafNode_DoesNothing ()
     {
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>();
 
-      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null);
-      var subNode = new ClassDefinitionBuilderNode(typeof(SubWithStorageGroup), baseNode);
+      var baseNode = new ClassDefinitionBuilderNode(typeof(Base), null, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
+      var subNode = new ClassDefinitionBuilderNode(typeof(SubWithStorageGroup), baseNode, Enumerable.Empty<InterfaceDefinitionBuilderNode>());
 
       baseNode.BeginBuildTypeDefinition(mappingObjectFactoryMock.Object);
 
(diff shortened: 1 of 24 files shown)


Remotion/Data/DomainObjects/Mapping/Builder/TypeDefinitionBuilderNode.cs
```diff
@@ -54,7 +54,7 @@ public void BeginBuildTypeDefinition (IMappingObjectFactory mappingObjectFactory
     /// <summary>
     /// Creates a <see cref="TypeDefinition"/> for this node and links it to its parents.
     /// </summary>
-    protected abstract void BuildTypeDefinition (IMappingObjectFactory mappingObjectFactory);
+    internal abstract void BuildTypeDefinition (IMappingObjectFactory mappingObjectFactory);
```
MichaelKetting, line 57: We should use internal only for testing completely internal types. making protected members internal means we're breaking encapsulation because we no longer declare this api as only-inside-this-type(hierarchy). 

For testing protected apis there's always the option of deriving the type in the tests as we do already and provide a public method for testing via the testable-type.

Remotion/Data/DomainObjects/Mapping/Builder/InterfaceDefinitionBuilderNode.cs
```diff
@@ -0,0 +1,144 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Mapping.Builder
+{
+  [DebuggerDisplay("Builder node for interface '{Type}'")]
+  public class InterfaceDefinitionBuilderNode : TypeDefinitionBuilderNode
```
MichaelKetting, line 26: xmldoc

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -105,5 +139,19 @@ private void AddDerivedClass (ClassDefinitionBuilderNode derivedClass)
...
       _derivedClasses.Add(derivedClass);
     }
+
+    private bool ImplementsInterface (InterfaceDefinitionBuilderNode interfaceDefinitionBuilderNode)
+    {
+      if (ImplementedInterfaces.Contains(interfaceDefinitionBuilderNode))
+        return true;
+
+      if (IsInheritanceRoot)
+        return false;
```
MichaelKetting, line 149: I think we can drop this. BaseClass already covers this case and removes all the questions about whether interitanceroot has interfaces or not (i.e. it's not relevant here?)
```suggestion
```

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -25,26 +27,48 @@ namespace Remotion.Data.DomainObjects.Mapping.Builder
...
         throw new ArgumentException("The specified base class must not be a constructed node.", "baseClass");
 
+      var implementedInterfacesArray = implementedInterfaces.ToArray();
+      if (implementedInterfacesArray.Any(e => e.IsConstructed))
+        throw new ArgumentException("The specified implemented interfaces must not be constructed.", "implementedInterfaces");
+
+      IsInheritanceRoot = ReflectionUtility.IsInheritanceRoot(type);
+
       BaseClass = baseClass;
+      ImplementedInterfaces = IsInheritanceRoot || baseClass == null
```
MichaelKetting, line 65:
```suggestion
      ImplementedInterfaces = baseClass == null
```
This should be identical?

Unless this is because baseClass can become not-null during processing. If that makes a difference for evaluating ImplementedInterfaces, I think we have a design problem because the code becomes strange.

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -64,13 +88,21 @@ protected override void BuildTypeDefinition (IMappingObjectFactory mappingObject
         return;
 
       ClassDefinition? baseClass = null;
-      if (BaseClass != null && !ReflectionUtility.IsInheritanceRoot(Type))
+      if (BaseClass != null && !IsInheritanceRoot)
```
MichaelKetting, line 91: If the other usages of IsInheritanceRoot are gone again, we can inline here.

Remotion/Data/DomainObjects/Mapping/TypeDefinitionFactory.cs
```diff
@@ -70,35 +70,56 @@ public TypeDefinition[] CreateTypeDefinitions (IEnumerable<Type> types)
...
         Dictionary<Type, TypeDefinitionBuilderNode> builderNodesLookup,
         Type type)
     {
       ArgumentUtility.CheckNotNull("type", type);
 
       if (type.IsClass)
       {
-        TypeDefinitionBuilderNode? baseClass = null;
+        if (!ReflectionUtility.IsDomainObject(type) || type == typeof(DomainObject))
+          return null;
```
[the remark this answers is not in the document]
MichaelKetting: Update: This was introduced because of the extra guards in the builder-ctors? Can the builder still make nullitems, now?

Remotion/Data/DomainObjects.UnitTests/Mapping/Builder/ClassDefinitionBuilderNodeTest.cs
```diff
@@ -14,86 +14,158 @@
...
     {
       var baseClassDefinition = ClassDefinitionObjectMother.CreateClassDefinition();
       var subClassDefinition = ClassDefinitionObjectMother.CreateClassDefinition(baseClass: baseClassDefinition);
       var mappingObjectFactoryMock = new Mock<IMappingObjectFactory>(MockBehavior.Strict);
 
-      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Base), null)).Returns(baseClassDefinition);
-      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition)).Returns(subClassDefinition);
+      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Base), null, It.Is<IEnumerable<InterfaceDefinition>>(f => !f.Any()))).Returns(baseClassDefinition);
+      mappingObjectFactoryMock.Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition, It.Is<IEnumerable<InterfaceDefinition>>(f => !f.Any())))
+          .Returns(subClassDefinition);
```
MichaelKetting, line 163:
```suggestion
      mappingObjectFactoryMock
          .Setup(_ => _.CreateClassDefinition(typeof(Sub), baseClassDefinition, It.Is<IEnumerable<InterfaceDefinition>>(f => !f.Any())))
          .Returns(subClassDefinition);
```

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -64,13 +88,21 @@ protected override void BuildTypeDefinition (IMappingObjectFactory mappingObject
...
       }
 
-      ClassDefinition = mappingObjectFactory.CreateClassDefinition(Type, baseClass);
+      var implementedInterfaces = new List<InterfaceDefinition>();
+      foreach (var implementedInterface in ImplementedInterfaces)
+      {
+        implementedInterface.BuildTypeDefinition(mappingObjectFactory);
+        if (implementedInterface.InterfaceDefinition != null)
+          implementedInterfaces.Add(implementedInterface.InterfaceDefinition);
+      }
```
MichaelKetting, line 103: am I missing the unit tests for this?

Remotion/Data/DomainObjects.UnitTests/Mapping/RelationDefinitionChecker.cs
```diff
@@ -69,7 +69,7 @@ public void Check (RelationDefinition expectedDefinition, RelationDefinition act
       CheckEndPointDefinitions(expectedDefinition, actualDefinition);
     }
 
-    private void CheckEndPointDefinitions (RelationDefinition expectedRelationDefinition, RelationDefinition actualRelationDefinition) // TODO R2I this needs to be rewritten for TypeDefinition
+    private void CheckEndPointDefinitions (RelationDefinition expectedRelationDefinition, RelationDefinition actualRelationDefinition)
```
MichaelKetting, line 72: what has changed that no rewrite/extension is needed?

Remotion/Data/DomainObjects.UnitTests/Mapping/RelationEndPointDefinitionChecker.cs
```diff
@@ -44,11 +44,10 @@ public class RelationEndPointDefinitionChecker
...
           expectedEndPointDefinition.PropertyName);
 
-      // TODO R2I Mapping: Do actually checking for TypeDefinition not only ClassDefinition
       Assert.AreEqual(
-          ((ClassDefinition)expectedEndPointDefinition.TypeDefinition).ID,
-          ((ClassDefinition)actualEndPointDefinition.TypeDefinition).ID,
-          "ClassDefinition of end point definitions (property name: '{0}') does not match.",
+          expectedEndPointDefinition.TypeDefinition.Type,
+          actualEndPointDefinition.TypeDefinition.Type,
+          "Type of end point definitions (property name: '{0}') does not match.",
```
MichaelKetting, line 50:
```suggestion
          "TypeDefinition of end point definitions (property name: '{0}') does not match.",
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -61,29 +68,20 @@ public void Check (TypeDefinition expectedDefinition, TypeDefinition actualDefin
...
-          }
-
-          CheckClassDefinition(expectedClassDefinition, actualClassDefinition);
-          CheckDerivedClasses(expectedClassDefinition, actualClassDefinition);
-        }
-        else
+        if (!actualDefinitions.TryGetValue(expectedDefinition.Type, out var actualDefinition))
         {
-          Assert.Fail("Unsupported type definition cannot be checked.");
+          Assert.Fail("The two type definitions must have a matching type.");
```
MichaelKetting, line 73: we should provide the names for diagnostic reasons. not fun, having this without infos.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -41,6 +41,13 @@ public void Check (TypeDefinition expectedDefinition, TypeDefinition actualDefin
 
         CheckClassDefinition(expectedClassDefinition, actualClassDefinition);
       }
+      else if (expectedDefinition is InterfaceDefinition expectedInterfaceDefinition)
+      {
+        if (actualDefinition is not InterfaceDefinition actualInterfaceDefinition)
+          throw new InvalidOperationException("The two type definitions must have a matching type.");
```
MichaelKetting, line 47: here and with classdef, the name of the two defs should be added.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
       Assert.AreEqual(
-          expectedDerivedClasses.Count(),
-          actualDerivedClasses.Count(),
-          "Number of derived classes of class definition '{0}' does not match.",
-          expectedClassDefinition.ID);
+          expectedCollection.Count,
+          actualCollection.Count,
+          "Number of {0}s of {1} definition '{2}' does not match",
+          itemDescription,
+          containingTypeDefinition is ClassDefinition ? "class" : "interface",
```
MichaelKetting, line 281:
```suggestion
          "Number of {0} of {1} definition '{2}' does not match",
          itemDescription,
          containingTypeDefinition is ClassDefinition ? "classes" : "interfaces",
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
+        throw new InvalidOperationException("The specified expected collection contains interface definitions with the same type.");
+      if (actualCollection.Select(e => e.ID).Distinct().Count() != actualCollection.Count)
+        throw new InvalidOperationException("The specified actual collection contains interface definitions with the same type.");
+
+      Assert.AreEqual(
+          expectedCollection.Count,
+          actualCollection.Count,
+          "Number of {0}es of {1} definition '{2}' does not match",
+          itemDescription,
+          containingTypeDefinition is ClassDefinition ? "class" : "interface",
```
MichaelKetting, line 337:
```suggestion
          "Number of {0} of {1} definition '{2}' does not match",
          itemDescription,
          containingTypeDefinition is ClassDefinition ? "classes" : "interfaces",
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
-      var actualDerivedClassesDictionary = actualDerivedClasses.ToDictionary(cd => cd.ID);
-      foreach (ClassDefinition expectedDerivedClass in expectedDerivedClasses)
+    private void CheckClassDefinitionCollection (
+        IReadOnlyCollection<ClassDefinition> expectedCollection,
+        IReadOnlyCollection<ClassDefinition> actualCollection,
+        TypeDefinition containingTypeDefinition,
+        string itemDescription)
+    {
+      if (expectedCollection.Select(e => e.ID).Distinct().Count() != expectedCollection.Count)
+        throw new InvalidOperationException("The specified expected collection contains interface definitions with the same type.");
```
MichaelKetting, line 328:
```suggestion
        throw new InvalidOperationException("The specified expected collection contains multiple interface definitions for the same type.");
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
+      if (actualCollection.Select(e => e.ID).Distinct().Count() != actualCollection.Count)
+        throw new InvalidOperationException("The specified actual collection contains interface definitions with the same type.");
```
MichaelKetting, line 330:
```suggestion
        throw new InvalidOperationException("The specified actual collection contains multiple interface definitions for the same type.");
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
-        IEnumerable<ClassDefinition> actualDerivedClasses,
-        ClassDefinition expectedClassDefinition)
+    private void CheckInterfaceDefinitionCollection (
+        IReadOnlyList<InterfaceDefinition> expectedCollection,
+        IReadOnlyList<InterfaceDefinition> actualCollection,
+        TypeDefinition containingTypeDefinition,
+        string itemDescription)
     {
+      if (expectedCollection.Select(e => e.Type).Distinct().Count() != expectedCollection.Count)
+        throw new InvalidOperationException("The specified expected collection contains interface definitions with the same type.");
```
MichaelKetting, line 272:
```suggestion
        throw new InvalidOperationException("The specified expected collection contains multiple interface definitions for the same type.");
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
+      if (actualCollection.Select(e => e.Type).Distinct().Count() != actualCollection.Count)
+        throw new InvalidOperationException("The specified actual collection contains interface definitions with the same type.");
```
MichaelKetting, line 274:
```suggestion
        throw new InvalidOperationException("The specified actual collection contains multiple interface definitions for the same type.");
```

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Mapping/TypeDefinitionFactory.cs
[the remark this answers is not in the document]
a-ctor: Previously, a builder node was created for `DomainObject` but its `ClassDefinition` was never created. With the introduction of the interfaces this behavior changed to not allow `IDomainObject`/`DomainObject`/or non-domain-object-types to have a builder node. I think that this is better as we make it more explicit that those types should not have a generated `TypeDefinition`.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionFactoryTest.cs
```diff
@@ -113,13 +139,13 @@ public void CreateTypeDefinitions_WithGenericBaseType_UsesGenericTypeDefinition
       var specificSubClass2 = ClassDefinitionObjectMother.CreateClassDefinition(classType: typeof(SpecificSubClass2));
 
       _mappingObjectFactoryMock
-          .Setup(mock => mock.CreateClassDefinition(typeof(GenericBaseClass<>), null))
+          .Setup(mock => mock.CreateClassDefinition(typeof(GenericBaseClass<>), null, It.Is<IEnumerable<InterfaceDefinition>>(f => !f.Any())))
```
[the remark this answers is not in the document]
a-ctor: In some cases I had to use `f` as `e` was already used. But in this case this is just a copy-paste error. I also discovered that I did not change all the existing mock setups to use `_ => _.xxx` so I fixed that as well.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/InterfaceReflector.cs
```diff
@@ -0,0 +1,63 @@
...
+          _type,
+          extendedInterfaces,
+          storageGroupType,
+          defaultStorageClass);
+
+      // todo R2I Mapping: Add properties to interface definition
+      interfaceDefinition.SetPropertyDefinitions(new PropertyDefinitionCollection());
+
+      // todo R2I Mapping: Add relation end points to interface definition
+      interfaceDefinition.SetRelationEndPointDefinitions(new RelationEndPointDefinitionCollection());
```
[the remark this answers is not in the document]
a-ctor: Yes it is done in a second step. I agree that the cut is kind of strange here but I did not know how much work the properties/relationships would require at the time.

Remotion/Data/DomainObjects/Mapping/Builder/TypeDefinitionBuilderNode.cs
a-ctor: The internal here is necessary because `ClassDefinitionBuilderNode` needs to call `BuildTypeDefinition` on the `InterfaceDefinitionBuilderNode`. Making it public would cause problems as we now have a `BuildTypeDefinition` method in addition to `Start/EndBuildTypeDefinition`

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
a-ctor: I don't think that we should drop this check here. If a node has `IsInteritanceRoot == true` then its `TypeDefinition` is not linked to the parent as is the case with the node. As such, the check in the node also should not ask the parent if it implements a certain interface. For our current scenario it might be okay to drop, but I think that the current check better represents the behavior we want.

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
a-ctor: No we need the check for `IsInheritanceRoot` here. Initially the node tree is built using just the base type/interface types information. So for an inheritance root it is possible that `baseClass != null`. In such a case we want to pull all the interfaces onto the inheritance root as the base class is not linked afterwards.

Remotion/Data/DomainObjects.UnitTests/Mapping/RelationDefinitionChecker.cs
a-ctor: The todo was just stale and the code part has already been changed

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionFactoryTest.cs
```diff
@@ -134,5 +160,96 @@ public void CreateTypeDefinitions_WithNonClassOrInterface_Throws ()
           Throws.InvalidOperationException
               .With.Message.EqualTo("Cannot create a builder node for type 'System.TypeCode' because it is not a class-type."));
     }
+
```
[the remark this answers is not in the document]
a-ctor: The null checks were added to private parts of the API and the behavior of the public API did not change. The existing test plus the tests that were added with this PR should cover the statements.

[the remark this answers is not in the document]
a-ctor: The rest of the functionality is covered by the `ClassDefinitionBuilderNode` + `InterfaceDefinitionBuilderNode`. The `TypeDefinitionFactory` is just using those correctly and as such I built the test to cover more specific use cases.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/MappingReflector.cs
```diff
@@ -129,7 +129,7 @@ public RelationDefinition[] GetRelationDefinitions (IDictionary<Type, TypeDefini
 
     private IEnumerable<Type> GetDomainObjectTypes ()
     {
-      return (from type in _typeDiscoveryService.GetTypes(typeof(DomainObject), excludeGlobalTypes: false).Cast<Type>()
+      return (from type in _typeDiscoveryService.GetTypes(typeof(IDomainObject), excludeGlobalTypes: false).Cast<Type>()
```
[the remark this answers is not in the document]
a-ctor: > There should be an assertion for this.

You mean a test?

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionFactoryTest.cs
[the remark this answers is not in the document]
MichaelKetting: This is such a big ball of yarn.  Even though it might be redundant, having a straight forward "Create Interface" and "Create Class + Interface" test does seem like the "correct" choice for the API. That way, typedeffactory can start its tests by doing basic stuff. It's a bit of an expectation problem: you expect certain tests for a SUT, if they're missing, it feels like problem.

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
MichaelKetting: I would like to discuss this live. Those rules are a bit hard to wrap the mind around.

Remotion/Data/DomainObjects/Mapping/Builder/TypeDefinitionBuilderNode.cs
a-ctor: I think you had the idea of introducing another non-virtual method that calls the virtual method? But with that we still have the same problem don't we? The new method would have to be public for this to work.

In this case I would leave the code as is. Making the `BuildTypeDefinition` method public or adding another public method that essential just redirects to this method harms the usability more than it improves it.

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -64,13 +88,21 @@ protected override void BuildTypeDefinition (IMappingObjectFactory mappingObject
...
-      if (BaseClass != null && !ReflectionUtility.IsInheritanceRoot(Type))
+      if (BaseClass != null && !IsInheritanceRoot)
       {
         BaseClass.BuildTypeDefinition(mappingObjectFactory);
         baseClass = BaseClass.ClassDefinition;
       }
 
-      ClassDefinition = mappingObjectFactory.CreateClassDefinition(Type, baseClass);
+      var implementedInterfaces = new List<InterfaceDefinition>();
+      foreach (var implementedInterface in ImplementedInterfaces)
```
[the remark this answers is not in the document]
a-ctor: Do you want to rename the properties themselves or just the variables?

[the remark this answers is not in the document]
MichaelKetting: Answer from mail:
I think we should update the properties, too. That way we have symmetric naming and can see the intended behavior without intellisense.

SharedSource/Development/UnitTesting/Enumerables/EnumerableExtensions.cs
```diff
@@ -54,5 +54,13 @@ public static T[] ForceEnumeration<T> (this IEnumerable<T> source)
 
       return source.ToArray();
     }
+
+    /// <summary>
+    /// Returns <see langword="true"/> if the specified <paramref name="source"/> is empty, or <see langword="false"/> if it is not empty.
+    /// </summary>
+    public static bool IsEmpty<T> (this IEnumerable<T> source)
+    {
```
MichaelKetting, line 62: argument null check
https://re-motion.atlassian.net/browse/RMTEAM-510

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/InterfaceReflector.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Linq;
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.ConfigurationLoader.ReflectionBasedConfigurationLoader
+{
+  /// <summary>
+  /// The <see cref="InterfaceReflector"/> is used to build a <see cref="InterfaceReflector"/>.
+  /// </summary>
+  /// <remarks>Derived classes must have a constructor with a matching the <see cref="InterfaceReflector"/>'s constructor signature.</remarks>
```
MichaelKetting, line 28: Why this constraint? is the IR instantiated via reflection?

Remotion/Data/DomainObjects.UnitTests/ReflectionUtilityTest.cs
```diff
@@ -355,9 +375,9 @@ public void IsDomainObject_DerivedFromDomainObject_ReturnsTrue ()
     }
 
     [Test]
-    public void IsDomainObject_IDomainObject_ReturnsFalse ()
+    public void IsDomainObject_IDomainObject_ReturnsTrue ()
```
MichaelKetting, line 378: Ah, yes, this is new. please make a breaking change ticket for this. (yes, there will be more tickets, but most of those will be signature changes that are easier to discover during a code-scan than a semantic change like this. => we need to document this right away)

Remotion/Data/DomainObjects.UnitTests/ReflectionUtilityTest.cs
```diff
@@ -138,6 +142,14 @@ public void IsInheritanceRoot_BaseTypeIsDomainObject ()
       Assert.That(ReflectionUtility.IsInheritanceRoot(type), Is.True);
     }
 
+    [Test]
+    public void IsInheritanceRoot_SupportsIDomainObjectInterfaces ()
+    {
+      var type = typeof(ITestDomainObject);
+
```
MichaelKetting, line 149: Please add the same checks and test-naming as is done in `IsInheritanceRoot_BaseTypeIsDomainObject`. Best to rename the class-based tests to include "class" or "DomainObject" and the interface-based tests to include "interface" or "IDomainObject".

e.g. IsInheritanceRoot_WithClass_AndBaseTypeIsDomainObject_ReturnsTrue (added the returns-suffix since that's good practice anyway)

Also, please add a test for a derived domainobject that introduces interface to show that this isn't an inheritance root. 

There's also more inheritance root tests for derived classes. please check those and add the appropriate itnerface cases, too.

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -37,19 +37,26 @@ public void Check (TypeDefinition expectedDefinition, TypeDefinition actualDefin
...
         if (actualDefinition is not ClassDefinition actualClassDefinition)
-          throw new InvalidOperationException("The two type definitions must have a matching type.");
+          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected a ClassDefinition but it was a '{actualDefinition.GetType().Name}'");
 
         CheckClassDefinition(expectedClassDefinition, actualClassDefinition);
       }
+      else if (expectedDefinition is InterfaceDefinition expectedInterfaceDefinition)
+      {
+        if (actualDefinition is not InterfaceDefinition actualInterfaceDefinition)
+          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected a InterfaceDefinition but it was a '{actualDefinition.GetType().Name}'");
```
MichaelKetting, line 47:
```suggestion
          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected '$(nameof(InterfaceDefinition)' but was '{actualDefinition.GetType().Name}'");
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -37,19 +37,26 @@ public void Check (TypeDefinition expectedDefinition, TypeDefinition actualDefin
       if (expectedDefinition is ClassDefinition expectedClassDefinition)
       {
         if (actualDefinition is not ClassDefinition actualClassDefinition)
-          throw new InvalidOperationException("The two type definitions must have a matching type.");
+          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected a ClassDefinition but it was a '{actualDefinition.GetType().Name}'");
```
MichaelKetting, line 40:
```suggestion
          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected '$(nameof(ClassDefinition)' but was '{actualDefinition.GetType().Name}'");
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -200,36 +198,155 @@ private void CheckClassDefinition (ClassDefinition expectedDefinition, ClassDefi
...
+    }
+
+    private void CheckClassDefinitionCollection (
+        IReadOnlyCollection<ClassDefinition> expectedCollection,
+        IReadOnlyCollection<ClassDefinition> actualCollection,
+        TypeDefinition containingTypeDefinition,
+        string itemDescription)
+    {
+      if (expectedCollection.Select(e => e.ID).Distinct().Count() != expectedCollection.Count)
+        throw new InvalidOperationException("The specified expected collection contains multiple interface definitions with the for type.");
```
MichaelKetting, line 328: Grammar "with the for type". Please reword.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectorTest.cs
```diff
@@ -111,13 +113,37 @@ public void Get_WithDuplicateAssembly ()
...
+    }
+
+    [Test]
+    public void GetTypeDefinitions_DiscoversInterfacesTypes ()
+    {
+      var typeDiscoveryServiceMock = new Mock<ITypeDiscoveryService>();
+      typeDiscoveryServiceMock
+          .Setup(_ => _.GetTypes(typeof(IDomainObject), false))
+          .Returns(Array.Empty<Type>())
+          .Verifiable();
```
MichaelKetting, line 132: I'm not 100% sure if that's the best way to test this. We already have a test for GetTypeDefintions, it's just that it only checks for not-empty right now. if we change this to assert that there's 1 or 2 represantative types in there, we could just add 1 or 2 interface types, too, and thus cover the correct usage of the api. Otherwise, we only test that the api was called, but not that it is used correctly.

Remotion/Data/DomainObjects/Mapping/Builder/InterfaceDefinitionBuilderNode.cs
```diff
@@ -0,0 +1,148 @@
...
+      Assertion.DebugAssert(!implementingClassNode.IsConstructed, "!implementingClass.IsConstructed");
+      Assertion.DebugAssert(!_implementingClassNodes.Contains(implementingClassNode), "!_implementingClasses.Contains(implementingClass)");
+
+      _implementingClassNodes.Add(implementingClassNode);
+    }
+
+    private void AddExtendingInterface (InterfaceDefinitionBuilderNode extendingInterfaceNode)
+    {
+      Assertion.DebugAssert(extendingInterfaceNode.ExtendedInterfaceNodes.Contains(this), "extendingInterface.ExtendedInterfaces.Contains(this)");
+      Assertion.DebugAssert(!_extendingInterfaceNodes.Contains(this), "!_extendingInterfaces.Contains(this)");
```
MichaelKetting, line 135: plz update messages

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -98,12 +133,23 @@ public override void EndBuildTypeDefinition ()
       return ClassDefinition;
     }
 
-    private void AddDerivedClass (ClassDefinitionBuilderNode derivedClass)
+    private void AddDerivedClass (ClassDefinitionBuilderNode derivedClassNode)
     {
-      Assertion.DebugAssert(derivedClass.BaseClass == this, "derivedClass.BaseClass == this");
-      Assertion.DebugAssert(!_derivedClasses.Contains(derivedClass), "!_derivedClasses.Contains(derivedClass)");
+      Assertion.DebugAssert(derivedClassNode.BaseClassNode == this, "derivedClass.BaseClass == this");
+      Assertion.DebugAssert(!_derivedClassNodes.Contains(derivedClassNode), "!_derivedClasses.Contains(derivedClass)");
```
MichaelKetting, line 139: plz update messages

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -25,32 +27,55 @@ namespace Remotion.Data.DomainObjects.Mapping.Builder
...
         throw new ArgumentException("The specified type must be a class.", "type");
-      if (baseClass?.IsConstructed == true)
-        throw new ArgumentException("The specified base class must not be a constructed node.", "baseClass");
-
-      BaseClass = baseClass;
-      BaseClass?.AddDerivedClass(this);
+      if (type == typeof(DomainObject))
+        throw new ArgumentException("Cannot create a builder node for the DomainObject type.", "type");
+      if (!ReflectionUtility.IsDomainObject(type))
+        throw new ArgumentException("Cannot create a builder node for a type that is not a domain object.", "type");
```
MichaelKetting, line 55: message: in arg checks, the behavior is usually not listed, the message focuses on the value itself. this reads more like an invalidopex.
 "specified type must not be DomainObject"
"specified  type must be derived from DomainObject"

Remotion/Data/DomainObjects/Mapping/Builder/InterfaceDefinitionBuilderNode.cs
```diff
@@ -0,0 +1,148 @@
...
+    public InterfaceDefinitionBuilderNode (Type type, IEnumerable<InterfaceDefinitionBuilderNode> extendedInterfaceNodes)
+        : base(type)
+    {
+      ArgumentUtility.CheckNotNull("extendedInterfaceNodes", extendedInterfaceNodes);
+      if (!type.IsInterface)
+        throw new ArgumentException("The specified type must be an interface.", "type");
+      if (type == typeof(IDomainObject))
+        throw new ArgumentException("Cannot create a builder node for the IDomainObject type.", "type");
+      if (!typeof(IDomainObject).IsAssignableFrom(type))
+        throw new ArgumentException("Cannot create a builder node for a type that is not a domain object.", "type");
```
MichaelKetting, line 50: see arg check for classdefbuildernode

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectorTest.cs
a-ctor: I removed the mock but changed the supplied type discovery service to use a fixed list of types supplied in the tests. This way, the types are specified as nested types in the test fixture. I don't want to add additional types to another test domain that is so far from the actual test. Especially since this test is a unit test.

Remotion/Data/DomainObjects.UnitTests/ReflectionUtilityTest.cs
```diff
@@ -139,7 +158,27 @@ public void IsInheritanceRoot_BaseTypeIsDomainObject ()
...
     [Test]
-    public void IsInheritanceRoot_TypeHasStorageGroupAttributeApplied_ReturnsTrue ()
+    public void IsInheritanceRoot_WithClassDerivedFromInheritanceRootAndImplementedInterface_ReturnsFalse ()
+    {
+      var type = typeof(AbstractClass);
+      Assertion.IsFalse(ReflectionUtility.IsTypeIgnoredForMappingConfiguration(type));
+      Assertion.IsTrue(type.BaseType == typeof(DomainObject));
+
+      Assert.That(ReflectionUtility.IsInheritanceRoot(type), Is.True);
+    }
```
MichaelKetting, line 168: Returntype-mismatch in test and name

Remotion/Data/DomainObjects.UnitTests/Mapping/TypeDefinitionChecker.cs
```diff
@@ -37,19 +37,26 @@ public void Check (TypeDefinition expectedDefinition, TypeDefinition actualDefin
...
         if (actualDefinition is not ClassDefinition actualClassDefinition)
-          throw new InvalidOperationException("The two type definitions must have a matching type.");
+          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected a {nameof(ClassDefinition)} but it was a '{actualDefinition.GetType().Name}'");
 
         CheckClassDefinition(expectedClassDefinition, actualClassDefinition);
       }
+      else if (expectedDefinition is InterfaceDefinition expectedInterfaceDefinition)
+      {
+        if (actualDefinition is not InterfaceDefinition actualInterfaceDefinition)
+          throw new InvalidOperationException($"The expected and actual type definition for '{expectedDefinition.Type}' differ. Expected a {nameof(InterfaceDefinition)} but it was a '{actualDefinition.GetType().Name}'");
```
MichaelKetting, line 47: I would prefer to go with quoted names instead of using the typename as part of the sentence because it makes the problem more apprent / focuses on the problem. If you do want to keep it like this, you need to fix the articels. "an InterfaceDefinition"

Remotion/Data/DomainObjects/Mapping/Builder/InterfaceDefinitionBuilderNode.cs
```diff
@@ -0,0 +1,148 @@
...
+    public InterfaceDefinitionBuilderNode (Type type, IEnumerable<InterfaceDefinitionBuilderNode> extendedInterfaceNodes)
+        : base(type)
+    {
+      ArgumentUtility.CheckNotNull("extendedInterfaceNodes", extendedInterfaceNodes);
+      if (!type.IsInterface)
+        throw new ArgumentException("The specified type must be an interface.", "type");
+      if (type == typeof(IDomainObject))
+        throw new ArgumentException($"The specified type must not be the {nameof(IDomainObject)} interface.", "type");
+      if (!typeof(IDomainObject).IsAssignableFrom(type))
+        throw new ArgumentException($"The specified type must be a domain object interface.", "type");
```
MichaelKetting, line 50:
```suggestion
        throw new ArgumentException($"The specified type must implement '{nameof(IDomainObject}'.", "type");
```

Remotion/Data/DomainObjects/Mapping/Builder/InterfaceDefinitionBuilderNode.cs
```diff
@@ -0,0 +1,148 @@
...
+    private readonly List<InterfaceDefinitionBuilderNode> _extendingInterfaceNodes = new();
+
+    public InterfaceDefinitionBuilderNode (Type type, IEnumerable<InterfaceDefinitionBuilderNode> extendedInterfaceNodes)
+        : base(type)
+    {
+      ArgumentUtility.CheckNotNull("extendedInterfaceNodes", extendedInterfaceNodes);
+      if (!type.IsInterface)
+        throw new ArgumentException("The specified type must be an interface.", "type");
+      if (type == typeof(IDomainObject))
+        throw new ArgumentException($"The specified type must not be the {nameof(IDomainObject)} interface.", "type");
```
MichaelKetting, line 48:
```suggestion
        throw new ArgumentException($"The specified type must not be the '{nameof(IDomainObject)}' interface.", "type");
```

Remotion/Data/DomainObjects/Mapping/Builder/ClassDefinitionBuilderNode.cs
```diff
@@ -25,32 +27,55 @@ namespace Remotion.Data.DomainObjects.Mapping.Builder
...
         throw new ArgumentException("The specified type must be a class.", "type");
-      if (baseClass?.IsConstructed == true)
-        throw new ArgumentException("The specified base class must not be a constructed node.", "baseClass");
-
-      BaseClass = baseClass;
-      BaseClass?.AddDerivedClass(this);
+      if (type == typeof(DomainObject))
+        throw new ArgumentException($"The specified type must not be the {nameof(DomainObject)} class.", "type");
+      if (!ReflectionUtility.IsDomainObject(type))
+        throw new ArgumentException($"The specified type must be a domain object.", "type");
```
MichaelKetting, line 55:
```suggestion
        throw new ArgumentException($"The specified type must be derived from '{nameof(DomainObject)}'.", "type");
```

MichaelKetting: Interesting tidbit: I only saw the updated text first, not my review notes and the new messages where my first thoughts on the subject. So, I guess I'm consistent with my thoughts on messages and do seem to feel strongly about them :/

MichaelKetting (changes requested):

MichaelKetting (approved):
you will want to fix the red tests now :)

PR #425, opened 2022-05-13
RM-8475 Initialize test database only on-demand

a-ctor:
https://re-motion.atlassian.net/browse/RM-8475


Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -118,5 +124,29 @@ public FakeDomainObjectsConfiguration GetDomainObjectsConfiguration ()
     {
       return new FakeDomainObjectsConfiguration(_mappingLoaderConfiguration, _storageConfiguration, _queryConfiguration);
     }
+
+    public void DisableConnections ()
+    {
+      foreach (var rdbmsProviderDefinition in _storageConfiguration.StorageProviderDefinitions.OfType<RdbmsProviderDefinition>())
+      {
+        var connectionString = rdbmsProviderDefinition.ConnectionString;
+        if (string.IsNullOrEmpty(connectionString))
```
MichaelKetting, line 133: You're using this to skip updating the connection strings. That's not really obvious from reading the code. It would be better to introduce a named flag for this condition. Also, continue vs statement-block: when we have  a nicely-named flag, we don't need `continue`.

About `continue`:
Use `break` when it's not possible to return. Use `continue` if the alternative is really ugly. It's derived from the intention of not using goto-label.

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -118,5 +124,29 @@ public FakeDomainObjectsConfiguration GetDomainObjectsConfiguration ()
...
+          continue;
+
+        _originalConnectionStrings[rdbmsProviderDefinition] = connectionString;
+        PrivateInvoke.SetNonPublicField(rdbmsProviderDefinition, "_connectionString", string.Empty);
```
MichaelKetting, line 137: This performance optimization depends on there being just a few switching between database /  no database. What's the execution time difference when caching getting the field instead of using private invoke?

And what's the difference when there is no IF, just the cached field and the set-field-value?

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -118,5 +124,29 @@ public FakeDomainObjectsConfiguration GetDomainObjectsConfiguration ()
...
+    }
+
+    public void EnableConnections ()
+    {
+      foreach (var rdbmsProviderDefinition in _storageConfiguration.StorageProviderDefinitions.OfType<RdbmsProviderDefinition>())
+      {
+        if (!_originalConnectionStrings.TryGetValue(rdbmsProviderDefinition, out var originalConnectionString))
+          continue;
+
+        PrivateInvoke.SetNonPublicField(rdbmsProviderDefinition, "_connectionString", originalConnectionString);
```
MichaelKetting, line 148:
```suggestion
        if (_originalConnectionStrings.TryGetValue(rdbmsProviderDefinition, out var originalConnectionString))
          PrivateInvoke.SetNonPublicField(rdbmsProviderDefinition, "_connectionString", originalConnectionString);
```

Remotion/Data/DomainObjects.UnitTests/Mapping/TestMappingConfiguration.cs
```diff
@@ -60,6 +60,9 @@ public static TestMappingConfiguration Instance
 
     public static void Initialize ()
```
MichaelKetting, line 61: Rename: `EnsureInitialized` -> makes it obvious on the callsite that this is allowed to be called multiple times.
(also update the exception message in line 55)

Remotion/Data/DomainObjects.UnitTests/DatabaseTest.cs
```diff
@@ -69,11 +70,14 @@ public virtual void TearDown ()
     [OneTimeSetUp]
     public virtual void OneTimeSetUp ()
     {
+      TestDomainDatabase.InitializeDatabase();
+      StandardConfiguration.Instance.EnableConnections();
```
MichaelKetting, line 74: Calling `StandardConfiguration.Instance` here without having `StandardConfiguration.Initialize()` is "wrong". `StandardConfiguration.Instance` would fail without `StandardConfiguration.Initialize()`.

Remotion/Data/DomainObjects.UnitTests/Factories/StandardConfiguration.cs
```diff
@@ -39,6 +39,7 @@ public static StandardConfiguration Instance
     public static void Initialize ()
     {
       s_instance = new StandardConfiguration();
+      s_instance.DisableConnections();
```
MichaelKetting, line 42: so, database access is an opt-in feature.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
```diff
@@ -48,6 +48,13 @@ public virtual void SetUp ()
       _typeDefinitions = typeDefinitions;
     }
 
+    [OneTimeSetUp]
+    public virtual void OneTimeSetUp ()
+    {
+      TestMappingConfiguration.Initialize();
+      StandardConfiguration.Initialize();
```
MichaelKetting, line 55: I take it, the reflection tests fail if there's not `StandardConfiguration`? Which tests are affected?

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
```diff
@@ -48,6 +48,13 @@ public virtual void SetUp ()
...
+    }
+
```
MichaelKetting, line 57: It feels like we should have an Uninitialize-API, too. Otherwise, the static field is always available in later tests. Unitialize (better name?) should name Instance no longer return the value and Initialize can in turn set the same instance once more. This would bring much better isolation.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+      }
+    }
+
+    public static void CleanupDatabase ()
+    {
+      if (!s_initialized)
+        return;
+
+      s_standardMappingDatabaseAgent.SetDatabaseReadWrite(DatabaseTest.DatabaseName);
```
MichaelKetting, line 100: We no longer call s_standardMappingDatabaseAgent.SetDatabaseReadOnly(). -> we can remove this here, too. Might result in a little more performance.

Remotion/Data/DomainObjects.UnitTests/SetupFixture.cs
```diff
@@ -15,87 +15,32 @@
...
+      LogManager.ResetConfiguration();
+      Assert.That(LogManager.GetLogger(typeof(LoggingClientTransactionListener)).IsDebugEnabled, Is.False);
     }
 
     [OneTimeTearDown]
     public void OneTimeTearDown ()
     {
-      var fields = PrivateInvoke.GetNonPublicStaticField(typeof(MappingConfiguration), "s_fields");
-      Assertion.IsNotNull(fields);
-      PrivateInvoke.SetPublicField(fields, "Current", _previousMappingConfigurationContainer);
```
MichaelKetting, line 95: You dropped this. Why do we not need it?

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,104 @@
...
+          var fields = PrivateInvoke.GetNonPublicStaticField(typeof(MappingConfiguration), "s_fields");
+          Assertion.IsNotNull(fields);
+          PrivateInvoke.SetPublicField(fields, "Current", throwingMappingConfigurationContainer);
+        }
+        catch (Exception ex)
+        {
+          Console.WriteLine("SetUpFixture failed: " + ex);
+          Console.WriteLine();
+          throw;
+        }
```
MichaelKetting, line 91: This should be part of SetupFixture, not the helper type.

Remotion/Data/DomainObjects.UnitTests/SetupFixture.cs
a-ctor: It is kind of pointless to set a public static field right before the test run ends anyways as this method is called after all the tests just before the shutdown of the test process.

Remotion/Data/DomainObjects.UnitTests/DatabaseTest.cs
a-ctor: In this case the initialization of the `StandardConfiguration` is done in `TestDomainDatabase.InitializeDatabase()`

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
a-ctor: All of the `When_XXX` tests fail if they haven't set up the `StandardConfiguration` but they also do not inherit from `DatabaseTest` so the initialization needs to be done here.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
a-ctor: While that is true, I don't think that we can do any better. Creating a `MappingConfiguration` costs time and re-creating it every time a fixture ends will increase the test runtime by a lot. The main point here is just that if you run one of the `When_XXX` tests in isolation the test would fail because the `StandardConfiguration` is not set up.

We have the same problem of uninitialization in the `DatabaseTest`. I think that this is not a problem as the configuration is always the same for all the tests and we stop the important mistake of accessing the database without inheriting from `DatabaseTest`.

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
a-ctor: The enable/disable code is only run per fixture setup/teardown and takes about 7.6us. With the time being so low I think we can just leave the code as is.

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
MichaelKetting: Miscummunication: I don't want to re-create it. Just block access to the field during the unitialized-phase / remove it from MappingConfig.Current, etc. That way, we don't leak. It wasn't possible before and if we can do it now cheaply, it would be good to do it right now. 

Otherwise, we may want to call it `EnsureInitialized`

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
MichaelKetting: Loving the consistency :D

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
MichaelKetting, line 137: Missed this earlier: please use a identifiable string instead of empty:
"DATABASE_DISABLED_IN_UNITEST" or something along those lines. Otherwise, the empty string might result in some headscratching down the line.

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
MichaelKetting, line 133: Given the PrivateInboke remains as is, we should just update the IF-structure to drop the `continue`

Remotion/Data/DomainObjects.UnitTests/DatabaseTest.cs
MichaelKetting: Then we should just do it twice (it's idempotent anyway) and thus make it obvious that the intiialization happens regardless of TestDomainDatabase.

I think this might also be a good example why an Uninitialize could be helpful. The DB-Setup needs the standardconfig, but after it's done, removing Instance until it's done explicitly in a test setup would remove potential sideeffects.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/Mapping/MappingReflectionIntegrationTests/MappingReflectionIntegrationTestBase.cs
a-ctor: We can do this but I would prefer to do it in another issue.

Remotion/Data/DomainObjects.UnitTests/DatabaseTest.cs
a-ctor: The initialization is not quite idempotent as it recreates a new instance of `StandardConfiguration`. But apart from that I don't think that there are any effects on the tests. As mentioned in the comment below I would prefer to do this in another issue.

MichaelKetting: Okay, so, separate ticket then (please create). btw: the idea is to keep the reference to this instance alive internally, so there is no new instance between swithcing it off and onagain..

a-ctor: Arrgg. Implicit base constructor calls paired with doing work in the constructor is painful. Tests took 1min longer when I put the `StandardConfiguration.Initialize()` into the `OneTimeSetUp`. Turns out, the base constructor is not quite indempotent and recreates the `MappingConfiguration` from the reflection....

I changed the `Initialize` methods to `EnsureInitialized` to prevent duplicate creation and I also found that the same `DisableConnection` stuff should be done on the `TableInheritanceConfiguration`.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+          var fields = PrivateInvoke.GetNonPublicStaticField(typeof(MappingConfiguration), "s_fields");
+          Assertion.IsNotNull(fields);
+          PrivateInvoke.SetPublicField(fields, "Current", throwingMappingConfigurationContainer);
+        }
+        catch (Exception ex)
+        {
+          Console.WriteLine("TestDomainDatabase.InitializeDatabase failed: " + ex);
+          Console.WriteLine();
+          throw;
+        }
```
MichaelKetting, line 91: Some mis-communication:
this try-catch block is needed in SetupFixture, because Nunit swallows meaningful error information that happens during setup fixtures. If we do this during regular testfixture setups, we should get the full details. please verify. It may be that we do not need it at all, now that we do testfixture.setupfxture instead of setupfixture.setup.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+    private static readonly object s_lock = new();
+
+    private static bool s_initialized;
+    private static StandardMappingDatabaseAgent s_standardMappingDatabaseAgent;
+
+    public static void InitializeDatabase ()
+    {
+      if (s_initialized)
+        return;
+
```
MichaelKetting, line 43:
```suggestion
```
There is no need to do doublechecked locking in the test code. 

And actually, the current code has a race-condition bug in case of multithreading: you set the s_initialzied to true first thing during the lock statement. this means if we actually do need the locking, statement, the other calls would return before the intiialization logic has completed, making this whole thing a moot point.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+      }
+    }
+
+    public static void CleanupDatabase ()
+    {
+      if (!s_initialized)
+        return;
+
+      SqlConnection.ClearAllPools();
```
MichaelKetting, line 100: shouldn't we add this in a lockstatement for s_lock, too?

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+
+          var throwingMappingConfigurationContainer = new DoubleCheckedLockingContainer<IMappingConfiguration>(
+              () =>
+              {
+                throw new InvalidOperationException(
+                    "This test failed to setup the mapping configuration. Did you forget to derive from StandardMappingTest or to call base.SetUp?");
+              });
+          var fields = PrivateInvoke.GetNonPublicStaticField(typeof(MappingConfiguration), "s_fields");
+          Assertion.IsNotNull(fields);
+          PrivateInvoke.SetPublicField(fields, "Current", throwingMappingConfigurationContainer);
```
MichaelKetting, line 84: the missing backup / restore logic we discussed earlier: yes, the SetupFixture was in the root and thus the restore didn't do anything. Now that we have a helper-api, I think the missing reset of the original mapping config is a problem if you start to use the helper in a more granular fashion.  

Alternative:
if you consider that this entire helper type should only be used once to initialize the database, we can actually drop the CleanupDatabase method because we don't need to clear the sql connections at the end of the testrun. the process ends anyway.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+  /// The initialization takes some time so it is only executed if the tests demand it (see <see cref="DatabaseTest"/>).
+  /// </summary>
+  public static class TestDomainDatabase
+  {
+    private static readonly object s_lock = new();
+
+    private static bool s_initialized;
+    private static StandardMappingDatabaseAgent s_standardMappingDatabaseAgent;
+
+    public static void InitializeDatabase ()
```
MichaelKetting, line 39: Since the API is new: please rename to EnsureDatabaseInitialized.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+  /// <summary>
+  /// Provides methods for setting up the test domain database for the integration tests.
+  /// The initialization takes some time so it is only executed if the tests demand it (see <see cref="DatabaseTest"/>).
+  /// </summary>
+  public static class TestDomainDatabase
+  {
+    private static readonly object s_lock = new();
+
+    private static bool s_initialized;
```
MichaelKetting, line 36: s_isInitialized.

flags should always try to use `is` / `has` prefixes.

Remotion/Data/DomainObjects.UnitTests/SetupFixture.cs
```diff
@@ -15,87 +15,32 @@
...
-        PrivateInvoke.SetPublicField(fields, "Current", throwingMappingConfigurationContainer);
-      }
-      catch (Exception ex)
-      {
-        Console.WriteLine("SetUpFixture failed: " + ex);
-        Console.WriteLine();
-        throw;
-      }
+      LogManager.ResetConfiguration();
+      Assert.That(LogManager.GetLogger(typeof(LoggingClientTransactionListener)).IsDebugEnabled, Is.False);
```
MichaelKetting, line 37: This code must remain inside the try-catch block to ensure helpful error output. unless nunit3 already provides helpful messages and stacktraces for exceptions during setupfixtures? (please verify)

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
```diff
@@ -0,0 +1,103 @@
...
+    private static StandardMappingDatabaseAgent s_standardMappingDatabaseAgent;
+
+    public static void InitializeDatabase ()
+    {
+      if (s_initialized)
+        return;
+
+      lock (s_lock)
```
MichaelKetting, line 44: why did you add a lockstatement here but not in the other helper types that where moved from setupfixture to other types.

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -118,5 +124,28 @@ public FakeDomainObjectsConfiguration GetDomainObjectsConfiguration ()
     {
       return new FakeDomainObjectsConfiguration(_mappingLoaderConfiguration, _storageConfiguration, _queryConfiguration);
     }
+
+    public void DisableConnectionsAccess ()
```
MichaelKetting, line 128: It doesn't need to be `DisableDatabaseAccess` from my own suggestion, however `DisableConnectionsAccess` doesn't work for me because it doesn't disable the access to the "connections", what a "connection" is. It disables the access to the databases identified by the connection strings. I would like to a have more targeted method name.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
a-ctor: The cleanup is still located in the `SetupFixture.OneTimeTearDown` so this code is not necessary. It would be better if we could cleanup the DB right after all the tests that needed it ran but this is not possible in NUnit.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
a-ctor: Because the database initialization takes a long time and it might be called from multiple threads if tests are parallelized. The `SetupFixture` had the implicit guarantee that it is called once at the start of the tests, but this guarantee is no longer there. Multiple inserts of the test data might cause weird test failures.

Remotion/Data/DomainObjects.UnitTests/TestDomainDatabase.cs
a-ctor: It is no longer needed in the `SetupFixture`. In this case it probably makes it easier to identify that the problem was a database one if we keep this message here.

MichaelKetting (approved):

PR #433, opened 2022-06-08
RM-8395 Package res folder based on modern practices in Nuget package

a-ctor:
https://re-motion.atlassian.net/browse/RM-8395

### DIFF
--- a/Build/NuSpec/Web.targets
+++ b/Build/NuSpec/Web.targets
@@ -9,16 +9,40 @@
   </ItemGroup>
 
   <PropertyGroup>
+    <!-- Old .NET Framwork ASP.NET projects are detected using the project type guid -->
     <WebApplicationProjectTypeGuid>{349c5851-65df-11da-9384-00065b846f21}</WebApplicationProjectTypeGuid>
-    <RemotionWebResource_IsWebApplication>$(ProjectTypeGuids.ToLower().Contains($(WebApplicationProjectTypeGuid)))</RemotionWebResource_IsWebApplication>
+    <_isNetFramworkWebApplicationProject>$(ProjectTypeGuids.ToLower().Contains($(WebApplicationProjectTypeGuid)))</_isNetFramworkWebApplicationProject>
+    <!-- .Contains returns 'True'/'False' so we need to normalize to lowercase here -->
+    <_isNetFramworkWebApplicationProject>$(_isNetFramworkWebApplicationProject.ToLower())</_isNetFramworkWebApplicationProject>
+    <!-- 'UsingMicrosoftNETSdkWeb' is set in the Sdk.props of the 'Microsoft.NET.Sdk.Web' SDK (all the ASP.NET Core projects) -->
+    <_isAspNetCoreWebApplicationProject>$(UsingMicrosoftNETSdkWeb)</_isAspNetCoreWebApplicationProject>
+    <_isWebApplicationProject>false</_isWebApplicationProject>
+    <_isWebApplicationProject Condition="'$(_isNetFramworkWebApplicationProject)' == 'true' OR '$(_isAspNetCoreWebApplicationProject)' == 'true'">true</_isWebApplicationProject>
+    <!-- Determine if we use symlinks (old) or contentFiles (new). For .NET Framework ASP.NET projects we choose symlinks, otherwise we use contentFiles. -->
+    <_useSymlink>false</_useSymlink>
+    <_useSymlink Condition="'$(_isNetFramworkWebApplicationProject)' == 'true'">true</_useSymlink>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


Build/NuSpec/Web.targets
```diff
@@ -9,16 +9,40 @@
   </ItemGroup>
 
   <PropertyGroup>
+    <!-- Old .NET Framwork ASP.NET projects are detected using the project type guid -->
     <WebApplicationProjectTypeGuid>{349c5851-65df-11da-9384-00065b846f21}</WebApplicationProjectTypeGuid>
-    <RemotionWebResource_IsWebApplication>$(ProjectTypeGuids.ToLower().Contains($(WebApplicationProjectTypeGuid)))</RemotionWebResource_IsWebApplication>
+    <_isNetFramworkWebApplicationProject>$(ProjectTypeGuids.ToLower().Contains($(WebApplicationProjectTypeGuid)))</_isNetFramworkWebApplicationProject>
+    <!-- .Contains returns 'True'/'False' so we need to normalize to lowercase here -->
+    <_isNetFramworkWebApplicationProject>$(_isNetFramworkWebApplicationProject.ToLower())</_isNetFramworkWebApplicationProject>
```
MichaelKetting, line 16: Wouldn't it be cleaner to do this all in one?

Build/NuSpec/Web.targets
```diff
@@ -9,16 +9,40 @@
...
+    <_isWebApplicationProject Condition="'$(_isNetFramworkWebApplicationProject)' == 'true' OR '$(_isAspNetCoreWebApplicationProject)' == 'true'">true</_isWebApplicationProject>
+    <!-- Determine if we use symlinks (old) or contentFiles (new). For .NET Framework ASP.NET projects we choose symlinks, otherwise we use contentFiles. -->
+    <_useSymlink>false</_useSymlink>
+    <_useSymlink Condition="'$(_isNetFramworkWebApplicationProject)' == 'true'">true</_useSymlink>
   </PropertyGroup>
 
-  <Target Name="RemotionWebResource_DeleteResFolder" Condition="'$(RemotionWebResource_IsWebApplication)' == 'True'">
+  <PropertyGroup>
+    <RemotionWebResource_IsWebApplication Condition="'$(RemotionWebResource_IsWebApplication)' == ''">$(_isWebApplicationProject)</RemotionWebResource_IsWebApplication>
+    <RemotionWebResource_TargetPath Condition="'$(RemotionWebResource_TargetPath)' == ''">res</RemotionWebResource_TargetPath>
```
MichaelKetting, line 28: Do we have logic duplication with `%(RemotionWebResource_Package.ProjectResFolder)%`?

MichaelKetting (commented):
Some questions.

Build/NuSpec/Web.targets
a-ctor: Problem is that I could not get the conversion from `bool` to `string` to work as `.ToLower` only works on string. This way, the conversion is done implicitly.

Build/NuSpec/Web.targets
a-ctor: Not really, the `RemotionWebResource_Package` is only needed for the symlink method and this only works for the contentFiles variant. In the case of contentFiles we do not need the complexity introduced by `RemotionWebResource_Package` as we do not need targets that could be overriden by another project

Build/ProjectTypes/WebLibrary.targets
```diff
@@ -5,6 +5,10 @@
       <Pack>true</Pack>
       <PackagePath>build\$(AssemblyName).targets</PackagePath>
     </Content>
+    <Content Include="$(MSBuildThisFileDirectory)..\NuSpec\Web.targets" CopyToOutputDirectory="Never">
+      <Pack>true</Pack>
+      <PackagePath>buildTransitive\$(AssemblyName).targets</PackagePath>
+    </Content>
```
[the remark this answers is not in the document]
a-ctor: - `build` works for SDK and non-SDK on the project that references the package
- `buildTransitive` works only for SDK and projects that transitively reference the package

We depend on the file name of the targets file so we couldn't separate the targets. We also have a bit of shared code that would need to be duplicated (as it is unclear which target file would execute first). Then there is the issue that all target files are loaded regardless of if it is SDK or not so we need to do the detection in both files...

I don't think that splitting up the code adds any benefit.

Build/NuSpec/Web.targets
MichaelKetting: Ah, yes. Personally, I probably would go with `.ToStirng().ToLower()` in that case. As you like?

Build/ProjectTypes/WebLibrary.targets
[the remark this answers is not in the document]
MichaelKetting: Hmm...it still feels fishy to me but here's to hoping for kicking .net full framework all together in the next major version so this is a temporary issue anyway.

MichaelKetting (approved):

PR #450, opened 2022-07-07
RM-8536 Remotion Web resources are no longer packaged by default in consuming projects

a-ctor:
https://re-motion.atlassian.net/browse/RM-8536

 - Added `Pack="true"`
 - Removed dead code

### DIFF
--- a/Build/NuSpec/Web.targets
+++ b/Build/NuSpec/Web.targets
@@ -29,11 +29,6 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(RemotionWebResource_IsWebApplication)' == 'true' AND '$(_useSymlink)' == 'false'">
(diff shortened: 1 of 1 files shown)


Build/NuSpec/Web.targets
```diff
@@ -29,12 +29,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(RemotionWebResource_IsWebApplication)' == 'true' AND '$(_useSymlink)' == 'false'">
-    <_remotionWebResourcesToCopy Remove="@(_remotionWebResourcesToCopy)" />
-    <_remotionWebResourcesToCopy Include="$(MSBuildThisFileDirectory)..\res\**\*" />
```
MichaelKetting, line 33: Is this just cleanup? if so, please split into separate commits

PR #452, opened 2022-07-08
feature/RM-7114-Support-for-RowHeaders-in-BocList

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-7114


Remotion/ObjectBinding/Web/UI/Controls/IBocRowHeaderColumnDefinition.cs
```diff
@@ -0,0 +1,29 @@
...
+// //
+using System;
+using Remotion.Web.UI.Controls;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls
+{
+  /// <summary>
+  /// Defines the required API for allowing a <see cref="BocColumnDefinition"/> to act as a header in a <see cref="BocList"/> row.
+  /// </summary>
+  public interface IBocRowHeaderColumnDefinition : IControlItem
```
MichaelKetting, line 25: @a-ctor I'm not completely happy with the name of this interface. It's meant to introduce the `IsRowHeader` flag and the only other name I can come up with is `IBocRowHeaderEnabledColumnDefinition`. A similar pattern exists with `IBocSortableColumnDefinition` and `IsSortable` but somehow, this feels more like it enables configuration while `IBocRowHeaderColumnDefinition` feels like it's always a row header, regardless of the flag. Thoughts?

a-ctor: `Sortable` sounds better because it is an adjective maybe it works if we use an adjective for the `IsRowHeader` as well.

e.g. `IBocRowHeadedColumnDefinition`, `IBocHeadedColumnDefinition`

I am not really happy but these are the only names I could come up with

MichaelKetting: Thanks for the ideas. To me, `rowheaded` doesn't work for me, maybe because it reads so much like 'bull-headed' :D How about `IBocRowHeaderSupportingColumnDefinition`? Or just admit defeat and make it `IBocColumnDefinitionWithRowHeaderSupport`?

a-ctor: I prefer `IBocColumnDefinitionWithRowHeaderSupport` over `IBocRowHeaderSupportingColumnDefinition`. We could also use the "is" in the name `IBocIsRowHeaderColumnDefinition`. But the `WithRowHeaderSupport` name is probably best.

PR #456, opened 2022-07-26
Feature/rm 8562 headless support

a-ctor:
https://re-motion.atlassian.net/browse/RM-8562

### DIFF
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocAutoCompleteReferenceValueControlObjectTest.cs
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocAutoCompleteReferenceValueControlObjectTest.cs
@@ -39,6 +39,7 @@ using Remotion.Web.Development.WebTesting.Utilities;
 namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
 {
   [TestFixture]
+  [RequiresUserInterface]
   public class BocAutoCompleteReferenceValueControlObjectTest : IntegrationTest
   {
     [Test]
--- a/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/ScreenshotCreation/ScreenshotTesting.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/ScreenshotCreation/ScreenshotTesting.cs
@@ -86,7 +86,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure.Sc
 
       var results = new List<SubTestResult>();
 
-      if (type.HasFlag(ScreenshotTestingType.Desktop))
+      if (type.HasFlag(ScreenshotTestingType.Desktop) && !helper.MainBrowserSession.Headless)
       {
         using (var diagnosticScreenshotBuilder = DiagnosticScreenshotBuilder.CreateDesktopScreenshot(helper.BrowserConfiguration.Locator))
         {
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
@@ -15,6 +15,8 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Diagnostics;
+using System.Linq;
 using Coypu;
 using NUnit.Framework;
 using NUnit.Framework.Interfaces;
@@ -52,8 +54,12 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
     [OneTimeSetUp]
     public void IntegrationTestOneTimeSetUp ()
     {
+      DriverConfigurationOverride driverConfigurationOverride = null;
+      if (Debugger.IsAttached || RequiresUserInterfaceAttribute.GetPropertyValue(TestContext.CurrentContext))
+        driverConfigurationOverride = new DriverConfigurationOverride { Headless = false };
+
       _webTestHelper = WebTestHelper.CreateFromConfiguration<CustomWebTestConfigurationFactory>();
-      _webTestHelper.OnFixtureSetUp(MaximizeMainBrowserSession);
+      _webTestHelper.OnFixtureSetUp(MaximizeMainBrowserSession, driverConfigurationOverride);
       s_webApplicationRoot = new Lazy<Uri>(
           () =>
           {
--- a/Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
+++ b/Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
@@ -49,6 +49,7 @@ namespace Remotion.Web.Development.WebTesting.Configuration
     private readonly ConfigurationProperty _chrome;
     private readonly ConfigurationProperty _edge;
     private readonly ConfigurationProperty _testSiteLayoutProperty;
+    private readonly ConfigurationProperty _headless;
 
     static WebTestConfigurationSection ()
     {
@@ -94,6 +95,7 @@ namespace Remotion.Web.Development.WebTesting.Configuration
       _testSiteLayoutProperty = new ConfigurationProperty("testSiteLayout", typeof(TestSiteLayoutConfigurationElement));
       _chrome = new ConfigurationProperty("chrome", typeof(ChromiumConfigurationElement));
       _edge = new ConfigurationProperty("edge", typeof(ChromiumConfigurationElement));
+      _headless = new ConfigurationProperty("headless", typeof(bool), false);
 
       _properties = new ConfigurationPropertyCollection
                     {
@@ -115,7 +117,8 @@ namespace Remotion.Web.Development.WebTesting.Configuration
                         _hostingProperty,
                         _testSiteLayoutProperty,
                         _chrome,
-                        _edge
+                        _edge,
+                        _headless
                     };
     }
 
@@ -286,6 +289,14 @@ namespace Remotion.Web.Development.WebTesting.Configuration
     public ChromiumConfigurationElement Edge
     {
       get { return (ChromiumConfigurationElement)this[_edge]; }
+    }
+
+    /// <summary>
+    /// Run the web browser without a user interface (headless mode).
+    /// </summary>
+    public bool Headless
+    {
+      get { return (bool)this[_headless]; }
     }
   }
 }
--- a/Remotion/Web/Development.WebTesting/ScreenshotCreation/Screenshot.cs
+++ b/Remotion/Web/Development.WebTesting/ScreenshotCreation/Screenshot.cs
@@ -16,8 +16,10 @@
 // 
 using System;
 using System.Drawing;
+using System.IO;
 using System.Linq;
 using System.Windows.Forms;
+using System.Windows.Media.Imaging;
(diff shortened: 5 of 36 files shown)


Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocAutoCompleteReferenceValueControlObjectTest.cs
```diff
@@ -38,6 +38,7 @@
 
 namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
 {
+  [RequiresUserInterface]
```
MichaelKetting, line 41: Consistency: most occurances of `TestFixture` have the TestFixture attribute as the first attribute. I would like to stick to this ordering.

Reasoning: The class is first and foremost a testfixture, so that's the preamble and it feels strange to not read this as the first information. 

I created a Roslyn ticket for this: https://re-motion.atlassian.net/browse/RMTEAM-528

Disclaimer: if there is a good reason to do it the other way, we should invert the Roslyn ticket.

PS: rest of review later

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -285,7 +286,8 @@ public void OnTearDown (bool hasSucceeded)
         Assertion.IsNotNull(_testName, "'{0}' should be set by the test infrastructure calling '{1}'", nameof(_testName), nameof(OnSetUp));
         var screenshotRecorder = new TestExecutionScreenshotRecorder(_testInfrastructureConfiguration.ScreenshotDirectory);
         screenshotRecorder.CaptureCursor();
-        screenshotRecorder.TakeDesktopScreenshot(_testName);
+        if (!MainBrowserSession.Headless)
```
MichaelKetting, line 289:
```suggestion
        if (!_mainBrowserSession.Headless)
```

Since you accessed the field earlier and we use fields in this method, too.

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -287,5 +290,13 @@ public ChromiumConfigurationElement Edge
     {
       get { return (ChromiumConfigurationElement)this[_edge]; }
     }
+
+    /// <summary>
+    /// Run the web browser without a user interface (headless mode).
+    /// </summary>
+    public bool Headless
+    {
+      get { return (bool)this[_headless]; }
```
MichaelKetting, line 299: The update in WebTestConfigurationSectionTest is missing. This will then also fail the missing schema update.

Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
```diff
@@ -52,8 +54,12 @@ protected IDriver Driver
     [OneTimeSetUp]
     public void IntegrationTestOneTimeSetUp ()
     {
+      DriverConfigurationOverride driverConfigurationOverride = null;
+      if (Debugger.IsAttached || RequiresUserInterfaceAttribute.GetPropertyValueFromAmbientTestContext())
```
MichaelKetting, line 58: Seeing this used, it feels kind of strange. Wouldn't an extension method be cleaner. or just a method that gets the testcontext injected?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Firefox/FirefoxConfiguration.cs
```diff
@@ -96,8 +96,10 @@ public FirefoxConfiguration ([NotNull] WebTestConfigurationSection webTestConfig
     public override ScreenshotTooltipStyle TooltipStyle => ScreenshotTooltipStyle.Firefox;
 
     /// <inheritdoc />
-    public virtual FirefoxOptions CreateFirefoxOptions ()
+    public virtual FirefoxOptions CreateFirefoxOptions (DriverConfiguration driverConfiguration)
```
MichaelKetting, line 99: passing the driverconfig as a parameter feels strange to me. It mixes the abtraction levels. The change is also missing the aspect that we can specify the headless default via WebTestConfig which is then supplied to the FirefoxConfig but ignored.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Firefox/FirefoxBrowserFactory.cs
```diff
@@ -50,12 +50,14 @@ public IBrowserSession CreateBrowser (DriverConfiguration driverConfiguration)
...
+      var extendedFirefoxOptions = _firefoxConfiguration.CreateFirefoxOptions(driverConfiguration);
+
       var firefoxDriverService = GetFirefoxDriverService();
-      var driver = new FirefoxDriver(firefoxDriverService, _firefoxConfiguration.CreateFirefoxOptions(), commandTimeout);
+      var driver = new FirefoxDriver(firefoxDriverService, extendedFirefoxOptions, commandTimeout);
       driver.Manage().Timeouts().AsynchronousJavaScript = driverConfiguration.AsyncJavaScriptTimeout;
       var session = new Coypu.BrowserSession(sessionConfiguration, new CustomSeleniumWebDriver(driver, Browser.Firefox));
 
-      return new FirefoxBrowserSession(session, _firefoxConfiguration, firefoxDriverService.ProcessId);
+      return new FirefoxBrowserSession(session, _firefoxConfiguration, firefoxDriverService.ProcessId, driverConfiguration.Headless);
```
MichaelKetting, line 60: Passing the driverConfig (for headless) into the options and then supplying the headless-flag explicitly for session splits the information into two separate parts that only make sense if they are together. is it possible that we need to inject the headless flag into the config / clone+update the config with the current headless flag to make this design clean?

better ideas welcome. 

This may also be why I originally thought of just splitting the cases. If there is no clean way to control the flag during the execution, we may need to initialize just once.

Remotion/Web/Development.WebTesting/ScreenshotCreation/Screenshot.cs
```diff
@@ -76,7 +108,7 @@ public static Screenshot TakePrimaryDesktopScreenshot ()
     }
 
     [NotNull]
-    private static Screenshot CreateBrowserScreenshot (IBrowserSession browserSession, IBrowserContentLocator locator)
+    private static Screenshot CreateBrowserScreenshotFromScreenContents (IBrowserSession browserSession, IBrowserContentLocator locator)
```
MichaelKetting, line 111: `CreateBrowserScreenshotBasedOnScreen`?

Remotion/Web/Development.WebTesting/ScreenshotCreation/Screenshot.cs
```diff
@@ -39,7 +41,37 @@ public static Screenshot TakeBrowserScreenshot ([NotNull] IBrowserSession browse
...
+        return TakeDriverScreenshot(browserSession);
+      else
+        return CreateBrowserScreenshotFromScreenContents(browserSession, locator);
+    }
+
+    /// <summary>
+    /// Takes a screenshot of the specified <paramref name="browserSession"/>.
+    /// </summary>
+    [NotNull]
+    public static Screenshot TakeDriverScreenshot (IBrowserSession browserSession)
```
MichaelKetting, line 54: `CreateBrowserScreenshotBasedOnDriver`?

MichaelKetting, line 54: Why is this one public but we still have a separate TakeBrowserScreenshot API? is the other one legacy or still a valid usecase?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/ScreenshotCreation/ScreenshotTesting.cs
```diff
@@ -97,19 +97,39 @@ public SubTestResult (bool hasSucceeded, string message, string imageSource, str
 
       if (type.HasFlag(ScreenshotTestingType.Browser))
       {
-        using (var diagnosticScreenshotBuilder = DiagnosticScreenshotBuilder.CreateBrowserScreenshot(helper.BrowserConfiguration.Locator, helper.MainBrowserSession))
+        if (helper.MainBrowserSession.Headless)
```
MichaelKetting, line 100: Why not move this into the DiagnosticScreenshotBuilder?

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
a-ctor: I think it is fine as the attribute is a NUnit specific attribute. Separating the logic from the attribute would make it more difficult to find the related code.

MichaelKetting: Yes, I can see how separating the types would be awkward (and why is c# so finicky about this...) What do you think about `RequiresUserInterfaceAttribute.GetPropertyValue(TestContext.Current)` ?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/DiagnosticScreenshotBuilder.cs
```diff
@@ -43,8 +45,12 @@ public static DiagnosticScreenshotBuilder CreateBrowserScreenshot ([NotNull] IBr
       ArgumentUtility.CheckNotNull("contentLocator", contentLocator);
       ArgumentUtility.CheckNotNull("browserSession", browserSession);
 
+      var screenshot = Screenshot.TakeBrowserScreenshot(browserSession, contentLocator);
+      if (browserSession.Headless)
+        contentLocator = new FixedBrowserContentLocator(new Rectangle(0, 0, screenshot.Image.Width, screenshot.Image.Height));
```
MichaelKetting, line 50: should we find a way to make this thing transparent (future improvment)?

MichaelKetting (approved):

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/DiagnosticScreenshotBuilder.cs
a-ctor: What do you mean with "transparent"?

MichaelKetting: right now, you need to post-process the result of Screenshot.TakeBrowserScreenshot manually inside your tests isntead of getting this done automatically. This seems like a possible cause for problems and needless complexity on the user-side.

a-ctor: After looking at it again I have concluded that the locator that is passed in is irrelevant. Since the screenshot is in the `Browser` coordinate system the locator is never needed. Thus, we have a few options:

 1. Remove `FixedBrowserContentLocator` and pass the provided `contentLocator`
 2. Always pass a `FixedBrowserContentLocator`
 3. Pass a new locator that always throws to make sure that nobody tries to call the locator.

I prefer option 1 as it keeps things simple

MichaelKetting: Option 1 sounds good.

PR #459, opened 2022-08-30
RM-8534 AutoCompleteReferenceValueInput: Validation message may not be entirely visible when control is too small

Frazerus:
https://re-motion.atlassian.net/browse/RM-8534

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
@@ -1770,7 +1770,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
         let contentWidth = parseInt(popUp.dataset['popUpContentWidth']!, 10);
         if (!isVisibe)
         {
-            contentWidth = Math.max(0, Math.max(...(Array.from(popUpDiv.children) as HTMLElement[]).map(function (el) { return el.offsetWidth + el.offsetLeft; })));
+            const htmlPopUpDiv = popUpDiv as HTMLElement;
+            contentWidth = Math.max(0, htmlPopUpDiv.offsetLeft + htmlPopUpDiv.offsetWidth);
             popUp.dataset['popUpContentWidth'] = '' + contentWidth;
         }
 


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1812,7 +1813,11 @@ namespace Remotion.BocAutoCompleteReferenceValue
         const minWidth = LayoutUtility.GetOuterWidth(reference);
         const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
         const requiredWidth = contentWidth + 30;
-        const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+
+        let elementWidth: number | string;
+        elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+        if(requiredHeight == maxHeightSafe)
+            elementWidth = "auto";
```
MichaelKetting, line 1820: Stylistic: I would prefer
```
if (...)
  elemWidth = ...
else
  elemWidth = ...
```
Reasoning, aside from symmetry: avoiding multiple assignments to the same variable whenever possible. Makes things more deterministic. 

Or, further refactoring, and considering the Javascript syntax needed / possibe:
`const elemWidth = reqHght == maxHeightSafe ? "auto" : ...;`

Sensible style change?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1812,7 +1813,11 @@ namespace Remotion.BocAutoCompleteReferenceValue
         const minWidth = LayoutUtility.GetOuterWidth(reference);
         const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
         const requiredWidth = contentWidth + 30;
-        const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+
+        let elementWidth: number | string;
```
MichaelKetting, line 1817: This code change also affects the result-list of the autocmplete (same calculation function is used). We need to test the expand/shrink-behavior of both the infobox and the resultlist. Do this tomorrow as a pair?

MichaelKetting (changes requested):
Changes, unless the suggestions aren't a good idea.

a-ctor (changes requested):
During the review I spotted a mistake in the calculation of the `contentWidth`:

https://github.com/re-motion/Framework/blob/9e36d1567606d5fa3c79ac390c2a305b74f2bd68/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts#L1774

```ts
const htmlPopUpDiv = popUpDiv as HTMLElement;
contentWidth = htmlPopUpDiv.offsetLeft + htmlPopUpDiv.offsetWidth;
```

Seems like this is enough to fix the problem. Then the popup and the result list use the same code again. The problem was that the current code depended on children to do the size calculation which fails in the case of the popup which has no children

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
Frazerus: With @a-ctor's change this has become redundant. We should more likely talk about whether this is the intended way the message should be displayed, as it now does not follow the design specified in the issue.

Remotion/Web/UnitTests/Core/UI/Controls/FontPreloadLinkTest.cs
```diff
@@ -48,6 +48,7 @@ public void Render ()
       _htmlHelper.AssertAttribute(element, "as", "font");
       _htmlHelper.AssertAttribute(element, "type", "font-type");
       _htmlHelper.AssertAttribute(element, "href", "myfont.ttf");
+      _htmlHelper.AssertAttribute(element, "crossorigin", "anonymous");
```
MichaelKetting, line 51: please rebase. (usually) good practice: always rebase on develop before (re-)requesting a PR

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1813,7 +1814,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
         const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
         const requiredWidth = contentWidth + 30;
         const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
-
+
```
MichaelKetting, line 1817: revert whitespace change

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1770,7 +1770,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
         let contentWidth = parseInt(popUp.dataset['popUpContentWidth']!, 10);
         if (!isVisibe)
         {
-            contentWidth = Math.max(0, Math.max(...(Array.from(popUpDiv.children) as HTMLElement[]).map(function (el) { return el.offsetWidth + el.offsetLeft; })));
+            const htmlPopUpDiv = popUpDiv as HTMLElement;
+            contentWidth = htmlPopUpDiv.offsetLeft + htmlPopUpDiv.offsetWidth;
```
MichaelKetting, line 1774: I'm surprised this works with the result list. am looking forward to seeing it in action

MichaelKetting (changes requested):

MichaelKetting (approved):

a-ctor (approved):

PR #462, opened 2022-09-14
Feature/rm 8501 provide a fallback when interacting with the href value of anchor based click targets

KuJo7:
https://re-motion.atlassian.net/browse/RM-8501

### DIFF
--- a/Remotion/Development/Web/UnitTesting/UI/Controls/FakeFallbackNavigationUrlProvider.cs
+++ b/Remotion/Development/Web/UnitTesting/UI/Controls/FakeFallbackNavigationUrlProvider.cs
@@ -1,0 +1,34 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using Remotion.ServiceLocation;
+using Remotion.Web.UI.Controls;
+
+namespace Remotion.Development.Web.UnitTesting.UI.Controls
+{
+  /// <summary>
+  /// Implements <see cref="IFallbackNavigationUrlProvider"/> interface with a default URL of "fakeFallbackUrl".
+  /// Only used for tests and therefore not part of the normal DI containers.
+  /// </summary>
+  /// <threadsafety static="true" instance="true" />
+  public class FakeFallbackNavigationUrlProvider : IFallbackNavigationUrlProvider
+  {
+    public string GetURL ()
+    {
+      return "fakeFallbackUrl";
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
@@ -21,6 +21,7 @@ using System.Xml;
 using Moq;
 using NUnit.Framework;
 using Remotion.Development.Web.UnitTesting.Resources;
+using Remotion.Development.Web.UnitTesting.UI.Controls;
 using Remotion.Development.Web.UnitTesting.UI.Controls.Rendering;
 using Remotion.FunctionalProgramming;
 using Remotion.Globalization;
@@ -33,6 +34,7 @@ using Remotion.Web;
 using Remotion.Web.Contracts.DiagnosticMetadata;
 using Remotion.Web.Infrastructure;
 using Remotion.Web.UI;
+using Remotion.Web.UI.Controls;
 using Remotion.Web.UI.Controls.Rendering;
 using Remotion.Web.Utilities;
 
@@ -361,7 +363,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocBooleanValueImplem
           RenderingFeatures.WithDiagnosticMetadata,
           new BocBooleanValueResourceSetFactory(resourceUrlFactory),
           new StubLabelReferenceRenderer(),
-          new StubValidationErrorRenderer());
+          new StubValidationErrorRenderer(),
+          new FakeFallbackNavigationUrlProvider());
       _renderer.Render(new BocBooleanValueRenderingContext(HttpContext, Html.Writer, _booleanValue.Object));
 
       return Html.GetResultDocument();
@@ -376,7 +379,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocBooleanValueImplem
           RenderingFeatures.Default,
           new BocBooleanValueResourceSetFactory(resourceUrlFactory),
           new StubLabelReferenceRenderer(),
-          new StubValidationErrorRenderer());
+          new StubValidationErrorRenderer(),
+          new FakeFallbackNavigationUrlProvider());
       _renderer.Render(new BocBooleanValueRenderingContext(HttpContext, Html.Writer, _booleanValue.Object));
       var document = Html.GetResultDocument();
       var outerSpan = Html.GetAssertedChildElement(document, "span", 0);
@@ -467,7 +471,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocBooleanValueImplem
       Html.AssertAttribute(link, "aria-checked", checkedState);
       if (isReadOnly)
         Html.AssertAttribute(link, "aria-readonly", "true");
-      Html.AssertAttribute(link, "href", "#");
+      Html.AssertAttribute(link, "href", "fakeFallbackUrl");
       if (!isReadOnly)
       {
         Html.AssertAttribute(link, "onclick", _booleanValue.Object.Enabled ? GetClickScript(isRequired, _booleanValue.Object.IsAutoPostBackEnabled) : _dummyScript);
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocCommandColumnRendererTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocCommandColumnRendererTest.cs
@@ -19,10 +19,12 @@ using System.Web.UI.WebControls;
 using NUnit.Framework;
 using Remotion.Development.Web.UnitTesting.Configuration;
 using Remotion.Development.Web.UnitTesting.Resources;
+using Remotion.Development.Web.UnitTesting.UI.Controls;
 using Remotion.Development.Web.UnitTesting.UI.Controls.Rendering;
 using Remotion.ObjectBinding.Web.Services;
 using Remotion.ObjectBinding.Web.UI.Controls;
 using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering;
+using Remotion.ServiceLocation;
 using Remotion.Web;
 using Remotion.Web.Configuration;
 using Remotion.Web.UI.Controls;
@@ -65,7 +67,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
     [Test]
     public void RenderBasicCell ()
     {
-      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition);
+      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition, new FakeFallbackNavigationUrlProvider());
       renderer.RenderDataCell(_renderingContext, CreateBocDataCellRenderArguments(rowIndex: 5));
 
       var document = Html.GetResultDocument();
@@ -76,7 +78,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
 
       var a = Html.GetAssertedChildElement(td, "a", 0);
       Html.AssertAttribute(a, "id", List.Object.ClientID + "_Column_0_Command_Row_10");
-      Html.AssertAttribute(a, "href", "#");
+      Html.AssertAttribute(a, "href", "fakeFallbackUrl");
       Html.AssertAttribute(a, "onclick", "postBackEventReference;BocList.OnCommandClick();return false;");
 
       Html.AssertTextNode(a, "TestCommand", 0);
@@ -86,7 +88,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
     public void RenderTextWebString ()
     {
       Column.Text = WebString.CreateFromText("Multiline\nText");
-      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition);
+      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition, new FakeFallbackNavigationUrlProvider());
       renderer.RenderDataCell(_renderingContext, CreateBocDataCellRenderArguments(rowIndex: 5));
 
       var document = Html.GetResultDocument();
@@ -97,7 +99,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
     [Test]
     public void RenderIconCell ()
     {
-      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition);
+      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition, new FakeFallbackNavigationUrlProvider());
       renderer.RenderDataCell(_renderingContext, CreateBocDataCellRenderArguments(showIcon: true));
 
       var document = Html.GetResultDocument();
@@ -107,7 +109,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
       Html.AssertAttribute(td, "role", "cell");
 
       var a = Html.GetAssertedChildElement(td, "a", 0);
-      Html.AssertAttribute(a, "href", "#");
+      Html.AssertAttribute(a, "href", "fakeFallbackUrl");
       Html.AssertAttribute(a, "onclick", "postBackEventReference;BocList.OnCommandClick();return false;");
 
       Html.AssertIcon(a, EventArgs.BusinessObject, null);
@@ -122,7 +124,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
       Column.Icon.Width = new Unit(16, UnitType.Pixel);
       Column.Icon.Height = new Unit(16, UnitType.Pixel);
 
-      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition);
+      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition, new FakeFallbackNavigationUrlProvider());
       renderer.RenderDataCell(_renderingContext, CreateBocDataCellRenderArguments());
 
       var document = Html.GetResultDocument();
@@ -132,7 +134,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
       Html.AssertAttribute(td, "role", "cell");
 
       var a = Html.GetAssertedChildElement(td, "a", 0);
-      Html.AssertAttribute(a, "href", "#");
+      Html.AssertAttribute(a, "href", "fakeFallbackUrl");
       Html.AssertAttribute(a, "onclick", "postBackEventReference;BocList.OnCommandClick();return false;");
 
       Html.AssertIcon(a, EventArgs.BusinessObject, Column.Icon.Url.TrimStart('~'));
@@ -145,7 +147,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
     {
       WebConfigurationMock.Current.Wcag.ConformanceLevel = WaiConformanceLevel.A;
 
-      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition);
+      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition, new FakeFallbackNavigationUrlProvider());
       renderer.RenderDataCell(_renderingContext, CreateBocDataCellRenderArguments());
 
       var document = Html.GetResultDocument();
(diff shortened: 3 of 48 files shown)


Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRendererTest.cs
```diff
@@ -361,7 +362,8 @@ private XmlDocument RenderBocBooleanValue ()
           RenderingFeatures.WithDiagnosticMetadata,
           new BocBooleanValueResourceSetFactory(resourceUrlFactory),
           new StubLabelReferenceRenderer(),
-          new StubValidationErrorRenderer());
+          new StubValidationErrorRenderer(),
+          new DefaultFallbackNavigationUrlProvider());
```
MichaelKetting, line 366: Since it looks like we have some time left, it would be great if you could do the fake-implementation now, instead of later. That way, we don't touch the code twice and have everything set up for future changes.

Step 1:
Class `FakeFallbackNavigationUrlProvider`, added to `DevelopmentWeb` project, in the `UI/Controls` namespace. Implementation also returns the "`#`". Assumption: this shouldn't take much time.

Step 2:
Update the stub to return "`fakeFallbackUrl`". This will require updating some of the assertions. I don't know how much work this will be, depends on how quickly/easily you can identify the necessary changes. So I recommend doing Step 1 first as a commit etc and then try to do Step 2 and we see if we can sensibly finish it in this sprint. Otherwise, we can move the Step 2 changes into a separate task. This doesn't hurt as much, there isn't any code overlap with the changes from this PR.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocCommandColumnRendererTest.cs
```diff
@@ -122,7 +123,7 @@ public void RenderCommandIconCell ()
       Column.Icon.Width = new Unit(16, UnitType.Pixel);
       Column.Icon.Height = new Unit(16, UnitType.Pixel);
 
-      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition);
+      IBocColumnRenderer renderer = new BocCommandColumnRenderer(new FakeResourceUrlFactory(), RenderingFeatures.Default, _bocListCssClassDefinition, SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>());
```
MichaelKetting, line 126: Here should be DefaultFallbackNavigationUrlProvider / FakeFallbackNavigationUrlProvider

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocCommandColumnRendererTest.cs
```diff
@@ -132,7 +133,7 @@ public void RenderCommandIconCell ()
       Html.AssertAttribute(td, "role", "cell");
 
       var a = Html.GetAssertedChildElement(td, "a", 0);
-      Html.AssertAttribute(a, "href", "#");
+      Html.AssertAttribute(a, "href", SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL());
```
MichaelKetting, line 136: Revert. The assertion should be hardcoded whenever possible (sensible)

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocListNavigationBlockRendererTest.cs
```diff
@@ -272,7 +278,7 @@ private void AssertActiveIcon (XmlNode link, string command, int pageIndex)
     {
       Html.AssertAttribute(link, "id", List.Object.ClientID + "_Navigation_" + command);
       Html.AssertAttribute(link, "onclick", string.Format("let element = document.getElementById('CurrentPageControl_UniqueID');element.value = {0};element.dispatchEvent(new Event('change'));return false;", pageIndex));
-      Html.AssertAttribute(link, "href", "#");
+      Html.AssertAttribute(link, "href", SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL());
```
MichaelKetting, line 281: revert

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/ReflectionBusinessObjectWebUIService.cs
```diff
@@ -57,7 +58,7 @@ public string GetToolTip (IBusinessObject obj)
       ArgumentUtility.CheckNotNull("businessObjectClass", businessObjectClass);
 
       return new HelpInfo(
-          "#",
+          SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL(),
```
MichaelKetting, line 61: Hmm... here we can (should) just use "fakeFallbackUrl" or not change it at all. It's a hardcoded value we inject and unless this is used as a reference value for comparison with actual results generated during the tests, the URL value is of no import.

Remotion/ObjectBinding/Web/UI/Controls/BocBooleanValueImplementation/Rendering/BocBooleanValueRenderer.cs
```diff
@@ -64,14 +64,16 @@ public enum ResourceIdentifier
...
         IGlobalizationService globalizationService,
         IRenderingFeatures renderingFeatures,
         IBocBooleanValueResourceSetFactory resourceSetFactory,
         ILabelReferenceRenderer labelReferenceRenderer,
-        IValidationErrorRenderer validationErrorRenderer)
+        IValidationErrorRenderer validationErrorRenderer,
+        IFallbackNavigationUrlProvider fallbackNavigationUrlProvider)
         : base(resourceUrlFactory, globalizationService, renderingFeatures)
     {
       ArgumentUtility.CheckNotNull("resourceSetFactory", resourceSetFactory);
```
MichaelKetting, line 79: I'm missing the argument check

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -53,11 +54,13 @@ public abstract class BocColumnRendererBase<TBocColumnDefinition> : IBocColumnRe
...
     private readonly IRenderingFeatures _renderingFeatures;
     private readonly BocListCssClassDefinition _cssClasses;
+    private readonly IFallbackNavigationUrlProvider _fallbackNavigationUrlProvider;
 
     protected BocColumnRendererBase (
         IResourceUrlFactory resourceUrlFactory,
         IRenderingFeatures renderingFeatures,
-        BocListCssClassDefinition cssClasses)
+        BocListCssClassDefinition cssClasses,
+        IFallbackNavigationUrlProvider fallbackNavigationUrlProvider)
```
MichaelKetting, line 63: please recheck all occurances for missing argument checks

Remotion/Web/Core/UI/Controls/CommandInfo.cs
```diff
@@ -157,7 +158,7 @@ private void AddDiagnosticMetadataAttributes (HtmlTextWriter writer)
 
     private bool IsTriggeringNavigation ()
     {
-      return _href != "#";
+      return _href != SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL();
```
MichaelKetting, line 161: I think this is breaking the intended semantics. It was probably hacky from the start but it worked due to lack of creative usage.

New semantics:
IsTriggeringNavigation tests
1. _href being not null and starting with "`#`"
2. _onClick being null or empty or not contains "`__doPostBack`"

IsTriggeringPostBack tests just for _onClick contains "__doPostback"

We assume that the developer correctly implements "__doPostback" with a "return false". This is a reasonable assumption when the intention is simply to detect the intended behavior, not to find / compensate for bugs or strange implementations. This may break the webtests. we need to check the result and and act if it does.

Remotion/Web/Core/UI/Controls/NavigationCommand.cs
```diff
@@ -29,24 +31,32 @@ namespace Remotion.Web.UI.Controls
   /// <summary> A <see cref="NavigationCommand"/> defines an action the user can invoke when navigating between pages. </summary>
   public class NavigationCommand : Command
   {
-    public NavigationCommand ()
-        : this(CommandType.Href, GetWebSecurityAdapter(), GetWxeSecurityAdapter())
+      private readonly IFallbackNavigationUrlProvider _fallbackNavigationUrlProvider;
+
+      public NavigationCommand ()
```
MichaelKetting, line 36: I think the indent is wrong here. please verify

Remotion/Web/Core/UI/Controls/NavigationCommand.cs
```diff
@@ -29,24 +31,32 @@ namespace Remotion.Web.UI.Controls
...
+        [CanBeNull] IWxeSecurityAdapter? wxeSecurityAdapter,
+        IFallbackNavigationUrlProvider fallbackNavigationUrlProvider)
         : base(defaultType, webSecurityAdapter, wxeSecurityAdapter)
     {
+        _fallbackNavigationUrlProvider = fallbackNavigationUrlProvider;
     }
 
+    private static IFallbackNavigationUrlProvider GetFallbackNavigationUrlProvider ()
+    {
+        return SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>();
```
MichaelKetting, line 58: indent is wrong.

Remotion/Web/Core/UI/Controls/NavigationCommand.cs
```diff
@@ -29,24 +31,32 @@ namespace Remotion.Web.UI.Controls
...
+    }
     /// <summary> Creates a <see cref="CommandInfo"/> for the <see cref="Command.WxeFunctionCommand"/>. </summary>
```
MichaelKetting, line 60: missing blank line

Remotion/Web/Core/UI/Controls/NavigationCommand.cs
```diff
@@ -29,24 +31,32 @@ namespace Remotion.Web.UI.Controls
...
 
     public NavigationCommand (
         CommandType defaultType,
         [CanBeNull] IWebSecurityAdapter? webSecurityAdapter,
-        [CanBeNull] IWxeSecurityAdapter? wxeSecurityAdapter)
+        [CanBeNull] IWxeSecurityAdapter? wxeSecurityAdapter,
+        IFallbackNavigationUrlProvider fallbackNavigationUrlProvider)
         : base(defaultType, webSecurityAdapter, wxeSecurityAdapter)
     {
+        _fallbackNavigationUrlProvider = fallbackNavigationUrlProvider;
```
MichaelKetting, line 53: indent is wrong

Remotion/Web/Test.Shared/MultiplePostBackCatching/TestControlGenerator.cs
```diff
@@ -69,7 +70,8 @@ public bool IsAlertHyperLink (Control control)
       if (control is HyperLink)
       {
         HyperLink hyperLink = (HyperLink)control;
-        if (hyperLink.NavigateUrl == "#" && hyperLink.Attributes["onclick"].Contains("alert"))
+        var url = SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL();
```
MichaelKetting, line 73: This is just a helper type used in our tests. we can stick with "#" to keep things simple and isolated.
(revert entire file)

Remotion/Web/UnitTests/Core/UI/Controls/CommandInfoTest.cs
```diff
@@ -61,7 +62,7 @@ public void CreateForPostBack ()
       Assert.That(commandInfo.Title, Is.EqualTo("TheTitle"));
       Assert.That(commandInfo.AccessKey, Is.EqualTo("A"));
       Assert.That(commandInfo.OnClick, Is.EqualTo("ClickHandler"));
-      Assert.That(commandInfo.Href, Is.EqualTo("#"));
+      Assert.That(commandInfo.Href, Is.EqualTo(SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL()));
```
MichaelKetting, line 65: Test for the actual value, not the IoC result

Remotion/Web/Test.Shared/UpdatePanelTests/SutUserControl.ascx.cs
```diff
@@ -50,14 +51,15 @@ protected override void OnInit (EventArgs e)
       _postBackEventHandler.PostBack += HandlePostBack;
       Controls.Add(_postBackEventHandler);
 
+      var localFallbackUrl = SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>().GetURL();
       string asyncPostBackCommandInsideUpdatePanelID = "AsyncPostBackCommandInsideUpdatePanel";
-      AsyncCommandInsideUpdatePanelHyperLink.NavigateUrl = "#";
+      AsyncCommandInsideUpdatePanelHyperLink.NavigateUrl = localFallbackUrl;
```
MichaelKetting, line 56: also a test type. here, there's no logic based on this (as opposed to the other test type), so I don't care which direction you choose. Personally, I probably would have left the "#" to reduce complexity.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -70,8 +70,9 @@ class DropDownMenu_ItemInfo
       public IsDisabled: boolean,
       public readonly Href: Nullable<string>,
       public readonly Target: Nullable<string>,
+      public readonly FallbackNavigationUrl: string,
       public readonly DiagnosticMetadata: Nullable<Dictionary<string | boolean>>,
-      public readonly DiagnosticMetadataForCommand: Nullable<Dictionary<string | boolean>>)
+      public readonly DiagnosticMetadataForCommand: Nullable<Dictionary<string | boolean>>)
```
MichaelKetting, line 75: trailing whitespace.

What I try to do is review the entire changeset when I do a commit to find lines that should not contain changes or have a formatting bug, left over TODO commetns etc

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -111,7 +112,7 @@ class DropDownMenu
   private static _updateFocus: boolean = true;
   private static _ignoreHoverMouseEvents: boolean = false;
   private static _lastPositionInformation: Nullable<string> = null;
-
+
```
MichaelKetting, line 115: revert

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -333,7 +334,6 @@ class DropDownMenu
   private static BeginOpenPopUp (menuID: string, context: HTMLElement, evt: Nullable<MouseEvent>): void
   {
     const menuButton = document.querySelector<HTMLAnchorElement> ('#' + menuID + ' a[aria-haspopup=menu]')!;
-
```
MichaelKetting, line 336: revert

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -825,11 +825,11 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement): boolean
+  private static OnItemClick (this: HTMLAnchorElement, ev:HTMLElementEventMap["click"]): boolean
```
MichaelKetting, line 828: @a-ctor is this the best way to write this in Typescript?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -825,11 +825,11 @@ class DropDownMenu
...
   {
+    ev.preventDefault();
     if (this.href == null || this.href === '')
       return false;
-
```
MichaelKetting, line 832: revert

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -825,11 +825,11 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement): boolean
+  private static OnItemClick (this: HTMLAnchorElement, ev:HTMLElementEventMap["click"]): boolean
   {
+    ev.preventDefault();
```
MichaelKetting, line 830: hmm... add a newline here, please, to separate the semantic parts.

MichaelKetting, line 830: I would like to pull this as a bugfix into a separate PR that gets integrated first and is linked as a blocking issue. The original code didn't result in an obvious problem but it was still incorrect behavior and isn't part of this change. By keeping things strictly separate, looking for bug-causes and why why have been introduced is much easier. Same goes for reasoning over a changeset in general. For instance "why is there more changes in DropDownMenu.ts than there is in ListMenu.ts? -> ah, okay, there was a bugfix that needs to get done. We don't have this bug over there? why? etc etc.

Remotion/Web/Core/UI/Controls/IFallbackNavigationUrlProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+namespace Remotion.Web.UI.Controls
+{
+  /// <summary>
+  /// Provides the fallback URL for clickable elements that should not be opened in a new tab
+  /// </summary>
+  /// <threadsafety static="true" instance="true" />
+  public interface IFallbackNavigationUrlProvider
```
MichaelKetting, line 23: Please add unit tests for the serivce registration and the implementation. 

As a convention, we also add a `<seealso>` to the interface and reference the implementation. it helps with the docs.

Remotion/Web/Core/UI/Controls/IFallbackNavigationUrlProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+  {
+    /// <summary>
+    /// A URL that can be used as a fallback if a user opens an href in a new tab
+    /// </summary>
+    /// <returns>A string representing a URL</returns>
+    string GetURL ();
```
MichaelKetting, line 29: please add a NotNullAttribute.

MichaelKetting (changes requested):
Please do the changes unrelated to FakeFallbackNavigationUrlProvider first and we check how much time we have left. It's possible that @a-ctor could take over some bits of the re-review tomorrow.

MichaelKetting:
@KuJo7  Oh, and please check TeamCity (better yet: run the unit tests before pushing).

Remotion/Web/Core/UI/Controls/CommandInfo.cs
Frazerus: We do not quite understand what is meant here. Will ask about it in standup :)

Remotion/Web/Test.Shared/MultiplePostBackCatching/TestControlGenerator.cs
Frazerus: Are you sure we dont want to include the Fallback Naviationprovider? There is no other way than the individual control tests for us to see the behavior of the service otherwise.

Remotion/Web/Core/UI/Controls/CommandInfo.cs
Frazerus: We now run into the problem of the tests not being green. We should probably have a quick talk about how to fix this.

Remotion/Web/Core/UI/Controls/NavigationCommand.cs
```diff
@@ -29,22 +31,33 @@ namespace Remotion.Web.UI.Controls
...
     {
+      ArgumentUtility.CheckNotNull("fallbackNavigationUrlProvider", fallbackNavigationUrlProvider);
+
+      _fallbackNavigationUrlProvider = fallbackNavigationUrlProvider;
+    }
+
+    private static IFallbackNavigationUrlProvider GetFallbackNavigationUrlProvider ()
+    {
+      return SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>();
     }
```
a-ctor, line 61: Static methods should be above the instance members. In this case, it should be the first member in this class.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -825,8 +826,10 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement): boolean
+  private static OnItemClick (this: HTMLAnchorElement, ev:HTMLElementEventMap["click"]): boolean
```
a-ctor, line 829: We should use `MouseEvent` instead:

```suggestion
  private static OnItemClick (this: HTMLAnchorElement, ev: MouseEvent): boolean
```

Did we extract the bugfix here or is this still TODO?

a-ctor (changes requested):
There are also a few of Michael's comments unresolved

Remotion/Web/Core/UI/Controls/CommandInfo.cs
MichaelKetting: @Frazerus discussed this about the general direction of the problem with @KuJo7. I'm sure you two will talk this through tomorrow :)

I'll do a final review/hint if there's still unresolved stuff.

Remotion/Web/Test.Shared/MultiplePostBackCatching/TestControlGenerator.cs
MichaelKetting: @a-ctor can show you the web-test testsites (if you don't find them) and you need more than the indiv-controls and the tabbed test. I think there's everything you need availble between those. The multi-postback stuff only tests some cases, much less than the other pages.

Remotion/Web/Core/UI/Controls/NavigationCommand.cs
MichaelKetting: This one here's a false positive: the private static method is just a private method without state. These methods remain with the instance methods. 

Historically, we had these methods as regular instance methods but R# recommends making them static and it was decided that it's good practice to clearly state a method isn't relying on state. So, private methods that are only used from instance methods get placed with the other private instance methods and only private static methods that are used from public/protected static methods are collected at the top of the class.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
Frazerus: Extracted

a-ctor (approved):
The other PR should be merged before this one tho, as we haven't fully extracted the changes.

Remotion/Web/UnitTests/Core/UI/Controls/IFallbackNavigationUrlProviderTest.cs
```diff
@@ -0,0 +1,37 @@
+﻿using NUnit.Framework;
```
MichaelKetting, line 1: License

Remotion/Development/Web/UnitTesting/UI/Controls/FakeFallbackNavigationUrlProvider.cs
```diff
@@ -0,0 +1,29 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
```
MichaelKetting, line 1: The two current commits can be flattened.

If separate commits should be done, they would be to separate the new classes and the integration.

Remotion/Development/Web/UnitTesting/UI/Controls/FakeFallbackNavigationUrlProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using Remotion.ServiceLocation;
+using Remotion.Web.UI.Controls;
+
+namespace Remotion.Development.Web.UnitTesting.UI.Controls
+{
+  public class FakeFallbackNavigationUrlProvider : IFallbackNavigationUrlProvider
```
MichaelKetting, line 22: Please add a comment. See other fake implementations in development.web

Remotion/Web/Core/UI/Controls/DefaultFallbackNavigationUrlProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+// 
+using Remotion.ServiceLocation;
+
+namespace Remotion.Web.UI.Controls
+{
+  /// <summary>
+  /// Implements <see cref="IFallbackNavigationUrlProvider"/> interface with a default URL of "#".
+  /// </summary>
+  /// <threadsafety static="true" instance="true" />
+  [ImplementationFor(typeof(IFallbackNavigationUrlProvider), Lifetime = LifetimeKind.Singleton, Position = 1)]
```
MichaelKetting, line 25: We actually do not need the Position-attribute since our own production code doesn't have alternatives, but if you do keep it to make overrding this in downstream projects eaiser/more deterministic, please check the pattern:

Add const field to the type:
```
    public const int Position = 0;
```

Use the const in the attribute:
```
[ImplementationFor(typeof(IFallbackNavigationUrlProvider), Lifetime = LifetimeKind.Singleton, Position = Position)]
```

This way, it allows an override by referencing the base value and adding an offset. 

Also, the base value is 0, not 1.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #463, opened 2022-09-15
RM-8589 DropDownMenu changes url fragment on open

Frazerus:
https://re-motion.atlassian.net/browse/RM-8589

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
@@ -825,8 +825,10 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement): boolean
+  private static OnItemClick (this: HTMLAnchorElement, ev: MouseEvent): boolean
   {
+    ev.preventDefault();
+
     if (this.href == null || this.href === '')
       return false;
 


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -825,8 +825,10 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement): boolean
+  private static OnItemClick (this: HTMLAnchorElement, ev: HTMLElementEventMap["click"]): boolean
```
MichaelKetting, line 828: @a-ctor already suggested a better option for the type in the other PR.

MichaelKetting (commented):
Intention looks good, I'll hand over the review to @a-ctor for the final go-ahread and check if we should add a webtest for this.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor:
```suggestion
  private static OnItemClick (this: HTMLAnchorElement, ev: MouseEvent): boolean
```

a-ctor (changes requested):

a-ctor (approved):

PR #464, opened 2022-09-16
RM-8590 Rewrite CommandInfo.IsTriggeringNavigation

Frazerus:
[Rewrite CommandInfo.IsTriggeringNavigation](https://re-motion.atlassian.net/browse/RM-8590)

Mike said that there would be some kind of hidden special case we missed in an earlier implementation. Please take a look at the unit tests for this and make sure we didnt do stupid things: 

https://github.com/re-motion/Framework/blob/develop/Remotion/Web/UnitTests/Core/UI/Controls/CommandInfoTest.cs

### DIFF
--- a/Remotion/Web/Core/UI/Controls/CommandInfo.cs
+++ b/Remotion/Web/Core/UI/Controls/CommandInfo.cs
@@ -157,12 +157,26 @@ namespace Remotion.Web.UI.Controls
 
     private bool IsTriggeringNavigation ()
     {
-      return _href != "#";
+      if (_href == null)
+        return false;
+      else if (_href.StartsWith("#"))
+        return false;
+      else if (_onClick == null)
+        return true;
+      else if (_onClick.Contains("__doPostBack"))
+        return false;
+      else
+        return true;
     }
 
     private bool IsTriggeringPostBack ()
     {
-      return !IsTriggeringNavigation() && _onClick!.Contains("__doPostBack"); // TODO RM-8104: Guard _onClick
+      if (_onClick == null)
+        return false;
+      else if (_onClick.Contains("__doPostBack"))
+        return true;
+      else
+        return false;
     }
   }
 }
--- a/Remotion/Web/UnitTests/Core/UI/Controls/CommandInfoTest.cs
+++ b/Remotion/Web/UnitTests/Core/UI/Controls/CommandInfoTest.cs
@@ -330,6 +330,25 @@ namespace Remotion.Web.UnitTests.Core.UI.Controls
     }
 
     [Test]
+    public void AddDiagnosticMetadataAttributes_FromCreateForLink_WithDoPostBackAndUrl ()
+    {
+      var commandInfo = CommandInfo.CreateForLink("TheTitle", null, "TheUrl", "TheTarget", "FrontGarbage__doPostBackBackGarbage");
+
+      var stringWriter = new StringWriter();
+      var htmlTextWriter = new HtmlTextWriter(stringWriter);
+      commandInfo.AddAttributesToRender(htmlTextWriter, RenderingFeatures.WithDiagnosticMetadata);
+
+      htmlTextWriter.RenderBeginTag(HtmlTextWriterTag.A);
+      htmlTextWriter.RenderEndTag();
+
+      var result = stringWriter.ToString();
+
+      Assert.That(result, Does.Contain(DiagnosticMetadataAttributes.ControlType + "=\"Command\""));
+      Assert.That(result, Does.Contain(DiagnosticMetadataAttributes.TriggersPostBack + "=\"true\""));
+      Assert.That(result, Does.Contain(DiagnosticMetadataAttributes.TriggersNavigation + "=\"false\""));
+    }
+
+    [Test]
     public void AddDiagnosticMetadataAttributes_FromCreateForLink_WithPureJavaScript ()
     {
       var commandInfo = CommandInfo.CreateForLink("TheTitle", null, "#", "TheTarget", "javascript:Foo();");


a-ctor (approved):
I can't validate that this logic is correct but I think you walked through it with MichaelKetting?

Remotion/Web/Core/UI/Controls/CommandInfo.cs
```diff
@@ -157,7 +157,22 @@ private void AddDiagnosticMetadataAttributes (HtmlTextWriter writer)
...
+      {
+        return false;
+      }
+
+      if (_href == null || _href.StartsWith("#"))
+      {
+        return false;
+      }
+
+      return true;
```
MichaelKetting, line 175: stylistic remarks:
* no braces in code structure like this (just simple return statements. it makes the code compacter without trading readability for it.
* this sort of logic isn't really a set of early exit conditions but a group of connected conditions that work better as an if/elseif/else. Can't promise there's not other places in re-motion with this semantics and style, but it's a bit clearer if done as a single block.

actual logic still has the bug I was looking to fix. The test for javascript is necessary but not sufficient when it comes to determining whether this is a navigation-triggering anchor. If the anchor does not have an HREF or the HREF starts with a "#", it will still not trigger a navigation. This means returning "true" if there's no javascript covers only part of the required conditions. You need to go with something like the following, starting with the most dominant conditions first.

```suggestion
      if (_href == null) // without an href, an anchor cannot do a navigation
        return false;
      else if (_href.StartsWith("#")) // when the href points to a page-local target, it's not considered a navigation
        return false;
      else if (_onClick == null) // if there is no onclick-handler and we have already confirmed that there is a href, it can only be a navigation. I dropped the empty-check since we only need to guard the null-reference and it makes for better symmetry.
        return true;
      else if (_onClick.Contains("__doPostBack")) // this could be here, it could also be at the beginning, if the null-check is part of the same condition.
        return false;
      else
        return true;
```

Please remove the comments, they're just instructional.

alternative:
```
      else if (_onClick?.Contains("__doPostBack"))
        return false;
      if (_href == null)
        return false;
      else if (_href.StartsWith("#"))
        return false;
      else
        return true;
```
This version would also work, but it has two conditions in the same IF. Given we have nullable reference types, this should be safe from null-reference execptions, but we still want to ensure that we have a test for both the null-value and the __doPostBack cases. When tracking code coverage, the first variant is easier to verify.

Remotion/Web/Core/UI/Controls/CommandInfo.cs
```diff
@@ -157,7 +157,22 @@ private void AddDiagnosticMetadataAttributes (HtmlTextWriter writer)
...
     }
 
     private bool IsTriggeringPostBack ()
```
MichaelKetting, line 178: This also needs an update now. 
```
  return _onClick?.Contains("__doPostBack");
```

Why not do it as separate if/elseif/else ? We can, it just seems overkill because there's just the simple contains check. But yes, doing this as separate statements would be the more consistent style so go for it :)

MichaelKetting (changes requested):
Please check the suggested semantics. This may break some unit tests that relied on the simpler rules from before but should not break the web tests.

Remotion/Web/Core/UI/Controls/CommandInfo.cs
```diff
@@ -157,12 +157,27 @@ private void AddDiagnosticMetadataAttributes (HtmlTextWriter writer)
...
     {
-      return !IsTriggeringNavigation() && _onClick!.Contains("__doPostBack"); // TODO RM-8104: Guard _onClick
+      if (IsTriggeringNavigation())
+        return false;
+      else if (_onClick!.Contains("__doPostBack"))
+        return true;
+      else
+        return false;
+      // TODO RM-8104: Guard _onClick
```
MichaelKetting, line 180:
```suggestion
```
That's done now. it referred to onclick being nullable.

Remotion/Web/Core/UI/Controls/CommandInfo.cs
```diff
@@ -157,12 +157,27 @@ private void AddDiagnosticMetadataAttributes (HtmlTextWriter writer)
...
 
     private bool IsTriggeringPostBack ()
     {
-      return !IsTriggeringNavigation() && _onClick!.Contains("__doPostBack"); // TODO RM-8104: Guard _onClick
+      if (IsTriggeringNavigation())
+        return false;
+      else if (_onClick!.Contains("__doPostBack"))
+        return true;
+      else
+        return false;
```
MichaelKetting, line 179:
```suggestion
      if (_onClick == null)
        return false;
      else if (_onClick.Contains("__doPostBack"))
        return true;
      else
        return false;
```

We no longer need the test for IsTriggeringNavigation because onclick starting a postback implies there's no navigation possible.

Remotion/Web/Core/UI/Controls/CommandInfo.cs
```diff
@@ -157,12 +157,27 @@ private void AddDiagnosticMetadataAttributes (HtmlTextWriter writer)
...
+      // TODO RM-8104: Guard _onClick
     }
   }
 }
```
MichaelKetting, line 183: We also need one more unit test:

```


    [Test]
    public void AddDiagnosticMetadataAttributes_FromCreateForLink_WithDoPostBackAndUrl ()
    {
      var commandInfo = CommandInfo.CreateForLink("TheTitle", null, "TheUrl", "TheTarget", "FrontGarbage__doPostBackBackGarbage");

      var stringWriter = new StringWriter();
      var htmlTextWriter = new HtmlTextWriter(stringWriter);
      commandInfo.AddAttributesToRender(htmlTextWriter, RenderingFeatures.WithDiagnosticMetadata);

      htmlTextWriter.RenderBeginTag(HtmlTextWriterTag.A);
      htmlTextWriter.RenderEndTag();

      var result = stringWriter.ToString();

      Assert.That(result, Does.Contain(DiagnosticMetadataAttributes.ControlType + "=\"Command\""));
      Assert.That(result, Does.Contain(DiagnosticMetadataAttributes.TriggersPostBack + "=\"true\""));
      Assert.That(result, Does.Contain(DiagnosticMetadataAttributes.TriggersNavigation + "=\"false\""));
    }
```

Please add it after `AddDiagnosticMetadataAttributes_FromCreateForLink_WithDoPostBack`.
The various tests for null cannot actually be tested thanks to guard conditions outside the scope of the private methods.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #468, opened 2022-09-23
RM-8559 Fix NovaViso Tabs to be left aligned

Frazerus:
https://re-motion.atlassian.net/browse/RM-8559

### DIFF
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabbedMenu.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabbedMenu.css
@@ -27,6 +27,20 @@ td.tabbedMainMenuCell
   background-color: var(--color-contrast);
 }
 
+div.tabbedMainMenu div.tabStripTabsPane ul li,
+div.tabbedSubMenu div.tabStripTabsPane ul li
+{
+  flex-grow: unset;
+}
+
+div.tabbedMainMenu .tabStripTabWrapper .tabStripTab,
+div.tabbedMainMenu .tabStripTabWrapper .tabStripTabSelected,
+div.tabbedSubMenu .tabStripTabWrapper .tabStripTab,
+div.tabbedSubMenu .tabStripTabWrapper .tabStripTabSelected
+{
+  justify-content: start;
+}
+
 td.tabbedMainMenuCell,
 td.tabbedSubMenuCell
 {


a-ctor (approved):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabStrip.css
```diff
@@ -17,16 +17,11 @@
   flex-grow: 99999;
```
MichaelKetting, line 17: The changes are applied to the wrong context. TabStrip is used standaline and with TabbedMenu and TabbedMultiView. The changes should only apply to TabbedMenu.

We need to reset the unwanted changes in TabbedMenu while leaving the original for TabStrip and TabbedMultiView.

The first element is the TabbedMenu, the second and third elements are TabbedMultiView and TabStrip. Only the TabbedMenu should be changed in this task.
image

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabbedMenu.css
```diff
@@ -27,6 +27,11 @@ td.tabbedMainMenuCell
   background-color: var(--color-contrast);
 }
 
+.tabbedMainMenu li, .tabbedSubMenu li
```
MichaelKetting, line 30: There's a cleaner solution in CSS available: you can use `flex-grow: unset` to clear the not-wanted value. For selector, please use the original selector and prefix with tabbedMainMenu / tabbedSubMenu. That way, the selector weight is better guaranteed.

We should also add the `justify-content: start` attribute to make things more explicit.

By using `display: flex` here, you start a new flexbox, which adds an additional nesting that is not needed and results more in a workaround. The best choice when overriding problematic rules is to find the correct way to remove them (either by unsetting them or by setting a new value that correctly expresses the intention).

Here's also good information on selectors:  https://www.w3schools.com/css/css_selectors.asp & https://www.w3schools.com/cssref/css_selectors.asp
and specificity: https://www.w3schools.com/css/css_specificity.asp

And iniitial/inhertit/unset of values: https://www.w3schools.com/cssref/css3_pr_all.asp

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TabbedMenu.css
```diff
@@ -27,6 +27,13 @@ td.tabbedMainMenuCell
   background-color: var(--color-contrast);
 }
 
+div.tabbedMainMenu div.tabStripTabsPane ul li,
+div.tabbedSubMenu div.tabStripTabsPane ul li
+{
+  flex-grow: unset;
+  justify-content: start;
```
MichaelKetting, line 34: Something got lost in translation here. The 'justify' needs to be applied to another selector (see ticket). let's do it together.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #469, opened 2022-09-23
RM-8560 Fix NovaViso DropDownMenuSelect to be not underlined

KuJo7:
https://re-motion.atlassian.net/browse/RM-8560


MichaelKetting (approved):
Please verify that the correct user display name is used when doing the PR-merge in github

PR #472, opened 2022-10-04
RM-8587 Additional SafeContext helper constructs are missing

a-ctor:
https://re-motion.atlassian.net/browse/RM-8587

### DIFF
--- a/Remotion/Core/Core/Context/SafeContext.Task.ContinueWith.cs
+++ b/Remotion/Core/Core/Context/SafeContext.Task.ContinueWith.cs
@@ -1,0 +1,341 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using SystemTask = System.Threading.Tasks.Task;
+
+namespace Remotion.Context
+{
+  public static partial class SafeContext
+  {
+    public static partial class Task
+    {
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task})" />
+      public static SystemTask ContinueWith (SystemTask task, Action<SystemTask> continuationAction) => ContinueWith(Instance, task, continuationAction);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task})" />
+      public static SystemTask ContinueWith (ISafeContextStorageProvider provider, SystemTask task, Action<SystemTask> continuationAction) => task.ContinueWith(CreateWrapper(provider, continuationAction));
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.CancellationToken)" />
+      public static SystemTask ContinueWith (SystemTask task, Action<SystemTask> continuationAction, CancellationToken cancellationToken) => ContinueWith(Instance, task, continuationAction, cancellationToken);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.CancellationToken)" />
+      public static SystemTask ContinueWith (ISafeContextStorageProvider provider, SystemTask task, Action<SystemTask> continuationAction, CancellationToken cancellationToken) => task.ContinueWith(CreateWrapper(provider, continuationAction), cancellationToken);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.Tasks.TaskScheduler)" />
+      public static SystemTask ContinueWith (SystemTask task, Action<SystemTask> continuationAction, TaskScheduler taskScheduler) => ContinueWith(Instance, task, continuationAction, taskScheduler);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.Tasks.TaskScheduler)" />
+      public static SystemTask ContinueWith (ISafeContextStorageProvider provider, SystemTask task, Action<SystemTask> continuationAction, TaskScheduler taskScheduler) => task.ContinueWith(CreateWrapper(provider, continuationAction), taskScheduler);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.Tasks.TaskContinuationOptions)" />
+      public static SystemTask ContinueWith (SystemTask task, Action<SystemTask> continuationAction, TaskContinuationOptions taskContinuationOptions) => ContinueWith(Instance, task, continuationAction, taskContinuationOptions);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.Tasks.TaskContinuationOptions)" />
+      public static SystemTask ContinueWith (ISafeContextStorageProvider provider, SystemTask task, Action<SystemTask> continuationAction, TaskContinuationOptions taskContinuationOptions) => task.ContinueWith(CreateWrapper(provider, continuationAction), taskContinuationOptions);
+
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task},System.Threading.CancellationToken,System.Threading.Tasks.TaskContinuationOptions,System.Threading.Tasks.TaskScheduler)" />
+      public static SystemTask ContinueWith (SystemTask task, Action<SystemTask> continuationAction, CancellationToken cancellationToken, TaskContinuationOptions taskContinuationOptions, TaskScheduler taskScheduler) => ContinueWith(Instance, task, continuationAction, cancellationToken, taskContinuationOptions, taskScheduler);
+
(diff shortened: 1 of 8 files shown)


Remotion/Core/Core/Context/SafeContext.Timers.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Remotion.Context
+{
+  public static partial class SafeContext
+  {
+    /// <summary>
+    /// Provides methods for setting a callback on <see cref="System.Timers.Timer"/> which are <see cref="SafeContext"/> aware.
+    /// </summary>
+    public static class Timers
+    {
+      public static ElapsedEventHandler AddElapsedEventHandler (System.Timers.Timer timer, ElapsedEventHandler handler)
```
MichaelKetting, line 31: Since you can't write
```
      System.Timers.Timer t = new();
      var x = t.Elapsed += (sender, args) => { };
```
I suggest we make the result void here and in line 39?

MichaelKetting, line 31: Use inherit doc?

Remotion/Core/Core/Context/SafeContext.Threading.cs
```diff
@@ -0,0 +1,65 @@
...
+  public static partial class SafeContext
+  {
+    /// <summary>
+    /// Provides methods for creating <see cref="System.Threading.Timer"/>s which are <see cref="SafeContext"/> aware.
+    /// </summary>
+    public static class Threading
+    {
+      public static Timer NewTimer (TimerCallback callback) => NewTimer(Instance, callback);
+
+      public static Timer NewTimer (ISafeContextStorageProvider provider, TimerCallback callback) => new Timer(CreateWrapper(provider, callback));
```
MichaelKetting, line 32: Why use `new Timer` here but `new(...)` the rest of the code?

Remotion/Core/Core/Context/SafeContext.Threading.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Remotion.Context
+{
+  public static partial class SafeContext
+  {
+    /// <summary>
+    /// Provides methods for creating <see cref="System.Threading.Timer"/>s which are <see cref="SafeContext"/> aware.
+    /// </summary>
+    public static class Threading
+    {
+      public static Timer NewTimer (TimerCallback callback) => NewTimer(Instance, callback);
```
MichaelKetting, line 30: use inerit doc?

Remotion/Core/Core/Context/SafeContext.Task.ContinueWith.cs
```diff
@@ -0,0 +1,325 @@
...
+using SystemTask = System.Threading.Tasks.Task;
+
+namespace Remotion.Context
+{
+  public static partial class SafeContext
+  {
+    public static partial class Task
+    {
+      /// <inheritdoc cref="M:System.Threading.Tasks.Task.ContinueWith(System.Action{System.Threading.Tasks.Task})" />
+      public static SystemTask ContinueWith (SystemTask task, Action<SystemTask> continuationAction) => ContinueWith(Instance, task, continuationAction);
```
MichaelKetting, line 13: I'm assuming the code was generated so no line-by-line check is needed.

MichaelKetting (commented):
Still need to review the tests

Remotion/Core/Core/Context/SafeContext.Timers.cs
MichaelKetting: okay, I see now why this is designed this way. There is no good way to deregister the event handler without getting a reference to it. I'm not 100% convided that the scenario is helpful/needed for practical applications. If someone really does need this, they can still do it themselves, but for the most part, there's no registration/deregistration need and us returning the new handler-instance may confuse the intended usage scenarios. thoughts?

a-ctor: We can do it either way but it is not easy to notice that the handler is being changed when you get a void result. And if remove the handler you might not notice that it did nothing and is still being called. Having a return type for the funtion makes that explicit and forces the consumers to make a conscious decision.

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+
+namespace Remotion.UnitTests.Context.SafeContextTests
+{
+  public partial class SafeContextTaskTest
+  {
+    [Flags]
+    public enum ContinueWithTestFlags
+    {
+      None = 0,
+      HasState = 1,
```
MichaelKetting, line 21: I can handle those bitflags, so no need to chenge this now. I did however find a nice way to make those more maintainable by using the bit-shift syntax (`1 << 2`, `1 << 3`, etc)

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+          WasCalled = true;
+          return TryExecuteTask(task);
+        }
+
+        return false;
+      }
+
+      protected override IEnumerable<Task> GetScheduledTasks ()
+      {
+        throw new NotImplementedException();
```
MichaelKetting, line 82: Even in tests, this should be NotSupportedException, not NotImplementedException.

The NIE is intended during development, not for completed types.

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+      }
+    }
+
+    private static readonly MethodInfo s_createTestInvocationMethod = typeof(SafeContextTaskTest).GetMethod(nameof(CreateTestInvocation), BindingFlags.NonPublic | BindingFlags.Static);
+
+    private readonly MethodInfo _runContinueWithTestMethod = typeof(SafeContextTaskTest).GetMethod(nameof(RunContinueWithTest), BindingFlags.NonPublic | BindingFlags.Instance);
```
MichaelKetting, line 88: this should be static and prefixed with `s`

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+
+      foreach (var method in methods)
+      {
+        var parameters = method.GetParameters();
+        var isImplicit = parameters[0].ParameterType != typeof(ISafeContextStorageProvider);
+        if (isImplicit != implicitSafeContextProvider)
+          continue;
+
+        var taskType = ResolveGenericParameters(parameters[isImplicit ? 0 : 1].ParameterType);
+        Debug.Assert(typeof(Task).IsAssignableFrom(taskType), "typeof(Task).IsAssignableFrom(taskType)");
```
MichaelKetting, line 133: Please use AssertionUtility.IsTrue/IsFalse, etc.

I avoid Debug.Assert in re-motion sources to not having to think about what causes/does-not-cause the message box to get displayed.

MichaelKetting: Added a task in the backlog https://re-motion.atlassian.net/browse/RMTEAM-594

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+        resultingTask.GetAwaiter().GetResult();
+      }
+      catch (OperationCanceledException ex)
+      {
+        Assert.That(resultingTask != null);
+        Assert.That(ex.CancellationToken, Is.EqualTo(cancellationToken));
+        Assert.That(resultingTask.Status, Is.EqualTo((flags & HasCancellationToken) != 0 ? TaskStatus.Canceled : TaskStatus.Faulted), "CancellationToken was not passed correctly.");
+      }
+
+      Assert.That(continuationCalled, "Continuation was not called.");
```
MichaelKetting, line 280: Please add Is.True to Assert.That. (several places)

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
+﻿using System;
```
MichaelKetting, line 1: License header

Remotion/Core/Core/Context/SafeContext.Task.ContinueWith.cs
```diff
@@ -0,0 +1,325 @@
+﻿using System;
```
MichaelKetting, line 1: License header

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+        return (TTask)Task.CompletedTask;
+
+      Debug.Assert(typeof(TTask) == typeof(Task<int>));
+      return (TTask)(object)Task.FromResult(1);
+    }
+
+    private static TContinuation CreateContinuation<TContinuation> (Action<Task, object> innerContinuation)
+        where TContinuation : Delegate
+    {
+      if (typeof(TContinuation) == typeof(Action<Task>))
```
MichaelKetting, line 297: style: this is a good example where we should do an if-elseif-else

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
```diff
@@ -0,0 +1,367 @@
...
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Remotion.Context;
+using static Remotion.UnitTests.Context.SafeContextTests.SafeContextTaskTest.ContinueWithTestFlags;
+
+namespace Remotion.UnitTests.Context.SafeContextTests
+{
+  public partial class SafeContextTaskTest
```
MichaelKetting, line 15: I did not try to backtrack the tests semantics, just a big picture sanity check. do you think we should add an explainatory comment block how the test works?

Remotion/Core/Core/Context/SafeContext.Timers.cs
a-ctor: I would add documentation that explains why a value is returned. The return type should force them to think about why something is returned, and they can look to the docs for the answer. This allows them to make their own decision if they want to keep the delegate or not.

Remotion/Core/UnitTests/Context/SafeContextTests/SafeContextTest.Task.ContinueWith.cs
a-ctor: The other option would be `0b0001`, `0b0010`, ... although I find this to be less readable overall.

MichaelKetting: yeah, bits are annoying. the bitshift variant as the advantage that it's the easiest to sanity check when you get to longer flag lists. I guess the sensible cutoff is at 1024 the latest.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #476, opened 2022-10-06
Feature/rm 7275 upgrade selenium to 4 5

KuJo7:
https://re-motion.atlassian.net/browse/RM-7275


a-ctor (changes requested):
Changes look good but we are not merging yet. We should run the CI build a few times to make sure that the changes do not break our tests.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityViolationAnalysisTest.cs
```diff
@@ -252,7 +252,7 @@ public void ConformanceLevel_Section508 ()
     private AccessibilityAnalyzer CreateAnalyzer (IAccessibilityConfiguration config)
     {
       return AccessibilityAnalyzer.CreateForRemoteWebDriver(
-          (RemoteWebDriver)Helper.MainBrowserSession.Driver.Native,
+          (OpenQA.Selenium.WebDriver)Helper.MainBrowserSession.Driver.Native,
```
MichaelKetting, line 255: The namespace "OpenQA.Selenium.Remote" is no longer needed. please check in all changed files.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
```diff
@@ -34,18 +34,18 @@ public void DeleteAllCookies_DeletesAllCookies ()
     {
       var browserSession = Start();
 
-      var allCookiesBeforeCreation = ((RemoteWebDriver)browserSession.Context.Browser.Driver.Native).Manage().Cookies.AllCookies;
+      var allCookiesBeforeCreation = ((OpenQA.Selenium.WebDriver)browserSession.Context.Browser.Driver.Native).Manage().Cookies.AllCookies;
```
MichaelKetting, line 37: Because Selenium updated it's type names, we no have a nameclash with our namespace "WebDriver", which means the "Selenium.WebDriver" type must be fully qualified. 

I think changing the namespace in re-motion would be a significant break, given the contents. So, @a-ctor  do you see any significant usage for Selenium.WebDriver" in downstream projects that would require additional consideration.

Please document this conflict as a breaking change ticket.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -42,7 +42,7 @@ public class AccessibilityAnalyzer
     /// </summary>
     [NotNull]
     public static AccessibilityAnalyzer CreateForRemoteWebDriver (
-        [NotNull] RemoteWebDriver remoteWebDriver,
+        [NotNull] OpenQA.Selenium.WebDriver remoteWebDriver,
```
MichaelKetting, line 45: Use IWebDriver instead, update parameter name, method name, and xmldocs.

Create breaking change ticket for the rename.

Also keep the original method with an an obsolete-attribute. Please check other usages of ObsoleteAttribute for how this should look like.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
MichaelKetting, line 37: It looks like we can simply switch to IWebDriver as the cast-type in this file.. Please do and verify.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
```diff
@@ -75,12 +75,13 @@ public void GeckoDriver_DoesNotSupportBrowserLogs ()
 
       var home = Start();
 
-      var selenium = (IWebDriver)home.Context.Browser.Driver.Native;
+      var js = JavaScriptExecutor.GetJavaScriptExecutor(home.Context.Browser);
```
MichaelKetting, line 78: I see you just took the variable name from the Test-method further down. No need to change the name now sinse it's the same as below, but just for learning: non-abbreviated names are preferred when writing new code.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
```diff
@@ -75,12 +75,13 @@ public void GeckoDriver_DoesNotSupportBrowserLogs ()
...
+      js.ExecuteScript($"console.error('Error')");
+      js.ExecuteScript($"console.warn('Warning')");
+
+      var logs = ((IWebDriver)home.Context.Browser.Driver.Native).Manage().Logs.GetLog(LogType.Browser);
 
-      Assert.That(
-          () => selenium.Manage().Logs.GetLog(LogType.Browser),
-          Throws.InstanceOf<NullReferenceException>().With.Message
-              .EqualTo("Object reference not set to an instance of an object."));
+      Assert.That(logs, Is.Empty);
```
MichaelKetting, line 84: I read something in the release notes about selenium fixing logging issues. The only fixed the exception when getting the logs, but firefox still doesn't return logs? Could you please check the release notes what they say about logging?

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityViolationAnalysisTest.cs
MichaelKetting, line 255: Becomes IWebDriver

Remotion/ObjectBinding/Web.Development.WebTesting/ScreenshotCreation/BocList/ScreenshotBocListColumn.cs
```diff
@@ -17,6 +17,7 @@
 using System;
 using System.Drawing;
 using JetBrains.Annotations;
+using OpenQA.Selenium;
 using OpenQA.Selenium.Internal;
```
MichaelKetting, line 21: Namespace obsolete

Remotion/Web/Development.WebTesting/Utilities/ActionsWithWaitSupport.cs
```diff
@@ -1,88 +0,0 @@
...
-using OpenQA.Selenium.Interactions;
-using OpenQA.Selenium.Support.UI;
-using Remotion.Utilities;
-
-namespace Remotion.Web.Development.WebTesting.Utilities
-{
-  /// <summary>
-  /// Enhanced version of Selenium's <see cref="Actions"/> class. It enables us to add WaitFor predicates after actions.
-  /// </summary>
-  public class ActionsWithWaitSupport : Actions
```
MichaelKetting, line 29: Why was this type removed? I do not see any usages in re-motion but it is problably introduced to help something downstream. Unless keeping this type would be harmful (i.e. using it with Selenium 4.5 would break things) we should not delete it at this time. 

Note; if we know that it's unused, we can make a task for marking it as obsolete in the next major release.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -378,7 +378,7 @@ public AccessibilityAnalyzer CreateAccessibilityAnalyzer ([NotNull] IBrowserSess
       ArgumentUtility.CheckNotNull("browserSession", browserSession);
 
       return AccessibilityAnalyzer.CreateForRemoteWebDriver(
-          (RemoteWebDriver)browserSession.Driver.Native,
+          (OpenQA.Selenium.WebDriver)browserSession.Driver.Native,
```
MichaelKetting, line 381: Cast to IWebDriver

MichaelKetting (changes requested):
+ @a-ctor FYI

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
a-ctor: I dont' think that this is a big problem. This only happens if you access the "native" Selenium APIs which should be a rare occurrence. And from the consumer perspective this problem only occurs if you import our `WebDriver` namespace. That should be an even rarer occurrence given that we have mostly browser specific classes in that namespace.

MichaelKetting: Okay, so we just need a breaking change documentation. Is OpenQA.Selenium.WebDriver a new type or was it simply never imported in the past?

Remotion/Web/Development.WebTesting/Utilities/ActionsWithWaitSupport.cs
a-ctor: The problem here is that parts of the `protected` `Actions` APIs are no longer available. In this case, the `AddAction` can no longer be accessed. We need to decide what to do with the class. The delete was just to remember that there is something left to do :D

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
a-ctor: It did not exist before but the `RemoteWebDriver` no longer exists so its more of a rename

Remotion/Web/Development.WebTesting/Utilities/ActionsWithWaitSupport.cs
MichaelKetting: Okay, I've checked in the project sources and found usages of ActionsWithWaitSupport -> the assumtion that everything is already done is not true, invalidating the estimate. => should we add a separate task to solve this? The impleemntation will still be in the main feature branch, but we may want to separate the estimate/planning for it.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
MichaelKetting: okay, brekaing change that the new api clashes with our namespace and we recommend using the interface whenever possible.

MichaelKetting: probably also add a breaking change that RemoteWebDriver has been replaced with WebDriver. These may be two separate breaking change tickets.

Remotion/Web/Development.WebTesting/Utilities/ActionsWithWaitSupport.cs
MichaelKetting: Note: it may be (please verify) that there's only one of dependency on this replicated several times. If there's a good alternative to solving the dependency, we don't need to implement this in re-motion the same way as it was done before.

a-ctor: `Actions` uses an `ActionBuilder` to append all the actions and then build and execute them. This builder is private and there is no way to access it. I don't quite understand why there is no extensibility on this type provided by Selenium.

My first approach would be to generate a delegate that can access the private field. Then we only need to change the `AddAction` to `GetBuilder().AddAction`. The delegate can be created once as part of static initialization and get pretty close to a normal field access (21t vs 19t). Here is an example code of that in action using a `StringBuilder` instead of `ActionBuilder` as I did not want to add the Selenium reference to my test project:

```c#
public class MyActions : Actions
{
  private static Func<Actions, StringBuilder> s_getter;

  static MyActions()
  {
    var dynamicMethod = new DynamicMethod(string.Empty, typeof(StringBuilder), new []{typeof(Actions)}, true);
    var ilGenerator = dynamicMethod.GetILGenerator();
    ilGenerator.Emit(OpCodes.Ldarg_0);
    ilGenerator.Emit(OpCodes.Ldfld, typeof(Actions).GetField("_stringBuilder", BindingFlags.NonPublic | BindingFlags.Instance));
    ilGenerator.Emit(OpCodes.Ret);

    s_getter = (Func<Actions, StringBuilder>)dynamicMethod.CreateDelegate(typeof(Func<Actions, StringBuilder>));
  }
  
  public void SayGoodbye()
  {
    s_getter(this).AppendLine("GOODBYE");
  }
}
```

`SayGoodbye` would be our `WaitFor` in this example. The biggest benefit of this approach is that we don't break any existing code. If we introduce the delegate on a helper class or introduce a base class we can help consumers that run into similar problems.

MichaelKetting: We don't need to do IL-Code here (and should use Expression instead, if we really want to), simply caching the field is more than enough in the webtest code.

Add an obsolete attribute (and breaking change doc that it is now obsolete) to this type and a follow-up ticket to remove it in re-motion v4.0. Please verify, but if I'm correct, there's other (also simple) ways to solve this ad-hoc on a project level and it removes our repsonsibility for working around Selenium design desicions.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
KuJo7: Yes, we can use IWebDriver.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserSessionTest.cs
KuJo7: FirefoxDriver doesn't implement ISupportLogs like other BrowserDrivers, so there is no support for reading logs at this point. The release notes only mention retrieving logs in relation to Chrome.

MichaelKetting: okay, so the only improvement was to get rid of the exception. okay.

Remotion/Web/Development.WebTesting.IntegrationTests/Web.Development.WebTesting.IntegrationTests.csproj
```diff
@@ -21,5 +21,8 @@
     <ProjectReference Include="..\Development.WebTesting.WebFormsControlObjects\Web.Development.WebTesting.WebFormsControlObjects.csproj" />
     <ProjectReference Include="..\Development.WebTesting\Web.Development.WebTesting.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Selenium.WebDriver" Version="4.5.0" />
+  </ItemGroup>
```
MichaelKetting, line 26: Why do we need to add this reference? it Wasn't needed before the version upgrade.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -36,10 +36,34 @@ public class AccessibilityAnalyzer
...
+      ArgumentUtility.CheckNotNull("logger", logger);
+
+      return new AccessibilityAnalyzer(webDriver, (IJavaScriptExecutor)webDriver, axeResultParser, configuration, sourceProvider, mapper, logger);
+    }
+
     /// <summary>
     /// Creates an <see cref="AccessibilityAnalyzer"/> that uses an instance of <see cref="RemoteWebDriver"/> for both <see cref="IWebDriver"/>
     /// and <see cref="IJavaScriptExecutor"/>.
     /// </summary>
+    [Obsolete("Use CreateForWebDriver(IWebDriver, IAxeResultParser, IAccessibilityConfiguration, IAxeSourceProvider, IAccessibilityResultMapper, ILog) instead. ", true)]
```
MichaelKetting, line 66:
```suggestion
    [Obsolete("Use CreateForWebDriver(IWebDriver, IAxeResultParser, IAccessibilityConfiguration, IAxeSourceProvider, IAccessibilityResultMapper, ILog) instead. (Version: 3.20.0)", false)]
```

I always add the version when the obsolete was introduced. It allows for efficient removal of obsolete-attributes that are old enough to no longer serve a purpose during migration. Typically, in the next major version.

The boolea flag specifies if the attribute is a warning or an error. Code that still works as expected should not be fitted with an error-level obsolete-attribute. And if you do add an error-level obsolete, the implementation should be replaced with a NotSupportedException (using the attribute's text as the message). That way, uneeded dependencies can be dropped and in general, no longer invokable code is removed.

Please check for things such as trailing spaces in messages.

MichaelKetting (changes requested):

MichaelKetting (approved):

Remotion/Web/Development.WebTesting/Web.Development.WebTesting.csproj
```diff
@@ -22,11 +22,10 @@
     <ProjectReference Include="..\Contracts.DiagnosticMetadata\Web.Contracts.DiagnosticMetadata.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Fork.Coypu" Version="3.1.1.2" />
-    <PackageReference Include="Microsoft.Edge.SeleniumTools" Version="3.141.1" />
+    <PackageReference Include="Fork.Coypu" Version="3.2.0.1" />
     <PackageReference Include="log4net" Version="2.0.12" />
-    <PackageReference Include="Selenium.Support" Version="3.141.0" />
-    <PackageReference Include="Selenium.WebDriver" Version="3.141.0" />
+    <PackageReference Include="Selenium.Support" Version="4.5.0" />
```
MichaelKetting, line 27: So we can manage this: please keep this conversation open until we have started v3.20 on the develop branch. Please check with @MichaelKetting at the start of November.

MichaelKetting (approved):

a-ctor (approved):

PR #483, opened 2022-10-12
Bugfix/rm 8596 port number missing

a-ctor:
https://re-motion.atlassian.net/browse/RM-8596

### DIFF
--- a/Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
+++ b/Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
@@ -144,42 +144,47 @@ namespace Remotion.Web.Development.WebTesting
 
       var stopwatch = Stopwatch.StartNew();
 
-#pragma warning disable SYSLIB0014
-      var webRequest = (HttpWebRequest)HttpWebRequest.Create(resolvedUri); // TODO RM-8492: Replace with HttpClient
-#pragma warning restore SYSLIB0014
-      webRequest.Method = WebRequestMethods.Http.Head;
-      webRequest.AllowAutoRedirect = true;
-      webRequest.Host = webApplicationRoot.Host;
-      webRequest.ServerCertificateValidationCallback += (sender, certificate, chain, errors) => true;
-
       HttpStatusCode statusCode = default;
       Assertion.DebugAssert(statusCode != HttpStatusCode.OK);
 
-      while (statusCode != HttpStatusCode.OK)
+      while (true)
       {
+#pragma warning disable SYSLIB0014
+        var webRequest = (HttpWebRequest)HttpWebRequest.Create(resolvedUri); // TODO RM-8492: Replace with HttpClient
+#pragma warning restore SYSLIB0014
+        webRequest.Method = WebRequestMethods.Http.Head;
+        webRequest.AllowAutoRedirect = true;
+        webRequest.Host = webApplicationRoot.GetComponents(UriComponents.Host | UriComponents.Port, UriFormat.UriEscaped);
+        webRequest.ServerCertificateValidationCallback += (sender, certificate, chain, errors) => true;
+
+        var remainingTimeout = (int)(applicationPingTimeout.TotalMilliseconds - stopwatch.Elapsed.TotalMilliseconds);
+        if (remainingTimeout <= 0)
+        {
+          s_log.Warn($"Checking the web application root '{webApplicationRoot}' timed out (HTTP status code: '{statusCode}').");
+          throw new WebException(
+              $"Checking the web application root '{webApplicationRoot}' did not return '{HttpStatusCode.OK}' in the defined {nameof(applicationPingTimeout)} ({applicationPingTimeout}). "
+              + $"Last received HTTP status code was: '{statusCode}'.",
+              WebExceptionStatus.Timeout);
+        }
+
         try
         {
-          var remainingTimeout = (int)(applicationPingTimeout.TotalMilliseconds - stopwatch.Elapsed.TotalMilliseconds);
-          webRequest.Timeout = Math.Max(remainingTimeout, 0);
+          webRequest.Timeout = remainingTimeout;
 
-          using (var response = (HttpWebResponse)webRequest.GetResponse())
-          {
-            statusCode = response.StatusCode;
-          }
+          using var response = (HttpWebResponse)webRequest.GetResponse();
+          s_log.Info($"Verified that '{resolvedUri}' is accessible after {stopwatch.Elapsed.TotalMilliseconds:N0} ms.");
+          return;
         }
         catch (WebException ex)
         {
-          CheckTimeout(webApplicationRoot, applicationPingTimeout, stopwatch, $"Failed with WebException '{ex.Message}'");
-        }
+          if (ex.Response is HttpWebResponse httpWebResponse)
+            statusCode = httpWebResponse.StatusCode;
 
-        CheckTimeout(webApplicationRoot, applicationPingTimeout, stopwatch, $"Failed with HttpStatusCode '{statusCode}'");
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -144,31 +144,37 @@ private void VerifyWebApplicationStarted (Uri webApplicationRoot, TimeSpan appli
...
+
+            using (var response = (HttpWebResponse)webRequest.GetResponse())
+            {
+              statusCode = response.StatusCode;
+            }
           }
         }
         catch (WebException ex)
         {
+          if (ex.Response is HttpWebResponse httpWebResponse)
```
MichaelKetting, line 175: Is there a valid scenario where this can actually happen or is it just to prevent infinite loops?

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -144,31 +144,37 @@ private void VerifyWebApplicationStarted (Uri webApplicationRoot, TimeSpan appli
...
+        webRequest.Host = webApplicationRoot.GetComponents(UriComponents.Host | UriComponents.Port, UriFormat.UriEscaped);
+        webRequest.ServerCertificateValidationCallback += (sender, certificate, chain, errors) => true;
+
         try
         {
           var remainingTimeout = (int)(applicationPingTimeout.TotalMilliseconds - stopwatch.Elapsed.TotalMilliseconds);
-          webRequest.Timeout = Math.Max(remainingTimeout, 0);
-
-          using (var response = (HttpWebResponse)webRequest.GetResponse())
+          if (remainingTimeout > 0)
```
MichaelKetting, line 163: I think we now have an infinite loop risk here. We need to set statusCode = Timeout in the else-branch.

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -144,31 +144,37 @@ private void VerifyWebApplicationStarted (Uri webApplicationRoot, TimeSpan appli
...
       Assertion.DebugAssert(statusCode != HttpStatusCode.OK);
 
       while (statusCode != HttpStatusCode.OK)
       {
+  #pragma warning disable SYSLIB0014
+        var webRequest = (HttpWebRequest)HttpWebRequest.Create(resolvedUri); // TODO RM-8492: Replace with HttpClient
+  #pragma warning restore SYSLIB0014
+        webRequest.Method = WebRequestMethods.Http.Head;
+        webRequest.AllowAutoRedirect = true;
+        webRequest.Host = webApplicationRoot.GetComponents(UriComponents.Host | UriComponents.Port, UriFormat.UriEscaped);
```
MichaelKetting, line 157: Do we need to specify UriFormat.UriEscaped at this point given we only use the existing hostname? I don't mind being safe, but if it's just overprotectiveness, it may reduce readability.

a-ctor: We have to supply a value, there is no overload with only one parameter and this is the default that is used in URI in other formatting scenarios.

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
MichaelKetting: Oh, yeah, now I see it.

a-ctor:
@MichaelKetting From my point of view all the comments are resolved? Is there something left to do?

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -144,31 +144,37 @@ private void VerifyWebApplicationStarted (Uri webApplicationRoot, TimeSpan appli
...
+            }
           }
         }
         catch (WebException ex)
         {
+          if (ex.Response is HttpWebResponse httpWebResponse)
+            statusCode = httpWebResponse.StatusCode;
+
           CheckTimeout(webApplicationRoot, applicationPingTimeout, stopwatch, $"Failed with WebException '{ex.Message}'");
         }
```
MichaelKetting, line 180: Not related to the current bugfix, but perhaps still something that should be fixed: if status code is 200 and no timeout happened, we still test for the timeout in line 181 and might trigger a timeout because the stopwatch is still running. And then we go to sleep for 500ms before we exit the operation. I think we should exit the method if the status code is successful (move the info log in line 188 into the getresponse in line 169.

thoughts?

MichaelKetting (approved):

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
a-ctor: I think we don't even need to check the status code. If `.GetResponse()` did not throw the server should be reachable so we can just log and return instead of setting the status code. This allows `201` or `202` status codes as well. Something like:

```c#
while (true)
{
#pragma warning disable SYSLIB0014
  var webRequest = (HttpWebRequest)HttpWebRequest.Create(resolvedUri); // TODO RM-8492: Replace with HttpClient
#pragma warning restore SYSLIB0014
  webRequest.Method = WebRequestMethods.Http.Head;
  webRequest.AllowAutoRedirect = true;
  webRequest.Host = webApplicationRoot.GetComponents(UriComponents.Host | UriComponents.Port, UriFormat.UriEscaped);
   webRequest.ServerCertificateValidationCallback += (sender, certificate, chain, errors) => true;

  try
  {
    var remainingTimeout = (int)(applicationPingTimeout.TotalMilliseconds - stopwatch.Elapsed.TotalMilliseconds);
    if (remainingTimeout > 0)
    {
      webRequest.Timeout = remainingTimeout;

      using (var response = (HttpWebResponse)webRequest.GetResponse())
      {
        s_log.Info($"Verified that '{resolvedUri}' is accessible after {stopwatch.Elapsed.TotalMilliseconds:N0} ms.");
        return;
      }
    }
  }
  catch (WebException ex)
  {
    if (ex.Response is HttpWebResponse httpWebResponse)
      statusCode = httpWebResponse.StatusCode;

    CheckTimeout(webApplicationRoot, applicationPingTimeout, stopwatch, $"Failed with WebException '{ex.Message}'");
  }

  Thread.Sleep(TimeSpan.FromMilliseconds(500));
}
```

If you want we can do this change in this issue. I see no point in rolling it up again

MichaelKetting: sounds good.

MichaelKetting: The new example is a bit buggy yet, if timeout becomes zero/negative outside of a request, it will infinite-loop. I reopen the review to give it a final sanity check.

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -144,42 +144,46 @@ private void VerifyWebApplicationStarted (Uri webApplicationRoot, TimeSpan appli
...
+#pragma warning restore SYSLIB0014
+        webRequest.Method = WebRequestMethods.Http.Head;
+        webRequest.AllowAutoRedirect = true;
+        webRequest.Host = webApplicationRoot.GetComponents(UriComponents.Host | UriComponents.Port, UriFormat.UriEscaped);
+        webRequest.ServerCertificateValidationCallback += (sender, certificate, chain, errors) => true;
+
+        var remainingTimeout = (int)(applicationPingTimeout.TotalMilliseconds - stopwatch.Elapsed.TotalMilliseconds);
+        if (remainingTimeout <= 0)
+        {
+          throw new WebException(
```
MichaelKetting, line 163: I think we should add a logstatement here, since we have one in the exception-case, too.

MichaelKetting (approved):

PR #485, opened 2022-10-12
 bugfix/RM-8615-DictionaryExtensions.GetOrCreateValue-Is-Incompatible-With-ConcurrentDictionary-v3.1

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-8615

### DIFF
--- a/Remotion/Core/Core/Collections/DictionaryExtensions.cs
+++ b/Remotion/Core/Core/Collections/DictionaryExtensions.cs
@@ -16,6 +16,7 @@
 // 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


Remotion/Core/Core/Collections/DictionaryExtensions.cs
```diff
@@ -108,11 +109,20 @@ public static class DictionaryExtensions
...
     }
 
+    [Obsolete("ConcurrentDictionary is not supported for GetOrCreateValue(...). (Version: 3.0)", true)]
+    public static TValue GetOrCreateValue<TKey, TValue> (this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> valueFactory)
+        where TKey : notnull
+    {
+      throw new NotSupportedException("ConcurrentDictionary is not supported for GetOrCreateValue(...).");
+    }
+
     public static TValue GetOrCreateValue<TKey, TValue> (this IDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> valueFactory)
```
a-ctor, line 119: @MichaelKetting Looking at the other develop PR we did not adapt the signature to `Dictionary<K, V>`. I assume this is intended? This way, there might be call sites that do not cause a compile exception but will throw in runtime (If the concurrent dictionary is stored as `IDictionary`)

MichaelKetting: @a-ctor yes, that's intentional. In the backport I tried to do the least-invasive version while still maintaining safety:

1. by adding the ConcurrentDictionary overload we can catch any strong-typed usages
2. by adding a runtime guard, we trip anyone who accidentally passes a ConcurrentDictionary via IDictionary. The intention is that it is better to fail consistently than to have an undiscovered concurrency issue in the code. I'll add a breaking change ticket, too.

a-ctor (approved):

PR #496, opened 2022-10-18
Feature/rm 8331 add analyzers that guard against possible safe context error suggesting the correct helper construct

Frazerus:
Added all analyzers except for the one of System.Timers.Time and the ones not included in the System Assembly.
https://re-motion.atlassian.net/browse/RM-8331

### DIFF
--- a/Build/Shared.build.props
+++ b/Build/Shared.build.props
@@ -58,6 +58,11 @@
 
   <ItemGroup>
     <PackageReference Include="Remotion.Infrastructure.Styles.Analyzer" Version="1.1.3" PrivateAssets="All" />
+    <ProjectReference Include="$(SolutionDir)Remotion\Core\Development.Analyzers\Core.Development.Analyzers.csproj"
+                      PrivateAssets="all"
+                      ReferenceOutputAssembly="false"
+                      SetTargetFramework="TargetFramework=netstandard2.0"
+                      OutputItemType="Analyzer" />
   </ItemGroup>
 
   <PropertyGroup>
--- a/Remotion/Core/Collections.Caching.UnitTests/ConcurrentCacheTest.cs
+++ b/Remotion/Core/Collections.Caching.UnitTests/ConcurrentCacheTest.cs
@@ -107,7 +107,7 @@ namespace Remotion.Collections.Caching.UnitTests
       var waitHandleThread1a = new ManualResetEvent(false);
       var waitHandleThread1b = new ManualResetEvent(false);
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
@@ -121,7 +121,7 @@ namespace Remotion.Collections.Caching.UnitTests
                 });
             waitHandleThread1b.Set();
           });
-
+#pragma warning restore RMCORE0001
       Assert.That(waitHandleThread1a.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
       var resultThread1 = _cache.GetOrCreateValue(
           "key1",
@@ -142,14 +142,14 @@ namespace Remotion.Collections.Caching.UnitTests
       var expectedThread1 = "T1";
       object resultThread2 = null;
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
             Assert.That(waitHandleThread2.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
             resultThread2 = _cache.GetOrCreateValue("key", key => throw new InvalidOperationException());
           });
-
+#pragma warning restore RMCORE0001
       var resultThread1 = _cache.GetOrCreateValue(
           "key",
           key =>
@@ -167,14 +167,14 @@ namespace Remotion.Collections.Caching.UnitTests
     {
       object resultThread2 = null;
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
             Assert.That(waitHandleThread2.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
             resultThread2 = _cache.GetOrCreateValue("key", key => throw new InvalidOperationException());
           });
-
+#pragma warning restore RMCORE0001
       var resultThread1 = _cache.GetOrCreateValue(
           "key",
           key =>
@@ -203,14 +203,14 @@ namespace Remotion.Collections.Caching.UnitTests
       var expectedThread1 = "T1";
       object resultThread2 = null;
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
             Assert.That(waitHandleThread2.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
             _cache.TryGetValue("key", out resultThread2);
           });
-
+#pragma warning restore RMCORE0001
       var resultThread1 = _cache.GetOrCreateValue(
           "key",
           key =>
@@ -522,11 +522,11 @@ namespace Remotion.Collections.Caching.UnitTests
 
       GC.Collect();
       GC.WaitForFullGCComplete();
-
+#pragma warning disable RMCORE0001
       var threads = new Thread[threadCount];
       for (int i = 0; i < threadCount; i++)
         threads[i] = new Thread(threadStart);
-
+#pragma warning restore RMCORE0001
       for (int i = 0; i < threadCount; i++)
         threads[i].Start(stopwatches[i]);
 
--- a/Remotion/Core/Collections.DataStore.UnitTests/ConcurrentDataStoreTest.cs
+++ b/Remotion/Core/Collections.DataStore.UnitTests/ConcurrentDataStoreTest.cs
@@ -468,7 +468,7 @@ namespace Remotion.Collections.DataStore.UnitTests
       var waitHandleThread1a = new ManualResetEvent(false);
       var waitHandleThread1b = new ManualResetEvent(false);
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
@@ -482,7 +482,7 @@ namespace Remotion.Collections.DataStore.UnitTests
                 });
             waitHandleThread1b.Set();
           });
-
+#pragma warning restore RMCORE0001
       Assert.That(waitHandleThread1a.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
       var resultThread1 = _store.GetOrCreateValue(
           "key1",
@@ -503,14 +503,14 @@ namespace Remotion.Collections.DataStore.UnitTests
       var expectedThread1 = "T1";
       object resultThread2 = null;
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
             Assert.That(waitHandleThread2.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
             resultThread2 = _store.GetOrCreateValue("key", key => throw new InvalidOperationException());
           });
-
+#pragma warning restore RMCORE0001
       var resultThread1 = _store.GetOrCreateValue(
           "key",
           key =>
@@ -528,14 +528,14 @@ namespace Remotion.Collections.DataStore.UnitTests
     {
       object resultThread2 = null;
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
             Assert.That(waitHandleThread2.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
             resultThread2 = _store.GetOrCreateValue("key", key => throw new InvalidOperationException());
           });
-
+#pragma warning restore RMCORE0001
       var resultThread1 = _store.GetOrCreateValue(
           "key",
           key =>
@@ -554,14 +554,14 @@ namespace Remotion.Collections.DataStore.UnitTests
       var expectedThread1 = "T1";
       object resultThread2 = null;
       var waitHandleThread2 = new ManualResetEvent(false);
-
+#pragma warning disable RMCORE0001
       var thread2 = Task.Run(
           () =>
           {
             Assert.That(waitHandleThread2.WaitOne(TimeSpan.FromSeconds(1)), "Dead lock detected.");
             _store.TryGetValue("key", out resultThread2);
           });
-
+#pragma warning restore RMCORE0001
       var resultThread1 = _store.GetOrCreateValue(
           "key",
           key =>
@@ -676,11 +676,11 @@ namespace Remotion.Collections.DataStore.UnitTests
 
       GC.Collect();
       GC.WaitForFullGCComplete();
-
+#pragma warning disable RMCORE0001
       var threads = new Thread[threadCount];
       for (int i = 0; i < threadCount; i++)
         threads[i] = new Thread(threadStart);
-
+#pragma warning restore RMCORE0001
       for (int i = 0; i < threadCount; i++)
         threads[i].Start(stopwatches[i]);
 
--- a/Remotion/Core/Core/Context/SafeContext.ExecutionContext.cs
+++ b/Remotion/Core/Core/Context/SafeContext.ExecutionContext.cs
@@ -13,7 +13,7 @@
 // 
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
-// 
+//
 using System.Threading;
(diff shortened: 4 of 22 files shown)


Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <LangVersion>9</LangVersion>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <DevelopmentDependency>true</DevelopmentDependency>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <AssemblyName>Remotion.Core.Development.Analyzers</AssemblyName>
+    <AssemblyTitle>Remotion Core Analyzers</AssemblyTitle>
+    <Description>Analyzer ensuring best practices when using the Remotion.Web assemblies.</Description>
```
a-ctor, line 13:
```suggestion
    <Description>Analyzer ensuring best practices when using the Remotion.Core assemblies.</Description>
```

Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
```diff
@@ -0,0 +1,29 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.8.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="$(OutputPath)\netstandard2.0\$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
+  </ItemGroup>
```
a-ctor, line 22: We are probably going to merge after #497 so we should do the same changes here. We want the analyzer project to not emit a nuget package, instead we want the Core project to embed the analyzer assembly. See the PR for the required changes

Remotion/Web/Development.Analyzers.IntegrationTests/Web.Development.Analyzers.IntegrationTests.csproj
```diff
@@ -20,6 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\Core\Core\Core.Core.csproj" />
```
a-ctor, line 23: Why do we need this?

Remotion/Core/Development.Analyzers.IntegrationTests/CSharpAnalyzerVerifier.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+using Remotion.Context;
+
+namespace Remotion.Core.Development.Analyzers.IntegrationTests
+{
+  public static class CSharpAnalyzerVerifier<TAnalyzer>
+      where TAnalyzer : DiagnosticAnalyzer, new()
+  {
+    private static Diagnostic _diagnosticToRun;
```
a-ctor, line 32: This is not needed?

Remotion/Core/Development.Analyzers.IntegrationTests/CSharpAnalyzerVerifier.cs
```diff
@@ -0,0 +1,63 @@
...
+                     SolutionTransforms = { (solution, id) =>
+                     {
+                       var project = solution.GetProject(id);
+                       if (withSafeContextReference)
+                       {
+                         project = project
+                             .AddMetadataReference(MetadataReference.CreateFromFile(contextAssemblyLocation));
+                       }
+                       return project.Solution;
+                     } }
```
a-ctor, line 56:
```suggestion
                     SolutionTransforms =
                     {
                       (solution, id) =>
                       {
                         var project = solution.GetProject(id);
                         if (withSafeContextReference)
                         {
                           project = project
                               .AddMetadataReference(MetadataReference.CreateFromFile(contextAssemblyLocation));
                         }
                         return project.Solution;
                       }
                     }
```

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -0,0 +1,236 @@
...
+using System.Threading.Tasks;
+using Core.Development.Analyzers;
+using NUnit.Framework;
+using Verifier = Remotion.Core.Development.Analyzers.IntegrationTests.CSharpAnalyzerVerifier<Core.Development.Analyzers.RMCORE0001_SafeContextAnalyzer>;
+using Analyzer = Core.Development.Analyzers.RMCORE0001_SafeContextAnalyzer;
+
+namespace Remotion.Core.Development.Analyzers.IntegrationTests
+{
+  [TestFixture]
+  public class RMCORE0001_SafeContextAnalyzerTests
```
a-ctor, line 11: There are a few formatting problems in this file,. Should be fixed with a reformat command.

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -0,0 +1,236 @@
...
+    );
+
+    Parallel.Invoke(() => Console.Write(""a""));
+  }
+}
+";
+      var diagnosticFor = Verifier.Diagnostic(Analyzer.AlternativeDescriptor)
+          .WithSpan(12, 5, 12, 77)
+          .WithMessage(
+              "'System.Threading.Tasks.Parallel.For' should not be used with SafeContext by itself, use 'provider.OpenSafeContextBoundary()' to create a safe context boundary.");
```
a-ctor, line 147: We have an API for parallel called `SafeContext.Parallel.OpenSafeContextBoundary` that should be used instead of `provider.OpenSafeContextBoundary` so we suggest that here

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -0,0 +1,236 @@
...
+  { 
+    SafeContext.Instance.SetData(""a"", 3);
+    var timer = new Timer(new TimerCallback(_ => { }));
+  }
+}
+";
+      var diagnosticFunc = Verifier.Diagnostic(Analyzer.AlternativeDescriptor)
+          .WithSpan(9, 17, 9, 55)
+          .WithMessage(
+              "'new System.Threading.Timer' should not be used with SafeContext by itself, use 'provider.OpenSafeContextBoundary()' to create a safe context boundary.");
```
a-ctor, line 208: `SafeContext.Threading.NewTimer`

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -0,0 +1,236 @@
...
+    {
+      const string input = @"
+using System.Timers;
+using Remotion.Context;
+public class A
+{
+  public void Test ()
+  { 
+    SafeContext.Instance.SetData(""a"", 3);
+    var timer = new Timer(new TimerCallback(_ => { }));
```
a-ctor, line 224: Does that compile? Because `System.Timers.Timer` does not have a callback constructor. So in this case we want to check for the elapsed event and suggest `SafeContext.Timers.AddElapsedEventHandler`

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -0,0 +1,277 @@
...
+      IMethodSymbol outMethodSymbol;
+      var errorMessageParameter = new ErrorMessageParameter();
+      var invocationSymbolInfo = ctx.SemanticModel.GetSymbolInfo(ctx.Node);
+
+      if (invocationSymbolInfo.Symbol == null)
+        return;
+
+      var invocationOriginalDefinitionSymbol = invocationSymbolInfo.Symbol.OriginalDefinition;
+
+      if (methodSymbols.TaskRunMethodSymbols != null && TryFindItem(methodSymbols.TaskRunMethodSymbols, s => SymbolEqualityComparer.Default.Equals(invocationOriginalDefinitionSymbol, s), out outMethodSymbol))
```
a-ctor, line 141: We are repeating a lot of the same diagnostic generation logic. We could change `MethodSymbolContext` to contain a list of `(IMethodSymbol method, Action<SyntaxNodeAnalysisContext , IMethodSymbol>)` and have all the "what error has what message" logic be moved into the `Initialize` method. This would simplify the logic here and probably also in `Initialize`. What do you think?

a-ctor (changes requested):

Remotion/Web/Development.Analyzers.IntegrationTests/Web.Development.Analyzers.IntegrationTests.csproj
Frazerus: We need this to reference Remotion.Core.Context in the verifier.

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
Frazerus: Not sure about that, reformatting only removed the `withSafeContextReference` parameter description

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
Frazerus: Do not know whether this should be the alternative or the message, did not find `Safecontext.Threading.NewTimer` or anything with a timer in safecontext, really.

a-ctor: You need to rebase to see the new classes regarding `ContinueWith`, and all the timers. The issue was still being merged when you started working on this issue.

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -0,0 +1,285 @@
...
+            var tasksTaskSymbol = compilationContext.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task");
+            var parallelSymbol = compilationContext.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Parallel");
+            var threadSymbol = compilationContext.Compilation.GetTypeByMetadataName("System.Threading.Thread");
+            var threadingTimerSymbol = compilationContext.Compilation.GetTypeByMetadataName("System.Threading.Timer");
+            var executionSymbol = compilationContext.Compilation.GetTypeByMetadataName("System.Threading.ExecutionContext");
+            var timingTimerSymbol = compilationContext.Compilation.GetTypeByMetadataName("System.Timers.Timer");
+
+            var listOfMethodsAndActions = new List<(IMethodSymbol, Action<SyntaxNodeAnalysisContext, ISymbol>)>();
+
+            if (tasksTaskSymbol != null)
```
a-ctor, line 50: I think we can cut down on the verbosity here still. How about we sit down together and take a look?

a-ctor (changes requested):

Build/Shared.build.props
```diff
@@ -58,6 +58,7 @@
 
   <ItemGroup>
     <PackageReference Include="Remotion.Infrastructure.Styles.Analyzer" Version="1.1.3" PrivateAssets="All" />
+    <PackageReference Include="Remotion.Core.Development.Analyzers" Version="1.0.0" PrivateAssets="All" />
```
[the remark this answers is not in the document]
Frazerus: @MichaelKetting Deine Meinung? :)

Remotion/Core/Core/Context/SafeContext.ExecutionContext.cs
```diff
@@ -25,6 +26,7 @@ public static partial class SafeContext
     /// Provides methods for running callbacks inside of a <see cref="System.Threading.ExecutionContext"/>s in a <see cref="SafeContext"/> aware manner.
     /// <see cref="SafeContext"/> values do not flow into callback.
     /// </summary>
+
```
a-ctor, line 29: revert

Remotion/Core/Core/Context/SafeContext.ExecutionContext.cs
```diff
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
+using System.Diagnostics.CodeAnalysis;
```
a-ctor, line 17: we don't need that anymore?

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -0,0 +1,258 @@
...
+
+
+      return typeSymbol
+          .GetMembers(name)
+          .Cast<IEventSymbol>()
+          .Where(s => s.AddMethod != null)
+          .Select(s => s.AddMethod!);
+    }
+
+    private static void AddRange (List<(IMethodSymbol, Action<SyntaxNodeAnalysisContext, ISymbol>)> listWithBoth, IEnumerable<IMethodSymbol> symbols, Action<SyntaxNodeAnalysisContext, ISymbol> action)
```
a-ctor, line 179: We need a better name for this. We might also pull in more of the logic to allow the main registration above to be a bit more streamlined. Ideally, one could add new methods just by adding a few lines in `RegisterCompilationStartAction` action above.

Build/Shared.build.props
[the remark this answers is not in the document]
MichaelKetting: ProjectReference is correct for unit tests. Within the same solution we should never have PackageReferences because it would be...weird and likely not work/behave strangely. (see the separate analyzers repository)

As for referencing in library projects, puh, if we want this guard to also check things in re-motion, I don't know how things should work. The question is, do we want/need this guard also inside remotion itself? If so, we probably should reference an earlier, already released version to avoid blowing us up during development. Of course, there's also far less risk in re-motion, I'd hope. We could probably make this a separate issue from unit testing the project? Also, how do we handle the web analyzers @a-ctor ?

@Frazerus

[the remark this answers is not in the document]
a-ctor: Using it in re-motion is not a problem. We do this with the Web.Analyzers already. The question is just which projects should receive the analyzer reference. If we do it in `Shared.build.props` everyone gets it, if we do it in a project type file only certain projects get the reference. I think it is fine to leave as is in `Shared.build.props`

Remotion/Core/Core/Context/SafeContext.Task.cs
```diff
@@ -27,6 +28,7 @@ public static partial class SafeContext
     /// Provides methods for running <see cref="SystemTask"/>s which are <see cref="SafeContext"/> aware.
     /// <see cref="SafeContext"/> values do not flow into the executed tasks.
     /// </summary>
+    [SuppressMessage("Usage", "RMCORE0001:Use SafeContext instead of typical API")]
```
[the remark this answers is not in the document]
MichaelKetting: @a-ctor 
cc @Frazerus 

I tend to lean towards SuppressMessageAttribute, mainly because I like attributes and it does seem the "correct" choice. I would like to do a quick discussion of the attribute and the pros and cons after the Standup and then document the findings here to make sure it's not just a gut desicion.

a-ctor (approved):

Remotion/Core/Collections.Caching.UnitTests/ConcurrentCacheTest.cs
```diff
@@ -26,6 +27,7 @@
 namespace Remotion.Collections.Caching.UnitTests
 {
   [TestFixture]
+  [SuppressMessage("Usage", "RMCORE0001: Use Safecontext instead of typical API")]
```
a-ctor, line 30: I think we should use `pragma` instead instead of `SupressMessage` in this test class. The idea is that any newly added code is reminded of the safe context problem instead of having the message already supressed. These tests probably won't have any SafeContext problems to begin with but I think it is cleaner this way.

The SafeContext related tests on the other hand should stay with `SupressMessage`. This is were almost every test would create a problem making pragma unattractive. Also since the tests are SafeContext related they any changes would need to consider SafeContext anyway so there is no need for the extra reminder

Remotion/Core/Collections.DataStore.UnitTests/ConcurrentDataStoreTest.cs
```diff
@@ -11,6 +12,7 @@
 namespace Remotion.Collections.DataStore.UnitTests
 {
   [TestFixture]
+  [SuppressMessage("Usage", "RMCORE0001: Use Safecontext instead of typical API")]
```
a-ctor, line 15: I think we should use `pragma` instead instead of `SupressMessage` in this test class.

a-ctor (changes requested):

a-ctor (approved):

PR #501, opened 2022-10-20
RM-8578 Update JetBrains Annotations to 2022.1.0

KUJO7:
https://re-motion.atlassian.net/browse/RM-8578


Remotion/Development/Core/Development.Core.csproj
```diff
@@ -12,9 +12,7 @@
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\..\..\SharedSource\Development\UnitTesting\**\*"
-             Link="SharedSource\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)"
-             Exclude="**\PEVerifyVersion.cs;**\NUnit\**" />
+    <Compile Include="..\..\..\SharedSource\Development\UnitTesting\**\*" Link="SharedSource\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)" Exclude="**\PEVerifyVersion.cs;**\NUnit\**" />
```
a-ctor, line 15: Those seem to be formatting changes? If so, we should reset them

Remotion/Globalization/UnitTests/Globalization.UnitTests.csproj
```diff
@@ -22,8 +22,7 @@
   </ItemGroup>
 
   <Target Name="CopyTestAssemblyFilesToOutputPath" AfterTargets="Build" Condition="'$(TargetFramework)' != ''">
-    <Copy SourceFiles="$(MSBuildProjectDirectory)\$(IntermediateOutputPath)Remotion.Globalization.UnitTests.TestDomain.Resources.ClassWithResources.resources"
-          DestinationFiles="$(MSBuildProjectDirectory)\$(OutputPath)\TestDomain\TestAssemblies\En\En.Globalization.ClassWithResources.resources" />
+    <Copy SourceFiles="$(MSBuildProjectDirectory)\$(IntermediateOutputPath)Remotion.Globalization.UnitTests.TestDomain.Resources.ClassWithResources.resources" DestinationFiles="$(MSBuildProjectDirectory)\$(OutputPath)\TestDomain\TestAssemblies\En\En.Globalization.ClassWithResources.resources" />
```
a-ctor, line 25: formatting changes? if so reset

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.NetFramework/Web.config
```diff
@@ -19,16 +19,15 @@
   </appSettings>
 
   <system.web>
-    <globalization culture="de-AT" uiCulture="en-US"/>
```
a-ctor, line 22: I think that all of the `Web.config` changes are not needed and just caused by the tooling during a Nuget package upgrade. So we can reset all of them.

a-ctor (changes requested):

MichaelKetting (changes requested):
Please alo update the embedded files in SharedSource
image

MichaelKetting:
> > Please alo update the embedded files in SharedSource image
> 
> Currently the Annotations are in separate files in "Remotion/SharedSource/ReSharperAnnotations". How should this be done? Delete all files and have only one single file?

Ah yes. It's a bit more complicated: stuff that's related to ASP.NET etc is ignored. The rest should be kept as individual files. The sources also have removed visibility modifiers to make them internal-by-default. This also needs to be kepts. So, it becomes a "paste type content" operation and add new/missing stuff. I don't want this to be busy work, so we should apply sensible thought to it to keep things efficient. Possibly, there's just some new stuff that's valuable? 

Second consideration / alternative approach:
There's only Core.Core, Mixins.Core, and NUnit2-Sandbox that use the sourcefile-version from within re-motion. Beyond that, it's only for other projects that want to have a NuGet source package with the attributes. The last time we did an update was in 2017. We could split this into a separate task with a new esitmate if you think doing this explodes the current task and put it off until the next release.

Also, maybe @a-ctor has good ideas?

MichaelKetting:
> > Please alo update the embedded files in SharedSource
> 
> Currently the Annotations are in separate files in "Remotion/SharedSource/ReSharperAnnotations". How should this be done? Delete all files and have only one single file?

I just realized, my earlier thoughts where driven from an outdated baseline. Years ago, we embedded the JetBrains Annotations as public API in Remotion.dll and also had the option to embedd them privately for assemblies that do not depend on Remotion.dll but shoudl also not introduce JetBrains Annotations as public API in oder to avoid potential dependency conflicts.

Fast forward to today: we use JetBrains via the NuGet package dependency / as an Assembly for most parts. Only the private (source) dependencies remain for some special cases. So, new approach: we can just drop the entire existing list of files and replace it with the `internal` visible generated official sources. It doesn't matter if we get more types than before since they're not shared / don't need patching anyway. I'm also considering (down the line) of replacing the external usages of the package with the pasted sources. One less NuGet package to maintain and no benefit in keeping it.

Please also update PackageInfo.ps1, there's the version number for JetBrains Annotations embedded in the file (line 312).

@KuJo7

MichaelKetting (approved):

a-ctor (approved):

PR #507, opened 2022-11-11
RM-8331 Add safecontext analyzer

Frazerus:
https://re-motion.atlassian.net/browse/RM-8331

The only difference to before is that now the project is also included in projects.props. Could pack it locally now.


Frazerus:
Hi! The nightly build worked now, but only after 2 errors in webtesting? 
I suppose those are flukes, but I'm not sure. 

@MichaelKetting any thoughts on the matter?

MichaelKetting:
> Hi! The nightly build worked now, but only after 2 errors in webtesting? I suppose those are flukes, but I'm not sure.
> 
> @MichaelKetting any thoughts on the matter?

@Frazerus 
TestInjectSourceToIFrames is flaky and already has a ticket
TestDownloadTwice_WithUnknownFileName_PreventsFileNameConflicts failed once, maybe flaky.

neither test is related to the current branch

a-ctor (approved):

PR #510, opened 2022-11-18
RM-8579 Use updated user agent information

KUJO7:
https://re-motion.atlassian.net/browse/RM-8579

### DIFF
--- a/Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
+++ b/Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
@@ -26,31 +26,41 @@ StyleUtility.CreateBorderSpans = function (selector)
 
 StyleUtility.AddBrowserSwitch = function ()
 {
-  var browser;
-  if (navigator.appVersion.indexOf ('Edge') !== -1)
+  function containsBrowser(brands, expectedBrand)
   {
-    browser = 'edge';
+    return brands.some(c => c.brand === expectedBrand);
   }
-  else if (navigator.appVersion.indexOf ('Chrome') !== -1)
+
+  let browser;
+  if (navigator.userAgentData)
   {
-    browser = 'blink';
+    const brands = navigator.userAgentData.brands;
+    if (containsBrowser(brands, 'Microsoft Edge'))
+      browser = 'blink';
+    else if (containsBrowser(brands, 'Google Chrome'))
+      browser = 'blink';
+    else if (containsBrowser(brands, 'Opera'))
+      browser = 'blink';
   }
-  else if (navigator.appVersion.indexOf ('OPR') !== -1)
+
+  // Fallback to the old api navigator.userAgent to prevent incorrect detection in future
+  // if a browser adds support for userAgentData
+  if(!browser)
   {
-    browser = 'blink';
+    if (navigator.userAgent.indexOf('Edg/') !== -1)
+      browser = 'blink';
+    else if (navigator.userAgent.indexOf('Chrome') !== -1)
+      browser = 'blink';
+    else if (navigator.userAgent.indexOf('OPR') !== -1)
+      browser = 'blink';
+    else if (navigator.userAgent.indexOf('WebKit') !== -1)
+      browser = 'webkit';
+    else if (navigator.userAgent.indexOf('Gecko') !== -1)
+      browser = 'mozilla';
+    else
+      browser = 'browserUnknown';
   }
-  else if (navigator.appVersion.indexOf('WebKit') !== -1)
-  {
-    browser = 'webkit';
-  }
-  else if (navigator.appName === 'Netscape')
-  {
-    browser = 'mozilla';
-  }
-  else
-  {
-    browser = 'browserUnknown';
-  }
+
   StyleUtility.AddPlatformSwitch();
 
   if (!document.body.classList.contains (browser))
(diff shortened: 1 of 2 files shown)


Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -27,23 +27,23 @@ StyleUtility.CreateBorderSpans = function (selector)
 StyleUtility.AddBrowserSwitch = function ()
 {
   var browser;
-  if (navigator.appVersion.indexOf ('Edge') !== -1)
+  if ((navigator.userAgent.indexOf ('Edge') !== -1) || (navigator.userAgent.indexOf ('Edg') !== -1))
   {
     browser = 'edge';
   }
```
a-ctor, line 33: After talking with mike we concluded to drop support for the legacy Edge. The new edge is then mapped to `blink` (Chromium engine).

```suggestion
  if (navigator.userAgent.indexOf ('Edg/') !== -1)
  {
    browser = 'blink';
  }
```

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -27,23 +27,23 @@ StyleUtility.CreateBorderSpans = function (selector)
 StyleUtility.AddBrowserSwitch = function ()
 {
```
a-ctor, line 28: The ticket mentioned an API that can be used to do the detection, but it had the wrong name. We found the API again and we should use it first if available. If not we fall back to the user agent detection.

The API is https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData and it works in Chrome, Edge, and Opera. We need to add logic for the Browser detection and the Platform detection (below).

a-ctor (changes requested):
Changes apply to both versions of StyleUtility

a-ctor (approved):

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -26,31 +26,31 @@ StyleUtility.CreateBorderSpans = function (selector)
...
   {
-    browser = 'mozilla';
+    const brands = navigator.userAgentData.brands;
+    if (searchBrowser(brands, 'Microsoft Edge') || searchBrowser(brands, 'Google Chrome') || searchBrowser(brands, 'Opera'))
+      browser = 'blink';
   }
-  else
+
+  if(!browser)
   {
```
MichaelKetting, line 43: Semantics of the feature detection: as far as I understand the situation, it's either the feature is avaible or it is not. the feature detection should therefor do an if/else based on the availability of the API, not based on whether a match was found. The fallback is only needed when the API is not available.

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -26,31 +26,31 @@ StyleUtility.CreateBorderSpans = function (selector)
...
-  }
-  else if (navigator.appName === 'Netscape')
+
+  let browser;
+  if (navigator.userAgentData)
   {
-    browser = 'mozilla';
+    const brands = navigator.userAgentData.brands;
+    if (searchBrowser(brands, 'Microsoft Edge') || searchBrowser(brands, 'Google Chrome') || searchBrowser(brands, 'Opera'))
+      browser = 'blink';
```
MichaelKetting, line 39: readability:
I would prefer separate if/elseif here because it makes it easier to follow (shorter lines). if you additionally want to prevent the duplication of the string-value, you can define a `const var blink='blink'` and use that during the assignment.

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -26,31 +26,31 @@ StyleUtility.CreateBorderSpans = function (selector)
...
   }
-  else
+
+  if(!browser)
   {
-    browser = 'browserUnknown';
+    if ((navigator.userAgent.indexOf('Edg/') !== -1) || (navigator.userAgent.indexOf('Chrome') !== -1) || (navigator.userAgent.indexOf('OPR') !== -1))
```
MichaelKetting, line 44: same as with line 38, please split the IF.

thought: you could declare all the browser values as const-vars if you like.

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -59,15 +59,28 @@ StyleUtility.AddBrowserSwitch = function ()
...
+    if (navigator.userAgentData.platform === 'Windows')
+      platform = 'win';
+    else if (navigator.userAgentData.platform === 'macOS')
+      platform = 'mac';
+    else if (navigator.userAgentData.platform === 'Linux')
+      platform = 'x11';
+  }
+
+  if(!platform)
+  {
```
MichaelKetting, line 74: same as with the other API, please do an IF/ELSE here

MichaelKetting (changes requested):

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
a-ctor: Yes, but if a browser adds the feature we won't detect them correctly and mistakenly identify it as `browserUnknown`. To make sure that does not happen we need to use the user agent detection as fallback.

MichaelKetting: oh, I get it. Okay, that's true. @KuJo7 sorry, I didn't realize this detail during the review. Maybe we should add a code comment for the intent to future-proof this?

Remotion/Web/Core/Res/Themes/NovaGray/HTML/StyleUtility.js
```diff
@@ -26,31 +26,41 @@ StyleUtility.CreateBorderSpans = function (selector)
...
+      browser = 'blink';
+    else if (containsBrowser(brands, 'Google Chrome'))
+      browser = 'blink';
+    else if (containsBrowser(brands, 'Opera'))
+      browser = 'blink';
   }
-  else
+
+  // Fallback to the old api navigator.userAgent to prevent incorrect detection in future
+  // if a browser add support for userAgentData
```
MichaelKetting, line 47:
```suggestion
  // if a browser adds support for userAgentData
```

Remotion/Web/Core/Res/Themes/NovaViso/HTML/StyleUtility.js
```diff
@@ -24,31 +24,41 @@ StyleUtility.CreateBorderSpans = function (selector)
...
+      browser = 'blink';
+    else if (containsBrowser(brands, 'Google Chrome'))
+      browser = 'blink';
+    else if (containsBrowser(brands, 'Opera'))
+      browser = 'blink';
   }
-  else
+
+  // Fallback to the old api navigator.userAgent to prevent incorrect detection in future
+  // if a browser add support for userAgentData
```
MichaelKetting, line 45:
```suggestion
  // if a browser adds support for userAgentData
```

MichaelKetting (approved):
Typo-fix

PR #540, opened 2022-11-30
RM-8656 Fix autocomplete not working correctly when editing boclist

Frazerus:
https://re-motion.atlassian.net/browse/RM-8656

The problem was that boclist.css added `display: flex !important;` to the results list, therefore the list was already visible when `applyPositionToPopup` was called. This lead to the `contentWidth` Attribute not being set on the element, leading to null values being calculated, which obviously cannot be applied as the proper position values.


a-ctor (approved):
I find it a bit strange that the BocList css sets the display value of its children but this is something that can't be undone now.

a-ctor:
> > I find it a bit strange that the BocList css sets the display value of its children but this is something that can't be undone now.
> > @a-ctor : Which css rule is the offending party?

```css
div.bocList table.bocListTable td.bocListDataCell > a:not([hidden]),
div.bocList table.bocListTable td.bocListDataCell > span:not([hidden]),
div.bocList table.bocListTable td.bocListDataCell > div:not([hidden])
{
  display: flex !important; /*overrides hardcoded enforce-width setting 'display:block' */
  align-items: center;
  box-sizing: border-box;
  width: 100%;
}
```

The problem is that the autocomplete list box is inserted besides directly under the `td` element because the boc list element is a `span` and cannot be inserted there.
 
image

MichaelKetting:
Thanks @Frazerus and @a-ctor for the analysis. I'm afraid we found something *fun*. Due to the BocList rendering SPAN instead of DIV, the injection is happening at the wrong place in the tree. Given all aspects of the problem, we should opt to fix the injection, not the effect. To do this, we could either implement the actual bugfix RM-1492 (which would be a breaking change right now and wouldn't want to backport this) or simply add a special case for the injection logic. I think the special case could be cheaper, but I did schedule the fix to be done as soon as we can a) accept the breaking change and b) have room in the planning for it (which could be in a day or two :) )

The plan for a non-breaking solution / backport to v3.10:
Update the injection logic to also match `span.bocListEditableCell > span.control`, next to `div`, etc and inject there. BocAutoCompleteReferenceValue.UI.ts lines 1175 and 1614. We also should fix the same issue in DropDownMenu.ts (also two places) to be completely safe. This can/should be done first on the v3.10 release-branch. That way, we provide the value to the reporter ASAP and can then focus on doing a clean solution on develop (see next point)

The plan when we implement RM-1492: BocList is changed to render DIV instead of SPAN and the associated CSS is updated. This could happen on devleop soonish, but will not be decided in the current sprint.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1611,7 +1611,9 @@ namespace Remotion.BocAutoCompleteReferenceValue
             this.element.style.position = 'fixed';
 
             LayoutUtility.Hide(this.element);
-            this.input.closest('div, td, th, body')!.appendChild(this.element);
+
+            this.input.closest('div, td, th, body, span.bocListEditableCell > span.control')!.appendChild(this.element);
+
 
             if (this.options.combobox.getAttribute('aria-labelledby') !== null) {
```
MichaelKetting, line 1618: Got some extra whitespace here. 
also since we we will now change the develop-version to 4.0, this non-breaking change will get rebased onto v3.10-branch (new branch name with -v3.10 suffix) and we do RM-1492 instead for develop, which will block RM-8656.

MichaelKetting (changes requested):

Frazerus:
Replaced with merge into `release-v3.10`

PR #541, opened 2022-11-30
RM-8666 BocAutoCompleteReference popup is more aligned with input field

Frazerus:
https://re-motion.atlassian.net/browse/RM-8666

The implementation creates the pictures as seen in the expected results of the jira issue.


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1809,11 +1809,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
             maxHeight = '';
         }
 
-        const availableWidth = position.left + LayoutUtility.GetOuterWidth(reference);
+        const bufferFromLeft = 30;
+        const availableWidth = position.left + LayoutUtility.GetOuterWidth(reference) - bufferFromLeft;
```
a-ctor, line 1813: If the input element is less than 30px away from the left side we still use the 30px as spacing. We should only reduce the left space if there is any. 
```suggestion
        const availableWidth = position.left + LayoutUtility.GetOuterWidth(reference) - Math.min(bufferFromLeft, position.left);
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1809,11 +1809,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
         const minWidth = LayoutUtility.GetOuterWidth(reference);
```
a-ctor, line 1814: Because we are changing the related code it would be good if we extract `LayoutUtility.GetOuterWidth(reference)` into a variable so that we do not call it three times in a row. The method is comparatively slow since it changes CSS properties of the element to calculate the width.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1809,11 +1809,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
         const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
-        const requiredWidth = contentWidth + 30;
-        const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+        const requiredWidth = contentWidth + 1; //to make sure that the inaccurate content width calculation is accurate enough
+        const maxAllowedWidth = Math.min(requiredWidth, maxWidth)
+        const elementWidth = Math.max(minWidth, maxAllowedWidth)
```
a-ctor, line 1818: If you want to split up the clamp here we need another name for `maxAllowedWidth` since that name does not fit the actual value.

a-ctor (changes requested):

a-ctor (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1809,11 +1809,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
-        const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
-        const requiredWidth = contentWidth + 30;
-        const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+        const referenceWidth = LayoutUtility.GetOuterWidth(reference);
+        const bufferFromLeft = 30;
+        const availableWidth = position.left + referenceWidth - bufferFromLeft;
+        const maxWidth = Math.min (isNaN (options.maxWidth) ? referenceWidth : options.maxWidth, availableWidth);
+        const requiredWidth = contentWidth + 1; //to make sure that the inaccurate content width calculation is accurate enough
+        const maxAllowedWidth = Math.min(requiredWidth, maxWidth)
+        const elementWidth = Math.max(referenceWidth, maxAllowedWidth)
```
MichaelKetting, line 1818: Okay, while testing this, I may have found the reason for the original `+ 30`. we may not actually have cared about having a left-margin. Instead, the 30px where meant to ensure whitespace for the scrollbar. Now, if you type `a`, you will see that there's a horizontal scrollbar instead of the popup opening up further to the left.

What would be needed to incorporate correct scollbar capacity? Basically, if the popup using overflow:hidden, no space is reserved and if overflow:auto, we need to reserve some space for the scrollbar. 20px is probably enough, should be named scrollbarWidth or some such (no more magic numbers).

```
.ac_informationPopUp div
{
  /* overflow is not specificed, we can change this to explicitly 'visible' if need be. */
  /* overflow: visible; */
}
.ac_results div
{
  overflow: auto;
}

```

With this new discovery, we can decide to ignore the +30 in line 1813. I don't think the margin was the original intent, just a misinterpretation of the code. Of course, it could still look okay with the new semantics.

MichaelKetting (changes requested):
Need to properly handle the result-popup since the code is used for both types of pop-up.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: We might be able to use https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter during measuring to include the scrollbar gutter in the size calculation

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1809,11 +1809,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
-        const availableWidth = position.left + LayoutUtility.GetOuterWidth(reference);
-        const minWidth = LayoutUtility.GetOuterWidth(reference);
-        const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
-        const requiredWidth = contentWidth + 30;
-        const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+        const referenceWidth = LayoutUtility.GetOuterWidth(reference);
+        const bufferFromLeft = 30;
+        const availableWidth = position.left + referenceWidth - bufferFromLeft;
+        const maxWidth = Math.min (isNaN (options.maxWidth) ? referenceWidth : options.maxWidth, availableWidth);
+        const requiredWidth = contentWidth + 1; //to make sure that the inaccurate content width calculation is accurate enough
```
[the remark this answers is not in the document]
Frazerus: I seems this was somehow related to something else in the code. I cannot reproduce the issue I found earlier when not adding the 1. Earlier this meant that if the required width becomes the element width, the content is too long for the field and a second row is created even though there is enough horizontal space.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1809,11 +1811,12 @@ namespace Remotion.BocAutoCompleteReferenceValue
             maxHeight = '';
         }
 
-        const availableWidth = position.left + LayoutUtility.GetOuterWidth(reference);
-        const minWidth = LayoutUtility.GetOuterWidth(reference);
-        const maxWidth = Math.min (isNaN (options.maxWidth) ? LayoutUtility.GetOuterWidth(reference) : options.maxWidth, availableWidth);
-        const requiredWidth = contentWidth + 30;
-        const elementWidth = Math.max (Math.min (requiredWidth, maxWidth), minWidth);
+        const referenceWidth = LayoutUtility.GetOuterWidth(reference);
```
MichaelKetting, line 1814: Something's still fishy:
image

It does work when the text is generally shorter than the input field
image

Maybe the padding isn't correct?

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #542, opened 2022-12-01
RM-8672 Remove Javascript Error when opening a DropDownMenu with Keyboard

Frazerus:
https://re-motion.atlassian.net/browse/RM-8672

Not sure whether that new method belongs where it is now, but it seems to only be needed in the dropdown, so it might stay there for now.

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
@@ -28,6 +28,13 @@ interface DropDownMenuPositioningDetails
   offsetY: number;
 }
 
+interface IKeyboardEventLike
+{
+  keyCode: number;
+  preventDefault(): void;
+  stopPropagation(): void;
+}
+
 type DropDownMenu_MenuInfo_LoadMenuItemsSuccessHandler = (itemInfos: DropDownMenu_ItemInfo[]) => void;
 type DropDownMenu_MenuInfo_LoadMenuItemsErrorHandler = () => void;
 type DropDownMenu_MenuInfo_LoadMenuItemsCallback = (successHandler: DropDownMenu_MenuInfo_LoadMenuItemsSuccessHandler, errorHandler: DropDownMenu_MenuInfo_LoadMenuItemsErrorHandler) => void;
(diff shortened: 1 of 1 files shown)


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -998,6 +998,12 @@ class DropDownMenu
     }
   }
 
+  private static SetEventKeyCode(event: KeyboardEvent, keyCode: number) {
+    Object.defineProperties(event, {
+      keyCode: {value: keyCode}
+    });
+  }
```
a-ctor, line 1005: I dislike that we are overriding an existing property of a browser object. While it seems to work fine in the relevant browsers there is no guarantee that it stays that way. Therefore, I would refrain from changing `Event`.

Instead I suggest creating a new object instead of changing the existing one. Since we only call `Options_OnKeyDown` with the changed event we can change the parameter type so that we can use structural typing for the properties that we need. We only need the three properties `keyCode`, `preventDefault`, and `stopPropagation`, which we can define in a interface/type. By using this new type as parameter type we can pass in an `Event` or custom object with those three properties.

Kind of difficult to explain, maybe we get together to change the code :)

a-ctor (changes requested):

a-ctor (approved):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -998,6 +1006,14 @@ class DropDownMenu
...
   }
 
+  private static CreateKeyboardEventLikeWithKeyCode(event: IKeyboardEventLike, keyCode: number): IKeyboardEventLike {
+    return {
+      keyCode: keyCode,
+      preventDefault: () => event.preventDefault(),
+      stopPropagation: () => event.stopPropagation()
+    };
+  }
+
```
MichaelKetting, line 1016: New-line format with `{` needs fixing

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -857,49 +864,50 @@ class DropDownMenu
...
-        event.preventDefault();
-        event.stopPropagation();
+        keyboardEvent.preventDefault();
+        keyboardEvent.stopPropagation();
         if (dropDownMenu !== DropDownMenu._currentMenu)
         {
           DropDownMenu.OnClick (dropDownMenu, dropDownMenu.id, getSelectionCount, null);
-          (event.keyCode as number) = 40; // always act as if the down-arrow was used when opening the drop down menu.
-          DropDownMenu.Options_OnKeyDown (event, dropDownMenu);
+          keyboardEvent = this.CreateKeyboardEventLikeWithKeyCode(keyboardEvent, 40);
```
MichaelKetting, line 880: add comment that 40 = down arrow

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -857,49 +864,50 @@ class DropDownMenu
...
       case 40: // down arrow
-        event.preventDefault();
-        event.stopPropagation();
+        keyboardEvent.preventDefault();
+        keyboardEvent.stopPropagation();
         if (dropDownMenu !== DropDownMenu._currentMenu)
         {
           DropDownMenu.OnClick (dropDownMenu, dropDownMenu.id, getSelectionCount, null);
-          (event.keyCode as number) = 40; // always act as if the down-arrow was used when opening the drop down menu.
+          keyboardEvent = this.CreateKeyboardEventLikeWithKeyCode(keyboardEvent, 40);
```
MichaelKetting, line 900: add comment that 40 = down arrow

MichaelKetting (changes requested):
Some small comment/format issues

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -998,6 +1006,15 @@ class DropDownMenu
     }
   }
 
+  private static CreateKeyboardEventLikeWithKeyCode(event: IKeyboardEventLike, keyCode: number): IKeyboardEventLike
+   {
+    return {
```
MichaelKetting, line 1011:
```suggestion
  private static CreateKeyboardEventLikeWithKeyCode(event: IKeyboardEventLike, keyCode: number): IKeyboardEventLike
  {
    return {
```

Look-see, there's been three spaces, now it's the correct two spaces to align with the function.

MichaelKetting (approved):

PR #546, opened 2022-12-14
RM-8679 Update CoreForms to v6.0.0-alpha.15

a-ctor:
https://re-motion.atlassian.net/browse/RM-8679

### DIFF
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Program.cs
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Program.cs
@@ -9,6 +9,8 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite
   {
     public static void Main (string[] args)
     {
+      LegacyAspNetInitialization.License = "I hereby confirm that I use CoreForms only for trial purposes and have read and accept the CoreForms Trial License.";
+
       var legacyAspNetInitializationOptions = new LegacyAspNetInitializationOptions("/", Environment.CurrentDirectory);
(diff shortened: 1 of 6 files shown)


Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Program.cs
```diff
@@ -9,6 +9,8 @@ public class Program
   {
     public static void Main (string[] args)
     {
+      LegacyAspNetInitialization.License = "I hereby confirm that I use CoreForms only as part of Software developed by RUBICON IT GmbH.";
```
MichaelKetting, line 12: Please switch to the trial-string. am not sure which would be best, but it feels like that's the way to go for non-prod strings

Remotion/Web/Test/Program.cs
```diff
@@ -9,7 +9,8 @@ public class Program
   {
     public static void Main (string[] args)
     {
-      Console.WriteLine(Environment.CurrentDirectory);
+      LegacyAspNetInitialization.License = "I hereby confirm that I use CoreForms only as part of Software developed by RUBICON IT GmbH.";
```
MichaelKetting, line 12: see other comment

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Program.cs
MichaelKetting, line 12: We only have two .net applications right now? (ust asking stupid questions)

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #550, opened 2022-12-21
RM-8690 Do not append resource prefix in Command.GetCommandInfoForHrefCommand if url is only parameters

a-ctor:
https://re-motion.atlassian.net/browse/RM-8690

### DIFF
--- a/Remotion/Web/UnitTests/Core/UI/Controls/CommandTests/HrefCommandTest.cs
+++ b/Remotion/Web/UnitTests/Core/UI/Controls/CommandTests/HrefCommandTest.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Specialized;
 using System.Web.UI;
 using System.Web.UI.WebControls;
+using Moq;
 using NUnit.Framework;
 using Remotion.Development.Web.UnitTesting.AspNetFramework;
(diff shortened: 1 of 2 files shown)


Remotion/Web/UnitTests/Core/UI/Controls/CommandTests/HrefCommandTest.cs
```diff
@@ -120,5 +121,31 @@ public void Render_WithAccessDenied ()
       Assert.AreEqual(HtmlTextWriterTag.A, _testHelper.HtmlWriter.Tag, "Wrong Tag");
       Assert.AreEqual(0, _testHelper.HtmlWriter.Attributes.Count, "Has Attributes");
     }
+
+    [Test]
+    public void GetCommandInfo_WithEmptyHrefAndParameters_DoesNotResolveClientUrl ()
```
MichaelKetting, line 126: I think we should update the comment message. true, we found it via the resoruce prefix, but the underlying issue is not related. Rather, something along the lines of "href with only parameters and no location is not resolved correctly"

MichaelKetting (approved):
please update commit message.

MichaelKetting:
Please check when this was introduced and update ticket.

PR #551, opened 2022-12-23
RM-8003 Split up web development webtesting testsite project

a-ctor:
https://re-motion.atlassian.net/browse/RM-8003


Remotion/Web/Development.WebTesting.TestSite.Shared/TabbedMultiViewTest.aspx
```diff
@@ -28,7 +29,7 @@
       <remotion:TabView ID="Tab2" Title="Tab2Title" runat="server">
         <span>Content2</span>
       </remotion:TabView>
-      <remotion:TabView runat="server" ID="Tab3" Title="Tab3Title" Icon-Url="~/Images/SampleIcon.gif" runat="server">
+      <remotion:TabView runat="server" ID="Tab3" Title="Tab3Title" Icon-Url="~/Image/SampleIcon.gif">
```
MichaelKetting, line 32: you dropped runat-server

MichaelKetting (commented):
partial review

Remotion/Web/Development.WebTesting.TestSite.Shared/ScreenshotTest.aspx
```diff
@@ -112,7 +113,7 @@
           </td>
         </tr>
       </table>
-      <iframe id="frame" src="ScreenshotTestFrame.aspx" >
+      <iframe id="frame" src="<%= ResolveUrl(".") %>/ScreenshotTestFrame.aspx" >
```
MichaelKetting, line 116:
```suggestion
      <iframe id="frame" src="<%= ResolveUrl("~/ScreenshotTestFrame.aspx") %>" >
```
?

Remotion/Web/Development.WebTesting.TestSite.Shared/WebButtonTest.aspx
```diff
@@ -29,16 +30,16 @@
...
+        <remotion:WebButton ID="MyWebButtonSupplemental3Href" Text="HrefButton" ButtonType="Supplemental" PostBackUrl="~/WebButtonTest.wxe" runat="server" />
       </div>
       <h3>Disabled WebButton</h3>
       <remotion:WebButton ID="MyDisabledWebButton" Text="DisabledWebButton" CommandName="Disabled" Enabled="False" runat="server"/>
       <remotion:WebButton ID="MyDisabledWebButtonPrimary" Text="DisabledWebButton" ButtonType="Primary" CommandName="Disabled" Enabled="False" runat="server"/>
       <remotion:WebButton ID="MyDisabledWebButtonSupplemental" Text="DisabledWebButton" ButtonType="Supplemental" CommandName="Disabled" Enabled="False" runat="server"/>
       <h3>WebButton with icon</h3>
-      <remotion:WebButton ID="MyWebButtonWithIcon" Text="This text should wrap correctly because it is too long to fit" Width="100px" Icon-Url="Images/SampleIcon.gif" CommandName="Disabled" runat="server"/>
+      <remotion:WebButton ID="MyWebButtonWithIcon" Text="This text should wrap correctly because it is too long to fit" Width="100px" CommandName="Disabled" runat="server"/>
```
MichaelKetting, line 42: `<%= ResolveUrl("~/Icon/SampleIcon.gif") %>`
or rather the new syntax then.

Remotion/Web/Development.WebTesting.TestSite.Shared/MultiWindowTest/Main.aspx
```diff
@@ -41,7 +42,7 @@
             <p><remotion:WebButton ID="LoadWindowFunctionInFrame" Text="Load WindowFunction in Frame" runat="server"/></p>
             <p><remotion:WebButton ID="LoadMainAutoRefreshingFrameFunctionInFrame" Text="Load Main-auto-refreshing FrameFunction in Frame" runat="server"/></p>
             <p><remotion:WebButton ID="LoadWindowFunctionInNewWindow" Text="Load WindowFunction in new Window" runat="server"/></p>
-            <p><asp:HyperLink ID="NavigateAway" Text="Navigate Away" NavigateUrl="~/Default.aspx" runat="server"/></p>
+            <p><asp:HyperLink ID="NavigateAway" Text="Navigate Away" runat="server"/></p>
```
MichaelKetting, line 45: via `<%=... %>` instead?
It's more a question of style/consistency then anything else.

Remotion/Web/Development.WebTesting.TestSite.Shared/MultiWindowTest/Main.aspx.cs
```diff
@@ -33,6 +35,11 @@ protected override void OnInit (EventArgs e)
       LoadWindowFunctionInFrame.Click += LoadWindowFunctionInFrameOnClick;
       LoadMainAutoRefreshingFrameFunctionInFrame.Click += LoadMainAutoRefreshingFrameFunctionInFrameOnClick;
       LoadWindowFunctionInNewWindow.Click += LoadWindowFunctionInNewWindowOnClick;
+
+      var defaultAspxResourceUrl = SafeServiceLocator.Current.GetInstance<IResourceUrlFactory>()
+          .CreateResourceUrl(typeof(MainFunction), TestResourceType.Root, "Default.aspx")
+          .GetUrl();
+      NavigateAway.NavigateUrl = defaultAspxResourceUrl;
```
MichaelKetting, line 42: Depending on the outcome of the change in Main.aspx.

Remotion/Web/Development.WebTesting.TestSite.Shared/GenericTest.aspx.cs
```diff
@@ -18,19 +18,27 @@
...
     protected override Dictionary<string, GenericTestPageParameter> Parameters { get; } = new Dictionary<string, GenericTestPageParameter>();
 
+    public string FrameTestFrameUrl { get; }
+
     public GenericTest ()
     {
+      FrameTestFrameUrl = SafeServiceLocator.Current.GetInstance<IResourceUrlFactory>()
+          .CreateResourceUrl(typeof(GenericTestFunction), TestResourceType.Root, "FrameTestFrame.aspx")
+          .GetUrl();
+
```
MichaelKetting, line 41: Given the way ASP.net works, it would be cleaner to make this lazy (and skip caching) and just have the property return the URL. Or we get a helper function that takes the file and trasforms it.
```
public static class Resolve
{
 string Root(string file);
 string Image(string file);
 string Html(string file);
}
```
(just an idea for a simple syntax)

Remotion/Web/Development.WebTesting.TestSite.Shared/GenericTest.aspx.cs
```diff
@@ -18,19 +18,27 @@
...
   public partial class GenericTest : GenericTestPageBase<GenericTestOptions>
   {
     protected override Dictionary<string, GenericTestPageParameter> Parameters { get; } = new Dictionary<string, GenericTestPageParameter>();
 
+    public string FrameTestFrameUrl { get; }
+
     public GenericTest ()
     {
+      FrameTestFrameUrl = SafeServiceLocator.Current.GetInstance<IResourceUrlFactory>()
+          .CreateResourceUrl(typeof(GenericTestFunction), TestResourceType.Root, "FrameTestFrame.aspx")
```
MichaelKetting, line 39: `TestResourceType.Root` => that's the one you need if you want to keep the download-files in the root of the res-folder for this assembly.

MichaelKetting (changes requested):
Basically, we only need to do the resource resolving, the rest looks good. I have a couple of comments on that and one or two design ideas. Am always open to better ones :)

Remotion/Web/Development.WebTesting.TestSite.Shared/TabbedMultiViewTest.aspx
a-ctor: It was a duplicate anyway :D

Remotion/Web/Development.WebTesting.TestSite.Shared/MultiWindowTest/Main.aspx
a-ctor: > Server tags cannot contain <% ... %> constructs.

Remotion/Web/Development.WebTesting.TestSite.Shared/ScreenshotTest.aspx.cs
```diff
@@ -18,15 +18,15 @@
...
+namespace Remotion.Web.Development.WebTesting.TestSite.Shared
 {
   public partial class ScreenshotTest : WxePage
   {
     protected override void OnInit (EventArgs e)
     {
       base.OnInit(e);
 
-      var node = new WebTreeNode("ItemA", WebString.CreateFromText("A"), "~/Images/SampleIcon.gif") { IsEvaluated = true, IsExpanded = true };
+      var node = new WebTreeNode("ItemA", WebString.CreateFromText("A"), "~/Image/SampleIcon.gif") { IsEvaluated = true, IsExpanded = true };
```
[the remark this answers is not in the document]
a-ctor: The problem with moving those resources is that its actually more complex than that. There are two main problems:

1. Most of the time it is not allowed to set control properties using `<%= %>`. Thus it is necessary to add code to the c# part, which makes splits up the property values and makes the ASPX harder to reader
2. I don't know how to retrieve the physical path from a mapped `res/.../...` path. `Context.Server.MapPath` does not work

I can move all the resources to the shared folder but it probably takes a few hours to move everything and check that stuff is working correctly. So my question is how important it is to us that we don't duplicate the resources. There are already examples where we have duplications, like the `objects` folder in ObjectBinding.WebTesting.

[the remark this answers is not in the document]
a-ctor: Sorry I mixed two different problems. Both are causes by the moving of the resources into the shared project:

1. `<%= %>` can't be used to set properties, which makes it necessary to use the codebehind to add the `CreateResourceUrl` code. 
2. Downloadables (zip file, text file, etc.) for the download tests require the full path of the file, which is now hidden behind a `res/...` virtual path

I question if it is worth moving the five files to the shared folder compared to the required work needed and benefit gained. Yes it is bit cleaner, but the readability takes a hit and if I am not mistaken we won't need the shared setup in the soon future? If we still want to do the clean separation I need a solution for the physical path as I have no clue how to get it.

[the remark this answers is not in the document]
MichaelKetting: ad 1) / question: readability: I'm not sure its actually less readable than now if we move all the construction of the WebTab/WebMenuItem to code behind. Markup is just more convenient, but we don't have typesafety, etc. It just wasn't necesarry to do it another way. Looking back, code behind would be my preferred way to do it from the start.

ad 1) / worth it: it feels like it's about 2h of work top. And and given all the retrofitting, am not sure we'll actually re-merge it because that would be extra effort. And it will exercise the primary usage scenario, which is code behind. If we do it cleanly, yes, it's worth it.

ad 2) okay, ok. Hmm..how about we just turn those into embedded resources and write the binary to the output via response.OutputStream or response.BinaryWrite(...)? This leaves just the image file.

ad 1a) oh, and I got another idea, if you really want to avoid the rewrite: try a dataurl. If you switch the image to the NullIcon.gif (and run it through an optimizer first), you get a nice, relatively short string: `data:image/gif;base64,R0lGODlhEAAQAPABAAAA/wAAACH5BAUAAAEALAAAAAAQABAAAAIdRIynGe0PgYG0Wiev3nl72n3iMY1faHLlubQJUAAAOw==` and the screenshots only care about it being 16x16 if I'm not mistaken. 

=> I think with 1a) + 2) we get the best of both worlds and a nice cheap solution?

Remotion/Web/Development.WebTesting.TestSite.Shared/WebButtonTest.aspx.cs
```diff
@@ -16,16 +16,20 @@
...
-namespace Remotion.Web.Development.WebTesting.TestSite
+namespace Remotion.Web.Development.WebTesting.TestSite.Shared
 {
   public partial class WebButtonTest : WxePage
   {
     protected override void OnInit (EventArgs e)
     {
       base.OnInit(e);
 
+      MyWebButtonWithIcon.Icon.Url = ResolveUrl(this.ResolveImageResource("SampleIcon.gif"));
```
MichaelKetting, line 31:
```suggestion
      MyWebButtonWithIcon.Icon.Url = this.ResolveImageResource("SampleIcon.gif");
```

Shouldn't this be enough?

Remotion/Web/Development.WebTesting.TestSite.Shared/TreeViewTest.aspx.cs
```diff
@@ -79,7 +79,7 @@ private void InitializeTreeView ()
 
     private void InitializeMyTreeView ()
     {
-      var root = new TreeNode("Root node", "RootValue", "~/Images/SampleIcon.gif");
+      var root = new TreeNode("Root node", "RootValue", this.ResolveImageResource("SampleIcon.gif"));
```
MichaelKetting, line 82: Would this also work with "Image/SampleIcon.gif"?
Not sure we should switch this, since we are in code behind and "Image" is a magic value that's normally resolved via the ResourceType. So, it's more a question of interest than a request for change :)

Unless we could completly revert the ResolveImageResource usage and drop the helper API again?

Remotion/Web/Development.WebTesting.TestSite.Shared/TreeViewTest.aspx.cs
```diff
@@ -118,13 +118,13 @@ private void InitializeMyTreeViewWithSpecialChildren ()
...
     {
-      var root = new TreeNode("With'SingleQuote", "RootValue", "~/Images/SampleIcon.gif");
+      var root = new TreeNode("With'SingleQuote", "RootValue", this.ResolveImageResource("SampleIcon.gif"));
       TreeViewWithOnlyRootWithSingleQuote.Nodes.Add(root);
     }
 
     private void InitializeTreeViewWithOnlyRootWithDoubleQuote ()
     {
-      var root = new TreeNode("With'SingleQuoteAndDouble\"Quote", "~/Images/SampleIcon.gif");
+      var root = new TreeNode("With'SingleQuoteAndDouble\"Quote", "RootValue", this.ResolveImageResource("SampleIcon.gif"));
```
MichaelKetting, line 127: "RootValue" here seems to be a copy&paste error?

Remotion/Web/Development.WebTesting.TestSite.Shared/ResourceResolvingPageExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+using Remotion.Web.Development.WebTesting.TestSite.Infrastructure;
+
+namespace Remotion.Web.Development.WebTesting.TestSite.Shared
+{
+  public static class ResourceResolvingPageExtensions
+  {
+    public static string ResolveRootResource (this IHttpHandler pageLike, string imageName)
+    {
+      return SafeServiceLocator.Current.GetInstance<IResourceUrlFactory>()
+          .CreateResourceUrl(GetOriginalPageType(pageLike.GetType()), TestResourceType.Root, imageName)
```
MichaelKetting, line 31: You could simplify this to `typeof(ResourceResolvingPageExtensions)` instead of GetOriginalPageType(pageLike.GetType()), Is there a deeper need to do it via discovery of the actual used type? We should(could?) make the Extensions-type internal and resolve any ambiguities via that.

Remotion/Web/Development.WebTesting.TestSite.Shared/ResourceResolvingPageExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+          .GetUrl();
+    }
+
+    public static string ResolveImageResource (this IHttpHandler pageLike, string imageName)
+    {
+      return SafeServiceLocator.Current.GetInstance<IResourceUrlFactory>()
+          .CreateResourceUrl(GetOriginalPageType(pageLike.GetType()), ResourceType.Image, imageName)
```
MichaelKetting, line 38: Same as line 31

Remotion/Web/Development.WebTesting.TestSite.Shared/ResourceResolvingPageExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Web;
+using System.Web.UI;
+using Remotion.ServiceLocation;
+using Remotion.Web.Development.WebTesting.TestSite.Infrastructure;
+
+namespace Remotion.Web.Development.WebTesting.TestSite.Shared
+{
+  public static class ResourceResolvingPageExtensions
+  {
+    public static string ResolveRootResource (this IHttpHandler pageLike, string imageName)
```
MichaelKetting, line 28: Is the parameter name incorrect? `imageName` -> `fileName`

Remotion/Web/Development.WebTesting.TestSite.Shared/ImageTest.aspx
```diff
@@ -14,13 +14,14 @@
  % You should have received a copy of the GNU Lesser General Public License
  % along with re-motion; if not, see http://www.gnu.org/licenses.
 --%>
-<%@ Page Language="C#" MasterPageFile="~/Layout.Master" AutoEventWireup="true" CodeBehind="ImageTest.aspx.cs" Inherits="Remotion.Web.Development.WebTesting.TestSite.ImageTest" %>
+<%@ Page Language="C#" MasterPageFile="Layout.Master" AutoEventWireup="true" CodeBehind="ImageTest.aspx.cs" Inherits="Remotion.Web.Development.WebTesting.TestSite.Shared.ImageTest" %>
+<%@ Import Namespace="Remotion.Web.Development.WebTesting.TestSite.Shared" %>
 <asp:Content ContentPlaceHolderID="body" runat="server">
   <h3>Image</h3>
-  <img id="MyImage" src="~/Images/SampleIcon.gif" alt="My alternative text" runat="server"/>
+  <img id="MyImage" src="data:image/gif;base64,R0lGODlhEAAQAPH/AAAAAAAAgMDAwP///yH5BAUAAAQALAAAAAAQABAAAANESLrcriBKEoIDImtQG8ZaVkmR8IXCOAznmXEBwJrzF1ExK68fu+SrIIvH2AmFgKJOJ/EtjEfic0mSJafR4BVS7T6+jQQAOw==" alt="My alternative text" runat="server"/>
```
MichaelKetting, line 21: I don't think those images actually need to runat-server? Would simply switch this to <%=ResolveUrl / Resource > (depending on what works best.

Remotion/Web/Development.WebTesting.TestSite.Shared/FileDownloadTest.aspx.cs
```diff
@@ -32,7 +34,7 @@ protected override void OnInit (EventArgs e)
     private void DownloadButtonOnClick (object sender, EventArgs eventArgs)
     {
       const string file = "SampleFile_06d6ff4d-c124-4d3f-9d96-5e4f2d0c7b0c.txt";
-      var fullFilePath = Server.MapPath("~/SampleFile.txt");
+      var fullFilePath = ((ResourceVirtualFile)HostingEnvironment.VirtualPathProvider.GetFile(this.ResolveRootResource("SampleFile.txt"))).PhysicalPath;
```
MichaelKetting, line 37: Okay, so you found the solution to not needing an embedded resource.

MichaelKetting (changes requested):
Some questions and one or two fixes.

Remotion/Web/Development.WebTesting.TestSite.Shared/TreeViewTest.aspx.cs
a-ctor: Actually, the code was broken before as image url was passed in as value so I inserted a new value based on the tree view above.

Remotion/Web/Development.WebTesting.TestSite.Shared/TreeViewTest.aspx.cs
a-ctor: Yes would also work without. We still have a lot of usages of the helper API so I would not drop it again. The repeating the `IResourceUrlFactory` is 🤢

Remotion/Web/Development.WebTesting.TestSite.Shared/ResourceResolvingPageExtensions.cs
MichaelKetting: Hmm...you mean via copy&paste to other projects? Or rather, to libraries / different use cases? Right now, I don't see a scenario where the more simplistic approach would be a problem, though I could be missing something here?

Remotion/Web/Development.WebTesting.TestSite.Shared/TreeViewTest.aspx.cs
MichaelKetting: Yeah, it's only a good idea if we can get rid of all the usages. At least those for image, though the benefit of that is already pretty small, so, okay, keep it.

MichaelKetting (commented):
Just finishing the discussion on ResolveResource

MichaelKetting (approved):

PR #560, opened 2023-01-03
RM-8686 Add localization properties to the DatePicker

a-ctor:
https://re-motion.atlassian.net/browse/RM-8686

Todo;

- [x] localization

[RM-8686]: https://re-motion.atlassian.net/browse/RM-8686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Remotion/Web/Core/UI/Controls/DatePickerPage.cs
+++ b/Remotion/Web/Core/UI/Controls/DatePickerPage.cs
@@ -19,12 +19,14 @@ using System.Web;
 using System.Web.UI;
 using System.Web.UI.HtmlControls;
 using System.Web.UI.WebControls;
+using Remotion.Globalization;
 using Remotion.Reflection;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
 using Remotion.Web.Infrastructure;
 using Remotion.Web.UI.Controls.DatePickerButtonImplementation.Rendering;
 using Remotion.Web.Utilities;
+using Calendar = System.Web.UI.WebControls.Calendar;
 
 namespace Remotion.Web.UI.Controls
 {
@@ -52,6 +54,26 @@ namespace Remotion.Web.UI.Controls
 /// </remarks>
 public class DatePickerPage : Page
 {
+  /// <summary> A list of control specific resources. </summary>
+  /// <remarks> 
+  ///   Resources will be accessed using 
+  ///   <see cref="M:Remotion.Globalization.IResourceManager.GetString(System.Enum)">IResourceManager.GetString(Enum)</see>. 
+  ///   See the documentation of <b>GetString</b> for further details.
+  /// </remarks>
+  [ResourceIdentifiers]
+  [MultiLingualResources("Remotion.Web.Globalization.DatePickerPage")]
+  public enum ResourceIdentifier
+  {
+    /// <summary>The tool tip text for the calendar popup.</summary>
+    TitleText,
+
+    /// <summary>The tool tip text for the select the previous month button.</summary>
+    PreviousMonthTitle,
+
+    /// <summary>The tool tip text for the select the next month button.</summary>
+    NextMonthTitle,
+  }
+
   public const string CultureParameterName = "Culture";
   public const string UICultureParameterName = "UICulture";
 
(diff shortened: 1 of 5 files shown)


Remotion/Web/Core/UI/Controls/DatePickerPage.cs
```diff
@@ -52,6 +54,34 @@ namespace Remotion.Web.UI.Controls
...
+  /// <remarks> 
+  ///   Resources will be accessed using 
+  ///   <see cref="M:Remotion.Globalization.IResourceManager.GetString(System.Enum)">IResourceManager.GetString(Enum)</see>. 
+  ///   See the documentation of <b>GetString</b> for further details.
+  /// </remarks>
+  [ResourceIdentifiers]
+  [MultiLingualResources("Remotion.Web.Globalization.DatePickerPage")]
+  public enum ResourceIdentifier
+  {
+    /// <summary> Todo </summary>
```
a-ctor, line 68: I realized that this had an open todo just as I clicked request review ^^

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/App.config
```diff
@@ -16,7 +16,11 @@
 
   <system.web>
     <globalization culture="de-AT" uiCulture="en-US"/>
-    <compilation debug="true" />
+    <compilation debug="true">
+      <assemblies>
+        <add assembly="System.Drawing.Primitives"/>
```
MichaelKetting, line 21: Is this really needed or an artifact from another branch?

Remotion/Web/Core/UI/Controls/DatePickerPage.cs
```diff
@@ -52,6 +54,34 @@ namespace Remotion.Web.UI.Controls
 /// </remarks>
 public class DatePickerPage : Page
 {
+#if !NETFRAMEWORK
```
MichaelKetting, line 57: Am thinking it would be easier to just make the resource value assignment conditional?

MichaelKetting (commented):
waiting for the resources. please re-request then :)

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/App.config
a-ctor: Seems like the date picker popup is broken in .NET because of this. I am going to create a bug ticket for it.

Remotion/Web/Core/UI/Controls/DatePickerPage.cs
```diff
@@ -108,6 +136,16 @@ override protected void OnInit (EventArgs e)
     //  Force the creation of the postback function
     Page.ClientScript.GetPostBackEventReference(this, "");
 
+#if !NETFRAMEWORK
+    var globalizationService = SafeServiceLocator.Current.GetInstance<IGlobalizationService>();
+    var resourceManager = globalizationService.GetResourceManager(typeof(ResourceIdentifier));
+    Calendar.TitleText = resourceManager.GetString(ResourceIdentifier.TitleText);
+    Calendar.SelectMonthTitle = resourceManager.GetString(ResourceIdentifier.SelectMonthTitle);
+    Calendar.SelectWeekTitle = resourceManager.GetString(ResourceIdentifier.SelectWeekTitle);
```
MichaelKetting, line 144: Are these two actually in use in re-motion? If not, we should drop the resource entries to remove the maintenance overhead.

You can test the DatePicker via
http://localhost:60500/res/Remotion.Web/Themes/NovaGray/UI/DatePickerForm.aspx?Culture=en-US&UICulture=de-AT

MichaelKetting (approved):
I have tested it and it looks like we may have overshot the target. (more resources than we actually need, see comment) Otherwise, we're done.

The missing System.Drawing.Primitives gets added to machine.config in a separate PR, correct?

Remotion/Web/Core/UI/Controls/DatePickerPage.cs
a-ctor: We don't have any usages of `SelectionMode` which would enable those two buttons but it would be possible to override the `DatePickerPage` and change that? I don't see the maintenance overhead tbh. We already have the localizations and the logic/localizations won't change

Yes primitives is going to be added in a new PR+issue

MichaelKetting: No, there's no override to enable this feature. Maintenance overhead here means that all strings are exported and get reviews and are re-evaluated for correctness or simply require translation if there's ever another language added. If we have completely unused/unneeded strings, all this adds to the total cost of ownership down the road.

PR #571, opened 2023-01-09
RM-8702 Upgrade Remotion.BuildScript to v3.0.0-alpha.8

a-ctor:
https://re-motion.atlassian.net/browse/RM-8702

@MichaelKetting do you want to backport the upgrade? -> please add the backport versions that you want. We probably want the same as the last backport?


MichaelKetting:
Will be backporting this to all re-motion v3.x branches (3.10-rc, 3.1-hotfix, 3.0-hotfix). We will also do a backport to buildscript v2.x and use that in re-motion 2.x-hotfix branches. Otherwise, running .net 4.6.x / 4.7.x configurations on Server 2022 agents will not work. These backports can wait until we are all done and convert the build agents to 2022, though. Maybe just setup an RMTEAM-task for now with fixversion v4.0 so we remember to do this?

MichaelKetting (approved):

PR #575, opened 2023-01-10
RM-8694 Fix TreeView problem where a click event is executed twice

a-ctor:
https://re-motion.atlassian.net/browse/RM-8694


Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -53,11 +60,22 @@ class WebTreeView
...
+            if (ev.target === linkElement)
+              return;
+
             linkElement.dispatchEvent (new PointerEvent (ev.type, ev));
             ev.preventDefault();
             ev.stopPropagation();
           });
         }
+
+        const expanderElement = node.querySelector(':scope > a:first-child');
```
MichaelKetting, line 73: Consistency question. In line 37, we have a cast to `Nullable<HTMLElement>`. Why do we not have this here, too?

MichaelKetting (approved):
I have one syntax question left, otherwise it's done.

Remotion/Web/ClientScript/Scripts/TreeView.ts
a-ctor: Removed the cast above

PR #591, opened 2023-01-17
Feature/rm-8715 clean up standard validators

martin-herwich:
https://re-motion.atlassian.net/browse/RM-8715


martin-herwich:
Maybe we should treat string[] differently from byte[], and both differently from other arrays, because byte[] conceptually is a single (binary) value, whereas string[] is a collection of lines of text, but in general an array is just a collection.
For string[], it would make sense to have a kind of NotEmptyOrWhitespace validator, which checks that at least one entry (line) contains at least one visible character.
For byte[], the current "NotEmpty" approach fits, because there is no concept analogous to "whitespace" for bytes; a byte[] that contains only zeroes should be valid, but a zero-length byte[] should not be (use null instead).
For other arrays, use NotEmptyListValidator, maybe rename to NotEmptyCollectionValidator.

MichaelKetting (changes requested):
I covered (alsmost) all of the implementation in the review. tests come in a round 2.

Remotion/Validation/Core/Validators/NotEmptyValueValidator.cs
```diff
@@ -15,24 +15,22 @@
...
-using JetBrains.Annotations;
 using Remotion.FunctionalProgramming;
 using Remotion.Utilities;
 using Remotion.Validation.Implementation;
 using Remotion.Validation.Results;
 
 namespace Remotion.Validation.Validators
 {
-  public class NotEmptyValidator : INotEmptyValidator
+  public class NotEmptyValueValidator : IPropertyValidator
```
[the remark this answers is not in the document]
martin-herwich: For now, `string[]` should be handled as a collection. When we introduce `NotEmptyOrWhitespaceValidator` in [RM-8680](https://re-motion.atlassian.net/browse/RM-8680), that validator should also support `string[]`.
I can see the value of splitting the `NotEmptyValueValidator` into `NotEmptyStringValidator` and `NotEmptyBinaryValidator`; the `NotEmptyStringValidator` would become `NotEmptyOrWhitespaceValidator` in RM-8680.
With that in mind, an interface `INotEmptyValueValidator` seems like overkill.

Remotion/Validation/UnitTests/Merging/OrderPrecedenceValidationRuleCollectorMergerTest.cs
```diff
@@ -42,7 +42,7 @@ public class OrderPrecedenceValidationRuleCollectorMergerTest
     private Expression<Func<Customer, string>> _firstNameExpression;
     private Expression<Func<Customer, string>> _lastNameExpression;
     private NotNullValidator _notNullValidator;
-    private NotEmptyValidator _notEmptyValidator;
+    private NotEmptyValueValidator _notEmptyValueValidator;
```
MichaelKetting, line 45: Note: many (all) of these unit and integration tests will get touched again when we update to NotEmptyOrWhitespaceValidator since we use string property validation so much in the test domain.

Remotion/Validation/UnitTests/Validators/NotEmptyValidatorTest.cs
```diff
@@ -15,7 +15,6 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 //
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
 using Remotion.Validation.Implementation;
```
MichaelKetting, line 20: Test needs to be renamed to NotEmptyValueValidatorTest

Remotion/Validation/UnitTests/Validators/NotEmptyValidatorTest.cs
```diff
@@ -77,8 +76,8 @@ public void Validate_WithPropertyValueIsEmptyCollection_ReturnsSingleValidationF
     [Test]
     public void Validate_WithPropertyValueIsNonEmptyCollection_ReturnsNoValidationFailures ()
```
MichaelKetting, line 77: Please check, there's probably a couple of tests that are not applicable here or must be added for byte[]. And when we split string and byte[], some tests get moved to just one of the testfixtures.

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
```diff
@@ -250,10 +249,36 @@ public void CreateValidationMessageForPropertyValidator_WithNotEmptyValidatorAnd
     }
 
     [Test]
-    public void CreateValidationMessageForPropertyValidator_WithNotEmptyValidatorAndPropertyIsIEnumerable_ReturnsLocalizedValidationMessage ()
+    public void CreateValidationMessageForPropertyValidator_WithNotEmptyValueValidatorAndPropertyIsIEnumerable_ReturnsLocalizedValidationMessage ()
```
MichaelKetting, line 252: IEnumerable is no longer correct, is it?

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
```diff
@@ -250,10 +249,36 @@ public void CreateValidationMessageForPropertyValidator_WithNotEmptyValidatorAnd
...
+      var validationMessage = _factory.CreateValidationMessageForPropertyValidator(validator, _propertyStub.Object);
+
+      Assert.That(validationMessage, Is.Not.Null);
+      Assert.That(validationMessage, Is.InstanceOf<ResourceManagerBasedValidationMessage>());
+
+      Assert.That(validationMessage.ToString(), Is.EqualTo("Add an entry."));
+    }
+
+    [Test]
+    public void CreateValidationMessageForPropertyValidator_WithNotEmptyListValidatorAndPropertyIsList_ReturnsLocalizedValidationMessage ()
```
MichaelKetting, line 278: Here we get extra tests for the different colleciton interfaces.

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
```diff
@@ -250,10 +249,36 @@ public void CreateValidationMessageForPropertyValidator_WithNotEmptyValidatorAnd
...
+      var validationMessage = _factory.CreateValidationMessageForPropertyValidator(validator, _propertyStub.Object);
+
+      Assert.That(validationMessage, Is.Not.Null);
+      Assert.That(validationMessage, Is.InstanceOf<ResourceManagerBasedValidationMessage>());
+
+      Assert.That(validationMessage.ToString(), Is.EqualTo("Add an entry."));
+    }
+
+    [Test]
+    public void CreateValidationMessageForPropertyValidator_WithNotEmptyValueValidatorAndPropertyIsArray_ReturnsLocalizedValidationMessage ()
```
MichaelKetting, line 265: This should be byte[]/binary, correct?

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
```diff
@@ -280,7 +280,7 @@ public void GetUnhandledValidationFailures_ForBusinessObject_ReturnsFilteredResu
       Assert.That(sortedUnhandledValidationFailures[0].ErrorMessage, Is.EqualTo("Localized Object Validation Error"));
 
       Assert.That(sortedUnhandledValidationFailures[1].ValidatedProperty, Is.SameAs(_customerNumberBusinessObjectProperty));
-      Assert.That(sortedUnhandledValidationFailures[1].ErrorMessage, Is.EqualTo("NotEmptyValidator: Validation error."));
+      Assert.That(sortedUnhandledValidationFailures[1].ErrorMessage, Is.EqualTo("NotEmptyValueValidator: Validation error."));
```
MichaelKetting, line 283: Am not 100% sure, but I think we may need to add a test case for NotEmptyListValidator to show every variant is covered?

MichaelKetting (changes requested):
Review done.

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
martin-herwich: I don't think so, because this test is about creating a `BusinessObjectValidationResult` from a `ValidationResult`, which is independent of the validators that create the `ValidationFailures` in the `ValidationResult`. The test has always used only a small subset of all possible validators.

Remotion/Validation/Core/Attributes/Validation/NotEmptyValidationAttribute.cs
```diff
@@ -38,18 +38,24 @@ protected override IEnumerable<IPropertyValidator> GetValidators (IPropertyInfor
       ArgumentUtility.CheckNotNull("property", property);
       ArgumentUtility.CheckNotNull("validationMessageFactory", validationMessageFactory);
 
-      NotEmptyValidator validator;
-      if (string.IsNullOrEmpty(ErrorMessage))
+      IPropertyValidator CreateValidator (Func<ValidationMessage, IPropertyValidator> factory)
```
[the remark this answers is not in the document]
MichaelKetting: You skipped the question - I would like to qualify the pros and cons of the local-function vs member method decision.

Remotion/Validation/Core/Attributes/Validation/NotEmptyValidationAttribute.cs
```diff
@@ -38,20 +61,34 @@ protected override IEnumerable<IPropertyValidator> GetValidators (IPropertyInfor
...
+      {
+        if (string.IsNullOrEmpty(ErrorMessage))
+        {
+          return PropertyValidatorFactory.Create(
+              property,
+              parameters => factory(parameters.ValidationMessage),
+              validationMessageFactory);
+        }
+
+        return factory(new InvariantValidationMessage(ErrorMessage));
```
MichaelKetting, line 90: Code style: we try to do balanced IFs when the code should be symmetrical: only guard/exit conditions get their stand-alone IF. Business logic that has two (or more) equal branches should also be set in an IF/ELSE to make this obvious.

Remotion/Validation/UnitTests/Attributes/Validation/NotEmptyValidationAttributeTest.cs
```diff
@@ -40,35 +40,99 @@ public void SetUp ()
...
+    {
+      var propertyInformation = PropertyInfoAdapter.Create(typeof(Customer).GetProperty(nameof(Customer.ShippingAddresses)));
+      _attribute.ErrorMessage = "CustomMessage";
+
+      var result = _attribute.GetPropertyValidators(propertyInformation, _validationMessageFactoryStub.Object).ToArray();
+
+      Assert.That(result.Length, Is.EqualTo(1));
+      Assert.That(((NotEmptyCollectionValidator)result[0]).ValidationMessage, Is.InstanceOf<InvariantValidationMessage>());
+      Assert.That(((NotEmptyCollectionValidator)result[0]).ValidationMessage.ToString(), Is.EqualTo("CustomMessage"));
     }
```
MichaelKetting, line 136: I blieve these tests only covers one of the 3 branches for the collection validator. could you please verify and either explain what I'm missing or add the missing tests?

Remotion/Validation/Core/DefaultValidatorExtensions.cs
```diff
@@ -30,10 +31,40 @@ public static class DefaultValidatorExtensions
       return ruleBuilder.SetValidator(p => new NotNullValidator(p.ValidationMessage));
     }
```
MichaelKetting, line 33: I know that I'm still missing the individual tests in DefaultValidatorExtensionsTest. SInce we've now added some generic extension methods that could apply to the same type, could you check if we either have integration tests that cover this possible ambiguity or add those as unit tests? If it's too much work to actually make the tests green / add good assertions, it's enough for now to add the calls and keep the test as Ignored so the compiler can at least tell us that we did all right.

Remotion/Validation/Core/Validators/NotEmptyBinaryValidator.cs
```diff
@@ -55,23 +53,10 @@ private bool IsValid (PropertyValidatorContext context)
...
-        return !collectionValue.Cast<object>().Any();
-
-      return false;
-    }
-
-    private bool IsEmptyString (object value)
-    {
-      if (value is string stringValue)
-        return stringValue.Length == 0;
+      if(propertyValue is byte[] binaryValue)
```
MichaelKetting, line 56:
```suggestion
      if (propertyValue is byte[] binaryValue)
```

Remotion/Validation/Core/Validators/NotEmptyCollectionValidator.cs
```diff
@@ -0,0 +1,81 @@
...
+      if (propertyValue is ICollection collection)
+        return collection.Count > 0;
+
+      if (propertyValue is IReadOnlyCollection<object> readOnlyCollection)
+        return readOnlyCollection.Count > 0;
+
+      if (propertyValue.GetType().CanAscribeTo(typeof(ICollection<>)))
+        return ((IEnumerable<object>)propertyValue).Any();
+
+      throw new NotSupportedException($"Only the following types are supported: IReadOnlyCollection<T>, ICollection<T>, and ICollection.");
```
MichaelKetting, line 68: The original implementation simply skipped incorrectly typed properties as valid. This behavior is true for all validators and should be preserved.

Remotion/Validation/UnitTests/Validators/NotEmptyCollectionValidatorTest.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+using Remotion.Validation.Implementation;
+using Remotion.Validation.Validators;
+
+namespace Remotion.Validation.UnitTests.Validators
+{
+  [TestFixture]
+  public class NotEmptyCollectionValidatorTest : ValidatorTestBase
```
MichaelKetting, line 27: pelase add tests for all collection property types. Right now, we don't have full coverage. We should also add an explicit test for empty IEnumerable<object> to show that that's ignored.

Remotion/Validation/Core/Validators/NotEmptyCollectionValidator.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    public IEnumerable<PropertyValidationFailure> Validate (PropertyValidatorContext context)
+    {
+      if (IsValid(context))
+        return Enumerable.Empty<PropertyValidationFailure>();
+
+      return EnumerableUtility.Singleton(CreateValidationError(context));
+    }
+
+    private bool IsValid (PropertyValidatorContext context)
```
MichaelKetting, line 52: Please remind me: what is the intended semantic if a developer applies the NotEmptyCollectionValidator to a property of type string or byte[]? Should it validate or ignore? Based on this we also need to add tests for byte[] and string to document the behavior.

Remotion/Validation/Core/Validators/NotEmptyStringValidator.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+      var propertyValue = context.PropertyValue;
+
+      if (propertyValue == null)
+        return true;
+
+      if(propertyValue is string stringValue)
+        return stringValue.Length > 0;
+
+      throw new NotSupportedException($"Only properties of type '{nameof(String)}' are supported.");
```
MichaelKetting, line 59: Same as with NotEmptyCollectionValidator: incorrect type should be a valid result.

MichaelKetting, line 59: Syntax:
can we just write
```
if (property value is not string)
```
instead of two separate return cases for null and not-string and would this improve legibility? It would still be two seaprate test cases though.

Remotion/Validation/Core/Validators/NotEmptyBinaryValidator.cs
```diff
@@ -55,23 +53,10 @@ private bool IsValid (PropertyValidatorContext context)
...
+        return binaryValue.Length > 0;
 
-      return false;
+      throw new NotSupportedException($"Only properties of type '{nameof(Byte)}[]' are supported.");
```
MichaelKetting, line 59: See NotEmptyStringValidator for review notes :)

Remotion/Validation/Globalization/LocalizedValidationMessageFactory.cs
```diff
@@ -52,7 +52,7 @@ public enum ResourceIdentifier
       ValueMustMatchPredicateValidationMessage,
       ValueMustMatchRegularExpressionValidationMessage,
       ValueMustNotBeEmptyCollectionValidationMessage,
-      ValueMustNotBeEmptyStringValidationMessage,
+      ValueMustNotBeEmptyValueValidationMessage,
```
MichaelKetting, line 55: Please split this up for byte[] and string. We have several cases (see null values) where we have the same message with different values but still used typed resources. That way, we can decide this on a case-by-case basis and don't have to rewrite the IDs just to customate the message for a specific type. it also allows types-specific overrides in project code so having types-specific messages is also a possible extension point / feature provided by the infrastrucuture.

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
```diff
@@ -217,43 +216,62 @@ public void CreateValidationMessageForPropertyValidator_WithMinimumLengthValidat
...
+      var validationMessage = _factory.CreateValidationMessageForPropertyValidator(validator, _propertyStub.Object);
+
+      Assert.That(validationMessage, Is.Not.Null);
+      Assert.That(validationMessage, Is.InstanceOf<ResourceManagerBasedValidationMessage>());
+
+      Assert.That(validationMessage.ToString(), Is.EqualTo("Add an entry."));
+    }
+
+    [Test]
+    public void CreateValidationMessageForPropertyValidator_WithNotEmptyCollectionValidator_WithGenericCollection_ReturnsLocalizedValidationMessage ()
```
MichaelKetting, line 271: I hope I'm not contradicting my first review here: for the MessageFactory, we don't need to cover individual property types supported by the validator, it should be enough to have one test for the validator-type to achieve coverage (the factory does not know about the property type).

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.Validation/DomainObjectAttributesBasedValidationPropertyRuleReflector.cs
```diff
@@ -170,15 +170,22 @@ public IEnumerable<IPropertyValidator> GetRemovablePropertyValidators ()
...
-          && typeof(IEnumerable).IsAssignableFrom(_implementationProperty.PropertyType)
-          && !ReflectionUtility.IsObjectList(_implementationProperty.PropertyType)
-          && !ReflectionUtility.IsIObjectList(_implementationProperty.PropertyType))
+      if (!_domainModelConstraintProvider.IsNullable(_implementationPropertyInformation))
       {
-        yield return PropertyValidatorFactory.Create(
-            _implementationPropertyInformation,
-            parameters => new NotEmptyValidator(parameters.ValidationMessage),
-            _validationMessageFactory);
+        if (typeof(string).IsAssignableFrom(_implementationProperty.PropertyType))
```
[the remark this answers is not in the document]
martin-herwich: In this case, I think the logic is more readable with the deeper nesting, because the separate IsNullable-check expresses that the entire nested block is about properties that are not nullable.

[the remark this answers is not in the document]
MichaelKetting: @a-ctor you agree, too?

Remotion/Validation/Core/Attributes/Validation/NotEmptyValidationAttribute.cs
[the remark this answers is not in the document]
martin-herwich: Sorry, I did not see the posting starting with "Question", only that above.
Converting the local function to a member method adds syntactic noise, because I have to pass two additional parameters in all 5 lines.
The most important information in this method is the mapping between `property.PropertyType` and the `factory` delegate. I could put that in a dictionary to make it explicit, and have the rest of the code looking up whether an entry exists and creating the `IPropertyValidator` from the entry's `factory` delegate.
What do you think?

Remotion/Validation/Core/Validators/NotEmptyStringValidator.cs
martin-herwich: Using a validator for an incompatible property is a programming error, isn't it? If so, the framework should support the developer in finding that he made a mistake.
Silently failing to validate anything and returning that the value is okay would hide the error. Why is that a better policy?

Remotion/Validation/Core/Attributes/Validation/NotEmptyValidationAttribute.cs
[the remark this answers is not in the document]
MichaelKetting: Ah, I see. That's actually an interesting idea! Bascially, we test for the type and use that as the key. We won't get rid of the if/else block because we still need to calculate the key based on the PropertyType though and I'm not sure if that works well. Perhaps just extract the "get a delegate" part not in a dictionary but in a separate code block that does just the mapping of PropertyType to delegate and then go from there. This could then be a separate method, perhaps just a local function or just a code block - not sure yet.

``` 
  var factory;
  if (type == string)
    factory = msg=> new NotEmptyStringValidator(msg);
  /* other conditions */
  else if (type.IsAssignableTo(ICollection<>)
    factory = msg => ...
  else
    throw ....
```

Also, noticed, we don't need `IsAsignable` with `string` and `byte[]`, just equality. I completely spaced on that while reading the code.

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
martin-herwich: It does contradict the earlier review.
`LocalizedValidationMessageFactory` get the `IPropertyInformation` as input parameter, so it can differentiate by property type. For `NotEmptyCollectionValidation`, there is only one path in the implementation, but it could be different (like for `LengthValidator` and `NotNullValidator`), wherefore the additional tests make sense.

Remotion/Validation/Core/Validators/NotEmptyStringValidator.cs
MichaelKetting: Yes, that's the conventinal wisdom. Thing is, right now there's prior art with the other validator types, so we should keep the existing semantics. If it turns out that it's a bad idea / we want something else, we should make a separate task to change all the validators in one go.

In practice: it's a guardrail for the infrastructure itself and anyone who's building their own factory infrastructure. 

Historically: this descision was inherited from FluentValidation. The big question is, should a null-value and an invalid type be treated differently? It does feel more consistent to ignore all invalid values, otherwise you need to question why NULL is special.

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
MichaelKetting: Ah, okay, yes, thank you. 

okay, so it comes down to if you do htis test first, you see there's no special logic for propertytype needed with collection validatior and you'd stop with one test. It's okay to cover all three, unless we think it could confuse someone reading the tests. (=> bascally, it doesn't matter if there's one or three in this particular instance)

Remotion/Data/DomainObjects.Validation/DomainObjectAttributesBasedValidationPropertyRuleReflector.cs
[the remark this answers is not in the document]
a-ctor: I think the nested version is more readable.

MichaelKetting, line 175: Same as with NotEmptyValidationAttribute:
```suggestion
        if (_implementationProperty.PropertyType == typeof(string))
```

Remotion/Data/DomainObjects.Validation/DomainObjectAttributesBasedValidationPropertyRuleReflector.cs
```diff
@@ -170,15 +170,22 @@ public IEnumerable<IPropertyValidator> GetRemovablePropertyValidators ()
...
+        {
+          yield return PropertyValidatorFactory.Create(
+              _implementationPropertyInformation,
+              parameters => new NotEmptyStringValidator(parameters.ValidationMessage),
+              _validationMessageFactory);
+        }
+        else if (typeof(byte[]).IsAssignableFrom(_implementationProperty.PropertyType))
```
MichaelKetting, line 182:
```suggestion
        else if (_implementationProperty.PropertyType == typeof(byte[]))
```

Remotion/Validation/Core/Attributes/Validation/NotEmptyValidationAttribute.cs
```diff
@@ -38,20 +65,34 @@ protected override IEnumerable<IPropertyValidator> GetValidators (IPropertyInfor
       ArgumentUtility.CheckNotNull("property", property);
       ArgumentUtility.CheckNotNull("validationMessageFactory", validationMessageFactory);
 
-      NotEmptyValidator validator;
+      Func<ValidationMessage, IPropertyValidator> validatorFactory;
+      if(typeof(string) == property.PropertyType)
```
MichaelKetting, line 69:
```suggestion
      if(property.PropertyType == typeof(string))
```
The compared-to data is on the right-side whenever syntactically possible to correspond with natural language / mental flow.

MichaelKetting (changes requested):
Some review remarks are left open.

And I did some inline edits that can simply be applied.

MichaelKetting (commented):
Somethings strange here, there's a bunch of unrelated changes in the PR. could you please check? Maybe you didn't push the updated position of `develop5` to your Github?

Remotion/Validation/UnitTests/Validators/NotEmptyCollectionValidatorTest.cs
martin-herwich: I added the tests, but I needed some test classes in order to separate the three interfaces; I doubt that the benefits of those tests are worth the effort, because the code we test here is trivial.

Remotion/Validation/Core/Validators/NotEmptyCollectionValidator.cs
martin-herwich: If you use a `NotEmptyCollectionValidator` on any collection (`ICollection`, `IReadOnlyCollection<T>`, or `ICollection<T>`)  the semantics is the same: the collection must not be empty. This also works for `byte[]` - but since this is not the semantics we usually need for `byte[]`. the `NotEmpty()` extension method creates a `NotEmptyByteArrayValidator`.
Since string is an `IEnumerable` and `IEnumerable<char>`, but not a **collection** (see above) of `char`, `NotEmptyCollectionValidator` would do nothing with a `string` value.

Tests added for empty `byte[]` and empty `string`.

Remotion/Validation/Core/DefaultValidatorExtensions.cs
martin-herwich: I have added tests for all extension methods.

Remotion/Validation/UnitTests/Validators/NotEmptyCollectionValidatorTest.cs
```diff
@@ -0,0 +1,261 @@
...
+
+      IEnumerator<T> IEnumerable<T>.GetEnumerator ()
+      {
+        return _list.GetEnumerator();
+      }
+
+      public void Add (T item)
+      {
+        _list.Add(item);
+      }
```
MichaelKetting, line 89: Syntax-question: do you want to shorten those oneliners using expression-bodied member syntax? (Resharper has a quick fix for that)

It's also okay to put them at the end of the file. That's become something we do in the tests.

```suggestion
      public void Add (T item) => _list.Add(item);
```

Remotion/Validation/UnitTests/DefaultValidatorExtensionsTest.cs
```diff
@@ -1,14 +1,416 @@
...
+    }
+
+    [Test]
+    public void ExactLength_ForString_ReturnsLengthValidator ()
+    {
+      var validationMessage = new InvariantValidationMessage("Fake message");
+      var initParameters = new PropertyValidationRuleInitializationParameters(validationMessage);
+      IPropertyValidator createdValidator = null;
+      var addingPropertyValidationRuleBuilder = new Mock<IAddingPropertyValidationRuleBuilder<object, string>>();
+      addingPropertyValidationRuleBuilder.Setup(_ => _.SetValidator(It.IsAny<Func<PropertyValidationRuleInitializationParameters, IPropertyValidator>>()))
```
MichaelKetting, line 193: Naming: please suffix those test doubles to keep with the existing naming conventions.
...Stub for stub
...Mock for mock

Here, it's all stubs since we don't do any expections.

Remotion/Validation/UnitTests/DefaultValidatorExtensionsTest.cs
```diff
@@ -1,14 +1,416 @@
...
+          .Callback<Func<PropertyValidationRuleInitializationParameters, IPropertyValidator>>(func => createdValidator = func.Invoke(initParameters));
```
MichaelKetting, line 194: Layout: we're trying to do those multiline statements for mock setups with one call per line. This helps with reading the setup syntax by giving the eye a single line to read.
```suggestion
      addingPropertyValidationRuleBuilder
          .Setup(_ => _.SetValidator(It.IsAny<Func<PropertyValidationRuleInitializationParameters, IPropertyValidator>>()))
          .Callback<Func<PropertyValidationRuleInitializationParameters, IPropertyValidator>>(func => createdValidator = func.Invoke(initParameters));
```

MichaelKetting (changes requested):
Reviewed the already done stuff.

Remotion/Validation/UnitTests/DefaultValidatorExtensionsTest.cs
martin-herwich: `addingPropertyValidationRuleBuilder` uses a `Setup` (which creates an expectation) and the call to `SetValidator` is also verified. Is it still a Stub? Should I remove the call to `Verify()`?

MichaelKetting: `Setup` is used for both stub and mock. Only by adding `Verifiable` at the end, it becomes mock semantics.

Since making a mock is only needed when we want to assert that void-calls are done correctly, we can stick with the existing stub semantics. The storing of the created validator during the callback and later asserting that value takes care of the verifications we need to do.

Remotion/Validation/UnitTests/DefaultValidatorExtensionsTest.cs
```diff
@@ -1,14 +1,416 @@
...
+
+      addingPropertyValidationRuleBuilder.Object.ExactLength(23);
+
+      addingPropertyValidationRuleBuilder.Verify();
+      Assert.That(createdValidator, Is.InstanceOf<LengthValidator>());
+      var lengthValidator = (LengthValidator)createdValidator;
+      Assert.That(lengthValidator.Min, Is.EqualTo(23));
+      Assert.That(lengthValidator.Max, Is.EqualTo(23));
```
MichaelKetting, line 202: Hmm...I think we should also test that the initParameters are used correctly by adding the following assertion at the end:
```
Assert.That(lengthValidator .ErrorMessage, Is.SameAs(validationMessage));
```

We should do this for each method. 

The intention of this is to show that the parameter is used correctly and no one creates a hardcoded message inline in the extension method. It goes in the same direction as testing that the other arguments are passed correctly, such the Min and Max values here.

MichaelKetting (changes requested):
Am sorry, I think I found a missing assertion while answering your question :o

Remotion/Validation/UnitTests/DefaultValidatorExtensionsTest.cs
martin-herwich: I see. I changed the names to "...Stub" and removed the calls to `Verify()`, except for the `predicateMock`, which I made `Verifiable` in order to check that the predicate I passed as a parameter is actually the one used in the `createdValidator`.

Remotion/Validation/UnitTests/DefaultValidatorExtensionsTest.cs
```diff
@@ -1,14 +1,448 @@
...
+
+      addingPropertyValidationRuleBuilderStub.Object.NotEmpty();
+
+      Assert.That(createdValidator, Is.InstanceOf<NotEmptyCollectionValidator>());
+      var notEmptyValidator = (NotEmptyCollectionValidator)createdValidator;
+      Assert.That(notEmptyValidator.ValidationMessage, Is.SameAs(validationMessage));
+    }
+
+    [Test]
+    public void NotEmpty_ForCollection_ReturnsNotEmptyReadOnlyCollectionValidator ()
```
MichaelKetting, line 126: As discussed, we also need to add tests for `List<T>` to show that types such as ObjectList<T> which implement  `IList`, `IList<T>`, and `IReadOnlyList<T>` work in the overload resolution.

MichaelKetting (changes requested):

Remotion/Validation/Core/DefaultValidatorExtensions.cs
```diff
@@ -143,13 +251,19 @@ public static class DefaultValidatorExtensions
...
+    /// </summary>
     public static IAddingPropertyValidationRuleBuilder<TValidatedType, string> Matches<TValidatedType> (
         this IAddingPropertyValidationRuleBuilder<TValidatedType, string> ruleBuilder,
         [JetBrains.Annotations.RegexPattern] string expression)
     {
       return ruleBuilder.SetValidator(p => new RegularExpressionValidator(new Regex(expression), p.ValidationMessage));
     }
 
+    /// <summary>
+    /// Adds a <see cref="PredicateValidator"/> with the given <paramref name="predicate"/> to the <paramref name="ruleBuilder"/>.
```
MichaelKetting, line 265: Thanks for the docs!

Remotion/Validation/UnitTests/Validators/NotEmptyCollectionValidatorTest.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+      private readonly List<object> _list = new();
+
+      public IEnumerator GetEnumerator () => _list.GetEnumerator();
+
+      public void CopyTo (Array array, int index)
+      {
+        var objectArray = array.Cast<object>().ToArray();
+        _list.CopyTo(objectArray, index);
+      }
```
MichaelKetting, line 174: Oh, there's a bug here. Since we're not using the API in the tests, it's not relevant, tough we should either fix make it through an exception instead. Otherwise, someone might find this and copy it to production.

The bug: the CopyTo method copies the list into the array. Since we create a new array from the parameter, the list is copied into the new array which is then discarded. the array from the parameter remains unchanged.

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/Validators/NotEmptyCollectionValidatorTest.cs
```diff
@@ -0,0 +1,218 @@
...
+    private class SimpleCollection : ICollection
+    {
+      private readonly ArrayList _list = new();
+
+      public IEnumerator GetEnumerator () => _list.GetEnumerator();
+
+      public void CopyTo (Array array, int index)
+      {
+        _list.CopyTo(array, index);
+      }
```
MichaelKetting, line 173: that works. great. I'll approve the PR, though recommend you make this into an expression. will remained approved.

MichaelKetting (approved):

PR #592, opened 2023-01-18
RM-7858 Replace NO_PEVERIFY with DISABLE_PEVERIFY

Frazerus:
https://re-motion.atlassian.net/browse/RM-7858

### DIFF
--- a/Build/Shared.build.props
+++ b/Build/Shared.build.props
@@ -35,7 +35,7 @@
     <PackageTags Condition="'$(ProductType)' == 'SecurityManager'">$(PackageTags) SecurityManager</PackageTags>
     <DisableTransitiveProjectReferences>true</DisableTransitiveProjectReferences>
     <DefineConstants Condition="'$(TargetFramework)' == 'net48'">$(DefineConstants);FEATURE_GAC;FEATURE_SERIALIZATION;FEATURE_ASSEMBLYBUILDER_SAVE;FEATURE_THREAD_ABORT;FEATURE_STRONGNAMESIGNING</DefineConstants>
-    <DefineConstants Condition="'$(TargetFramework)' != 'net48'">$(DefineConstants);NO_PEVERIFY</DefineConstants>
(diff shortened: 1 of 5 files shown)


a-ctor (approved):

Build/Shared.build.props
```diff
@@ -35,7 +35,7 @@
     <PackageTags Condition="'$(ProductType)' == 'SecurityManager'">$(PackageTags) SecurityManager</PackageTags>
     <DisableTransitiveProjectReferences>true</DisableTransitiveProjectReferences>
     <DefineConstants Condition="'$(TargetFramework)' == 'net48'">$(DefineConstants);FEATURE_GAC;FEATURE_SERIALIZATION;FEATURE_ASSEMBLYBUILDER_SAVE;FEATURE_THREAD_ABORT;FEATURE_STRONGNAMESIGNING</DefineConstants>
-    <DefineConstants Condition="'$(TargetFramework)' != 'net48'">$(DefineConstants);NO_PEVERIFY</DefineConstants>
+    <DefineConstants Condition="'$(TargetFramework)' != 'net48'">$(DefineConstants);DISABLE_PEVERIFY</DefineConstants>
```
MichaelKetting, line 38: Question @a-ctor: since the goal is to improve legiblity of the code:

would ENABLE_PEVERIFY actually be a better option?

Every usage uses the !DISABLE_PEVERIFY notation, so it's a double negative when reading. 

We have to decide if it's okay to not have PEVERIFY unless we set the flag. This goes hand-in-hand with the feature flags.

Thoughts? @a-ctor & @Frazerus ?

a-ctor: Enable would be better IMHO.

MichaelKetting (changes requested):
As discussed: Rename / invert boolean for the flag

MichaelKetting (approved):
Please update the ticket when done.

PR #599, opened 2023-01-19
RM-8698 Remove RegisterResizeOnElement()

KuJo7:
https://re-motion.atlassian.net/browse/RM-8699

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/SmartPage.ts
+++ b/Remotion/Web/ClientScript/Scripts/SmartPage.ts
@@ -37,7 +37,6 @@ type SmartPage_OnAbortEventHandler = (this: null, hasSubmitted: boolean, isCache
 type SmartPage_OnUnloadEventHandler = (this: null) => void;
 type SmartPage_OnPostbackEventHandler = (this: null, eventTarget: Nullable<string>, eventObject: string) => void;
 type SmartPage_OnScrollEventHandler = (this: null) => void;
-type SmartPage_OnResizeEventHandler = (this: null) => void;
 
 type SmartPage_EventMap = {
   onloading: SmartPage_OnLoadingEventHandler,
@@ -47,7 +46,6 @@ type SmartPage_EventMap = {
   onunload: SmartPage_OnUnloadEventHandler,
   onpostback: SmartPage_OnPostbackEventHandler,
   onscroll: SmartPage_OnScrollEventHandler,
-  onresize: SmartPage_OnResizeEventHandler,
 };
 type SmartPage_Event = keyof SmartPage_EventMap;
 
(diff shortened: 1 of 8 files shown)


Remotion/Web/Core/UI/SmartPageEvents.cs
```diff
@@ -34,8 +34,6 @@ public enum SmartPageEvents
     OnAbort,
     /// <summary> Raised when the user scrolls the page. Signature: <c>void Function ()</c> </summary>
     OnScroll,
```
a-ctor, line 36: I think we still need to create the issue to remove on scroll?

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -821,14 +816,6 @@ class SmartPage_Context
     this.ExecuteEventHandlers('onscroll');
   };
 
-  // Event handler for Window.OnResize.
-  public OnResize(): void
-  {
-    if (this._statusMessageWindow != null)
-      this.AlignStatusMessage(this._statusMessageWindow);
```
a-ctor, line 828: We need to check if the status message is still getting aligned correctly on resize. If no we should try and fix it using css so that we can remove the code here.

Remotion/Web/Core/Utilities/IScriptUtility.cs
```diff
@@ -27,20 +27,10 @@ namespace Remotion.Web.Utilities
  public interface IScriptUtility
   {
     /// <summary>
-    /// Registers the include files required for the calls to <see cref="RegisterResizeOnElement"/>.
```
a-ctor, line 30: We cannot remove all of the description. Maybe something like `Registers the client side scripts for the current theme.`

a-ctor (changes requested):

Remotion/Web/Core/UI/SmartPageEvents.cs
KuJo7: https://re-motion.atlassian.net/browse/RM-8717

Remotion/Web/ClientScript/Scripts/SmartPage.ts
KuJo7: I tested this by doing a postback and then clicking the go back button, which triggers a message. I did this before the change and after the change. In both cases, the message is getting aligned correctly on resize.
@MichaelKetting Is there another issue that we have missed?

MichaelKetting: did you test in both NovaViso theme and NovaGray theme?

You can also trigger it via `SmartPage_Context.Instance.ShowStatusIsSubmittingMessage()` in the browser's Javascript console.

KuJo7: Yes, it works for both themes!

a-ctor (approved):

Remotion/Web/Core/UI/SmartPageEvents.cs
```diff
@@ -34,8 +34,6 @@ public enum SmartPageEvents
...
-    /// <summary> Raised when the user resizes the page. Signature: <c>void Function ()</c> </summary>
-    OnResize,
```
MichaelKetting, line 38: Considering there are (very) few usages, but not none, we should keep the value and mark it as Obsolete(error=true). Message: `"SmartPage.OnResize event is no longer supported. Use CSS (recommendation) or implement a resize handler inside application code. (Version 4.0.0)"`

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #601, opened 2023-01-25
RM-7347 Add LicenseHeader to packed Nuget

KuJo7:
https://re-motion.atlassian.net/browse/RM-7347

### DIFF
--- a/Remotion/ObjectBinding/Web/ObjectBinding.Web.csproj
+++ b/Remotion/ObjectBinding/Web/ObjectBinding.Web.csproj
@@ -19,6 +19,9 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
+    <None Update="..\..\Remotion LGPL.licenseheader" Pack="true" PackagePath="res/" Visible="false" />
+  </ItemGroup>
+  <ItemGroup>
     <ProjectReference Include="..\Web.Contracts.DiagnosticMetadata\ObjectBinding.Web.Contracts.DiagnosticMetadata.csproj" />
     <ProjectReference Include="..\..\Core\Core\Core.Core.csproj" />
     <ProjectReference Include="..\..\Core\Reflection\Core.Reflection.csproj" />
--- a/SecurityManager/Clients.Web/SecurityManager.Clients.Web.csproj
+++ b/SecurityManager/Clients.Web/SecurityManager.Clients.Web.csproj
@@ -12,6 +12,9 @@
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
   <ItemGroup>
+    <None Update="..\SecurityManager AGPL.licenseheader" Pack="true" PackagePath="res/" Visible="false" />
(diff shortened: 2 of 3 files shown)


SecurityManager/Clients.Web/SecurityManager.Clients.Web.csproj
```diff
@@ -11,6 +11,9 @@
     <AssemblyName>Remotion.SecurityManager.Clients.Web</AssemblyName>
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
+  <ItemGroup>
+    <None Update="..\SecurityManager AGPL.licenseheader" Pack="true" PackagePath="res/HTML" Link="SecurityManager AGPL.licenseheader"/>
```
a-ctor, line 15:
```suggestion
    <None Update="..\SecurityManager AGPL.licenseheader" Pack="true" PackagePath="res/HTML" />
```

a-ctor (approved):

Remotion/ObjectBinding/Web/ObjectBinding.Web.csproj
```diff
@@ -18,6 +18,9 @@
       <PackagePath>res/HTML</PackagePath>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <None Update="..\..\Remotion LGPL.licenseheader" Pack="true" PackagePath="res/HTML" />
```
a-ctor, line 22: We should change the package path to `res/` in all of the cases. I don't think we have any other files with licenseheaders in any other folder than `HTML` but it can't hurt.

MichaelKetting, line 22: We also should apply `Visible=false`.
see https://re-motion.atlassian.net/browse/RMSRCBUILD-181 for reference.

MichaelKetting:
@KuJo7 
cc @a-ctor 
I don't know what went wrong, I did not yet mark it as approved. Maybe just one approver is needed no matter how many have been requested? Anyway, no nightly build has run, so I couldn't yet review the final package output which means the PR is still open.

KuJo7:
> @KuJo7 cc @a-ctor I don't know what went wrong, I did not yet mark it as approved. Maybe just one approver is needed no matter how many have been requested? Anyway, no nightly build has run, so I couldn't yet review the final package output which means the PR is still open.

new PR: https://github.com/re-motion/Framework/pull/603

Remotion/ObjectBinding/Web/ObjectBinding.Web.csproj
a-ctor: Since we only update the item (and not add it) we should evaluate if `Visible="False"` is really wanted here. The licenseheader is added in `Shared.Build.props` and now it is invisible only in this project.

If we want it invisible we should change it in `Shared.Build.props` otherwise we should remove the attribute again.

PR #602, opened 2023-01-27
RM-8725 Fix indent size for .cs Files

KuJo7:
https://re-motion.atlassian.net/browse/RM-8725


a-ctor (approved):

MichaelKetting:
> @MichaelKetting Do we want to merge this in the next version or can we do it now since it is already done?

@a-ctor we can merge it now.

PR #607, opened 2023-02-02
RM-8726 AssemblyFinderTypeDiscoveryService.GetTypes performs very bad

a-ctor:
https://re-motion.atlassian.net/browse/RM-8706

### DIFF
--- a/Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
+++ b/Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
@@ -39,8 +39,8 @@ namespace Remotion.Reflection.TypeDiscovery
       using (StopwatchScope.CreateScope(s_log.Value, LogLevel.Debug, string.Format("Built BaseTypeCache. Time taken: {{elapsed}}")))
       {
         // Note: there is no measurable impact when switching this code to parallel execution.
-        var classes = new HashSet<KeyValuePair<Type, Type>>();
-        var interfaces = new HashSet<KeyValuePair<Type, Type>>();
+        var classes = new List<KeyValuePair<Type, Type>>();
+        var interfaces = new List<KeyValuePair<Type, Type>>();
 
         var typesToProcess = new Queue<Type>();
         var enqueuedTypes = new HashSet<Type>();
@@ -51,6 +51,7 @@ namespace Remotion.Reflection.TypeDiscovery
             typesToProcess.Enqueue(type);
         }
 
+        var processedInterfaces = new HashSet<Type>();
         while (typesToProcess.Count > 0)
         {
           var type = typesToProcess.Dequeue();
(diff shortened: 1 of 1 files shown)


Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
```diff
@@ -69,13 +70,19 @@ public static BaseTypeCache Create (IEnumerable<Type> types)
             } while (nextBaseType != null);
           }
 
-          foreach (var interfaceType in type.GetInterfaces())
+          var implementedInterfaces = type.GetInterfaces();
+          if (implementedInterfaces.Length > 0)
```
MichaelKetting, line 74: Optimization vs code question:
Is there a benefit in guarding the entry? The clear-operation and the loop aren't really an issue for empty lists but we got more conditions so the code is more complex. thoughts?

Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
```diff
@@ -69,13 +70,19 @@ public static BaseTypeCache Create (IEnumerable<Type> types)
...
-              typesToProcess.Enqueue(normalizedInterfaceType);
+            processedInterfaces.Clear();
+            foreach (var implementedInterface in implementedInterfaces)
+            {
+              var normalizedInterfaceType = GetNormalizedType(implementedInterface);
+              if (enqueuedTypes.Add(normalizedInterfaceType))
+                typesToProcess.Enqueue(normalizedInterfaceType);
 
-            interfaces.Add(new KeyValuePair<Type, Type>(normalizedInterfaceType, type));
+              if (processedInterfaces.Add(normalizedInterfaceType))
```
MichaelKetting, line 83: What did you learn about the performance characteristicts for small Hashsets?

MichaelKetting (changes requested):
Got questions about simplifying things and Perf-results

Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
a-ctor: I checked and removed the if

Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
a-ctor: Too much variability to make a decisive decision so I choose the simpler one with the hash set.

Remotion/Core/Core/Reflection/TypeDiscovery/BaseTypeCache.cs
```diff
@@ -69,13 +70,15 @@ public static BaseTypeCache Create (IEnumerable<Type> types)
             } while (nextBaseType != null);
           }
 
-          foreach (var interfaceType in type.GetInterfaces())
+          processedInterfaces.Clear();
+          foreach (var implementedInterface in type.GetInterfaces())
           {
-            var normalizedInterfaceType = GetNormalizedType(interfaceType);
+            var normalizedInterfaceType = GetNormalizedType(implementedInterface);
```
MichaelKetting, line 76: last question: change `implementedInterface` back to `interfaceType` to reduce change set?

MichaelKetting (approved):

MichaelKetting (changes requested):
Forgot something: do we have a test for the multiple-generic interfaces situation?

a-ctor:
> Forgot something: do we have a test for the multiple-generic interfaces situation?

We have a test that covers the scenario. That's why I introduced the `HashSet`'s in the first place :p

MichaelKetting (approved):

PR #618, opened 2023-02-14
Default NotEmptyOrWhitespace validation for all string properties

martin-herwich:
https://re-motion.atlassian.net/browse/RM-8714


MichaelKetting (approved):
Looks great, including the hisotry :)

Please don't forget the Breaking Change tickets for the dropped NotEmpty behavior in re-store and the dropped API / change in the NotEmptyAttribute.

PR #619, opened 2023-02-16
RM-7808 Integrate the RoslynCodeDomProvider and renable the AssemblyCompiler tests

a-ctor:
https://re-motion.atlassian.net/browse/RM-7808

### DIFF
--- a/Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
+++ b/Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
@@ -25,14 +25,14 @@ using Remotion.Development.UnitTesting.Compilation;
 using Remotion.Reflection.TypeDiscovery.AssemblyFinding;
 using Remotion.Reflection.TypeDiscovery.AssemblyLoading;
 using Remotion.Utilities;
+#if !NETFRAMEWORK
+using Remotion.Development.UnitTesting.IsolatedCodeRunner;
+#endif
 
 namespace Remotion.UnitTests.Reflection.TypeDiscovery.AssemblyFinding
 {
   [TestFixture]
   [Serializable]
-#if !NETFRAMEWORK
-  [Ignore("TODO RM-7808: Integrate the RoslynCodeDomProvider and renable the AssemblyCompiler tests")]
-#endif
   public class AssemblyFinderIntegrationTest
   {
 #if !NETFRAMEWORK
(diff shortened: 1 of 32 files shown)


Remotion/Development/UnitTesting.Compilation/Roslyn/CSharpRoslynCodeProvider.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) RUBICON IT GmbH, www.rubicon.eu
```
MichaelKetting, line 1: Please add this to PackageInfo.ps1 (see other embedded sources) and to dependency docs in wiki

Remotion/Development/UnitTesting.Compilation/Roslyn/Util/IRoslynProviderOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// Source code from https://github.com/aspnet/RoslynCodeDomProvider
+// Checkout commit:  213c6bdf00e7f56557024e3b2b1287f8529c7155
+
+using System;
+using System.Collections.Generic;
+
+namespace Remotion.Development.UnitTesting.Compilation.Roslyn.Util {
```
MichaelKetting, line 30: Since we touch the namespace anyway, I would suggest making this "Utilities" or just keep them in the root?

Remotion/Development/UnitTesting.Compilation/Roslyn/CSharpRoslynCodeProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.CodeDom.Compiler;
+using System.ComponentModel;
+using Remotion.Development.UnitTesting.Compilation.Roslyn.Util;
+
+namespace Remotion.Development.UnitTesting.Compilation.Roslyn {
+    /// <summary>
+    /// Provides access to instances of the .NET Compiler Platform C# code generator and code compiler.
+    /// </summary>
+    [DesignerCategory("code")]
+    public sealed class CSharpRoslynCodeProvider : Microsoft.CSharp.CSharpCodeProvider {
```
MichaelKetting, line 39: do we need the Roslyn types to be public? If not, we should make them internal because the embedded MS code is not something we should expose. If we need to update, etc, we inherit breaking changes. We are also more limited if we find a better way to do this in the future (e.g. MS provides package for this)

Remotion/Development/UnitTesting.Compilation/Roslyn/Util/RoslynCompilationUtil.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        internal static bool IsDebuggerAttached
+        {
+            get {
+                return IsDebuggerPresent() || Debugger.IsAttached;
+            }
+        }
+
+        [DllImport("kernel32.dll")]
```
MichaelKetting, line 184: Meh, not something we fix now, but in the future this might bite us if we do linux support. Of course, then we also need a Linux-based roslyn or just can't do those tests in Linux...

Remotion/Development/UnitTesting.Compilation/Roslyn/CSharpRoslynCodeProvider.cs
MichaelKetting, line 1: Please remind me: that's the least-code version to do this, correct? Aside from stripping out lines of code, we can't remove any of those Roslyn files.

Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyLoading/ApplicationAssemblyLoaderFilterTest.cs
```diff
@@ -129,5 +115,24 @@ public void ApplicationAssemblyInclusion_DependsOnAttribute ()
           FileUtility.DeleteAndWaitForCompletion(compiledAssemblyPath);
       }
     }
+
+    static void Test (string[] args)
```
MichaelKetting, line 119:
```suggestion
    private static void Test (string[] args)
```

Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
```diff
@@ -260,16 +375,14 @@ private void ExecuteInSeparateAppDomain (CrossAppDomainDelegate test)
 
     private void InitializeDynamicDirectory ()
     {
-      _dynamicDirectoryBuildOutputManager.CopyAllGeneratedAssembliesToNewDirectory(
 #if NETFRAMEWORK
-          AppDomain.CurrentDomain.DynamicDirectory
+      _dynamicDirectoryBuildOutputManager.CopyAllGeneratedAssembliesToNewDirectory(AppDomain.CurrentDomain.DynamicDirectory);
 #else
-          null
+      throw new InvalidOperationException("Dynamic directory is not supported under .NET.");
```
MichaelKetting, line 381: Wouldn't thsi be PlatformNotSupportedException?

MichaelKetting (changes requested):

Remotion/Development/UnitTesting.Compilation/Development.UnitTesting.Compilation.csproj
```diff
@@ -16,5 +16,8 @@
   <ItemGroup>
     <PackageReference Include="System.CodeDom" Version="7.0.0" />
```
MichaelKetting, line 17: Is that dependency still needed?

MichaelKetting (changes requested):

Remotion/Development/UnitTesting.Compilation/Roslyn/Util/IRoslynProviderOptions.cs
a-ctor: I would leave them as is as they mirror the original file structure

MichaelKetting: okay... especially if we can make the stuff internal :)

Remotion/Development/UnitTesting.Compilation/Development.UnitTesting.Compilation.csproj
a-ctor: Yes because the Rosyln stuff extends the code DOM provider infrastructure

MichaelKetting (changes requested):
Please check the resolved comments, they seem to be resolved but no changes where integrated?

Please add a 3rd party notices file, see CoreForms.

a-ctor:
@MichaelKetting I think I have addressed all comments. Could you point out the comments you feel are not addressed?

MichaelKetting (changes requested):
History clenaup & the 3rd party notices and we're done.

Remotion/Development/UnitTesting.Compilation/Development.UnitTesting.Compilation.csproj
```diff
@@ -16,5 +16,10 @@
...
   </ItemGroup>
+  <ItemGroup>
+    <None Include="build\*.*" Pack="true" PackagePath="build\%(Filename)%(Extension);buildTransitive\%(Filename)%(Extension)" />
```
MichaelKetting, line 20: oh, something else was missing?

MichaelKetting (approved):
Please review the nightly build before merging to check that the package is correct.

Remotion/Development/UnitTesting.Compilation/Development.UnitTesting.Compilation.csproj
a-ctor: I think that is just an artifact caused by reordering commits

PR #645, opened 2023-02-27
RM-8054 WebButton Icon url is not resolved before render

a-ctor:
https://re-motion.atlassian.net/browse/RM-8054


MichaelKetting (changes requested):
Please check RM-1109. Is there good reasons not to build this based on the "dedicated task" option mentioned in RM-1109 (the answer/descision depends on the cause-effect situation and should be documented in the PR). There is also the missing consideration for the `Icon` css class.

PR #649, opened 2023-03-01
RM-8759 Add an ASP.NET error page detection

a-ctor:
https://re-motion.atlassian.net/browse/RM-8759

### DIFF
--- a/Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
+++ b/Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
@@ -1,0 +1,68 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System.Runtime.Serialization;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Support.Extensions;
+using Remotion.Web.Development.WebTesting.Utilities;
+
+namespace Remotion.Web.Development.WebTesting
+{
+  /// <summary>
+  /// Provides methods for detecting an ASP.NET error page.
+  /// </summary>
+  public static class AspNetErrorPageDetection
+  {
+    [DataContract]
+    private class ErrorObject
+    {
+      [DataMember]
+      public string Message { get; set; } = default!;
+
+      [DataMember]
+      public string Details { get; set; } = default!;
+    }
+
+    private const string c_errorPageDetectionJs = @"
+let result = null;
+// The following CSS selector matches specific parts of the ASP.NET error page, which make it improbable that a non error page will match it
+if (document.querySelector(""body[bgcolor='white']>span:first-child>h1:first-child>hr[color='silver']:last-child"")) {
+  result = {};
+  result.Message = document.querySelector(""body>span>h2"").innerText;
+  result.Details = document.querySelector(""body>font"").innerText;
+}
+return JSON.stringify(result);
+";
+
+    public static void ThrowOnErrorPage (PageObject pageObject)
+    {
+      ThrowOnErrorPage((IWebDriver)pageObject.Driver.Native);
+    }
+
+    public static void ThrowOnErrorPage (IWebDriver webDriver)
+    {
+      var resultJson = webDriver.ExecuteJavaScript<string>(c_errorPageDetectionJs);
+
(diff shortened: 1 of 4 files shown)


Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    private const string c_errorPageDetectionJs = @"
+let result = null;
+// The following CSS selector matches specific parts of the ASP.NET error page, which make it improbable that a non error page will match it
+if (document.querySelector(""body[bgcolor='white']>span:first-child>h1:first-child>hr[color='silver']:last-child"")) {
+  result = {};
+  result.Message = document.querySelector(""body>span>h2"").innerText;
+  result.Details = document.querySelector(""body>font"").innerText;
+}
+return JSON.stringify(result);
```
MichaelKetting, line 47: Formatting for readability.
```suggestion
let result = null;
// The following CSS selector matches specific parts of the ASP.NET error page, which make it improbable that a non error page will match it
if (document.querySelector(""body[bgcolor='white'] > span:first-child > h1:first-child > hr[color='silver']:last-child"")) {
  result = {};
  result.Message = document.querySelector(""body > span > h2"").innerText;
  result.Details = document.querySelector(""body > font"").innerText;
}
return JSON.stringify(result);
```

Stylistic question: I usually use single-quote for Javascript to simplify C# strings and then only do the nested strings with double quotes. 

BTW: span>h1 isn't valid HTML because H1 is a blocklevel element. Fun times...

Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
```diff
@@ -0,0 +1,68 @@
...
+";
+
+    public static void ThrowOnErrorPage (PageObject pageObject)
```
MichaelKetting, line 50: please add argument checks.

Remotion/Web/Development.WebTesting/AspNetErrorPageException.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+  /// <summary>
+  /// Thrown when an ASP.NET error page was detected.
+  /// Additional information rendered onto the error page can be found in <see cref="Details"/>.
+  /// </summary>
+  public class AspNetErrorPageException : Exception
+  {
+    public string Details { get; }
+
+    public AspNetErrorPageException (string message, string details)
```
MichaelKetting, line 29: given this is a specific exception, please add argument checks. We don't plan on creating an exception with a null or empty message/details.

Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+      var resultJson = webDriver.ExecuteJavaScript<string>(c_errorPageDetectionJs);
+
+      // Shortcut the JSON deserialization for the common case to improve performance
+      if (resultJson is null or "null")
+        return;
+
+      var resultObject = DataContractJsonSerializationUtility.Deserialize<ErrorObject>(resultJson);
+      if (resultObject != null)
+        throw new AspNetErrorPageException(resultObject.Message, resultObject.Details);
```
MichaelKetting, line 65: When we add arg-checks here, we might want to do a fallback message and details to indicate that no data was found and thus prevent an exception. We should also do this if there are no arg-checks added.

Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
```diff
@@ -0,0 +1,68 @@
...
+using OpenQA.Selenium;
+using OpenQA.Selenium.Support.Extensions;
+using Remotion.Web.Development.WebTesting.Utilities;
+
+namespace Remotion.Web.Development.WebTesting
+{
+  /// <summary>
+  /// Provides methods for detecting an ASP.NET error page.
+  /// </summary>
+  public static class AspNetErrorPageDetection
```
MichaelKetting, line 27: Since the detection cannot be used with async-postbacks on Smartpage at this time, we should document that this api is only intended for synchronous postbacks or GET-requests at this time. And we should add a ticket to support async-postbacks to be resolved at a later date.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/AspNetErrorPageDetection.cs
a-ctor: I tested errors and asynchronous postbacks, and it just did nothing -> there was no update. Is there even a visible error page that can be detected on a failed asynchronous postback?

MichaelKetting: Yes, if you use the SmartPage or WxePage as a baseclass. The plain page create a console error (or swallows the error by default)

You can repro this by opening the test page for BocAutoComplete, and in the first BocAutoComplete (has an auto-postback) select a value without an icon (those generate exceptions). Or you build a testpage off SmartPage :)

Remotion/Web/Development.WebTesting/AspNetErrorPageException.cs
```diff
@@ -0,0 +1,39 @@
...
+        : base(message)
+    {
+      ArgumentUtility.CheckNotNull("message", message))
+      ArgumentUtility.CheckNotNull("details", details);
```
MichaelKetting, line 34: Here and at the usage site: I feel that we should use NotNullOrEmpty instead of just NotNull. these values are equivalent in their meaninglessness and should thus be handled the same. THoughts?

a-ctor:
Closing since the scope of the issue has changed as we already have an existing infrastructure that was missed before.

PR #651, opened 2023-03-06
RM-8759 Improve the ASP.NET error page detection

a-ctor:
https://re-motion.atlassian.net/browse/RM-8763

### DIFF
--- a/Remotion/Web/Development.WebTesting/RequestErrorDetectionStrategies/AspNetRequestErrorDetectionParser.cs
+++ b/Remotion/Web/Development.WebTesting/RequestErrorDetectionStrategies/AspNetRequestErrorDetectionParser.cs
@@ -32,16 +32,6 @@ namespace Remotion.Web.Development.WebTesting.RequestErrorDetectionStrategies
   /// </summary>
   public sealed class AspNetRequestErrorDetectionParser
   {
-    [DataContract]
-    private class ErrorObject
-    {
-      [DataMember]
-      public string? Message { get; set; }
-
-      [DataMember]
-      public string? StackTrace { get; set; }
-    }
-
     public sealed class Result
     {
       public static Result CreateErrorResult ([NotNull] string message, [NotNull] string stacktrace)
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting/RequestErrorDetectionStrategies/AspNetRequestErrorDetectionParser.cs
```diff
@@ -27,6 +32,16 @@ namespace Remotion.Web.Development.WebTesting.RequestErrorDetectionStrategies
   /// </summary>
   public sealed class AspNetRequestErrorDetectionParser
   {
+    [DataContract]
```
MichaelKetting, line 35: Please move the ErrorObject after the public nested type to follow visibility practices.

Remotion/Web/Development.WebTesting/RequestErrorDetectionStrategies/AspNetRequestErrorDetectionParser.cs
```diff
@@ -71,33 +86,54 @@ public string Stacktrace
...
+function fixNewlines(text) {
+  // Ensure strings have a CRLF endings
+  return text.replaceAll(/(?<!\r)\n/g, '\r\n');
+}
 
-      var pageRequestManagerException = scope.FindCss(".SmartPageErrorBody:first-child", Options.NoWait);
+let result = null;
 
-      var startSelector = GetRootSelectorForErrorInformation(pageRequestManagerException);
+const smartPageErrorBody = document.querySelector('.SmartPageErrorBody:first-child');
```
MichaelKetting, line 97: considering what we discovered about performance, we can improve this initial detection:
`body > div#SmartPageServerErrorMessage` first and then drill into `:scope .SmartPageErrorBody:first-child > div` to get the `errorPageTarget` for async requests.

Please verify to make sure I'm right about that :)

MichaelKetting (changes requested):
some minor optimization requests

Remotion/Web/Development.WebTesting/RequestErrorDetectionStrategies/AspNetRequestErrorDetectionParser.cs
a-ctor: Right now the detection takes 2ms, which includes the JS execution + JSON deserialization. I don't think it is worth trying to optimize anymore as it might change the existing behavior.

MichaelKetting: Did you measure the 2ms in async, too? Also with a big (deep) page or just demo pages. Thing is, when there's a ton of stuff on the page, things might look different. Anyway, the updated version would be much more precise/safer actually since we detect & capture based on the root element and only then select what we need.

MichaelKetting (approved):

PR #653, opened 2023-03-07
RM-8763 Automatically check for ASP.NET error pages on page object creation

a-ctor:
https://re-motion.atlassian.net/browse/RM-8763

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/WxeCompletionDetectionHelpersTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/WxeCompletionDetectionHelpersTest.cs
@@ -82,7 +82,13 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
 
     private WxePageObject Start ()
     {
-      return Start<WxePageObject>("AspNetRequestErrorDetectionParserStaticPages/CustomErrorDefaultErrorPage.html");
+      // We cannot start the correct page directly as it would throw because of the error detection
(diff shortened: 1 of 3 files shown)


MichaelKetting (approved):
Okay, so we actually had a missing case in the production code, not just the generic tests infrastructure?

Remotion/Web/Development.WebTesting.IntegrationTests/WxeCompletionDetectionHelpersTest.cs
```diff
@@ -82,7 +82,13 @@ public void WxeCompletionDetectionHelpers_GetWxeFunctionToken_CallsRequestErrorD
...
     private WxePageObject Start ()
     {
-      return Start<WxePageObject>("AspNetRequestErrorDetectionParserStaticPages/CustomErrorDefaultErrorPage.html");
+      // We cannot start the correct page directly as it would throw because of the error detection
+      // So instead, we start an empty page, navigate to the actual page and return the empty-page
+      // page object to ensure that the error detection is correctly set in the context
+      var pageObject = Start<WxePageObject>("Empty.aspx");
+
+      StartWithoutRequestErrorDetection<WxePageObject>("AspNetRequestErrorDetectionParserStaticPages/CustomErrorDefaultErrorPage.html");
+      return pageObject;
```
MichaelKetting, line 91: We should add in comment that we are doing this only because we want the error-dectection in there, too. Otherwise, we could just use / return `StartWithoutRequestErrorDetection`

MichaelKetting (approved):
Please update code comments a bit, see review notes.

PR #659, opened 2023-03-16
Feature/rm 8600 update axe version to 3.x (latest minor)

Frazerus:
https://re-motion.atlassian.net/browse/RM-8600


MichaelKetting (commented):
waiting for update

Frazerus:
https://dequeuniversity.com/rules/axe/3.4/label
This changed from 3.3.2 to 4.1.2 it seems, I do not know exactly why. 

https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html
https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html

MichaelKetting:
> https://dequeuniversity.com/rules/axe/3.4/label This changed from 3.3.2 to 4.1.2 it seems, I do not know exactly why.
> 
> https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html

@Frazerus Indeed. The description indicates more that it's 3.3.4, not 4.1.2. 
https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html
https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html

Since the test itself only focuses on testing whether the contract works, not that each rule is "correct", I think we don't need to investigate this. On a personal note, this particular change (since we noticed it) does seem interesting. If you want to, it would be graet if you could ceck out the axe repository if you can quickly identify the commit where this was changed and the associated ticket (or vice versa).

MichaelKetting (approved):

Frazerus:
Located the issue in question, but no reasoning as to why they changed it other than that it should be changed...
https://github.com/dequelabs/axe-core/issues/2014

PR #669, opened 2023-03-24
RM-8773 FillWithAction uses separate calls to SendKeys with GeckoDriver

Frazerus:
https://re-motion.atlassian.net/browse/RM-8773

### DIFF
--- a/Remotion/Web/Development.WebTesting/CoypuElementScopeFillInWithAndSendKeysExtensions.cs
+++ b/Remotion/Web/Development.WebTesting/CoypuElementScopeFillInWithAndSendKeysExtensions.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Linq;
 using Coypu;
+using Coypu.Drivers;
 using JetBrains.Annotations;
 using log4net;
 using OpenQA.Selenium;
@@ -105,11 +106,26 @@ namespace Remotion.Web.Development.WebTesting
       value = value.Replace("\r", "");
 
       var clearTextBoxWithoutTriggeringPostBack = Keys.Control + "a" + Keys.Control + Keys.Delete;
-      value = clearTextBoxWithoutTriggeringPostBack + value;
 
-      s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}'.", value);
+      if (scope.Browser.IsFirefox())
+      {
+        // GeckoDriver cannot handle inputting the value after receiving the delete command, if both were sent in the same SendKeys.
+        // This occurs after Firefox v99 and would lead to an empty field.
+        s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}' followed by '{1}'.", clearTextBoxWithoutTriggeringPostBack, value);
 
-      scope.SendKeys(value);
+        scope.SendKeys(clearTextBoxWithoutTriggeringPostBack);
+        scope.SendKeys(value);
+      }
+      else
+      {
+        // The Firefox code has been updated recently (March 2023) and should work for other browsers as well.
+        // However, due to previous problems concerning the insertion of values, we leave the Chromium-based implementation unchanged to avoid the risk of regressions.
+        value = clearTextBoxWithoutTriggeringPostBack + value;
+
+        s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}'.", value);
+
+        scope.SendKeys(value);
+      }
     }
 
     private static string GetFillInJavaScriptCommand ([NotNull] string id)


Remotion/Web/Development.WebTesting/CoypuElementScopeFillInWithAndSendKeysExtensions.cs
```diff
@@ -105,11 +105,24 @@ private static void SetValueUsingSendKeys ([NotNull] this ElementScope scope, [N
       value = value.Replace("\r", "");
 
       var clearTextBoxWithoutTriggeringPostBack = Keys.Control + "a" + Keys.Control + Keys.Delete;
-      value = clearTextBoxWithoutTriggeringPostBack + value;
 
-      s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}'.", value);
+      //GeckoDriver cannot handle the delete command and inputting the value in the same SendKeys after Firefox v99 (would lead to empty field).
+      if (scope.Browser == Browser.Firefox)
```
MichaelKetting, line 110: Please use BrowserConfigurationExtensions instead
```suggestion
      if (scope.Browser.IsFirefox())
```
That way we avoid logic duplication.
(Disclaimer: unless there's technical reason against this?)

Remotion/Web/Development.WebTesting/CoypuElementScopeFillInWithAndSendKeysExtensions.cs
```diff
@@ -105,11 +105,24 @@ private static void SetValueUsingSendKeys ([NotNull] this ElementScope scope, [N
...
+      {
+        s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}'.", clearTextBoxWithoutTriggeringPostBack);
+        s_log.DebugFormat("FillInWith without triggering PostBack on inputting value: '{0}'.", value);
```
MichaelKetting, line 113: I think we should combine this into one message. 
"FillInWith without triggering PostBack on clear: '{0}' followed by '{1}'."

Reasoning: we have two operations that are meant to work together and you need to either move the log-statements interleaved with the sendkey operations and update the text to be a bit more aware of the split operation or we just make it one message and acknowldge that this is just for firefix

MichaelKetting (changes requested):
Some stylistic stuff.

Frazerus:
Another thought about this whole ordeal:
In my opinion, the sending of the delete operation should be a separate call to the API from when the value is inserted, _in all browsers_. This would be cleaner and would prevent such problems altogether if they would occur in the future in another browser. 

The main reason I even looked into the direction of separating the API calls was because adding the delete sequence and the value seemed off to me. 

Your thoughts?

Remotion/Web/Development.WebTesting/CoypuElementScopeFillInWithAndSendKeysExtensions.cs
```diff
@@ -105,11 +106,26 @@ private static void SetValueUsingSendKeys ([NotNull] this ElementScope scope, [N
...
       var clearTextBoxWithoutTriggeringPostBack = Keys.Control + "a" + Keys.Control + Keys.Delete;
-      value = clearTextBoxWithoutTriggeringPostBack + value;
 
-      s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}'.", value);
-
-      scope.SendKeys(value);
+      if (scope.Browser.IsFirefox())
+      {
+        //GeckoDriver cannot handle inputting the value after receiving the delete command, IF both were sent in the same SendKeys.
+        //This occurs after Firefox v99 and would lead to an empty field.
```
MichaelKetting, line 113:
```suggestion
        // GeckoDriver cannot handle inputting the value after receiving the delete command, iF both were sent in the same SendKeys.
        // This occurs after Firefox v99 and would lead to an empty field.
```

Remotion/Web/Development.WebTesting/CoypuElementScopeFillInWithAndSendKeysExtensions.cs
```diff
@@ -105,11 +106,26 @@ private static void SetValueUsingSendKeys ([NotNull] this ElementScope scope, [N
...
+        s_log.DebugFormat("FillInWith without triggering PostBack on clear: '{0}' followed by '{1}'.", clearTextBoxWithoutTriggeringPostBack, value);
+
+        scope.SendKeys(clearTextBoxWithoutTriggeringPostBack);
+        scope.SendKeys(value);
+      }
+      else
+      {
+        //The firefox code should work for other browsers as well.
+        //BUT due to previous problems concerning insertion of values, we leave the chromium based implementation alone.
```
MichaelKetting, line 122:
```suggestion
        // The Firefox code has been updated recently (March 2023) and should work for other browsers as well.
        // However, due to previous problems concerning the insertion of values, we leave the Chromium-based implementation unchanged to avoid the risk of regressions.
```

MichaelKetting (approved):
Some text updated to the comments. Please integrate and we're done.

PR #671, opened 2023-03-31
RM-8774 NovaViso ReadoOnly BocText and BocReferenceValue enforce line…

Frazerus:
…breaks if there is not enough space

https://re-motion.atlassian.net/browse/RM-8774

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
@@ -42,6 +42,8 @@ span.bocTextValue textarea
   padding: var(--remotion-themed-control-padding-readonly);
   border: var(--remotion-themed-border-readonly);
   border-radius: var(--remotion-themed-border-radius);
+  overflow-wrap: anywhere;
+  word-break: normal;
 }
 
 .bocTextValue.readOnly .content input:read-only:focus-visible + span,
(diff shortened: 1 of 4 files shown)


MichaelKetting (changes requested):
BocAutoCompleteReferenceValue and BocMultilineTextValue don't have changed files in the PR. I didn't verify it yet and my gut feeling says: "something's missing". Please check/confirm.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -21,6 +21,8 @@ span.bocTextValue textarea
   display: inline-flex;
   align-items: center;
   width: fit-content;
+  overflow-wrap: anywhere;
+  word-break: normal;
```
MichaelKetting, line 25: I recommend moving this down to line 37-ish, to the section where the input+span / textarea+span styles are set. Makes it more consistent with the rest of the styles

MichaelKetting (changes requested):
see comment

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
```diff
@@ -21,6 +21,7 @@ span.bocTextValue textarea
   display: inline-flex;
   align-items: center;
   width: fit-content;
+
```
MichaelKetting, line 24:
```suggestion
```

MichaelKetting (approved):
Looks correct. foudn a new ticket triggered by this change, though it's not really caused by it:
image
Could you please create a bug ticket that the drop down menu in referecnevalue/autocompetereferencevalue isn't top aligned when the value linewrapped. this is incorrect in NovaViso.

Alternatively, we could look into fixing it in this PR, not sure if that's best, though. I leave it up to you.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocTextValue.css
MichaelKetting: Altrernatively, we could look into fixign it in here, not sure if that's best, though. I leave it up to you.

PR #674, opened 2023-04-12
feature/RM-8639-Initialize-non-nullable-reference-type-properties-in-DomainObjects-with-null

martin-herwich:
https://re-motion.atlassian.net/browse/RM-8639


MichaelKetting (commented):
Partial review done. More to follow.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/PropertyDefaultValueProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+  {
+    public object? GetDefaultValue (IPropertyInformation propertyInfo, bool isNullable)
+    {
+      if (isNullable)
+        return null;
+
+      var propertyType = propertyInfo.PropertyType;
+      if (propertyType.IsEnum)
+      {
+        var firstValueOrNull = EnumUtility.GetEnumMetadata(propertyType).OrderedValues.FirstOrDefault();
```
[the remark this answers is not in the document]
martin-herwich: Since this issue is only about reference types, the semantics for enum properties should not change. However, because the DefaultValue is now created when instantiating the PropertyDefinition instead of the first access to the DefaultValue property, it is important that we do not throw exceptions, so that the mapping validation can find the error. This is why even the LegacyPropertyDefaultValueProvider creates a zero value if no enum value is defined.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+        var firstValueOrNull = EnumUtility.GetEnumMetadata(propertyType).OrderedValues.FirstOrDefault();
+        if (firstValueOrNull == null)
+          firstValueOrNull = Activator.CreateInstance(propertyType);
+
+        return firstValueOrNull;
+      }
+
+      if (ExtensibleEnumUtility.IsExtensibleEnumType(propertyType))
+      {
+        // TODO if there are no values, should be return a dummy value?
```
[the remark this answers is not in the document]
martin-herwich: We have to decide whether to return `null ` or make up a value. The previous implementation of `PropertyDefinition.DefaultValue` threw an exception, which we should not do (see comment on https://github.com/re-motion/Framework/pull/674/files/69a0a3e16236e5b720a96313306a1bcdfbd70cf1#diff-758b70ed37a08eceb6a718ee93ead95d66319b5e419730933d29554173d4b27d)

While it is possible to return `null` here, it seems more consistent with the legacy behavior to create a dummy value. In the grand scheme of things it should not matter because the mapping validation will throw an error.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/PropertyDefaultValueProvider.cs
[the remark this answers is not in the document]
MichaelKetting: I see. Follow up: do you think we should update the Enum in a separate ticket now that we are touching things here anyway and update the PropertyDefaultValueProvider to use the 0-value for Enum to better represent the .net default / unitialized backing field behavior?

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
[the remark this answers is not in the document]
MichaelKetting: Good. So, then the decision is to use a dummy value based on the IExtensibleEnum interface. I think we have a name suggestion for that somewhere in the ticket?

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/PropertyDefaultValueProvider.cs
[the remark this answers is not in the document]
martin-herwich: I lean towards yes we should do it, but no we should not create a different issue. This is because the behavior here is almost irrelevant because of the mapping validation. We can re-write the ticket to better reflect what we now know about the interaction between the PropertyDefaultValueProvider and the mapping validation.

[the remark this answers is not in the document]
MichaelKetting: that's fine by me. I agree that the scope change is negligible in terms of the budget. we do should keep it as a separate commit though.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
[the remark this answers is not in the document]
MichaelKetting: Ah, because we have a typeguard in there. I.C. 

>We could use FormatterServices.GetSafeUninitializedObject(Type type) like we do for DomainObjects, since ExtensibleEnum<T> is [Serializable], and then use PopulateObjectMembers() to set ValueName to "Uninitialized".
Sounds good to me.

> Or we could have a repository of those objects within the LegacyPropertyDefaultValueProvider, because we only intend to use it there.
I prefer that one.

MichaelKetting (changes requested):
Review almost done. (the PropertyDefaultValue-Test fixtures are left)

Remotion/Data/DomainObjects.UnitTests/Mapping/PropertyDefaultValueProviderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    [Test]
+    public void DefaultValue_NotNullable_ExtensibleEnumWithoutValues ()
+    {
+      _propertyInformationStub.Setup(stub => stub.PropertyType).Returns(typeof(ExtensibleEnumNotDefiningAnyValues));
+      var defaultValue = _defaultValueProvider.GetDefaultValue(_propertyInformationStub.Object, false);
+      Assert.That(defaultValue, Is.Null);
+    }
+
+    [Test]
+    public void DefaultValue_NotNullable_OtherType ()
```
MichaelKetting, line 93:
```suggestion
    public void DefaultValue_ValueType ()
```

Remotion/Data/DomainObjects.UnitTests/Mapping/LegacyPropertyDefaultValueProviderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    [Test]
+    public void DefaultValue_NotNullable_ExtensibleEnumWithoutValues ()
+    {
+      _propertyInformationStub.Setup(stub => stub.PropertyType).Returns(typeof(ExtensibleEnumNotDefiningAnyValues));
+      var defaultValue = _defaultValueProvider.GetDefaultValue(_propertyInformationStub.Object, false);
+      Assert.That(defaultValue, Is.Null);
+    }
+
+    [Test]
+    public void DefaultValue_NotNullable_OtherType ()
```
MichaelKetting, line 93:
```suggestion
    public void DefaultValue_ValueType ()
```

Remotion/Data/DomainObjects.UnitTests/Mapping/LegacyPropertyDefaultValueProviderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+      var defaultValue = _defaultValueProvider.GetDefaultValue(_propertyInformationStub.Object, false);
+      Assert.That(defaultValue, Is.EqualTo(Color.Values.Blue()));
+    }
+
+    [Test]
+    public void DefaultValue_NotNullable_ExtensibleEnumWithoutValues ()
+    {
+      _propertyInformationStub.Setup(stub => stub.PropertyType).Returns(typeof(ExtensibleEnumNotDefiningAnyValues));
+      var defaultValue = _defaultValueProvider.GetDefaultValue(_propertyInformationStub.Object, false);
+      Assert.That(defaultValue, Is.Null);
```
MichaelKetting, line 89: This will check the dummy enum values, unlesss I'm mistaken.

MichaelKetting (commented):
now I'm "done".

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/IPropertyDefaultValueProvider.cs
```diff
@@ -0,0 +1,16 @@
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Reflection;
+
+namespace Remotion.Data.DomainObjects.ConfigurationLoader.ReflectionBasedConfigurationLoader
+{
+  /// <summary>
+  /// Provides the default value returned by <see cref="PropertyDefinition"/>. <see cref="PropertyDefinition.DefaultValue"/>.
+  /// </summary>
+  public interface IPropertyDefaultValueProvider
```
MichaelKetting, line 9: Oops, the unit test for the IoC setup is missing.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/PropertyDefaultValueProvider.cs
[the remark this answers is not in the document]
martin-herwich: For the record, we agreed upon not including this in the scope of this issue, because changing the behavior to initializing an invalid value would necessitate that we get rid of the value check when setting the property value and some other changes. This should be done in a separate issue.

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Unload/UnloadDataTest.cs
```diff
@@ -129,7 +129,12 @@ public void UnloadData_Computer_WithUnsynchronizedOppositeEndPoint ()
       employee.EnsureDataAvailable();
 
       var unsynchronizedComputerID =
-          RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction<Computer, Employee>(employee.ID, (ot, o) => ot.Employee = o);
+          RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction<Computer, Employee>(employee.ID, (ot, o) =>
+          {
+            ot.SerialNumber = "12345";
+            o.Name = "Employee";
```
MichaelKetting, line 135: please move those to lines 127ff.
This delegate is just for the relation assignment and we shouldn't muddle the waters to keep the intended effects easily understandable.

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/VirtualEndPointGarbageCollectionTest.cs
```diff
@@ -72,7 +72,12 @@ public void UnloadUnsynchronizedFK_LeavesCompleteEmptyCollection ()
       var unsynchronizedSubordinateID =
           RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction<Employee, Employee>(
             employee.ID,
-            (subOrdinate, e) => subOrdinate.Supervisor = e);
```
MichaelKetting, line 75: marked file (please move property initialization up to object load/initialization)

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/SyncStateTest.cs
```diff
@@ -52,7 +52,11 @@ public void CollectionItems_Unsynchronized_WithUnload ()
       var order = DomainObjectIDs.Order1.GetObject<Order>();
       order.OrderItems.EnsureDataComplete();
 
-      var orderItemID = RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction<OrderItem, Order>(order.ID, (oi, o) => oi.Order = o);
```
MichaelKetting, line 55: marked file (please move property initialization up to object load/initialization)

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/RelationInconsistenciesTestBase.cs
```diff
@@ -43,7 +46,11 @@ protected void SetIndustrialSectorInOtherTransaction (ObjectID companyID, Object
 
     protected ObjectID CreateComputerAndSetEmployeeInOtherTransaction (ObjectID employeeID)
     {
-      return RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction<Computer, Employee>(employeeID, (c, e) => c.Employee = e);
+      return RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction<Computer, Employee>(employeeID, (c, e) =>
```
MichaelKetting, line 49: marked file (please move property initialization up to object load/initialization)

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+  /// <summary>
+  /// Implements the legacy behavior for <see cref="PropertyDefinition.DefaultValue"/>, which sets non-nullable reference-<see cref="Type"/> properties to
+  /// non-<see langword="null"/> values. This affects all <see cref="Array"/>, <see cref="string"/>, and <see cref="ExtensibleEnum{T}"/> properties.
+  /// </summary>
+  /// <threadsafety static="true" instance="true" />
+  public class LegacyPropertyDefaultValueProvider : IPropertyDefaultValueProvider
+  {
+    private static readonly ConcurrentDictionary<Type, IExtensibleEnum> s_uninitializedValues = new();
+    private static readonly object?[] s_extensibleEnumValues = { "Uninitialized" };
```
MichaelKetting, line 20: Are we still missing the mapping validation rules, or is the following pre-existing test actually covering this scenario and if so, how was this doable before our rework?
see Remotion.Data.DomainObjects.UnitTests.Mapping.Validation.Logical.MandatoryExtensibleEnumTypeHasValuesDefinedValidationRuleTest
and NotSupportedPropertyTypesIntegrationTests.PropertyTypeOfExtensibleEnum_WithoutValues_Mandatory

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      return propertyType.IsValueType ? Activator.CreateInstance(propertyType) : null;
+    }
+
+    private IExtensibleEnum CreateUninitializedExtensibleEnumValue (Type extensibleEnumType)
+    {
+      var uninitializedObject = FormatterServices.GetSafeUninitializedObject(extensibleEnumType);
+      var members = FormatterServices.GetSerializableMembers(extensibleEnumType).Where(m=>m.Name.Contains(nameof(IExtensibleEnum.ValueName))).ToArray();
+      FormatterServices.PopulateObjectMembers(uninitializedObject, members, s_extensibleEnumValues);
```
MichaelKetting, line 60: I'm not sure about the impelmentation here. 
a) there's no guarantee that ExtensibleEnum is implemented as serializable (don't know if that's requred). Anyway, I would suggest using Type.GetFields instead.
b) There's no guarantee that the fieldname matches. I would suggest we simply go and initalize all string fields on the enum.

MichaelKetting (changes requested):

Remotion/Development/Core/Development.Core.csproj
```diff
@@ -12,12 +12,11 @@
...
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\..\..\SharedSource\Development\UnitTesting\**\*"
-             Link="SharedSource\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)"
-             Exclude="**\PEVerifyVersion.cs;**\NUnit\**" />
+    <Compile Include="..\..\..\SharedSource\Development\UnitTesting\**\*" Link="SharedSource\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)" Exclude="**\PEVerifyVersion.cs;**\NUnit\**" />
     <Compile Include="..\..\..\SharedSource\Core\RoslynAnnotations\NullableAnnotations.cs" Link="SharedSource\RoslynAnnotations\NullableAnnotations.cs" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\Core\ExtensibleEnums\Core.ExtensibleEnums.csproj" />
```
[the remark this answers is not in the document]
martin-herwich: I originally had the ExtensibleEnumObjectMother in Development.Core next to BooleanObjectMother, which necessitated this new reference; this is no longer the case.

Remotion/Data/DomainObjects.UnitTests/ObjectMothers/ExtensibleEnumObjectMother.cs
```diff
@@ -0,0 +1,27 @@
...
+  {
+    private static readonly Random s_random = new Random();
+
+    /// <summary>
+    /// Gets a random <see cref="ExtensibleEnum{T}"/> value for the given <typeparamref name="T"/>. This is used by unit tests when they need
+    /// code to work with arbitrary extensible enum values. Rather than duplicating the test, once for each value, the test is written once and is executed 
+    /// any value chosen at random.
+    /// </summary>
+    /// <returns>A random <typeparamref name="T"/> value.</returns>
+    public static T GetRandomValue<T> ()
```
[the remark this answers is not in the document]
martin-herwich: Our `TestExtensibleEnum` only has two values like `bool`, but I get that it is not the same situation. I will use `Value2()`, since that is not the value provided by any `IPropertyDefaultValueProvider`.

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/VirtualEndPointGarbageCollectionTest.cs
martin-herwich: I can move the initialization of the existing object out of this delegate, but since `CreateObjectAndSetRelationInOtherTransaction` creates an object generically, I have to initialize that object in a delegate given to the generic method.
Can I use this delegate, should I create a separate one, or do we rework the whole approach to DomainObject creation in the tests?

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/SyncStateTest.cs
martin-herwich: Same problem as above: I cannot set `OrderItem.Product` before creating the `OrderItem`, which is done in `RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction()`.

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/RelationInconsistenciesTestBase.cs
martin-herwich: Same problem as above: I cannot set properties before creating the respective `DomainObject`, which is done in RelationInconcsistenciesTestHelper.CreateObjectAndSetRelationInOtherTransaction().

MichaelKetting: yep, same answer, too :)

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
martin-herwich: The mapping validation was always there, so the existing tests cover that it works.
Before extracting `LegacyPropertyDefaultValueProvider`, a call to `PropertyDefinition.DefaultValue` would throw an exception if there were no defined values; this was okay because calls to that property would have only occurred after the mapping validation had already found the problem and thrown a better exception.
Since we now inject the default value into `PropertyDefinition` at creation, that timing has changed; we no longer want to throw an exception, because that would preempt the mapping validation.

MichaelKetting: Good, same conclusion I had. Original validation covers everything, we just needed to prevent the exception during initialization of the object graph. No extra work. all good.

Remotion/Data/DomainObjects/ConfigurationLoader/ReflectionBasedConfigurationLoader/LegacyPropertyDefaultValueProvider.cs
martin-herwich: How do I initialize all string fields if I don't know what they mean? I only know `ValueName` and `DeclarationSpace`, because those are defined on `ExtensibleEnum<T>`.
We might be overthinking this, since this "uninitialized" value only exists so we don't have to throw an exception here, but let the mapping validation handle it.

MichaelKetting: The problem is, we can't make sure we identify the correct fields, it's just a heuristic that could be correct. Since we don't use the enum object for more than a dummy value and have the system fail via validation anyway, we can also just skip the entire exercise and leave the object with null-values in their fields. less code, same result :)

martin-herwich: Because `ValueName` is not nullable, I thought we cannot leave it at `null`, but now that you mention it - why not?
Do you think the `ConcurrentDictionary` is valuable, or should we get rid of that as well?

MichaelKetting: ad 1) ValueName: exactly, no reason to not leave it null
ad 2) ConDict: Since we don't need to test/detect them, I think we can drop it.

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/RelationInconsistenciesTestBase.cs
```diff
@@ -118,12 +131,13 @@ protected IndustrialSector CreateIndustrialSectorInDatabaseAndLoad ()
       return objectID.GetObject<IndustrialSector>();
     }
 
-    protected T CreateObjectInDatabaseAndLoad<T> () where T : DomainObject
+    protected T CreateObjectInDatabaseAndLoad<T> ([CanBeNull] Action<T> initialize = null) where T : DomainObject
```
MichaelKetting, line 134: The CanBeNull feels like it should be redundant? It's more a curiosity though, than a change-this.

Remotion/Development/Core/Development.Core.csproj
```diff
@@ -12,9 +12,7 @@
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\..\..\SharedSource\Development\UnitTesting\**\*"
-             Link="SharedSource\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)"
-             Exclude="**\PEVerifyVersion.cs;**\NUnit\**" />
+    <Compile Include="..\..\..\SharedSource\Development\UnitTesting\**\*" Link="SharedSource\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)" Exclude="**\PEVerifyVersion.cs;**\NUnit\**" />
```
MichaelKetting, line 15: please revert

MichaelKetting (approved):

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Synchronization/RelationInconsistenciesTestBase.cs
martin-herwich: I tried to make the parameter nullable, but nullable reference types are disabled in the project, so it is `[CanBeNull]` instead. Rider does not mark it as redundant, even though it is clear that a `null` default must mean that the parameter can also be `null`. My guess is that Rider/Resharper does not infer `[CanBeNull]` from the presence of a default value.

PR #675, opened 2023-04-13
RM-8780 DropDownMenu in BocReferenceValue and BocAutoCompleteReferenc…

Frazerus:
…eValue is now always top aligned
https://re-motion.atlassian.net/browse/RM-8780


MichaelKetting (approved):
I like it. I wasn't able to break this solution while testing it in the lab :)

PR #677, opened 2023-04-14
RM-8776 Fix DropDownMenu ButtonType rendering when custom css is set

a-ctor:
https://re-motion.atlassian.net/browse/RM-8776


MichaelKetting (changes requested):
Update to implementation using BocRendererBase as a template as disussed

PR #678, opened 2023-04-18
Feature/rm 8369 retrieve boc list column definitions

a-ctor:
https://re-motion.atlassian.net/browse/RM-8369

### DIFF
--- a/Remotion/Core/Core/Collections/ReadOnlyListExtensions.cs
+++ b/Remotion/Core/Core/Collections/ReadOnlyListExtensions.cs
@@ -1,0 +1,57 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using Remotion.Utilities;
+
+namespace Remotion.Collections
+{
+  /// <summary>
+  /// Provides useful extension methods for <see cref="IReadOnlyList{T}"/>.
+  /// </summary>
+  public static class ReadOnlyListExtensions
+  {
+    /// <summary>
+    /// Searches for the specified <paramref name="value"/> and returns the index of its
+    /// first occurence in <paramref name="list"/>.
+    /// </summary>
+    /// <param name="list">The list to search <paramref name="value"/> in. Must not be <see langword="null"/>.</param>
+    /// <param name="value">The value to search for.</param>
+    /// <returns>The index of the first occurence of <paramref name="value"/> in <paramref name="list"/>, or <c>-1</c> if not found.</returns>
+    public static int IndexOf<T> (this IReadOnlyList<T> list, T value)
+    {
+      ArgumentUtility.CheckNotNull("list", list);
+
+      for (var i = 0; i < list.Count; i++)
+      {
+        var item = list[i];
+        if (item == null)
+        {
+          if (value == null)
+            return i;
+        }
+        else
+        {
+          if (item.Equals(value))
+            return i;
+        }
+      }
+
+      return -1;
+    }
+  }
+}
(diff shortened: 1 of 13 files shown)


MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -1886,7 +1902,9 @@ public BocListSortingOrderEntry[] GetSortingOrder ()
             }
             else
             {
-              var newIndex = Array.IndexOf(columns, entry.Column);
+              var newIndex = entry.Column is BocColumnDefinition bocColumnDefinition
+                ? columns.IndexOf(bocColumnDefinition)
+                : -1;
```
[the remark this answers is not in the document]
a-ctor: I have change all the usages from `ReadOnlyCollection<T>` to `IReadOnlyList<T>` and introduced an extension method for the `IndexOf<T>`. I would keep the existing `is BocColumnDefinition` pattern over the hardcast because `column.Entry` is nullable. Using `is` will prevent an unnecessary `IndexOf` in the case of `null`.

Remotion/Core/Core/Collections/ReadOnlyListExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+using Remotion.Utilities;
+
+namespace Remotion.Collections
+{
+  /// <summary>
+  /// Provides useful extension methods for <see cref="IReadOnlyList{T}"/>.
+  /// </summary>
+  public static class ReadOnlyListExtensions
```
MichaelKetting, line 9: Unit tests, please.

Remotion/Core/Core/Collections/ReadOnlyListExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+  {
+    public static int IndexOf<T> (this IReadOnlyList<T> list, T value)
```
MichaelKetting, line 11: XML DOC please

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -1886,7 +1903,9 @@ public BocListSortingOrderEntry[] GetSortingOrder ()
             }
             else
             {
-              var newIndex = Array.IndexOf(columns, entry.Column);
+              var newIndex = entry.Column is BocColumnDefinition bocColumnDefinition
```
MichaelKetting, line 1906: given the old semantic, wouldn't this now be `columns.IndexOf((BocColumnDefinition)entry.Column)` ?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -120,7 +122,9 @@ private void PrepareSorting (IDictionary<int, SortingDirection> sortingDirection
 
         foreach (var entry in SortingOrder)
         {
-          var columnIndex = Array.IndexOf(_columnDefinitions, entry.Column);
+          var columnIndex = entry.Column is BocColumnDefinition bocColumnDefinition
```
MichaelKetting, line 125: `var columnIndex = _columnDefinitions.IndexOf((BocColumnDefinition)entry.Column)`

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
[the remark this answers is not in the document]
MichaelKetting: > I have change all the usages from ReadOnlyCollection<T> to IReadOnlyList<T> and introduced an extension method for the IndexOf<T>. I would keep the existing is BocColumnDefinition pattern over the hardcast because column.Entry is nullable. Using is will prevent an unnecessary IndexOf in the case of null.

If you want to do "as" instead of "is" we more cloesly match the old semantics. I wouldn't worry about empty sort entries which are already skipped in this implementation. Down in ArrayBuilder, yeah, it could happen, but I don't think special logic is warrented for that case. To me, the soft cast indicates more a "can be another type" not "can also be null" and I don't want to imply the type semantics.

MichaelKetting (changes requested):

Remotion/Core/Core/Collections/ReadOnlyListExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+  /// <summary>
+  /// Provides useful extension methods for <see cref="IReadOnlyList{T}"/>.
+  /// </summary>
+  public static class ReadOnlyListExtensions
+  {
+    /// <summary>
+    /// Searches for the specified <paramref name="value"/> and returns the index of its
+    /// first occurence in <paramref name="list"/>.
+    /// </summary>
+    /// <param name="list">The list to search <paramref name="value"/> in.</param>
```
MichaelKetting, line 32:
```suggestion
    /// <param name="list">The list to search <paramref name="value"/> in. Must not be <see langword="null" /></param>
```

(we have a live template for that phrase)

MichaelKetting (approved):

PR #681, opened 2023-04-19
RM-8776 Fix DropDownMenu ButtonType rendering when custom css is set (Backport 4.0)

a-ctor:
https://re-motion.atlassian.net/browse/RM-8776

backport of #677


MichaelKetting (approved):

MichaelKetting (changes requested):
Something got lost along the way. The backport to v4.0 was only requeested because of the backport to v3.10. Since we dropped the v3.10 backport, the v4.0 backport is also dropped.

PR #687, opened 2023-04-20
RM-8788 Revert source linking to the old way for .NET Framework to fix problem where sources cannot be retrieved

a-ctor:
https://re-motion.atlassian.net/browse/RM-8788


MichaelKetting (approved):

a-ctor:
@MichaelKetting the build failed for some reason even if the `Condition="..."` element should prevent the target from running on .NET Core. Now I have extracted the source link feature into its own file, which is only included if source linking is needed. I don't know what exactly caused the first version to fail but now it works :D

MichaelKetting (approved):

PR #688, opened 2023-04-20
Bugfix/rm 5437 tabbed multi view empty tab is not removed when tabs are recreated

Frazerus:
https://re-motion.atlassian.net/browse/RM-5437

Will still need to remove the relevant testing changes or rework them so that they better fit, will be done after we talk.

### DIFF
--- a/Remotion/ObjectBinding/Web.Test/TestTabbedForm.aspx.cs
+++ b/Remotion/ObjectBinding/Web.Test/TestTabbedForm.aspx.cs
@@ -73,19 +73,27 @@ public class TestTabbedForm : TestWxeBasePage
     TabView activeView = (TabView)MultiView.Views[activeViewIndex];
     if (activeViewIndex > 0)
       MultiView.SetActiveView(activeView);
+    if (IsPostBack)
+    {
+      var a = (WebTab)PagesTabStrip.Tabs[2];
+      a.IsVisible = false;
+      PagesTabStrip.Tabs.RemoveAt(1);
+
+      ((WebTab)PagesTabStrip.Tabs[4]).IsDisabled = true;
+      PagesTabStrip.Tabs.RemoveAt(3);
+    }
   }
 
   private void LoadUserControls ()
   {
-    // add tabs
 
-    //    AddMainMenuTab ("1", "Main Tab 1", null);
-    //    AddMainMenuTab ("2", "Main Tab 2 foo bar", null);
-    //    AddMainMenuTab ("3", "Main Tab 3 foo", null);
-    //    AddMainMenuTab ("4", "Main Tab 4 foo foo bar", null);
-    //    AddMainMenuTab ("5", "Main Tab 5", null);
-    //    AddMainMenuTab ("6", "Main Tab 6 foo", null);
-    //    AddMainMenuTab ("7", "Main Tab 7 foo foo bar", null);
+       AddTab("1", "Main Tab 1", null);
+       AddTab("2", "Main Tab 2 foo bar", null);
+       AddTab("3", "Main Tab 3 foo", null);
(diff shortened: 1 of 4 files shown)


Remotion/ObjectBinding/Web.Test/TestTabbedForm.aspx.cs
```diff
@@ -73,26 +73,27 @@ protected override void OnLoad (EventArgs e)
...
-
-    //    AddMainMenuTab ("1", "Main Tab 1", null);
-    //    AddMainMenuTab ("2", "Main Tab 2 foo bar", null);
-    //    AddMainMenuTab ("3", "Main Tab 3 foo", null);
-    //    AddMainMenuTab ("4", "Main Tab 4 foo foo bar", null);
-    //    AddMainMenuTab ("5", "Main Tab 5", null);
-    //    AddMainMenuTab ("6", "Main Tab 6 foo", null);
-    //    AddMainMenuTab ("7", "Main Tab 7 foo foo bar", null);
+
+       AddTab("1", "Main Tab 1", null);
```
MichaelKetting, line 90: nesting here seems off?

Remotion/Web/Core/UI/Controls/TabbedMultiView.cs
```diff
@@ -268,11 +268,13 @@ private void OnTabViewInserted (TabView view)
       tab.Target = view.ID;
       _tabStrip.Tabs.Add(tab);
 
-      if (Views.Count == 2 && Views.IndexOf(_placeHolderTabView) > 0)
+      if (Views.Count == 2)
         Views.Remove(_placeHolderTabView);
-
-      if (Views.Count == 1)
+      else if (Views.Count == 1)
```
MichaelKetting, line 273: since you combined the IFs into IF/ELSE, both blocks need braces. Didn't we implement an analyzer for this last fall?

MichaelKetting, line 273: We can combine those IFs because the OnTabViewRemoved handler takes are of the logic, correct? I think we should add assertions in the IF block that activeViewIndex == 0 and the first view is selected.

Remotion/Web/Core/UI/Controls/TabbedMultiView.cs
```diff
@@ -288,8 +290,6 @@ private void OnTabViewRemove (TabView view)
         {
           if (MultiViewInternal.Controls.Count > 1)
             _newActiveTabAfterRemove = (TabView)MultiViewInternal.Controls[index - 1];
-          else // No Tabs left after this tab
-            _newActiveTabAfterRemove = _placeHolderTabView;
```
MichaelKetting, line 292: this, you need to explain to me, please, because I can't verify the semantics in my head.

Remotion/Web/Core/UI/Controls/WebTabStrip.cs
```diff
@@ -384,9 +384,20 @@ internal void SetSelectedTabInternal (WebTab? tab)
...
+        {
+          for (int i = Tabs.IndexOf(_selectedTab); i < Tabs.Count; i++)
+          {
+            if (Tabs[i].IsVisible && !Tabs[i].IsDisabled)
+            {
+              _selectedTab = Tabs[i];
+              break;
+            }
+          }
+        }
```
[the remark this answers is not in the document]
Frazerus: Works regardless

Frazerus:
The properties in the code behind (enabled & visible) did not do anything when trying to create an invisible tab or when trying to make the button non interactive when on specific tabs (or in general for that matter).

MichaelKetting (changes requested):
some code formatting please :)

PR #692, opened 2023-05-03
Change DomainObjectTransactionContext to ValueType to reduce the allocations needed during property access

Frazerus:
https://re-motion.atlassian.net/browse/RM-5615
V3 was chosen due to comment.


Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextStruct.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using Remotion.Data.DomainObjects.DataManagement;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Infrastructure
+{
+  public readonly struct DomainObjectTransactionContextStruct
```
a-ctor, line 7: This struct and all its members need XML docs. We can take the docs from `IDomainObjectTransactionContext` since we probably delete this anyway.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContext.cs
```diff
@@ -21,48 +21,37 @@
 namespace Remotion.Data.DomainObjects.Infrastructure
 {
   /// <summary>
-  /// Provides the default implementation of the <see cref="IDomainObjectTransactionContext"/> interface.
+  /// Provides actual implementation for the <see cref="DomainObjectTransactionContextStruct"/>.
   /// Represents the context of a <see cref="DomainObject"/> that is associated with a specific <see cref="ClientTransaction"/>.
   /// </summary>
-  public class DomainObjectTransactionContext : IDomainObjectTransactionContext
+  public class DomainObjectTransactionContext
```
a-ctor, line 27: Since we call it strategy in a few places we might want to rename this to have a `Strategy` suffix. This would then match a bit better with its intent.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextStruct.cs
```diff
@@ -0,0 +1,43 @@
...
+  {
+
+    /// <exception cref="ClientTransactionsDifferException">The object cannot be used in the given transaction.</exception>
+    public DomainObjectTransactionContextStruct (DomainObject domainObject, ClientTransaction clientTransaction, DomainObjectTransactionContext strategy)
+    {
+      ArgumentUtility.CheckNotNull("associatedTransaction", clientTransaction);
```
a-ctor, line 13: parameter name is wrong

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContext.cs
a-ctor, line 27: Depending on the definition of public API we might need XML docs and arg checks in this file. But that is something that mike should decide once we have all the other stuff in order.

Remotion/Data/DomainObjects/DomainObject.cs
```diff
@@ -723,7 +729,7 @@ protected internal virtual void OnDeleted (EventArgs args)
 
     private void CheckInitializeEventNotExecuting ()
     {
-      if (_isReferenceInitializeEventExecuting)
+      if (_strategy != null && _strategy.IsInitializing)
         throw new InvalidOperationException("While the OnReferenceInitializing event is executing, this member cannot be used.");
     }
```
a-ctor, line 734: Since we throw the same exception in the strategy we can consider moving this method to the strategy as public API.

a-ctor (changes requested):
CI build failing with tests which also need to be addressed

Remotion/Data/DomainObjects/DomainObject.cs
```diff
@@ -507,15 +512,16 @@ protected PropertyIndexer Properties
     /// </summary>
     internal void RaiseReferenceInitializatingEvent ()
     {
-      _isReferenceInitializeEventExecuting = true;
+      _strategy ??= new DomainObjectTransactionContext(this);
```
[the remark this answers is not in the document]
a-ctor: My assumption was wrong. We need to keep it as is as not all instances are created using the constructor

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextStruct.cs
```diff
@@ -0,0 +1,43 @@
...
+
+      Strategy = strategy;
+      ClientTransaction = clientTransaction;
+    }
+
+    public DomainObject DomainObject => Strategy.GetDomainObject();
+
+    public ClientTransaction ClientTransaction { get; }
+
+    public DomainObjectState State => Strategy.GetState(ClientTransaction);
```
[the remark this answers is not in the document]
Frazerus: Do you want more or less space? This shows alright spacing already IMO. Fixed the area to have proper spacing though.

Remotion/Data/DomainObjects/DomainObject.cs
Frazerus: This method would be so small afterwards, can't we just throw it away and use the null propagated call afterwards?

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextStruct.cs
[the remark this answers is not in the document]
a-ctor: I meant newlines between members. Sorry for the confusion :)

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextIndexer.cs
```diff
@@ -26,25 +27,21 @@ namespace Remotion.Data.DomainObjects.Infrastructure
   public struct DomainObjectTransactionContextIndexer
   {
     private readonly DomainObject _domainObject;
-    private readonly bool _isInitializedEventExecuting;
+    private readonly DomainObjectTransactionContextStrategy _strategy;
 
-    public DomainObjectTransactionContextIndexer (DomainObject domainObject, bool isInitializedEventExecuting)
+    public DomainObjectTransactionContextIndexer (DomainObject domainObject, DomainObjectTransactionContextStrategy strategy)
```
a-ctor, line 32: We don't need domain object here anymore

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextStrategy.cs
```diff
@@ -0,0 +1,87 @@
...
+using Remotion.Data.DomainObjects.DataManagement;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Infrastructure
+{
+  /// <summary>
+  /// Provides actual implementation for the <see cref="DomainObjectTransactionContextExecutor"/>.
+  /// Represents the context of a <see cref="DomainObjects.DomainObject"/> that is associated with a specific <see cref="ClientTransaction"/>.
+  /// </summary>
+  [Serializable]
```
a-ctor, line 27: Why do we need `[Serializable]`?

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextExecutor.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+      ArgumentUtility.CheckNotNull("strategy", strategy);
+      ArgumentUtility.CheckNotNull("clientTransaction", clientTransaction);
+      DomainObjectCheckUtility.CheckIfRightTransaction(strategy.DomainObject, clientTransaction);
+
+      _strategy = strategy;
+      ClientTransaction = clientTransaction;
+    }
+
+        /// <summary>
```
a-ctor, line 69:
```suggestion
    /// <summary>
```

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextExecutor.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// <summary>
+    /// Gets the timestamp used for optimistic locking when the object is committed to the database.
+    /// </summary>
+    /// <value>The timestamp of the object.</value>
+    /// <exception cref="ObjectInvalidException">The object is invalid in the associated <see cref="ClientTransaction"/>.</exception>
+    /// <exception cref="ClientTransactionsDifferException">The object cannot be used in the given transaction.</exception>
+    public object? Timestamp => _strategy.GetTimestamp(ClientTransaction);
+
+    private readonly DomainObjectTransactionContextStrategy _strategy;
```
a-ctor, line 55: private stuff before the public

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextExecutor.cs
```diff
@@ -0,0 +1,152 @@
...
+  /// <summary>
+  /// Represents the transaction execution context of a <see cref="DomainObject"/> that is associated with a specific <see cref="ClientTransaction"/>.
+  /// </summary>
+  public readonly struct DomainObjectTransactionContextExecutor
+  {
+    /// <summary>
+    /// Gets the <see cref="DomainObject"/> this context is associated with.
+    /// </summary>
+    /// <value>The domain object.</value>
+    public DomainObject DomainObject => _strategy.DomainObject;
```
a-ctor, line 33: This is not an auto property and it belongs below the constructor as it does not represent any actual state of the class

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextExecutor.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// <exception cref="ClientTransactionsDifferException">The object cannot be used in the given transaction.</exception>
+    public DomainObjectState State => _strategy.GetState(ClientTransaction);
+
+    /// <summary>
+    /// Gets the timestamp used for optimistic locking when the object is committed to the database.
+    /// </summary>
+    /// <value>The timestamp of the object.</value>
+    /// <exception cref="ObjectInvalidException">The object is invalid in the associated <see cref="ClientTransaction"/>.</exception>
+    /// <exception cref="ClientTransactionsDifferException">The object cannot be used in the given transaction.</exception>
+    public object? Timestamp => _strategy.GetTimestamp(ClientTransaction);
```
a-ctor, line 53: move below the constructor

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextExecutor.cs
```diff
@@ -0,0 +1,152 @@
...
+using Remotion.Data.DomainObjects.DataManagement;
+using Remotion.Data.DomainObjects.Persistence;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Infrastructure
+{
+  /// <summary>
+  /// Represents the transaction execution context of a <see cref="DomainObject"/> that is associated with a specific <see cref="ClientTransaction"/>.
+  /// </summary>
+  public readonly struct DomainObjectTransactionContextExecutor
```
a-ctor, line 27: Executor does not fit well together with design IMO. We need to find a better name or even revert back to `DomainObjectTransactionContext`. Another todo that we should talk with mike about.

Remotion/Data/DomainObjects/DomainObject.cs
```diff
@@ -170,7 +170,7 @@ public class DomainObject : IDomainObject
     private ObjectID _id;
     private ClientTransaction _rootTransaction;
     private bool _needsLoadModeDataContainerOnly; // true if the object was created by a constructor call or OnLoaded has already been called once
-    private bool _isReferenceInitializeEventExecuting; // true only while OnReferenceInitializing is executed
+    private DomainObjectTransactionContextStrategy? _strategy;
```
a-ctor, line 173: The field name cannot stay as is, `_strategy` is not very helpful. We should change that once we decided on the names.

Remotion/Data/DomainObjects.UnitTests/DomainObjectTest.cs
```diff
@@ -231,12 +231,12 @@ public void RaiseReferenceInitializatingEvent_CallsReferenceInitializing_Current
...
 
-    [Test]
+    /*[Test]
     public void RaiseReferenceInitializatingEvent_CallsReferenceInitializing_TransactionContextIsRestricted ()
     {
       var result = _transaction.ExecuteInScope(() => DomainObjectTestHelper.ExecuteInReferenceInitializing_NewObject(o => o.DefaultTransactionContext));
-      Assert.That(result, Is.TypeOf(typeof(InitializedEventDomainObjectTransactionContextDecorator)));
-    }
+      Assert.That(result.Strategy.IsReferenceInitializeEventExecution, Is.True);
+    }*/
```
a-ctor, line 239: Either we rewrite this test or we delete it if it does not make any sense anymore. Comments are not an option :p

Remotion/Data/DomainObjects.UnitTests/DomainObjectTest.cs
```diff
@@ -558,7 +558,7 @@ public void TransactionContext ()
       var transactionContextIndexer = order.TransactionContext;
 
       Assert.That(transactionContextIndexer, Is.InstanceOf(typeof(DomainObjectTransactionContextIndexer)));
-      Assert.That(((DomainObjectTransactionContext)transactionContextIndexer[_transaction]).DomainObject, Is.SameAs(order));
+      Assert.That((transactionContextIndexer[_transaction]).DomainObject, Is.SameAs(order));
```
a-ctor, line 561:
```suggestion
      Assert.That(transactionContextIndexer[_transaction].DomainObject, Is.SameAs(order));
```

a-ctor (changes requested):
@MichaelKetting a few things need discussion

- [ ] Should the `IDomainObjectTransactionContext` be removed since we don't have any implementations anymore?
- [ ] Do we consider the `DomainObjectTransactionContextStrategy` public API? -> do we need arg checks and XML docs?
- [ ] We need to make a decision on the names of the classes
- [ ]

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextStrategy.cs
Frazerus: Great question, I have no clue. Probably something I shouldve talked about during standup -> Basically all of the tests that were going red on TC were due to serializing errors, something wants to serialize the strategy, but I'm not sure what, where and why.

a-ctor: Okay got the info now, we need `[Serializable]` as `DomainObject`s can be serialized so every store member needs to be serializable as well.

a-ctor (dismissed):

Remotion/Data/DomainObjects/DomainObject.cs
```diff
@@ -463,7 +468,7 @@ public override string ToString ()
     /// <remarks>To perform custom actions when a <see cref="DomainObject"/> is deleted <see cref="OnDeleting"/> and <see cref="OnDeleted"/> should be overridden.</remarks>
     protected void Delete ()
     {
-      CheckInitializeEventNotExecuting();
+      _strategy?.ThrowIfReferenceInitializeEventExecuting();
```
MichaelKetting, line 471: we should move `_strategy.ThrowIfReferenceInitializeEventExecuting()` into `CheckInitializeEventNotExecuting`. This way, we keep the semantic simple and in case further requirements are added in the future, have an existing helper method to fall back on. Law of Demeter can also apply inside a class.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextImplementation.cs
```diff
@@ -0,0 +1,103 @@
...
+  [Serializable]
+  public sealed class DomainObjectTransactionContextImplementation
+  {
+    public DomainObjectTransactionContextImplementation (DomainObject domainObject)
+    {
+      ArgumentUtility.CheckNotNull(nameof(domainObject), domainObject);
+      DomainObject = domainObject;
+    }
+
+    public bool IsReferenceInitializeEventExecution { get; set; }
```
MichaelKetting, line 36: state above ctor, please.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextImplementation.cs
```diff
@@ -0,0 +1,103 @@
...
+      ArgumentUtility.CheckNotNull(nameof(clientTransaction), clientTransaction);
+      DomainObjectCheckUtility.CheckIfRightTransaction(DomainObject, clientTransaction);
+
+      return clientTransaction.TryEnsureDataAvailable(DomainObject.ID);
+    }
+
+    public void ThrowIfReferenceInitializeEventExecuting ()
+    {
+      if (IsReferenceInitializeEventExecution)
+        throw new InvalidOperationException("While the OnReferenceInitializing event is executing, this member cannot be used.");
```
MichaelKetting, line 100:
```suggestion
        throw new InvalidOperationException("While the DomainObject.OnReferenceInitializing event is executing, this member cannot be used.");
```

MichaelKetting: @a-ctor we should play with branch prediction and check if placing the exception into the else-branch causes better performance. it's not clear if we can measure this, we may need to check the assembler. we want a version where no jump is needed in the event that no exception is thrown because that's the standard case. It may be that .net is already optimizing for this.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextIndexer.cs
```diff
@@ -25,26 +26,20 @@ namespace Remotion.Data.DomainObjects.Infrastructure
...
-    private readonly bool _isInitializedEventExecuting;
+    private readonly DomainObjectTransactionContextImplementation _implementation;
 
-    public DomainObjectTransactionContextIndexer (DomainObject domainObject, bool isInitializedEventExecuting)
+    public DomainObjectTransactionContextIndexer (DomainObjectTransactionContextImplementation implementation)
     {
-      ArgumentUtility.CheckNotNull("domainObject", domainObject);
-      _domainObject = domainObject;
-      _isInitializedEventExecuting = isInitializedEventExecuting;
+      ArgumentUtility.CheckNotNull("domainObject", implementation.DomainObject);
```
MichaelKetting, line 33: arg check should be for impllementation, not implementation.domainobject. it may be okay to also make that check debug-only.

Remotion/Data/DomainObjects/DomainObject.cs
```diff
@@ -208,6 +208,7 @@ protected DomainObject ()
       RaiseReferenceInitializatingEvent();
 
       _needsLoadModeDataContainerOnly = true;
+      _strategy = new DomainObjectTransactionContextImplementation(this);
```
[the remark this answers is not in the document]
Frazerus: Since it isn't set in the constructor, the field kicks out errors that it isn't initialized through the constructor even though it is not nullable. Am I missing something?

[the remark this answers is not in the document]
Frazerus: Never mind, Mike's comments further down explain the situation.

Remotion/Data/DomainObjects.UnitTests/DomainObjectExtensionsTest.cs
```diff
@@ -81,15 +81,14 @@ public void GetDefaultTransactionContext ()
...
 
       var expectedTransactionContext =
-          (DomainObjectTransactionContext)domainObject.TransactionContext[domainObject.RootTransaction.ActiveTransaction];
+          domainObject.TransactionContext[domainObject.RootTransaction.ActiveTransaction];
 
       Assert.That(actualTransactionContext.ClientTransaction, Is.EqualTo(expectedTransactionContext.ClientTransaction));
       Assert.That(actualTransactionContext.DomainObject, Is.EqualTo(expectedTransactionContext.DomainObject));
-      Assert.That(actualTransactionContext, Is.Not.EqualTo(expectedTransactionContext), "DomainObjectTransactionContext cannot be compared by reference at the moment.");
-
+      Assert.That(actualTransactionContext, Is.EqualTo(expectedTransactionContext));
```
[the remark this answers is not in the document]
Frazerus: Since the `transactionContext` was a reference type previously, the equals implementation would not return true. The new implementation uses structs, which by default implement value equality -> we actually only need the last Assertion, but for readability I would probably keep the other 2 as well.

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextIndexerTest.cs
```diff
@@ -29,24 +30,14 @@ public void Item ()
...
-      var order = DomainObjectIDs.Order1.GetObject<Order>();
-      var tx = order.RootTransaction.CreateSubTransaction();
-      var indexer = new DomainObjectTransactionContextIndexer(order, true);
 
-      var item = indexer[tx];
-      Assert.That(item, Is.InstanceOf(typeof(InitializedEventDomainObjectTransactionContextDecorator)));
       Assert.That(item.ClientTransaction, Is.SameAs(tx));
+      var transactionContextImplementation = PrivateInvoke.GetNonPublicField(item, "_transactionContextImplementation");
+      Assert.That(transactionContextImplementation, Is.Not.Null);
+      Assert.That(transactionContextImplementation, Is.InstanceOf<DomainObjectTransactionContextImplementation>());
```
MichaelKetting, line 40: that second check is redunant because there's only one implementation now.

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextIndexerTest.cs
```diff
@@ -29,24 +30,14 @@ public void Item ()
...
-    {
-      var order = DomainObjectIDs.Order1.GetObject<Order>();
-      var tx = order.RootTransaction.CreateSubTransaction();
-      var indexer = new DomainObjectTransactionContextIndexer(order, true);
 
-      var item = indexer[tx];
-      Assert.That(item, Is.InstanceOf(typeof(InitializedEventDomainObjectTransactionContextDecorator)));
       Assert.That(item.ClientTransaction, Is.SameAs(tx));
+      var transactionContextImplementation = PrivateInvoke.GetNonPublicField(item, "_transactionContextImplementation");
+      Assert.That(transactionContextImplementation, Is.Not.Null);
```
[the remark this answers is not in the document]
MichaelKetting: Please disregard, I thought `order` and `item` were the same instance

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextImplementation.cs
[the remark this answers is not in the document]
MichaelKetting: @a-ctor yeah, I can see that and given the other comments about possible more complete/clean approach, that would actually not help with the future changes. As for just the immediate: yes, I also considered a using with a scope object and came to the same conclusion - it's just architectural overhead for what we need here.

Remotion/Data/DomainObjects/DomainObject.cs
```diff
@@ -392,9 +390,10 @@ protected void BaseGetObjectData (SerializationInfo info, StreamingContext conte
     /// <exception cref="ArgumentNullException">The <paramref name="id"/> or <paramref name="rootTransaction"/> parameter is null.</exception>
     /// <exception cref="InvalidOperationException">This <see cref="DomainObject"/> has already been initialized.</exception>
     /// <remarks>This method is always called exactly once per <see cref="DomainObject"/> instance by the framework. It sets the object's 
-    /// <see cref="ID"/> and enlists it with the given <see cref="DomainObjects.ClientTransaction"/>.</remarks>
+    /// <see cref="ID"/> and enlists it with the given <see cref="DomainObjects.ClientTransaction"/> as well as the proper <see cref="DomainObjectTransactionContextImplementation"/>.</remarks>
```
MichaelKetting, line 393:
```suggestion
    /// <see cref="ID"/> and enlists it with the given <see cref="DomainObjects.ClientTransaction"/>.</remarks>
```
That's an implementation detail we don't need to document here.

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextIndexerTest.cs
```diff
@@ -29,24 +30,13 @@ public void Item ()
     {
       var order = DomainObjectIDs.Order1.GetObject<Order>();
       var tx = order.RootTransaction.CreateSubTransaction();
-      var indexer = new DomainObjectTransactionContextIndexer(order, false);
+      var indexer = new DomainObjectTransactionContextIndexer(new DomainObjectTransactionContextImplementation(order));
```
MichaelKetting, line 33: After reading this test for the third or so time, I realized we can solve this without the private invoke.

Step 1:
```suggestion
      var transactionContextImplementation = new DomainObjectTransactionContextImplementation(order)
      var indexer = new DomainObjectTransactionContextIndexer(transactionContextImplementation);
```

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextIndexerTest.cs
```diff
@@ -29,24 +30,13 @@ public void Item ()
...
-    {
-      var order = DomainObjectIDs.Order1.GetObject<Order>();
-      var tx = order.RootTransaction.CreateSubTransaction();
-      var indexer = new DomainObjectTransactionContextIndexer(order, true);
 
-      var item = indexer[tx];
-      Assert.That(item, Is.InstanceOf(typeof(InitializedEventDomainObjectTransactionContextDecorator)));
       Assert.That(item.ClientTransaction, Is.SameAs(tx));
+      var transactionContextImplementation = PrivateInvoke.GetNonPublicField(item, "_transactionContextImplementation");
+      Assert.That(transactionContextImplementation, Is.Not.Null);
```
MichaelKetting, line 39: Step 2:
```suggestion
      Assert.That(item, Is.EqualTo(new DomainObjectTranscationContext(transactionContextImplementation, tx));
```

This works because DOTxCtx is a value type so there is implicit state equality.

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextIndexerTest.cs
```diff
@@ -18,6 +18,7 @@
 using NUnit.Framework;
 using Remotion.Data.DomainObjects.Infrastructure;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain;
+using Remotion.Development.UnitTesting;
```
MichaelKetting, line 21:
```suggestion
```

Remotion/Data/DomainObjects.UnitTests/DomainObjectTest.cs
```diff
@@ -558,7 +563,25 @@ public void TransactionContext ()
...
+      var transactionContextIndexer = order.TransactionContext;
+
+      Assert.That(transactionContextIndexer, Is.InstanceOf(typeof(DomainObjectTransactionContextIndexer)));
+      Assert.That(transactionContextIndexer[_transaction].DomainObject, Is.SameAs(order));
+
+      var deserializedOrder = Serializer.SerializeAndDeserialize(order);
+      var newTransactionContextIndexer = deserializedOrder.TransactionContext;
+
+      Assert.That(transactionContextIndexer[_transaction].DomainObject, Is.SameAs(order));
+      Assert.That(transactionContextIndexer[_transaction].DomainObject, Is.Not.SameAs(deserializedOrder));
```
MichaelKetting, line 584:
```suggestion
      Assert.That(deserializedTransactionContextIndexer [_transaction].DomainObject, Is.SameAs(deserializedOrder));
```
We don't need to re-check the original indexer because it is immutable.

Remotion/Data/DomainObjects.UnitTests/DomainObjectTest.cs
```diff
@@ -558,7 +563,25 @@ public void TransactionContext ()
...
+      Assert2.IgnoreIfFeatureSerializationIsDisabled();
+
+      var order = _transaction.ExecuteInScope(() => Order.NewObject());
+      var transactionContextIndexer = order.TransactionContext;
+
+      Assert.That(transactionContextIndexer, Is.InstanceOf(typeof(DomainObjectTransactionContextIndexer)));
+      Assert.That(transactionContextIndexer[_transaction].DomainObject, Is.SameAs(order));
+
+      var deserializedOrder = Serializer.SerializeAndDeserialize(order);
+      var newTransactionContextIndexer = deserializedOrder.TransactionContext;
```
MichaelKetting, line 581:
```suggestion
      var deserializedTransactionContextIndexer = deserializedOrder.TransactionContext;
```

MichaelKetting (changes requested):
Some small test improvments plus the open question for @a-ctor in TxCtxImplementation.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextIndexer.cs
```diff
@@ -25,27 +26,16 @@ namespace Remotion.Data.DomainObjects.Infrastructure
   /// </summary>
   public struct DomainObjectTransactionContextIndexer
```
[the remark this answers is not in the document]
a-ctor: If we can we should use `readonly struct` as it allows compiler optimizations. If all the members are read-only there is no reason not to make the whole struct readonly

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextImplementation.cs
```diff
@@ -0,0 +1,124 @@
...
+      return clientTransaction.TryEnsureDataAvailable(DomainObject.ID);
+    }
+
+    public void ThrowIfReferenceInitializeEventExecuting ()
+    {
+      if (IsReferenceInitializeEventExecution)
+        throw new InvalidOperationException("While the DomainObject.OnReferenceInitializing event is executing, this member cannot be used.");
+    }
+
+    public void ExecutingOnReferenceInitializedEvent ()
```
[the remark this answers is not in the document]
a-ctor: I think that begin and end is better here since it conveys the intention that if you begin it you should end it afterwards.

Remotion/Data/DomainObjects/Infrastructure/DomainObjectTransactionContextImplementation.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+      ArgumentUtility.DebugCheckNotNull(nameof(clientTransaction), clientTransaction);
+      DebugCheckIfRightTransaction(clientTransaction);
+
+      ThrowIfReferenceInitializeEventExecuting();
+
+      return clientTransaction.TryEnsureDataAvailable(DomainObject.ID);
+    }
+
+    public void ThrowIfReferenceInitializeEventExecuting ()
```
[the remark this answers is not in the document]
a-ctor: Sounds sensible. We need to have the talk about why we can't call the default constructor of `DomainObject` would have also solved the `true` initialization problem.

Frazerus:
The serializable Attribute is still required on the DOTctxImplementation, as the serialize tests fail otherwise.

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextTest.cs
```diff
@@ -288,6 +290,83 @@ private void DeleteOrder (Order order)
       order.Delete();
     }
 
+    [Test]
+    public void AllowedMembers ()
```
MichaelKetting, line 294: We need to slighlty tweak the tests that were moved over from the deleted `Initialzing`-state version of the Context.
```suggestion
    public void AllowedMembers_WhileReferenceInitializingIsActive ()
```

Remotion/Data/DomainObjects.UnitTests/Infrastructure/DomainObjectTransactionContextTest.cs
```diff
@@ -288,6 +290,83 @@ private void DeleteOrder (Order order)
...
+      var ctx = new DomainObjectTransactionContext(transactionContextImplementation, TestableClientTransaction);
+
+      Assert.That(
+          () => ctx.Timestamp,
+          Throws.InvalidOperationException
+              .With.Message.EqualTo("While the DomainObject.OnReferenceInitializing event is executing, this member cannot be used."));
+    }
+
+    [Test]
+    public void ForbiddenMembers_RegisterForCommit ()
```
MichaelKetting, line 332:
```suggestion
    public void BeginDomainObjectReferenceInitializing_ForbiddenMembers_RegisterForCommit ()
```

MichaelKetting (changes requested):
Sorry,some test improvements and then we're done.

MichaelKetting (approved):
I inline did some last optimizations for the tests. please review the commits before flattening them to check what I did :) I also left some detailed notes in the commit description part of the commit message.

PR #693, opened 2023-05-04
RM-8799 Add column failure matching to the BocList

Frazerus:
https://re-motion.atlassian.net/browse/RM-8799

PR #695, opened 2023-05-05
RM-8800 Add a repository for storing validation errors in the BocList

Frazerus:
https://re-motion.atlassian.net/browse/RM-8800
These changes are already based on the changes for https://github.com/re-motion/Framework/pull/693, please take a look at them first.

### DIFF
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
@@ -1,0 +1,498 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Linq;
+using Moq;
+using NUnit.Framework;
+using Remotion.ObjectBinding.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class BocListValidationFailureRepositoryTest
+  {
+    private Mock<IBusinessObject> _rowObjectStub;
+    private Mock<IBocColumnDefinitionWithValidationSupport> _columnDefinitionStub;
+
+    [SetUp]
+    public void Setup ()
+    {
+      _rowObjectStub = new Mock<IBusinessObject>();
+      _columnDefinitionStub = new Mock<IBocColumnDefinitionWithValidationSupport>();
+    }
+
+    [Test]
+    public void AddValidationFailuresForBocList_ContainedInRepository ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForBocList(failures);
+
+      var result = repository.GetUnhandledValidationFailuresForBocList(false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void AddValidationFailuresForDataRow_ContainedInRepository ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataRow(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void AddValidationFailuresForDataCell_ContainedInRepository ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void ClearAllValidationFailures_NothingContainedInRepository ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      var listFailure = BusinessObjectValidationFailure.Create("List failure");
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForBocList(new[] { listFailure });
+
+      repository.ClearAllValidationFailures();
+
+      var result = repository.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(false);
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataCell_ReturnsCellFailures ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      var listFailure = BusinessObjectValidationFailure.Create("List failure");
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForBocList(new[] { listFailure });
+
+      var result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
+      Assert.That(result, Is.EquivalentTo(new[] { cellFailure }));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataCell_MarkAsHandled_ValidationFailureIsNotReturnedAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, true);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataCell_DontMarkAsHandled_ReturnsValidationFailureInSubsequentCalls ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRow_ReturnsRowFailures ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      var listFailure = BusinessObjectValidationFailure.Create("List failure");
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForBocList(new[] { listFailure });
+
+      var result = repository.GetUnhandledValidationFailuresForDataRow(_rowObjectStub.Object, false).ToArray();
+      Assert.That(result, Does.Not.Contain(cellFailure));
+      Assert.That(result, Does.Contain(rowFailure));
+      Assert.That(result, Does.Not.Contain(listFailure));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRow_MarkAsHandled_NotContainedInRepositoryAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataRow(_rowObjectStub.Object, true);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForDataRow(_rowObjectStub.Object, false);
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRow_DontMarkAsHandled_ContainedInRepositoryAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataRow(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForDataRow(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRowAndContainingDataCells_ReturnsCellAndRowFailures ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      var listFailure = BusinessObjectValidationFailure.Create("List failure");
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForBocList(new[] { listFailure });
+
+      var result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, false).ToArray();
+      Assert.That(result, Does.Contain(cellFailure));
+      Assert.That(result, Does.Contain(rowFailure));
+      Assert.That(result, Does.Not.Contain(listFailure));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRowAndContainingDataCells_MarkAsHandled_NotContainedInRepositoryAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, true);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, false);
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRowAndContainingDataCells_DontMarkAsHandled_ContainedInRepositoryAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, failures);
+
+      var result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForDataRowAndContainingDataCells_WithCellHandledSeparately_OnlyContainsUnhandledFailures ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var rowFailure = BusinessObjectValidationFailure.Create("A failure");
+      var unmarkedFailure = BusinessObjectValidationFailure.Create("Unmarked Failure");
+      var markedFailure = BusinessObjectValidationFailure.Create("Marked Failure");
+      var markedColumnStub = new Mock<IBocColumnDefinitionWithValidationSupport>();
+
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { unmarkedFailure });
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, markedColumnStub.Object, new[] { markedFailure });
+
+      var result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EquivalentTo(new[] { rowFailure, unmarkedFailure, markedFailure }));
+
+      result = repository.GetUnhandledValidationFailuresForDataCell(_rowObjectStub.Object, markedColumnStub.Object, true);
+      Assert.That(result, Is.EqualTo(new[] { markedFailure }));
+
+      result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(_rowObjectStub.Object, false);
+      Assert.That(result, Is.EquivalentTo(new[] { rowFailure, unmarkedFailure }));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForBocList_ReturnsListFailures ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      var listFailure = BusinessObjectValidationFailure.Create("List failure");
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForBocList(new[] { listFailure });
+
+      var result = repository.GetUnhandledValidationFailuresForBocList(false);
+      Assert.That(result, Does.Not.Contain(cellFailure));
+      Assert.That(result, Does.Not.Contain(rowFailure));
+      Assert.That(result, Does.Contain(listFailure));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForBocList_MarkAsHandled_NotContainedInRepositoryAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForBocList(failures);
+
+      var result = repository.GetUnhandledValidationFailuresForBocList(true);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForBocList(false);
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForBocList_DontMarkAsHandled_ContainedInRepositoryAfterwards ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var failures = new[]
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddValidationFailuresForBocList(failures);
+
+      var result = repository.GetUnhandledValidationFailuresForBocList(false);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetUnhandledValidationFailuresForBocList(false);
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells_ReturnsCellAndRowAndListFailures ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      var listFailure = BusinessObjectValidationFailure.Create("List failure");
+
+      repository.AddValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+      repository.AddValidationFailuresForDataRow(_rowObjectStub.Object, new[] { rowFailure });
+      repository.AddValidationFailuresForBocList(new[] { listFailure });
+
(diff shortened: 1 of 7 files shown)


Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundValidationFailureMatcher.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetMatchingValidationFailures (IBusinessObject rowObject, IBusinessObjectValidationResult validationResult)
+    {
+      var validationFailures = new List<BusinessObjectValidationFailure>();
+      foreach (var propertyPath in _propertyPaths)
+      {
+        validationFailures.AddRange( GetMatchingValidationFailures(propertyPath, rowObject, validationResult));
+      }
+      return validationFailures;
```
KuJo7, line 40:
```suggestion
      
      return validationFailures;
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
```diff
@@ -0,0 +1,424 @@
...
+      Assert.That(result, Is.EqualTo(failures));
+    }
+
+    [Test]
+    public void HasRowValidationFailures_WithoutFailures_ReturnsFalse ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var result = repository.HasRowValidationFailures(_rowObjectStub.Object);
+
```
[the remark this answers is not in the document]
Frazerus: I don't think we should remove this empty line. Tests are typically grouped into:

- Setup: In this case -> `var repository =...`
- Act: The thing we actually want to have happen -> `repository.HasRowValidationFailures(...)`
- Assert: Did the act part actually fulfill its role?

We typically want to have these parts easily visible and separated, which is given with the empty lines between the different parts.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
```diff
@@ -0,0 +1,424 @@
...
+    [Test]
+    public void HasRowValidationFailures_WithOnlyCellFailures_ReturnsTrue ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var cellFailure = BusinessObjectValidationFailure.Create("Cell failure");
+      repository.AddCellValidationFailures(_rowObjectStub.Object, _columnDefinitionStub.Object, new[] { cellFailure });
+
+      var result = repository.HasRowValidationFailures(_rowObjectStub.Object);
+
```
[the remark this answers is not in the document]
Frazerus: See above

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
```diff
@@ -0,0 +1,424 @@
...
+    [Test]
+    public void HasRowValidationFailures_WithOnlyRowFailures_ReturnsTrue ()
+    {
+      var repository = new BocListValidationFailureRepository();
+
+      var rowFailure = BusinessObjectValidationFailure.Create("Row failure");
+      repository.AddRowValidationFailures(_rowObjectStub.Object, new[] { rowFailure });
+
+      var result = repository.HasRowValidationFailures(_rowObjectStub.Object);
+
```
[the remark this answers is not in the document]
Frazerus: See above

Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundValidationFailureMatcher.cs
Frazerus: Hmm, interesting. I find the code to be readable in both ways, aesthetically I prefer my version. 

Is there any specific reasoning other than readability for this change? And should we argue about it? ;)

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+      {
+        return _rowFailures is { Count: > 0 } || _cellFailures != null && _cellFailures.Values.SelectMany(e => e).Any();
+      }
+    }
+
+    private List<HandleableFailure>? _listValidationFailures;
+    private Dictionary<IBusinessObject, RowFailureContainer>? _rowValidationFailures;
+
+    /// <summary>
+    /// Adds 0..N <see cref="BusinessObjectValidationFailure"/>s that belong to the <see cref="BocList"/> as a whole.
```
MichaelKetting, line 131: Descriptions: "Adds something" is enough. mulitplicity doesn't need documetnation unless there's somehting special

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+    /// <summary>
+    /// Adds 0..N <see cref="BusinessObjectValidationFailure"/>s that belong to an <see cref="IBocColumnDefinitionWithValidationSupport"/> cell
+    /// inside an <see cref="IBusinessObject"/> row inside the <see cref="BocList"/>.
+    /// </summary>
+    public void AddCellValidationFailures (IBusinessObject rowObject, IBocColumnDefinitionWithValidationSupport cellDefinition, IEnumerable<BusinessObjectValidationFailure> validationFailures)
+    {
+      ArgumentUtility.CheckNotNull(nameof(validationFailures), validationFailures);
+
+      var failures = GetOrCreateRowFailureContainer(rowObject);
+      failures.AddCellFailures(cellDefinition, validationFailures);
```
MichaelKetting, line 161: naming, in AddRowValidationFailures  we use `failureContainer`, here `failure`

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+          handleableRowFailure.MarkAsHandled();
+      }
+
+      return failures;
+    }
+
+    /// <summary>
+    /// Returns a collection of all <see cref="BusinessObjectValidationFailure"/>s belonging to a <see cref="BocList"/>.
+    /// </summary>
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetAllUnhandledListValidationFailures (bool markAsHandled)
```
MichaelKetting, line 188: What is difference between `GetUnhandledListValidationFailures` and `GetAllUnhandledListValidationFailures`. Docs say, only one of them has cell and row failures included? -> we need better naming.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+      if (_listValidationFailures == null)
+        return Array.Empty<BusinessObjectValidationFailure>();
+
+      var failures = new List<BusinessObjectValidationFailure>();
+      foreach (var handleableRowFailure in _listValidationFailures.Where(e => !e.Handled))
+      {
+        failures.Add(handleableRowFailure.Failure);
+        if (markAsHandled)
+          handleableRowFailure.MarkAsHandled();
+      }
```
MichaelKetting, line 199: This part duplciates GetUnhandledListValidationFailures 
 
Remove duplication and make a helper method from the rest of this this here and then use both methods inside GetAllUnhandledListValidationFailures

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+      var failures = new List<BusinessObjectValidationFailure>();
+      failureContainer.GetRowFailures(failures, markAsHandled);
+
+      return failures;
+    }
+
+    /// <summary>
+    /// Returns a collection of all <see cref="BusinessObjectValidationFailure"/>s belonging to an <see cref="IBusinessObject"/> row.
+    /// </summary>
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetAllUnhandledRowValidationFailures (IBusinessObject rowObject, bool markAsHandled)
```
MichaelKetting, line 233: naming: specificity - I think actually the naming might be inverse. Methods named "All" would return the failures for the  context object (e.g. the List or the Row) and all dependent objects (e.g. the Cells of the Row or the Rows of the List). The termn "All" is used for deciding if we want all failures or just some. Here, we also have failure catorization (e.g. the Row-asociated failures). I'm not sure if we don't want longer names that specify what type of failures are returned. Right now "GetAllUnhandledRowValidationFailures " means to me "row failures for all rows" and not "failures for this row including the cells". It's very possilbe, we always only want to include all nested failures when showing "unhandled failures" which makes the "all" naming redundant.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+        _rowFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void AddCellFailures (IBocColumnDefinitionWithValidationSupport cellDefinition, IEnumerable<BusinessObjectValidationFailure> failures)
+      {
+        var cellFailures = GetOrCreateListOfCellFailures(cellDefinition);
+        cellFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void GetAllFailures (ICollection<BusinessObjectValidationFailure> failures, bool markAsHandled)
```
MichaelKetting, line 62: Get should always return the result, not use a collecting parameter. (semantic naming). In our case, the collecting isn't even helpful because we don't check for duplicates (or it would be O(n*m) runtime) and we don't remove etc, so there is no need to pass the existing list.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+
+      public void AddRowFailures (IEnumerable<BusinessObjectValidationFailure> failures)
+      {
+        _rowFailures ??= new List<HandleableFailure>();
+        _rowFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void AddCellFailures (IBocColumnDefinitionWithValidationSupport cellDefinition, IEnumerable<BusinessObjectValidationFailure> failures)
+      {
+        var cellFailures = GetOrCreateListOfCellFailures(cellDefinition);
```
MichaelKetting, line 58: I think we might want to inline GetOrCreateListOfCellFailures. call it a gut feeling that it is easier to read.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+      private List<HandleableFailure>? _rowFailures;
+      private Dictionary<IBocColumnDefinitionWithValidationSupport, List<HandleableFailure>>? _cellFailures;
+
+      public void AddRowFailures (IEnumerable<BusinessObjectValidationFailure> failures)
+      {
+        _rowFailures ??= new List<HandleableFailure>();
+        _rowFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void AddCellFailures (IBocColumnDefinitionWithValidationSupport cellDefinition, IEnumerable<BusinessObjectValidationFailure> failures)
```
MichaelKetting, line 56: the api supports adding extra validation failures to an existing item. if we need it, this is okay. I don't know it's needed.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,306 @@
...
+        cellFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void GetAllFailures (ICollection<BusinessObjectValidationFailure> failures, bool markAsHandled)
+      {
+        GetRowFailures(failures, markAsHandled);
+        GetAllCellFailuresForRow(failures, markAsHandled);
+      }
+
+      public void GetRowFailures (ICollection<BusinessObjectValidationFailure> failures, bool markAsHandled)
```
MichaelKetting, line 68: Querstion: are these mthods used in multiple places with markashandled semantics or just once. it's possible we don't need mark as handled in the nested type but can keep that in the parent class. this would make the nested class a stupid collecton struture without business logic. might be easier to maintain.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
Frazerus: I don't think the validators should be required to add all the failures beforehand so that one validator does not remove the failures input by another validator. 

Therefore, I think we should leave this in here, just because it makes using the repository much more failsafe.

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <summary>
+    /// Adds <see cref="BusinessObjectValidationFailure"/>s that belong to an <see cref="IBocColumnDefinitionWithValidationSupport"/> cell
+    /// inside an <see cref="IBusinessObject"/> row inside the <see cref="BocList"/>.
+    /// </summary>
+    public void AddCellValidationFailures (IBusinessObject rowObject, IBocColumnDefinitionWithValidationSupport cellDefinition, IEnumerable<BusinessObjectValidationFailure> validationFailures);
+
+    /// <summary>
+    /// Returns a collection of all <see cref="BusinessObjectValidationFailure"/>s belonging to a <see cref="BocList"/>.
+    /// </summary>
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledListValidationFailures (bool markAsHandled);
```
MichaelKetting, line 28: We need to add parameter docs for markAsHandled. basically:
> If set to <see langword="true"/>, validation failures will not be returned again by any of the <c>GetUnhandled*</c> methods of this repository.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+    public BocListValidationFailureRepository ()
+    {
+      _listValidationFailures = new List<HandleableFailure>();
+      _rowValidationFailures = new Dictionary<IBusinessObject, RowFailureContainer>();
+    }
+
+    /// <summary>
+    /// Adds <see cref="BusinessObjectValidationFailure"/>s that belong to the <see cref="BocList"/> as a whole.
+    /// </summary>
+    public void AddListValidationFailures (IEnumerable<BusinessObjectValidationFailure> validationFailures)
```
MichaelKetting, line 129: Tip / please do: since we have an interface, you don't need to add method doc on the implementation. if you do want to make sure there's docs, you can use <inhertdoc />. Of course, if there's important add-ons that are needed, things are different. for the most part, that's not the case :)

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+
+      public IReadOnlyCollection<HandleableFailure> GetRowFailures ()
+      {
+        if (_rowFailures == null)
+          return Array.Empty<HandleableFailure>();
+
+        return _rowFailures.Where(e => !e.Handled).ToList();
+      }
+
+      private IReadOnlyCollection<HandleableFailure> GetAllCellFailuresForRow ()
```
MichaelKetting, line 92: this private method should be after the public methods.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+      private List<HandleableFailure>? _rowFailures;
+      private Dictionary<IBocColumnDefinitionWithValidationSupport, List<HandleableFailure>>? _cellFailures;
+
+      public void AddRowFailures (IEnumerable<BusinessObjectValidationFailure> failures)
+      {
+        _rowFailures ??= new List<HandleableFailure>();
+        _rowFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void AddCellFailures (IBocColumnDefinitionWithValidationSupport cellDefinition, IEnumerable<BusinessObjectValidationFailure> failures)
```
MichaelKetting, line 59: cellDefinition => columnDefinition. I am sure I missed a couple, please check.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+      {
+        List<HandleableFailure> cellFailures;
+        _cellFailures ??= new Dictionary<IBocColumnDefinitionWithValidationSupport, List<HandleableFailure>>();
```
MichaelKetting, line 62: I don't like this naming abiguity. the field and the variable are named the same but mean different things.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+      public void MarkAsHandled ()
+      {
+        Handled = true;
+      }
+    }
+
+    private class RowFailureContainer
+    {
+      private List<HandleableFailure>? _rowFailures;
+      private Dictionary<IBocColumnDefinitionWithValidationSupport, List<HandleableFailure>>? _cellFailures;
```
MichaelKetting, line 51: After much deliberation, I think we can just init those fields here and simplify the rest. Why? A RowFailureContainer is only created if there's validation failures to be added anyway. in particular the cellfailures are most likely to occur. And that leaves just the rowFaiures and saving on a single list-object? nah, we can simplify.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+
+      public IReadOnlyCollection<HandleableFailure> GetCellFailures (IBocColumnDefinitionWithValidationSupport cellDefinition)
+      {
+        if (_cellFailures == null)
+          return Array.Empty<HandleableFailure>();
+
+        if (!_cellFailures.TryGetValue(cellDefinition, out var handleableFailures))
+          return Array.Empty<HandleableFailure>();
+
+        return handleableFailures.Where(e => !e.Handled).ToList();
```
MichaelKetting, line 108: we should move the handled-check from RowFailure to the call site. it will simplify the code and have the same type (i.e. the outer type in this case) always be responsible for the Handled check. Right now, it's mixed and that doesn't help readability. It will also mean that the naming on RowFailure does not need to include "Unhandled"

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,315 @@
...
+        else
+        {
+          cellFailures = new List<HandleableFailure>();
+          _cellFailures[cellDefinition] = cellFailures;
+        }
+
+        cellFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public IReadOnlyCollection<HandleableFailure> GetRowAndCellFailures ()
```
MichaelKetting, line 76: please check, I think we can replace all the IReadOnlyCollection<> types with IEnumerable<> 

This allows us to use Concat in this method, preventing the copy operation. And since we're iterating the results anyway, we can get rid of all the listbuilding and make things completely lazy until returned from the outer/public api.

MichaelKetting (changes requested):
Okay, I got a couple of design improvement notes for the repostiory class. Please check, the basic rule is consistency, aligned naming, expressive APIs and reduced lines of code. 

I may find more optimizations after this round. 

@a-ctor you might also want to check my thoughts to make sure they are sensible.

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using Remotion.ObjectBinding.Validation;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  public interface IBocListValidationFailureRepository
```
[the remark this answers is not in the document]
Frazerus: I don't think the naming for the children methods is telling what is actually happening, as `...DataRowAndDataCells`  says nothing about the relationship between the row and the data cells. 

I would rather elongate the name to `*ValidationFailuresForDataRowAndContainingDataCells`

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,265 @@
...
+        List<HandleableFailure> cellFailureForRow;
+        if (_cellFailures.TryGetValue(columnDefinition, out var existingFailureList))
+        {
+          cellFailureForRow = existingFailureList;
+        }
+        else
+        {
+          cellFailureForRow = new List<HandleableFailure>();
+          _cellFailures[columnDefinition] = cellFailureForRow;
+        }
```
MichaelKetting, line 70: Possible simplification. May or may not feel better.
```suggestion
        List<HandleableFailure> cellFailureForRow;
        if (!_cellFailures.TryGetValue(columnDefinition, out var cellFailureForRow))
        {
          cellFailureForRow = new List<HandleableFailure>();
          _cellFailures.Add(columnDefinition, cellFailureForRow);
        }
```
Eitehr way, we should use Add not indexer for setting the value. Add has the semantic to fail if the item exists, indexer will overwrite. Since we want to add, not overwrite, the `Add` api should be used for expressiveness.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        cellFailureForRow.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public IEnumerable<HandleableFailure> GetValidationFailuresForDataRowAndDataCells ()
+      {
+        var failures = GetValidationFailuresForDataRow();
+        return failures.Concat(GetAllCellFailuresForRow());
+      }
```
MichaelKetting, line 79: I think we should inline this method. There is just one usage and it makes things more explicit while saving a couple of lines code - meaning the semantic is now only in the outer class (i.e. in a single place)

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,265 @@
...
+        return _rowFailures;
+      }
+
+      public IEnumerable<HandleableFailure> GetValidationFailuresForDataCell (IBocColumnDefinitionWithValidationSupport columnDefinition)
+      {
+        if (!_cellFailures.TryGetValue(columnDefinition, out var handleableFailures))
+          return Array.Empty<HandleableFailure>();
+
+        return handleableFailures;
+      }
```
MichaelKetting, line 92: Simplification
```suggestion
      public IEnumerable<HandleableFailure> GetValidationFailuresForDataCell (IBocColumnDefinitionWithValidationSupport columnDefinition)
      {
        return _cellFailures.GetValueOrDefault(columnDefinition, Array.Empty<HandleableFailure>());
      }
```

MichaelKetting (changes requested):
Prod code. Test comes next

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
```diff
@@ -0,0 +1,440 @@
...
+      var result = repository.GetValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, true);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void GetValidationFailuresForDataCell_DontMarkAsHandled_ContainedInRepositoryAfterwards ()
```
MichaelKetting, line 145:
```suggestion
    public void GetValidationFailuresForDataCell_DontMarkAsHandled_ReturnsValidationFailureInSubsequentCalls ()
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
```diff
@@ -0,0 +1,440 @@
...
+                     {
+                       BusinessObjectValidationFailure.Create("A failure")
+                     };
+
+      repository.AddCellValidationFailures(_rowObjectStub.Object, _columnDefinitionStub.Object, failures);
+
+      var result = repository.GetValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, true);
+      Assert.That(result, Is.EqualTo(failures));
+
+      result = repository.GetValidationFailuresForDataCell(_rowObjectStub.Object, _columnDefinitionStub.Object, false);
```
MichaelKetting, line 140: From a semantic point of view, the most important operation is that more general calls also do not include this failure. I think we should add to the tests that we add a failure for a second column and then use the parent opertion (e.g. the row operation) and verify that only the unhandled results are returned.

This applies to the other APIs as well.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,265 @@
...
+
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetValidationFailuresForDataRow (IBusinessObject rowObject, bool markAsHandled)
+    {
+      ArgumentUtility.CheckNotNull(nameof(rowObject), rowObject);
+
+      var failureContainer = GetRowFailureContainer(rowObject);
+      if (failureContainer == null)
+        return Array.Empty<BusinessObjectValidationFailure>();
+
+      var rowFailures = failureContainer.GetValidationFailuresForDataRow().Where(f => !f.Handled).ToList();
```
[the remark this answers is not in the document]
Frazerus: This would create multiple enumeration, is that preferred here?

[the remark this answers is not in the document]
Frazerus: I see, the next comment fixes that.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,265 @@
...
+
+      if(markAsHandled)
+      {
+        foreach (var failure in rowFailures)
+        {
+          failure.MarkAsHandled();
+        }
+      }
+
+      return rowFailures.Select(f => f.Failure).ToList();
```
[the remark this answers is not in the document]
Frazerus: When I implemented this, I was looking for applySideeffect but didnt find it. Thanks!

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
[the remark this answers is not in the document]
MichaelKetting: Correct and true. We never want multiple enums, sometimes it's a detail that gets missed in review suggestions :) Anyway, since the updated design solves this anyway, the point is moot :)

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,234 @@
...
+
+      return newFailureContainer;
+    }
+
+    private IEnumerable<HandleableFailure> GetUnhandledValidationFailuresForBocListInternal ()
+    {
+
+      var failures= _listValidationFailures.Where(e => !e.Handled);
+      return failures;
+    }
```
MichaelKetting, line 232: given the rest of the code, we can now inline this. There's is a lot of similiarity with _rowValidationFailures and those also have the handled-filter inline. It will make the code actually more compact / easier to read because we are able to unify the handled-calls.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,234 @@
...
+      failures = failures.Concat(failureContainer.GetAllCellFailuresForRow().Where(f => !f.Handled));
+
+      if (markAsHandled)
+        failures = failures.ApplySideEffect(f => f.MarkAsHandled());
+
+      return failures.Select(f => f.Failure).ToList();
+
+    }
+
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailuresForDataCell (IBusinessObject rowObject, IBocColumnDefinitionWithValidationSupport columnDefinition, bool markAsHandled)
```
MichaelKetting, line 186: The line is to long as far as I can tell without opening the code in the IDE. please verify and add linewraps. Please also check the rest of the files.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,234 @@
...
+        return Array.Empty<BusinessObjectValidationFailure>();
+
+      var failures = failureContainer.GetValidationFailuresForDataRow().Where(f => !f.Handled);
+      failures = failures.Concat(failureContainer.GetAllCellFailuresForRow().Where(f => !f.Handled));
+
+      if (markAsHandled)
+        failures = failures.ApplySideEffect(f => f.MarkAsHandled());
+
+      return failures.Select(f => f.Failure).ToList();
+
```
MichaelKetting, line 183: found extra blank line.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,234 @@
...
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailuresForDataRowAndContainingDataCells (IBusinessObject rowObject, bool markAsHandled)
+    {
+      ArgumentUtility.CheckNotNull(nameof(rowObject), rowObject);
+
+      var failureContainer = _rowValidationFailures.GetValueOrDefault(rowObject);
+      if (failureContainer == null)
+        return Array.Empty<BusinessObjectValidationFailure>();
+
+      var failures = failureContainer.GetValidationFailuresForDataRow().Where(f => !f.Handled);
+      failures = failures.Concat(failureContainer.GetAllCellFailuresForRow().Where(f => !f.Handled));
```
MichaelKetting, line 177:
```suggestion
      var failures = failureContainer
          .GetValidationFailuresForDataRow()
          .Concat(failureContainer.GetAllCellFailuresForRow())
          .Where(f => !f.Handled);
```

We don't need to rework it, it's more a design-study question. Personally, I prefer this version because it removes the duplication.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,234 @@
...
+
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetValidationUnhandledFailuresForBocListAndContainingDataRowsAndDataCells (bool markAsHandled)
+    {
+      var failures = GetUnhandledValidationFailuresForBocListInternal();
+
+      foreach (var failureContainer in _rowValidationFailures.Values)
+      {
+        failures = failures.Concat(failureContainer.GetValidationFailuresForDataRow().Where(f => !f.Handled));
+        failures = failures.Concat(failureContainer.GetAllCellFailuresForRow().Where(f => !f.Handled));
+      }
```
MichaelKetting, line 144:
```suggestion
      var failures = _listValidationFailures
          .Concat(_rowValidationFailures.Values.SelectMany(row => row .GetValidationFailuresForDataRow().Concat(row.GetAllCellFailuresForRow()))
          .Where(f => !f.Handled);
```

Another design-study for your consideration. The exact number of line breaks is something that can be subject to preference. I think this version is the most easiest to read because it has three top level operations and three lines. adding formatting linebreaks in line 2 could also work, it's a matter of preference.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,234 @@
...
+        _rowFailures.AddRange(failures.Select(e => new HandleableFailure(e)));
+      }
+
+      public void AddCellFailures (IBocColumnDefinitionWithValidationSupport columnDefinition, IEnumerable<BusinessObjectValidationFailure> failures)
+      {
+        List<HandleableFailure>? cellFailureForRow;
+        if (!_cellFailures.TryGetValue(columnDefinition, out cellFailureForRow))
+        {
+          cellFailureForRow = new List<HandleableFailure>();
+          _cellFailures.Add(columnDefinition,cellFailureForRow);
```
MichaelKetting, line 66:
```suggestion
          _cellFailures.Add(columnDefinition, cellFailureForRow);
```
the whitespace stuff is a bit strange. I figured the tooling would take care of that. @a-ctor, ideas?

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationFailureRepositoryTest.cs
```diff
@@ -0,0 +1,498 @@
...
+      Assert.That(result, Is.EqualTo(new[] { markedCellFailure }));
+
+      result = repository.GetValidationUnhandledFailuresForBocListAndContainingDataRowsAndDataCells(false);
+      Assert.That(result, Is.EquivalentTo(new[] { unmarkedCellFailure, unmarkedRowFailure, markedRowFailure, markedHiddenCellFailure }));
+
+      result = repository.GetUnhandledValidationFailuresForDataRowAndContainingDataCells(markedRowStub.Object, true);
+      Assert.That(result, Is.EquivalentTo(new[] { markedRowFailure, markedHiddenCellFailure }));
+
+      result = repository.GetValidationUnhandledFailuresForBocListAndContainingDataRowsAndDataCells(false);
+      Assert.That(result, Is.EquivalentTo(new[] { unmarkedCellFailure, unmarkedRowFailure }));
```
MichaelKetting, line 446: I like the updated naming. It is really helping with undestanding the expected test results  without needing to reason over the entire test and test name. Basically, I can trust in that by using easily to understand expected-values, there's a much lower chance of typo-style mistakes in the expectations because you would stumble over your own naming while setting up the expectations.

"trust & review": https://blog.ploeh.dk/2023/03/20/on-trust-in-software-development/
TL;DR: everyone makes mistakes, so we do reviews to spot them

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureRepository.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </remarks>
+    /// <param name="markAsHandled">If set to true, validation failures will not be returned again by any of the <c>GetUnhandled*</c> methods of this repository.</param>
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailuresForBocList (bool markAsHandled);
+
+    /// <summary>
+    /// Returns a collection of all unhandled <see cref="BusinessObjectValidationFailure"/>s associated directly with the <see cref="BocList"/>
+    /// itself and any of the (nested) property values associated with the <see cref="BusinessObject"/>s of individual rows and cells.
+    /// </summary>
+    /// <param name="markAsHandled">If set to true, validation failures will not be returned again by any of the <c>GetUnhandled*</c> methods of this repository.</param>
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetValidationUnhandledFailuresForBocListAndContainingDataRowsAndDataCells (bool markAsHandled);
```
MichaelKetting, line 45:
```suggestion
    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells (bool markAsHandled);
```

MichaelKetting (approved):
Summary: Some stylistic ideas / personal preferences and one typo in the method names that needs fixing. I don't think I need to see the changes again, unless you got questions?

a-ctor:
@Frazerus please change the base branch to `feature/RMTEAM-519-BocList-Validation` before merging the PR. This way, we can clean up the commits later on if we decide we are missing some functionality.

PR #697, opened 2023-05-11
RM-8810 Implement tests for CompoundValidationMessageFactory

Frazerus:
https://re-motion.atlassian.net/browse/RM-8810

### DIFF
--- a/Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
+++ b/Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
@@ -14,7 +14,11 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
+using Moq;
 using NUnit.Framework;
+using Remotion.Reflection;
+using Remotion.Validation.Implementation;
+using Remotion.Validation.Validators;
 
 namespace Remotion.Validation.UnitTests.Implementation
 {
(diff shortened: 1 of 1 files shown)


Remotion/Validation/UnitTests/Implementation/TestDomain/FakeValidationMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Remotion.Validation.UnitTests.Implementation.TestDomain
+{
+  public class FakeValidationMessage: ValidationMessage
+  {
+    private readonly string _message;
+
+    public FakeValidationMessage (string message)
+    {
+      _message = message;
```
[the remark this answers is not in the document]
Frazerus: It  is only a FakeValidation message. I don't think argument checks are useful when testing some completely different behavior.

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,142 @@
...
-    [Ignore("RM-5906")]
-    public void CreateValidationMessageForPropertyValidator ()
+    public void CreateValidationMessageForPropertyValidator_ContainsMessageFromSavedFactory ()
     {
+      var validatorStub = new Mock<IPropertyValidator>();
+      var propertyStub = new Mock<IPropertyInformation>();
+
+      var expectedMessage = new FakeValidationMessage("Oh no, a validation error!");
+      var factoryStub = new Mock<IValidationMessageFactory>();
+      factoryStub.Setup(_ => _.CreateValidationMessageForPropertyValidator(validatorStub.Object, propertyStub.Object))
```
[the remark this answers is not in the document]
Frazerus: I think the connection between the stub and the setup corresponding to it makes sense, but I can reduce the size of the block by moving the validation message away, I think this has the same effect but reduces crossovers. Is this alright?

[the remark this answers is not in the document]
KuJo7: This is what Mike told me regarding the formatting:

_Formatting:
when we write multi-line mock statements, each part is on a new line, inluding the Setup call. That way, we can read it in visual alignment. Historically (with RhinoMocks), the Setup-call was actuall either Stub or Expect so the alignment was even more imporant to quickly understand the test._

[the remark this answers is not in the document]
Frazerus: Ahhh I was thinking about the wrong thing, sorry about that. Will fix it :)

Remotion/Validation/UnitTests/Implementation/TestDomain/FakeValidationMessage.cs
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public override string Format (CultureInfo culture, IFormatProvider formatProvider, params object[] parameters)
+    {
+      throw new NotImplementedException();
```
MichaelKetting, line 18: I think we should build this as
return parameters.Aggregate(new StringBuilder(_message), (p, sb) => sb.AppendFormat(" | '{0}')).ToString();

still relatively simple, no exception, all information available.

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,160 @@
...
+      var factoryStub = new Mock<IValidationMessageFactory>();
+      factoryStub
+          .Setup(_ => _.CreateValidationMessageForPropertyValidator(validatorStub.Object, propertyStub.Object))
+          .Returns(expectedMessage);
+
+      var factory = new CompoundValidationMessageFactory(new[] { factoryStub.Object });
+
+      var result = factory.CreateValidationMessageForPropertyValidator(validatorStub.Object, propertyStub.Object);
+
+      Assert.That(result, Is.EqualTo(expectedMessage));
```
MichaelKetting, line 46: just a note: when I expect fake objects, I often (when I think of it, feel like it ;) ) test for reference equality using Is.SameAs instead of semantic equality using Is.EqualTo. It feels more precise, removes ambiguity.

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,160 @@
...
 
 namespace Remotion.Validation.UnitTests.Implementation
 {
   [TestFixture]
   public class CompoundValidationMessageFactoryTest
   {
     [Test]
-    [Ignore("RM-5906")]
-    public void CreateValidationMessageForPropertyValidator ()
+    public void CreateValidationMessageForPropertyValidator_ContainsMessageFromSavedFactory ()
```
MichaelKetting, line 30: Naming
CreateValidationMessageForPropertyValidator_ReturnsMessageFromFactory

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,160 @@
...
+
+      var factory = new CompoundValidationMessageFactory(new[] { factoryStub.Object });
+
+      var result = factory.CreateValidationMessageForPropertyValidator(validatorStub.Object, propertyStub.Object);
+
+      Assert.That(result, Is.EqualTo(expectedMessage));
+    }
+
+    [Test]
+    public void CreateValidationMessageForPropertyValidator_WithSeveralFactories_ContainsMessageFromFirstFactory ()
```
MichaelKetting, line 50:
```suggestion
    public void CreateValidationMessageForPropertyValidator_WithMultipleFactories_ReturnsMessageFromFirstFactory ()
```

Multiple vs Several: stylistic thing, we don't use "several", I think.

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,160 @@
...
   {
     [Test]
-    [Ignore("RM-5906")]
-    public void CreateValidationMessageForPropertyValidator ()
+    public void CreateValidationMessageForPropertyValidator_ContainsMessageFromSavedFactory ()
     {
+      var validatorStub = new Mock<IPropertyValidator>();
+      var propertyStub = new Mock<IPropertyInformation>();
+
+      var expectedMessage = new FakeValidationMessage("Oh no, a validation error!");
```
MichaelKetting, line 35: Intention question: we got the InvariantValidationMessage that has the same call-site overhead in the tests. Why not just use that one instead? I do think I also went down the FakeValidationMessage route first, then had the InvariantValiationMessage and figured that's good enough and we don't need the extra weight of a fake type.

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,160 @@
...
+
+      var factory = new CompoundValidationMessageFactory(new[] { factory1Stub.Object, factory2Stub.Object });
+
+      var result = factory.CreateValidationMessageForPropertyValidator(validatorStub.Object, propertyStub.Object);
+
+      Assert.That(result, Is.EqualTo(expectedMessage));
+    }
+
+    [Test]
+    public void CreateValidationMessageForPropertyValidator_WithFirstFactoryReturningNullMessage_ContainsMessageFromSecondFactory ()
```
MichaelKetting, line 77:
```suggestion
    public void CreateValidationMessageForPropertyValidator_WithFirstFactoryReturningNullMessage_ReturnsMessageFromSecondFactory ()
```

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
```diff
@@ -14,23 +14,160 @@
...
+
+      var result = factory.CreateValidationMessageForPropertyValidator(validatorStub.Object, propertyStub.Object);
+
+      Assert.That(result, Is.EqualTo(expectedMessage));
     }
 
     [Test]
-    [Ignore("RM-5906")]
-    public void CreateValidationMessageForObjectValidator ()
+    public void CreateValidationMessageForObjectValidator_ContainsMessageFromSavedFactory ()
```
MichaelKetting, line 102: my mistake: I didn't highlight all occurances of naming updates. Next time, I'll mention it (so long as I don't forget).

=> please check the reminaing tests and apply naming guidelines from the original review

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/Implementation/CompoundValidationMessageFactoryTest.cs
MichaelKetting: Could you please add the reasoning / answer for this question? Thanks!

Frazerus: At the point of creating the FakeValidationMessage I did not understand what the InvariantValidationMessage is used for. Therefore I simply thought going safe and creating an extra Fake to make sure that I don't misuse some functionality.

MichaelKetting (approved):

PR #698, opened 2023-05-11
RM-8811 Implement tests for ObjectValidationRule

Frazerus:
https://re-motion.atlassian.net/browse/RM-8811

### DIFF
--- a/Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
+++ b/Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
@@ -1,4 +1,27 @@
-﻿using NUnit.Framework;
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using Moq;
+using NUnit.Framework;
+using Remotion.Validation.Results;
+using Remotion.Validation.Rules;
+using Remotion.Validation.UnitTests.TestDomain;
+using Remotion.Validation.Validators;
 
 namespace Remotion.Validation.UnitTests.Rules
 {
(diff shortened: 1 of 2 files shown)


Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
     [Test]
-    [Ignore("TODO RM-5906")]
-    public void IsActive ()
+    public void Validate_TryToValidateNull_ReturnsEmpty ()
     {
-      // consider better name than "IsActive"
+      bool condition (object o)
+      {
+        return o != null;
+      }
```
KuJo7, line 52:
```suggestion
Func<object, bool> condition = o => o != null;
```
Since you have it like this in the other tests

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void Validate_WithUnfulfilledCondition_ReturnsEmpty ()
+    {
+      Func<int, bool> condition = i => i > 0;
+      var context = new ValidationContext(-1);
```
KuJo7, line 78:
```suggestion
      
      var context = new ValidationContext(-1);
```

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void IsActive_TryToValidateNull_ReturnsFalse ()
+    {
+      Func<object, bool> condition = o => true;
+      var context = new ValidationContext(null);
```
KuJo7, line 90:
```suggestion
      
      var context = new ValidationContext(null);
```

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void Validate_WithNullCondition_ReturnsEmpty ()
+    {
+      Func<object, bool> condition = null;
+      var context = new ValidationContext(new object());
```
KuJo7, line 66:
```suggestion
      
      var context = new ValidationContext(new object());
```

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
   [TestFixture]
   public class ObjectValidationRuleTest
   {
     [Test]
-    [Ignore("TODO RM-5906")]
-    public void Validate ()
+    public void Validate_ReturnsValidationFailures ()
     {
+      Func<int, bool> condition = i => true;
+      var context = new ValidationContext(1);
```
KuJo7, line 18:
```suggestion
      
      var context = new ValidationContext(1);
```

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+      var context = new ValidationContext(new object());
+      var rule = new ObjectValidationRule<object>(condition, new List<IObjectValidator>());
+
+      var result = rule.IsActive(context);
+
+      Assert.That(result, Is.True);
+    }
+
+    [Test]
+    public void IsActive_ReturnsValueFromCondition ()
```
KuJo7, line 112: better name than "IsActive"?

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+          "15 is not a number",
+          "Localized validation message: 15 is not a number.");
+
+      var failure2 = new ObjectValidationFailure(
+          42,
+          "42 is not a number",
+          "Localized validation message: 42 is not a number.");
+
+      var validatorMock1 = new Mock<IObjectValidator>();
+      validatorMock1.Setup(_ => _.Validate(It.IsAny<ObjectValidatorContext>()))
```
MichaelKetting: If we use 
```suggestion
      validatorMock1.Setup(_ => _.Validate(context))
```
the code is just as easy to maintain/robust and more compact. it also ensured that we don't pass null instead of the actual object.

MichaelKetting: formatting: please add a new line before Setup. We always align the mock syntax. (a matching roslyn analyzer is in the backlog ;)

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
Frazerus: No clue, we should talk about it with the others.

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+          .Returns(new[] { failure1 })
+          .Verifiable();
+
+      var validatorMock2 = new Mock<IObjectValidator>();
+      validatorMock2.Setup(_ => _.Validate(It.IsAny<ObjectValidatorContext>()))
+          .Returns(new[] { failure2 })
+          .Verifiable();
+
+      var rule = new ObjectValidationRule<int>(condition, new[] { validatorMock1.Object, validatorMock2.Object });
```
MichaelKetting:
```suggestion
      var rule = new ObjectValidationRule<int>(null, new[] { validatorMock1.Object, validatorMock2.Object });
```

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
 {
   [TestFixture]
   public class ObjectValidationRuleTest
   {
     [Test]
-    [Ignore("TODO RM-5906")]
-    public void Validate ()
+    public void Validate_ReturnsValidationFailures ()
     {
+      Func<int, bool> condition = i => true;
```
MichaelKetting, line 17:
```suggestion
```

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
MichaelKetting, line 18: we can use an int here, some object would be more appropriate though for the intended semantics ob "Object Validation" since it refers to validating objects (entities), e.g. a Person.

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,132 @@
...
+          .Setup(_ => _.Validate(It.IsAny<ObjectValidatorContext>()))
+          .Returns(new[] { failure2 })
+          .Verifiable();
+
+      var rule = new ObjectValidationRule<int>(condition, new[] { validator1Mock.Object, validator2Mock.Object });
+
+      var result = rule.Validate(context);
+      Assert.That(result, Is.EquivalentTo(new[] { failure1, failure2 }));
+      validator2Mock.Verify();
+      validator1Mock.Verify();
```
MichaelKetting, line 47: We don't need them verifiable, isntead stub semantics is sufficient because the result can only satisfy the assertion if both validators where called correctly. please also update the name to Stub

MichaelKetting (changes requested):
I may have missed some bits, and because there's bigger requests anyway, I'll submit it for now to get things moving.

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,128 @@
...
+      var context = new ValidationContext(new object());
+      var rule = new ObjectValidationRule<object>(condition, new List<IObjectValidator>());
+
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void Validate_WithUnfulfilledCondition_ReturnsEmpty ()
```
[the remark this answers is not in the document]
Frazerus: This test name is fine though, right? If we actually have a condition, and that condition returns false we do not want to have anything returned by the validators as they should not be used in this case. 

Therefore, the method should return empty given an unfulfilled condition.

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
Frazerus: Sadly, this doesnt work. The context for the object validators is created inside the rule, we cannot create it beforehand and pass it through.

MichaelKetting: That's true - different objects. My bad.

So, updated request: we should have a test that verifies that ObjectValidationContext is correctly constructured. we can do the following:

ObjectValidatioNContext actualCtx
validatorMock2.Setup(_ => _.Validate(It.IsAny<ObjectValidatorContext>()))
          .WhenCalled(ctx => actualCtx = ctx)
          .Returns(new[] { failure2 })
          .Verifiable();
Assert.that (actualCtx, Is.Not.Null)
Assert.that(actualCtx.Parent, Is.SameAs(context))
Assert.that(actualCtx.InstanceToValidate, Is.SameAs(contxt.InstanceToValidate))

Doing it this way instead of just having a predicate in It.Is<ObjectValidationContext>() decouples the mock from the assertion (always nice when done cheaply) and gives us better diagnostic when there do is an error. And it also hihglihgts the actual test we want to do (here: test that the ObjctValidatioNContext is built correctly)

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,176 @@
...
+      var context = new ValidationContext(new object());
+      var rule = new ObjectValidationRule<object>(null, new List<IObjectValidator>());
+
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void Validate_WithNullAsInstanceToValidate_ReturnsEmptyResultAsValidatorsAreNotCalled ()
```
MichaelKetting, line 95: Tip: "AsValidatorsAreNotCalled` is something we usually put to the left or leave out. it's possible I wrote this myself during the review so...yeah. Nothing to fuss oer, though.

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,176 @@
...
+      var context = new ValidationContext(new object());
+      var rule = new ObjectValidationRule<object>(null, new List<IObjectValidator>());
+
+      var result = rule.IsActive(context);
+
+      Assert.That(result, Is.True);
+    }
+
+    [Test]
+    public void IsActive_ReturnsValueFromCondition ()
```
MichaelKetting, line 160: This would be best in two tests, or using TestCase attribute or using BooleanObjectMother -> we try to avoid having two system under test isntances in the same test method.

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,176 @@
...
+
+      var rule = new ObjectValidationRule<object>(condition, new [] {validatorStub.Object});
+
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.Empty);
+    }
+
+    [Test]
+    public void IsActive_WithoutInstanceToValidate_ReturnsFalse ()
```
MichaelKetting, line 136: I think we should make the condition null in this test. WithoutInstanceToValidate has nothing to do with the consditionand making the condition null is more explcit aobut this. This should be updatd, please, but I don't need to see it after.

MichaelKetting (approved):
two idas for future tests and one request but i already approved since it's mini

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
Frazerus: Problem here is that this test is about `IsActive,` which always returns true given a null condition. You may have thought that this is the `Validate` method?

Given this, we need the condition to return true to actually show a different output given the missing instance to validate.

MichaelKetting: oh, yeah. thaks. gotcha! no change then

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,176 @@
...
+      var inactiveRule = new ObjectValidationRule<object>(alwaysFalseCondition, new List<IObjectValidator>());
+
+      var resultActiveRule = activeRule.IsActive(context);
+      var resultInactiveRule = inactiveRule.IsActive(context);
+
+      Assert.That(resultActiveRule, Is.True);
+      Assert.That(resultInactiveRule, Is.False);
     }
   }
 }
```
MichaelKetting, line 176: Reading the PropertyValidationRule, I realzied that I forgot a test:
ToString => string.Format("ObjectValidationRule (${typeof(ValidatedType.GetFullNameSafe)})"

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/Rules/ObjectValidationRuleTest.cs
```diff
@@ -1,21 +1,193 @@
...
+
+    [Test]
+    public void ToString_WithObjectType_ReturnsFullyQualifiedName ()
+    {
+      const string expectedResult = "ObjectValidationRule (System.Object)";
+      var rule = new ObjectValidationRule<object>(null, new List<IObjectValidator>());
+
+      var result = rule.ToString();
+
+      Assert.That(result, Is.EqualTo(expectedResult));
```
MichaelKetting, line 190: Stylistic things:
When there's no special cases, we don't need the "WithSomething" phrase in the test title. Here' it would just be "ToString_ReturnsValidatorTypeAndFullNameOfValidatedType". The existing option is good enough, though.

In case of the expected result, unless there's sepecial logic involved, it's inlined at the assertion. Usually, the more simple the situation, the more is inlined. That way, you see it when you look at the Assert instead of needing to hunt down the right variable inside the test.

Using System.Object instead of somehting test-specifc like Person from the test domain makes the test less safe. In particular, System.Object, you could easily use that one accidentally in production code and never notice it via a red test. That's why I try to use either obsucre .NET BCL types or types from the actual test assembly, whenever I can.

PR #703, opened 2023-05-17
RM-8812 Implement tests for PropertyValidationRule

KuJo7:
https://re-motion.atlassian.net/browse/RM-8812

### DIFF
--- a/Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
+++ b/Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
@@ -1,6 +1,27 @@
-﻿using NUnit.Framework;
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using Moq;
+using NUnit.Framework;
+using Remotion.Reflection;
+using Remotion.Validation.Results;
 using Remotion.Validation.Rules;
 using Remotion.Validation.UnitTests.TestDomain;
+using Remotion.Validation.Validators;
 
 namespace Remotion.Validation.UnitTests.Rules
 {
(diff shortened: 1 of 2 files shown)


Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+      Assert.That(res, Is.Empty);
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
-    public void IsActive ()
+    public void Validate_WithConditionIsFalse_ReturnsEmptyResult ()
     {
-      // consider better name than "IsActive"
+      var context = new ValidationContext("test");
```
MichaelKetting, line 56: an object instance instead of a string, please. Reasoning is that object is intended to be a domain type. We can use e.g. `Person` from the test domain to make it more expressive.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+
+      var rule = new PropertyValidationRule<object, int>(propertyInformationMock.Object, propertyFunc, conditionFunc, new[] { validatorMock.Object });
+
+      var res = rule.Validate(context);
+
+      Assert.That(res, Is.Empty);
+    }
+
+    [Test]
+    public void Validate_ReturnsListOfValidationFailures ()
```
MichaelKetting, line 76: am missing the condition-true case.

MichaelKetting, line 76:
```suggestion
    public void Validate_WithoutCondition_ReturnsValidationResults ()
```

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+      var context = new ValidationContext("test");
+      var propertyInformationMock = new Mock<IPropertyInformation>();
+
+      var propertyFailure = new PropertyValidationFailure("test", propertyInformationMock.Object, "propval", "err", "validmessage");
+
+      var validatorMock = new Mock<IPropertyValidator>();
+      validatorMock
+          .Setup(_ => _.Validate(It.IsAny<PropertyValidatorContext>()))
+          .Returns(new[] { propertyFailure })
+          .Verifiable();
```
MichaelKetting, line 87: We don't need this verifiable. If Validate isn't called, the result assertion wouldn't match.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+    public void Validate_ReturnsListOfValidationFailures ()
+    {
+      var context = new ValidationContext("test");
+      var propertyInformationMock = new Mock<IPropertyInformation>();
+
+      var propertyFailure = new PropertyValidationFailure("test", propertyInformationMock.Object, "propval", "err", "validmessage");
+
+      var validatorMock = new Mock<IPropertyValidator>();
+      validatorMock
+          .Setup(_ => _.Validate(It.IsAny<PropertyValidatorContext>()))
```
MichaelKetting, line 85: we should capture the PropertyValidatorContext using `Callback` and assert it at the end of the test.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+          .Returns(new[] { propertyFailure });
```
MichaelKetting, line 110:
```suggestion
```

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+
+      Func<object, object> propertyFunc = i => i;
```
MichaelKetting, line 112:
```suggestion
      Func<object, object> propertyFunc = obj => 42;
```

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+    }
+
+    [Test]
+    public void IsActive_WithNullAsInstanceToValidate_ReturnsFalse ()
+    {
+      var context = new ValidationContext(null);
+      var propertyInformationMock = new Mock<IPropertyInformation>();
+
+      Func<object, object> propertyFunc = p => p;
+      Func<object, bool> conditionFunc = c => c != null;
```
MichaelKetting, line 128:
```suggestion
      Func<object, bool> conditionFunc = obj => throw new AssertionException("Should not be called.");
```

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+      Assert.That(res, Is.True);
+    }
+
+    [Test]
+    public void IsActive_WithNullAsInstanceToValidate_ReturnsFalse ()
+    {
+      var context = new ValidationContext(null);
+      var propertyInformationMock = new Mock<IPropertyInformation>();
+
+      Func<object, object> propertyFunc = p => p;
```
MichaelKetting, line 127:
```suggestion
      Func<object, object> propertyFunc = obj => 42;
```

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+    }
+
+    [Test]
+    public void IsActive_ReturnsTrueFromCondition ()
+    {
+      var context = new ValidationContext("test");
+      var propertyInformationMock = new Mock<IPropertyInformation>();
+
+      Func<object, object> propertyFunc = p => p;
+      Func<object, bool> conditionFunc = _ => true;
```
MichaelKetting, line 144: You need to capture the object.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+
+      var rule = new PropertyValidationRule<object, int>(propertyInformationMock.Object, propertyFunc, null, new[] { validatorMock.Object });
+
+      var res = rule.IsActive(context);
+
+      Assert.That(res, Is.True);
+    }
+
+    [Test]
+    public void IsActive_WithNullAsInstanceToValidate_ReturnsFalse ()
```
MichaelKetting, line 122:
```suggestion
    public void IsActive_WithCondition_WithoutInstanceToValidate_ReturnsFalse ()
```

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,30 +1,194 @@
...
+      var rule = new PropertyValidationRule<object, int>(propertyInformationMock.Object, propertyFunc, null, new[] { validatorMock.Object });
+
+      var res = rule.Validate(context);
+
+      Assert.That(res, Is.EqualTo(new[] { propertyFailure }));
+      validatorMock.Verify();
+    }
+
+    [Test]
+    public void IsActive_WithConditionIsNull_ReturnsTrue ()
```
MichaelKetting, line 100:
```suggestion
    public void IsActive_WithoutCondition_ReturnsTrue ()
```

I think we are missing `public void IsActive_WithoutCondition_WithoutInstanceToValidate_ReturnsTrue ()`

MichaelKetting (changes requested):
Please check for consistency of test names, test structure based on review notes. Please also check that tests are grouped semantically.

Frazerus:
Theres currently an error here. Because one of my old branches was named wrongly, my current branch keeps track of this remote branch instead of the actual one I want it to keep track of. I fixed it now, so it shouldn't happen again.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,250 @@
...
+          .Returns(new[] { propertyFailure });
+
+
+      object propertyObj = null;
+      object propertyValueObj = null;
+      Func<object, object> propertyFunc = obj =>
+      {
+        propertyObj = obj;
+        propertyValueObj = "test";
+        return propertyValueObj;
```
Frazerus, line 98: You could just return "test" here, since `propertyValueObj` seems to not have any other uses.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,250 @@
...
```
Frazerus, line 142: Same here

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,250 @@
...
+      Assert.That(result, Is.EqualTo(new[] { propertyFailure }));
+      Assert.That(context.InstanceToValidate, Is.SameAs(propertyValidatorContext.Instance));
+      Assert.That(context.InstanceToValidate, Is.SameAs(propertyObj));
+      Assert.That(propertyValueObj, Is.EqualTo("test"));
+    }
+
+    [Test]
+    public void IsActive_WithoutCondition_WithoutInstanceToValidate_ReturnsTrue ()
+    {
+      var context = new ValidationContext(null);
```
Frazerus, line 159: This might be nitpicking at a completely over the top level, but I think you should probably move context further down in the tests where you don't need it earlier. 

It would make it easier to see that it is only used when calling the method under test.

Frazerus (approved):

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,242 @@
...
+
+      propertyInformationMock
+          .Setup(p => p.DeclaringType)
+          .Returns(declaringTypeMock.Object);
+
+      Func<object, object> propertyFunc = obj => obj;
+      Func<object, bool> conditionFunc = _ => false;
+
+      PropertyValidationRule<Customer, string> validationRule =
+          new PropertyValidationRule<Customer, string>(propertyInformationMock.Object, propertyFunc, conditionFunc, Array.Empty<IPropertyValidator>());
```
MichaelKetting, line 235:
```suggestion
      PropertyValidationRule<Customer, string> validationRule =
          new PropertyValidationRule<Customer, string>(propertyInformationMock.Object, obj => null, null, new IPropertyValidator[0]);
```
I recommend dropping unsed code.

And the array syntax: passing an empty array instead of using Array.Empty makes the code idomatically easier to read. It may be personal preference or simply due to the fact that empty arrays have been arround longer than Array.Empty so there's more pattern recognition, though. To me it simply pops more when I see the empty array during the review than Array.Empty. It's different in prod-code. There, the Array.Empty is valuable because it saves a memory allocation.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,242 @@
...
+
+      Assert.That(result, Is.True);
+    }
+
+    [Test]
+    public void IsActive_WithCondition_WithoutInstanceToValidate_ReturnsFalse ()
+    {
+      var propertyInformationMock = new Mock<IPropertyInformation>();
+
+      Func<object, object> propertyFunc = obj => 42;
```
MichaelKetting, line 170: clean up the other occurances, too?

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,242 @@
...
+
+      var rule = new PropertyValidationRule<object, int>(propertyInformationMock.Object, propertyFunc, null, new[] { validatorMock.Object });
+
+      var context = new ValidationContext(new Person());
+      var result = rule.Validate(context);
+
+      Assert.That(result, Is.EqualTo(new[] { propertyFailure }));
+      Assert.That(context.InstanceToValidate, Is.SameAs(actualPropertyValidatorContext.Instance));
+      Assert.That(context.InstanceToValidate, Is.SameAs(propertyObj));
+      Assert.That("test", Is.EqualTo(actualPropertyValidatorContext.PropertyValue));
```
MichaelKetting, line 145: please check. you flipped actual and expected. recommend you add `actual` (or `result`, depending on context) to each variable name.

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,242 @@
...
+
+      var validatorMock = new Mock<IPropertyValidator>();
+
+      PropertyValidatorContext actualPropertyValidatorContext = null;
+      validatorMock
+          .Setup(_ => _.Validate(It.IsAny<PropertyValidatorContext>()))
+          .Callback((PropertyValidatorContext ctx) => { actualPropertyValidatorContext = ctx; })
+          .Returns(new[] { propertyFailure });
+
+      object propertyObj = null;
```
MichaelKetting, line 130: `actualInstanceToValidateFromPropertyFunc`

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/Rules/PropertyValidationRuleTest.cs
```diff
@@ -1,33 +1,222 @@
...
+      object actualInstanceToValidateFromPropertyFunc = null;
+      Func<object, object> propertyFunc = obj =>
+      {
+        actualInstanceToValidateFromPropertyFunc = obj;
+        return "test";
+      };
+
+      var rule = new PropertyValidationRule<object, int>(propertyInformationMock.Object, propertyFunc, null, new[] { validatorMock.Object });
+
+      var context = new ValidationContext(new Person());
```
MichaelKetting, line 134: see Validate_WithConditionIsTrue_ReturnsValidationFailures

MichaelKetting (approved):

MichaelKetting:
@Frazerus see my last comment. a quick review when all is done should suffice.

Frazerus (approved):

PR #705, opened 2023-05-24
RM-8807 Implement tests for ObjectValidatorFactory

KuJo7:
https://re-motion.atlassian.net/browse/RM-8807

### DIFF
--- a/Remotion/Validation/Core/RuleCollectors/DeferredInitializationValidationMessage.cs
+++ b/Remotion/Validation/Core/RuleCollectors/DeferredInitializationValidationMessage.cs
@@ -1,5 +1,22 @@
-﻿using System;
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
 using System.Globalization;
+using Remotion.Utilities;
 using Remotion.Validation.Implementation;
 
 namespace Remotion.Validation.RuleCollectors
@@ -14,6 +31,8 @@ namespace Remotion.Validation.RuleCollectors
 
     public void Initialize (ValidationMessage validationMessage)
     {
+      ArgumentUtility.CheckNotNull("validationMessage", validationMessage);
+
       if (_validationMessage != null)
         throw new InvalidOperationException("Validation message has already been initialized.");
 
--- a/Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
+++ b/Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
@@ -15,29 +15,80 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using Moq;
 using NUnit.Framework;
+using Remotion.Reflection;
+using Remotion.Validation.Implementation;
+using Remotion.Validation.Rules;
+using Remotion.Validation.UnitTests.TestDomain;
+using Remotion.Validation.UnitTests.TestDomain.Validators;
+using Remotion.Validation.UnitTests.TestHelpers;
+using Remotion.Validation.Validators;
 
 namespace Remotion.Validation.UnitTests.Implementation
 {
   [TestFixture]
   public class ObjectValidatorFactoryTest
   {
-    [Ignore("RM-5906")]
     [Test]
-    public void Create ()
+    public void Create_ReturnsValidator ()
     {
+      var fakeCustomerValidator = new FakeCustomerValidator();
+      ObjectValidationRuleInitializationParameters actualInitializationParameters = null;
+      Func<ObjectValidationRuleInitializationParameters, IObjectValidator> validatorFactoryFunc = parameters =>
+      {
+        actualInitializationParameters = parameters;
+        return fakeCustomerValidator;
+      };
+
+      var validatedType = TypeAdapter.Create(typeof(Person));
+
+      var validationMessageFactoryMock = new Mock<IValidationMessageFactory>(MockBehavior.Strict);
+      validationMessageFactoryMock
+          .Setup(_ => _.CreateValidationMessageForObjectValidator(fakeCustomerValidator, validatedType))
+          .Returns(new InvariantValidationMessage("test"));
+
+      var result = ObjectValidatorFactory.Create<IObjectValidator>(validatedType, validatorFactoryFunc, validationMessageFactoryMock.Object);
+
+      Assert.That(result, Is.SameAs(fakeCustomerValidator));
+      Assert.That(actualInitializationParameters, Is.Not.Null);
+      Assert.That(actualInitializationParameters.ValidationMessage.ToString(), Is.EqualTo("test"));
     }
 
-    [Ignore("RM-5906")]
     [Test]
     public void Create_WithFactoryDelegateReturnsNull_ThrowsInvalidOperationException ()
     {
+      Func<ObjectValidationRuleInitializationParameters, IObjectValidator> factoryDelegate = r => null;
+
+      var validatedType = TypeAdapter.Create(typeof(Person));
+
+      var validationMessageFactoryMock = new Mock<IValidationMessageFactory>();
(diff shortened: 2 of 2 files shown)


Remotion/Validation/Core/RuleCollectors/DeferredInitializationValidationMessage.cs
```diff
@@ -14,6 +15,8 @@ public DeferredInitializationValidationMessage ()
 
     public void Initialize (ValidationMessage validationMessage)
     {
+      ArgumentUtility.CheckNotNull("validationMessage", validationMessage);
```
Frazerus, line 18: There is also the option to use `nameof(validationMessage)` which would be beneficial in the case of wanting to rename the parameter. I haven't seen it used that much in re-motion though, but it might be useful to know. 

@MichaelKetting is there a reason for not using `nameof`?

Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
```diff
@@ -15,29 +15,88 @@
...
+          .Setup(_ => _.FullName)
+          .Returns(It.IsAny<string>());
+
+      var validationMessageFactoryMock = new Mock<IValidationMessageFactory>();
+      validationMessageFactoryMock
+          .Setup(_ => _.CreateValidationMessageForObjectValidator(null, validatedTypeMock.Object))
+          .Returns(It.IsAny<ValidationMessage>());
+
+      Assert.That(() => ObjectValidatorFactory.Create<IObjectValidator>(validatedTypeMock.Object, factoryDelegate, validationMessageFactoryMock.Object),
+          Throws.TypeOf<InvalidOperationException>().With.Message.EqualTo("validatorFactory evaluated and returned null."));
```
Frazerus, line 76: Im used to seeing `Throws.InstanceOf` instead of `TypeOf` from my own code.

I researched and found that instanceOf also includes derivatives, while typeOf does not. Just thought I should tell you as well, in case you were wondering. 😊 (I will probably use TypeOf now instead of instanceOf...)

Frazerus (approved):

Remotion/Validation/Core/RuleCollectors/DeferredInitializationValidationMessage.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
```
Frazerus, line 1: Missing file header

Frazerus (approved):

Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
MichaelKetting: There's no clear precedent in the code base. Both are uses roughly the same. The dicsussion is correct, TypeOf is more precise and the question is, should you care when writing the assertion. For exceptions I think it's not relevant in most cases because you only care about it being a specific base type, not something more specific unless you explicitly want to test for that. This means, InstanceOf makes the test more robust.

For this particular instance, though, I recommend the shorthand `Throws.InvalidOperationException` though (underneath, it uses TypeOf, but nobody is using InvalidOpEx as a basetype)

And please move the delegate in line 75 into its own line, too. If we shart line wrapping, each parameter gets its own line, including the first one.

Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
```diff
@@ -15,29 +15,88 @@
...
     {
+      var stub = new StubObjectValidator();
+      Func<ObjectValidationRuleInitializationParameters, IObjectValidator> factoryDelegate = r => stub;
+
+      var validatedTypeMock = new Mock<ITypeInformation>();
+      validatedTypeMock
+          .Setup(_ => _.FullName)
+          .Returns("Person");
+
+      var typeAdapter = TypeAdapter.Create(typeof(Person));
```
MichaelKetting, line 90: right now, you got both variants but you use only one.

MichaelKetting (changes requested):
for completeness' sake: I only marked the first occurances when the same issue occurs multiple times.

Remotion/Validation/Core/RuleCollectors/AddingPropertyValidationRuleCollector1.cs
```diff
@@ -1,4 +1,20 @@
-﻿using System;
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
```
KuJo7, line 1: I added here also the missing file header since it was in the same namespace `Remotion.Validation.RuleCollectors`. @MichaelKetting Is that okay?

MichaelKetting: Is it just a random one and there's others or was this the last file? I do would prefer to keep those in separate PRs (don't need to review those :) so if you find typos or missing cleanup in unrelated files, you can just make a ticket and a PR and fix it that way. it keeps the original PR clean and still allows for cleanup.

Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
```diff
@@ -15,29 +15,77 @@
...
+      var validatedType = TypeAdapter.Create(typeof(Person));
+
+      var validationMessageFactoryMock = new Mock<IValidationMessageFactory>();
+      validationMessageFactoryMock
+          .Setup(_ => _.CreateValidationMessageForObjectValidator(factoryDelegate(It.IsAny<ObjectValidationRuleInitializationParameters>()), validatedType))
+          .Returns((ValidationMessage)null);
+
+      Assert.That(() => ObjectValidatorFactory.Create<IObjectValidator>(validatedType, factoryDelegate, validationMessageFactoryMock.Object),
+          Throws.TypeOf<InvalidOperationException>().With.Message
+              .EqualTo("The IValidationMessageFactory did not return a result for StubObjectValidator applied to type 'Remotion.Validation.UnitTests.TestDomain.Person'."));
```
MichaelKetting, line 88:
```suggestion
      Assert.That(
          () => ObjectValidatorFactory.Create<IObjectValidator>(validatedType, factoryDelegate, validationMessageFactoryMock.Object),
          Throws.InvalidOperationException
              .With.Message.EqualTo("The IValidationMessageFactory did not return a result for StubObjectValidator applied to type 'Remotion.Validation.UnitTests.TestDomain.Person'."));
```

Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
```diff
@@ -15,29 +15,77 @@
...
+
+      var validatedType = TypeAdapter.Create(typeof(Person));
+
+      var validationMessageFactoryMock = new Mock<IValidationMessageFactory>();
+      validationMessageFactoryMock
+          .Setup(_ => _.CreateValidationMessageForObjectValidator(null, validatedType))
+          .Returns(It.IsAny<ValidationMessage>());
+
+      Assert.That(() => ObjectValidatorFactory.Create<IObjectValidator>(validatedType, factoryDelegate, validationMessageFactoryMock.Object),
+          Throws.TypeOf<InvalidOperationException>().With.Message.EqualTo("validatorFactory evaluated and returned null."));
```
MichaelKetting, line 70:
```suggestion
      Assert.That(
          () => ObjectValidatorFactory.Create<IObjectValidator>(validatedType, factoryDelegate, validationMessageFactoryMock.Object),
          Throws.InvalidOperationException.With.Message.EqualTo("validatorFactory evaluated and returned null."));
```

Remotion/Validation/UnitTests/Implementation/ObjectValidatorFactoryTest.cs
```diff
@@ -15,29 +15,77 @@
...
+
+      var validationMessageFactoryMock = new Mock<IValidationMessageFactory>(MockBehavior.Strict);
+      validationMessageFactoryMock
+          .Setup(_ => _.CreateValidationMessageForObjectValidator(fakeCustomerValidator, validatedType))
+          .Returns(new InvariantValidationMessage("test"));
+
+      var result = ObjectValidatorFactory.Create<IObjectValidator>(validatedType, validatorFactoryFunc, validationMessageFactoryMock.Object);
+
+      Assert.That(result, Is.SameAs(fakeCustomerValidator));
+      Assert.That(actualInitializationParameters.ValidationMessage.ToString(), Is.EqualTo("test"));
```
MichaelKetting, line 54:
```suggestion
      Assert.That(result, Is.SameAs(fakeCustomerValidator));
      Assert.That(actualInitializationParameters, Is.Not.Null);
      Assert.That(actualInitializationParameters.ValidationMessage.ToString(), Is.EqualTo("test"));
```

This is actually a good practice in these cases becasue it gives a clear error in case the call didn't happen / the parameter was left null. Otherwise, we get a NullRefEx in the build and that's more annoying to figure out.

MichaelKetting (changes requested):

Remotion/Validation/Core/RuleCollectors/AddingPropertyValidationRuleCollector1.cs
KuJo7: yes I think it was the last one.  I will revert this change and make a ticket

PR #706, opened 2023-05-25
RM-8803 Dispatch BocList validation failures to the repository

Frazerus:
https://re-motion.atlassian.net/browse/RM-8803

### DIFF
--- a/Remotion/ObjectBinding/Core/Validation/IBusinessObjectValidationResult.cs
+++ b/Remotion/ObjectBinding/Core/Validation/IBusinessObjectValidationResult.cs
@@ -28,8 +28,9 @@ namespace Remotion.ObjectBinding.Validation
         [NotNull] IBusinessObjectProperty businessObjectProperty,
         bool markAsHandled);
 
-    IReadOnlyCollection<UnhandledBusinessObjectValidationFailure> GetUnhandledValidationFailures (
+    IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailures (
         [NotNull] IBusinessObject businessObject,
-        bool includePartiallyHandledFailures = false);
+        bool includePartiallyHandledFailures = false,
+        bool markAsHandled = false);
   }
 }
--- a/Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
+++ b/Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
@@ -225,6 +225,117 @@ namespace Remotion.ObjectBinding.Validation.UnitTests
     }
 
     [Test]
+    public void GetValidationFailuresForBusinessObject_ReturnsMatchingValidationFailures ()
+    {
+      var person = new Person();
+      person.FirstName = null;
+      person.LastName = "";
+      person.PhoneNumber = null;
+      ((ICustomer)person).CustomerNumber = "";
+
+      var validationResult = _personValidator.Validate(person);
+      Assert.That(validationResult.IsValid, Is.False);
+      Assert.That(validationResult.Errors.Count, Is.EqualTo(4));
+      Assert.That(
+          validationResult.Errors.OfType<PropertyValidationFailure>().Select(f => f.ValidatedProperty).Distinct().Count(),
+          Is.EqualTo(4));
+
+      var businessObjectValidationResult = BusinessObjectValidationResult.Create(validationResult);
+
+      var businessObjectValidationFailures =
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, false, false).ToArray();
+
+      var firstNamePropertyFailures = businessObjectValidationFailures.Single(f => f.ValidatedProperty == _firstNameBusinessObjectProperty);
+      Assert.That(firstNamePropertyFailures.ValidatedObject, Is.SameAs(person));
+      Assert.That(firstNamePropertyFailures.ValidatedProperty, Is.SameAs(_firstNameBusinessObjectProperty));
+      Assert.That(firstNamePropertyFailures.ErrorMessage, Is.EqualTo("NotNullValidator: Validation error."));
+
+      var lastNamePropertyFailures = businessObjectValidationFailures.Single(f => f.ValidatedProperty == _lastNameBusinessObjectProperty);
+      Assert.That(lastNamePropertyFailures.ValidatedObject, Is.SameAs(person));
+      Assert.That(lastNamePropertyFailures.ValidatedProperty, Is.SameAs(_lastNameBusinessObjectProperty));
+      Assert.That(lastNamePropertyFailures.ErrorMessage, Is.EqualTo("NotEmptyOrWhitespaceValidator: Validation error."));
+
+      var phoneNumberValidationFailures = businessObjectValidationFailures.Single(f => f.ValidatedProperty == _phoneNumberBusinessObjectProperty);
+      Assert.That(phoneNumberValidationFailures.ValidatedObject, Is.SameAs(person));
+      Assert.That(phoneNumberValidationFailures.ValidatedProperty, Is.SameAs(_phoneNumberBusinessObjectProperty));
+      Assert.That(phoneNumberValidationFailures.ErrorMessage, Is.EqualTo("NotNullValidator: Validation error."));
+
+      var customerNumberValidationFailures = businessObjectValidationFailures.Single(f => f.ValidatedProperty == _customerNumberBusinessObjectProperty);
+      Assert.That(customerNumberValidationFailures.ValidatedObject, Is.SameAs(person));
+      Assert.That(customerNumberValidationFailures.ValidatedProperty, Is.SameAs(_customerNumberBusinessObjectProperty));
+      Assert.That(customerNumberValidationFailures.ErrorMessage, Is.EqualTo("NotEmptyOrWhitespaceValidator: Validation error."));
+    }
+
+    [Test]
+    public void GetValidationFailuresForBusinessObject_WithPartiallyHandledAndMarkAsHandledTrueOrFalse_ReturnsMatchingValidationFailuresOnEachCall ()
+    {
+      var person = new Person();
+      person.FirstName = null;
+      person.LastName = "";
+      person.PhoneNumber = null;
+      ((ICustomer)person).CustomerNumber = "";
+
+      var validationResult = _personValidator.Validate(person);
+      Assert.That(validationResult.IsValid, Is.False);
+      Assert.That(validationResult.Errors.Count, Is.EqualTo(4));
+      Assert.That(
+          validationResult.Errors.OfType<PropertyValidationFailure>().Select(f => f.ValidatedProperty).Distinct().Count(),
+          Is.EqualTo(4));
+
+      var businessObjectValidationResult = BusinessObjectValidationResult.Create(validationResult);
+
+      var businessObjectValidationFailures =
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, true, markAsHandled: false).ToArray();
+      Assert.That(businessObjectValidationFailures.Length, Is.EqualTo(4));
+      Assert.That(businessObjectValidationFailures[0].ValidatedObject, Is.SameAs(person));
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, true, markAsHandled: false),
+          Is.EquivalentTo(businessObjectValidationFailures));
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, true, markAsHandled: false),
+          Is.EquivalentTo(businessObjectValidationFailures));
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, true, markAsHandled: true),
+          Is.EquivalentTo(businessObjectValidationFailures));
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, true, markAsHandled: true),
+          Is.Empty);
+    }
+
+    [Test]
+    public void GetValidationFailuresForBusinessObject_WithoutPartiallyHandledAndMarkAsHandledTrueOrFalse_ReturnsOnlyUnmarkedValidationFailures ()
+    {
+      var person = new Person();
+      person.FirstName = null;
+      person.LastName = "";
+      person.PhoneNumber = null;
+      ((ICustomer)person).CustomerNumber = "";
+
+      var validationResult = _personValidator.Validate(person);
+      Assert.That(validationResult.IsValid, Is.False);
+      Assert.That(validationResult.Errors.Count, Is.EqualTo(4));
+      Assert.That(
+          validationResult.Errors.OfType<PropertyValidationFailure>().Select(f => f.ValidatedProperty).Distinct().Count(),
+          Is.EqualTo(4));
+
+      var businessObjectValidationResult = BusinessObjectValidationResult.Create(validationResult);
+
+      var businessObjectValidationFailures =
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, false, markAsHandled: false).ToArray();
+      Assert.That(businessObjectValidationFailures.Length, Is.EqualTo(4));
+      Assert.That(businessObjectValidationFailures[0].ValidatedObject, Is.SameAs(person));
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, false, markAsHandled: true),
+          Is.EquivalentTo(businessObjectValidationFailures));
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, false, markAsHandled: false),
+          Is.Empty);
+      Assert.That(
+          businessObjectValidationResult.GetUnhandledValidationFailures(person, false, markAsHandled: true),
+          Is.Empty);
+    }
+
+    [Test]
     public void GetUnhandledValidationFailures_ForBusinessObject_ReturnsFilteredResult ()
     {
       var person1 = new Person();
--- a/Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
+++ b/Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
@@ -169,27 +169,34 @@ namespace Remotion.ObjectBinding.Validation
         result.Add(businessObjectValidationFailure);
 
         if (markAsHandled)
-        {
-          _unhandledValidationFailures.Remove((businessObjectValidationFailure, validationFailure));
-          _handledValidationFailures.Add(validationFailure);
-        }
+          MoveFailureToHandled(businessObjectValidationFailure, validationFailure);
       }
 
       return result;
     }
 
-    public IReadOnlyCollection<UnhandledBusinessObjectValidationFailure> GetUnhandledValidationFailures (
-        IBusinessObject businessObject,
-        bool includePartiallyHandledFailures = false)
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailures (
+        [NotNull] IBusinessObject businessObject,
+        bool includePartiallyHandledFailures = false,
+        bool markAsHandled = false)
     {
-      ArgumentUtility.CheckNotNull("businessObject", businessObject);
+      ArgumentUtility.CheckNotNull(nameof(businessObject), businessObject);
 
-      return _unhandledValidationFailures
+      var validationFailures = _unhandledValidationFailures
           .Where(f => Equals(f.BusinessObjectValidationFailure.ValidatedObject, businessObject))
           .Where(f => includePartiallyHandledFailures || !_handledValidationFailures.Contains(f.ValidationFailure))
-          .Select(f => f.BusinessObjectValidationFailure)
-          .Select(f => new UnhandledBusinessObjectValidationFailure(f.ErrorMessage, f.ValidatedProperty))
           .ToArray();
+
+      var result = new List<BusinessObjectValidationFailure>();
+      foreach (var (businessObjectValidationFailure, validationFailure) in validationFailures)
+      {
+        result.Add(businessObjectValidationFailure);
+
+        if (markAsHandled)
+          MoveFailureToHandled(businessObjectValidationFailure, validationFailure);
+      }
+
+      return result;
     }
 
     public IReadOnlyCollection<ValidationFailure> GetUnhandledValidationFailures (bool includePartiallyHandledFailures = false)
@@ -199,6 +206,12 @@ namespace Remotion.ObjectBinding.Validation
           .Select(f => f.ValidationFailure)
           .Distinct()
           .ToArray();
+    }
+
+    private void MoveFailureToHandled (BusinessObjectValidationFailure businessObjectValidationFailure, ValidationFailure validationFailure)
+    {
+      _unhandledValidationFailures.Remove((businessObjectValidationFailure, validationFailure));
+      _handledValidationFailures.Add(validationFailure);
     }
   }
 }
--- a/Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx
+++ b/Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx
@@ -220,9 +220,12 @@
   <asp:button id="ChildrenListSetPageButton" runat="server" Text="Set Page to 1st"></asp:button>
   <asp:button id="ChildrenListAddRowButton" runat="server" Text="Add"></asp:button>
   <asp:button id="ChildrenListRemoveRowsButton" runat="server" Text="Remove Selected"></asp:button>
+  <asp:button id="ChildrenAddValidationErrorsButton" runat="server" Text="Add validation errors"></asp:button>
 </p>
 <p>
   <asp:checkbox id=ChildrenListEventCheckBox runat="server" Text="ChildrenList Event raised" enableviewstate="False" Enabled="False"></asp:checkbox>
+  <br>
+  <asp:Label id="UnhandledValidationErrorsLabel" runat="server" ForeColor="Red" style="white-space: pre"></asp:Label>
 </p>
 <p>
   <asp:label id=ChildrenListEventArgsLabel runat="server" enableviewstate="False"></asp:label>
--- a/Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
+++ b/Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
@@ -16,14 +16,19 @@
 // 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Web.UI.HtmlControls;
 using System.Web.UI.WebControls;
 using Remotion.ObjectBinding;
 using Remotion.ObjectBinding.Sample;
+using Remotion.ObjectBinding.Validation;
 using Remotion.ObjectBinding.Web.UI.Controls;
-using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
 using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+using Remotion.Reflection;
+using Remotion.Utilities;
+using Remotion.Validation.Results;
 using Remotion.Web;
+using Remotion.Web.UI;
 using Remotion.Web.UI.Controls;
 using Remotion.Web.UI.Globalization;
 
@@ -45,8 +50,10 @@ public class BocListUserControl : BaseUserControl
   protected Button ChildrenListSetPageButton;
   protected Button ChildrenListAddRowButton;
   protected Button ChildrenListRemoveRowsButton;
+  protected Button ChildrenAddValidationErrorsButton;
   protected CheckBox ChildrenListEventCheckBox;
   protected Label ChildrenListEventArgsLabel;
+  protected Label UnhandledValidationErrorsLabel;
   protected FormGridManager FormGridManager;
   protected BindableObjectDataSourceControl EmptyDataSourceControl;
   protected HtmlTable FormGrid;
@@ -66,6 +73,8 @@ public class BocListUserControl : BaseUserControl
     ChildrenListSetPageButton.Click += ChildrenListSetPageButton_Click;
     ChildrenListAddRowButton.Click += ChildrenListAddRowButton_Click;
     ChildrenListRemoveRowsButton.Click += ChildrenListRemoveRowsButton_Click;
+
+    ChildrenAddValidationErrorsButton.Click += ChildrenAddValidationErrorsButtonClick;
 
     ChildrenList.ListItemCommandClick += new BocListItemCommandClickEventHandler(ChildrenList_ListItemCommandClick);
     ChildrenList.MenuItemClick += new WebMenuItemClickEventHandler(ChildrenList_MenuItemClick);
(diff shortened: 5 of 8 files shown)


Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -33,10 +33,17 @@ public BocListValidationResultDispatchingValidator ()
...
+        DispatchPreviousImplementation(validationResult, bocListControl);
+      else
+        DispatchNewImplementation(validationResult, bocListControl);
+    }
 
+    private void DispatchPreviousImplementation (IBusinessObjectValidationResult validationResult, BocList bocListControl)
+    {
       var validatorsMatchingToControls = EnumerableUtility.SelectRecursiveDepthFirst(
               bocListControl as Control,
               child => child.Controls.Cast<Control>().Where(item => !(item is INamingContainer)))
```
KuJo7, line 49:
```suggestion
              child => child.Controls.Cast<Control>().Where(item => item is not INamingContainer))
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -46,6 +53,44 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+        {
+          var matched = columnAndMatcher.Matcher.GetMatchingValidationFailures(rowObject, validationResult);
+          if (matched.Count > 0)
+            validationFailureRepository.AddCellValidationFailures(rowObject, columnAndMatcher.Column, matched);
+        }
+
+        var otherRowFailures = validationResult.GetValidationFailures(rowObject);
+        if (otherRowFailures.Count > 0)
+          validationFailureRepository.AddRowValidationFailures(rowObject, otherRowFailures);
+      }
```
KuJo7, line 83:
```suggestion
      }
      
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -46,6 +53,44 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+      var businessObject = bocListControl.DataSource?.BusinessObject;
+      var property = bocListControl.Property;
+      if (businessObject != null && property != null)
+      {
+        var listPropertyFailures = validationResult.GetValidationFailures(businessObject, property, true);
+        if(listPropertyFailures.Count > 0)
```
KuJo7, line 89:
```suggestion
        if (listPropertyFailures.Count > 0)
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -33,10 +33,17 @@ public BocListValidationResultDispatchingValidator ()
 
     public void DispatchValidationFailures (IBusinessObjectValidationResult validationResult)
     {
-      ArgumentUtility.CheckNotNull("validationResult", validationResult);
+      ArgumentUtility.CheckNotNull(nameof(validationResult), validationResult);
```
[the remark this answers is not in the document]
Frazerus: Can do here, but I really want to start using the nameof syntax ;)

@MichaelKetting  should I create a Task to fix all argument checks to use the nameof syntax? And should this include some sort of update to the templates? Or is this just overkill?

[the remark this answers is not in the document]
MichaelKetting: yeah, you can use it here. and yes, that would include the update of the template. please create a task. there's re-sharper structural refactoring or a roslyn rewrite. there will be a couple of failures (we have some special names used until now) but I think it's a pretty safe & straight forward update. you want to check with @a-ctor because it will make merge conflicts with the relations to interface branch so we might want to wait till that's integrated (or not, you two need to figure that out). and we need to scedule it so that it's not gonna break the rest of the release timeline because I'm sure it's not just 1h of work. don't know if it's 4h, 8h or several days, though.

KuJo7 (approved):

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,79 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+
+      var validationFailureRepository = bocListControl.ValidationFailureRepository;
+      validationFailureRepository.ClearAllValidationFailures();
+
+      foreach (var rowObject in rowObjects)
+      {
+        foreach (var columnAndMatcher in columnsAndMatchers)
+        {
+          var matched = columnAndMatcher.Matcher.GetMatchingValidationFailures(rowObject, validationResult);
+          BocColumnDefinition columnDefinition = columnAndMatcher.Column switch
```
a-ctor, line 82: We don't need to do any fancy checks here we can cast to `BocColumnDefinition` directly. Even better, we can add the cast when we create the anonymous object `new { Column = (BocColumnDefinition)c, ... }`. We know that this cast is guaranteed to succeed as `.GetColumnDefinitions()` returns an `IReadOnlyList<BocColumnDefinition>`.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -151,6 +151,32 @@ private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObject
       _handledValidationFailures = new HashSet<ValidationFailure>();
     }
 
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetValidationFailures (
```
a-ctor, line 154: we should add tests for this new method.

a-ctor (changes requested):

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
```diff
@@ -392,6 +400,186 @@ private void ChildrenListRemoveRowsButton_Click (object sender, EventArgs e)
...
+                    "Bad last child.Partner",
+                    "Localized bad last child.Partner"),
+                new ObjectValidationFailure(
+                    lastChild,
+                    new[]
+                    {
+                        new ValidatedProperty(lastChild, personLastNameProperty),
+                    },
+                    "Bad last child name",
+                    "Localized bad last child names"),
```
a-ctor, line 547: Here is the answer to the problem we kept debugging. Please check if we have other cases like this. Apart from that the reporting we experienced is correct. As we have a compound validation failure below, some validation failures are (correctly) reported multiple times.

```suggestion
                    "Localized bad last child name"),
```

a-ctor (approved):

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
```diff
@@ -224,6 +224,121 @@ public void GetValidationFailures_WithMarkHasHandledTrueOrFalse_ReturnsMatchingV
...
+      person.FirstName = null;
+      person.LastName = "";
+      person.PhoneNumber = null;
+      ((ICustomer)person).CustomerNumber = "";
+
+      var validationResult = _personValidator.Validate(person);
+      Assert.That(validationResult.IsValid, Is.False);
+      Assert.That(validationResult.Errors.Count, Is.EqualTo(4));
+      Assert.That(
+          validationResult.Errors.OfType<PropertyValidationFailure>().Select(f => f.ValidatedProperty).Distinct().Count(),
```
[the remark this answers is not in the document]
Frazerus: I think distinctBy is not supported by dotnet framework :(

Remotion/ObjectBinding/Core/Validation/IBusinessObjectValidationResult.cs
```diff
@@ -23,6 +23,11 @@ namespace Remotion.ObjectBinding.Validation
   /// <threadsafety static="true" instance="false" />
   public interface IBusinessObjectValidationResult
   {
+    IReadOnlyCollection<BusinessObjectValidationFailure> GetValidationFailuresForBusinessObject (
+        [NotNull] IBusinessObject businessObject,
+        bool includePartiallyHandledFailures = false,
+        bool markAsHandled = true);
+
```
MichaelKetting, line 30: The new API doesn't fit into the existing contract:

Naming: I think `GetValidationFailures` should be sufficient considering the existing methods on this interface. Not specifying a BusinessObjectProperty implies we get all failures for the BusienssObject, regardless of the property (see `GetUnhandledValidationFailures`)

`GetValidationFailuresForBusinessObject` should not have a parameter `includePartiallyHandledFailures`. Instead, it will always return all validation failures. If, on the other hand, the intention is to get unhandled failures, we should add this into the name, see `GetUnhandledValidationFailures` and drop `markAsHandled`. 

It's very likely that this comment doesn't cover the actually intended scenario / design and we need to discuss how to best design the API so it works within the parameters of the existing API or how the existing API should/need to be updated.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx
```diff
@@ -220,9 +220,12 @@
   <asp:button id="ChildrenListSetPageButton" runat="server" Text="Set Page to 1st"></asp:button>
   <asp:button id="ChildrenListAddRowButton" runat="server" Text="Add"></asp:button>
   <asp:button id="ChildrenListRemoveRowsButton" runat="server" Text="Remove Selected"></asp:button>
+  <asp:button id="ChildrenAddValidationErrorsButton" runat="server" Text="Add validation errors"></asp:button>
 </p>
 <p>
   <asp:checkbox id=ChildrenListEventCheckBox runat="server" Text="ChildrenList Event raised" enableviewstate="False" Enabled="False"></asp:checkbox>
+  <br>
+  <asp:Label id="UnhandledValidationErrorsLabel" runat="server" ForeColor="Red" style="white-space: pre"></asp:Label>
```
MichaelKetting, line 228: I'm assuming this is just a temporary fix for something and will get replaced later? We should add a todo-comment.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
```diff
@@ -392,6 +400,186 @@ private void ChildrenListRemoveRowsButton_Click (object sender, EventArgs e)
...
+    {
+      return failure switch
+      {
+        PropertyValidationFailure propertyValidationFailure => $"Unhandled property validation failure '{propertyValidationFailure.ErrorMessage}'"
+                                                               + $" for property '{propertyValidationFailure.ValidatedProperty.Name}'"
+                                                               + $" on object '{propertyValidationFailure.ValidatedObject}'.",
+        ObjectValidationFailure objectValidationFailure => $"Unhandled object validation failure '{objectValidationFailure.ErrorMessage}'"
+                                                           + $" for properties '{string.Join(", ", objectValidationFailure.ValidatedProperties.Select(e => e.Property.Name))}'"
+                                                           + $" on object '{objectValidationFailure.ValidatedObject}'.",
+        _ => $"Unhandled property validation failure '{failure.ErrorMessage}' on object '{failure.ValidatedObject}'."
```
MichaelKetting, line 578: We don't yet have a pre-existing code style for this particular case. Based on other rules, I think there should be a line break after the arrow (most likely it is "after the arrow" and not "before the arrow", given how we format anonymous method declarations with a statement-block body.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
```diff
@@ -392,6 +400,186 @@ private void ChildrenListRemoveRowsButton_Click (object sender, EventArgs e)
...
+                    {
+                        new ValidatedProperty(lastChild, personLastNameProperty),
+                        new ValidatedProperty(lastChild, personFullNameProperty)
+                    },
+                    "Bad last child names",
+                    "Localized bad last child names"),
+            }));
+
+    PrepareValidation();
+    FormGridManager.PrepareValidation();
```
MichaelKetting, line 560: I don't think we should have to use FormGridManager.PrepareValidation(). I don't use it in other places, either, so it should be covered by this.PrepareValidation(). What we are missign is Page.PrepareValidation() since you're doing this from inside the UserControl. I would recommend trying

```suggestion
    ((ISmartPage)Page).PrepareValidation();
```

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
```diff
@@ -392,6 +400,186 @@ private void ChildrenListRemoveRowsButton_Click (object sender, EventArgs e)
...
+    DataSourceValidationResultDispatchingValidator.DispatchValidationFailures(validationResult);
+    DataSourceValidationResultDispatchingValidator.Validate();
+
+    UnhandledValidationErrorsLabel.Text = string.Join(
+        Environment.NewLine,
+        validationResult.GetUnhandledValidationFailures().Select(FormatValidationFailure));
+
+    static string FormatValidationFailure (ValidationFailure failure)
```
MichaelKetting, line 568: This method doesn't seem to get used?

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,71 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+      if (businessObject == null)
+        return;
+      if (property == null)
+        return;
+
+      var listPropertyFailures = validationResult.GetValidationFailures(businessObject, property, true);
+      if (listPropertyFailures.Count > 0)
+        validationFailureRepository.AddValidationFailuresForBocList(listPropertyFailures);
+    }
+
```
MichaelKetting, line 103: This looks like extra whitespace?

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,71 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
 
       var bocListControl = GetControlToValidate();
+      var dispatchToValidationFailureRepository = bocListControl
+          .GetColumnDefinitions()
+          .Any(cd => cd is BocValidationErrorIndicatorColumnDefinition);
 
+      if (dispatchToValidationFailureRepository)
+        DispatchToValidationFailureRepository(validationResult, bocListControl);
+      else
+        DispatchToValidationResultDispatchers(validationResult, bocListControl);
```
MichaelKetting, line 46: I'm not 100% sure if this is correct. I think we may need to dispatch both to the nested controls and to the repostiory? Without the dispatch to the nested controls, how will the inline edit controls get their validation errors?

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,71 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+      validationFailureRepository.ClearAllValidationFailures();
+
+      var rowObjects = bocListControl.Value?.Cast<IBusinessObject>();
+
+      if (rowObjects != null)
+      {
+        var columnsAndMatchers = bocListControl
+            .GetColumnDefinitions()
+            .OfType<IBocColumnDefinitionWithValidationSupport>()
+            .Select(c => new { Column = (BocColumnDefinition)c, Matcher = c.GetValidationFailureMatcher() })
```
MichaelKetting, line 72: These days  (modern C# etc), should we use ValueTuple isntead of Tuple in cases such as this to optimize allocatons a bit?

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,71 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+          {
+            var matchedFailures = columnAndMatcher.Matcher.GetMatchingValidationFailures(rowObject, validationResult);
+
+            if (matchedFailures.Count > 0)
+              validationFailureRepository.AddValidationFailuresForDataCell(rowObject, columnAndMatcher.Column, matchedFailures);
+          }
+
+          var otherRowFailures = validationResult.GetValidationFailuresForBusinessObject(rowObject);
+          if (otherRowFailures.Count > 0)
+            validationFailureRepository.AddValidationFailuresForDataRow(rowObject, otherRowFailures);
```
MichaelKetting, line 87: Formitting consistency: I think we should remove the blank line in line 80. It looks weirder if we have a blank line in line 80 and line 86 than without blank lines.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,71 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
+
+      var businessObject = bocListControl.DataSource?.BusinessObject;
+      var property = bocListControl.Property;
+
+      if (businessObject == null)
+        return;
+      if (property == null)
+        return;
+
+      var listPropertyFailures = validationResult.GetValidationFailures(businessObject, property, true);
```
MichaelKetting, line 99: please use a parameter name to better document the code. it's somehting I try to do with any non-obvious boolean these days (modern c#)

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -36,16 +36,71 @@ public void DispatchValidationFailures (IBusinessObjectValidationResult validati
...
       foreach (var validator in validatorsMatchingToControls)
         validator.DispatchValidationFailures(validationResult);
     }
 
+    private void DispatchToValidationFailureRepository (IBusinessObjectValidationResult validationResult, BocList bocListControl)
+    {
+      var validationFailureRepository = bocListControl.ValidationFailureRepository;
+      validationFailureRepository.ClearAllValidationFailures();
+
+      var rowObjects = bocListControl.Value?.Cast<IBusinessObject>();
```
MichaelKetting, line 65: Why is the cast necessary? The Value property already returns an IReadOnlyList<IBusinessObject> ?

Remotion/ObjectBinding/Validation.UnitTests/BusinessObjectValidationResultTest.cs
```diff
@@ -224,6 +224,117 @@ public void GetValidationFailures_WithMarkHasHandledTrueOrFalse_ReturnsMatchingV
...
+      Assert.That(
+          businessObjectValidationResult.GetValidationFailuresForBusinessObject(person, true, markAsHandled: true),
+          Is.EquivalentTo(businessObjectValidationFailures));
+      Assert.That(
+          businessObjectValidationResult.GetValidationFailuresForBusinessObject(person, true, markAsHandled: true),
+          Is.EquivalentTo(businessObjectValidationFailures));
+    }
+
+   [Test]
+    public void GetValidationFailuresForBusinessObject_WithoutPartiallyHandledAndMarkAsHandledTrueOrFalse_ReturnsOnlyUnmarkedValidationFailures ()
```
MichaelKetting, line 306: Formatting

MichaelKetting (changes requested):
@a-ctor I got a couple of design questions.

Remotion/ObjectBinding/Core/Validation/IBusinessObjectValidationResult.cs
a-ctor: I discussed this issue with mike and we create a new design for the functionality we need. Right now, the method is very orthogonal to the rest of the design. To fix this, we are going to extend the `GetUnhandledValidationFailures` methods with the `markAsHandled` functionality.

Another problem is the fact that `GetUnhandledValidationFailures` returns an `UnhandledBusinessObjectValidationFailure` and not a `BusinessObjectValidationFailure` that we need. We decided to simply change the return type here as the same properties are available so the only break are explicit variable type usages.

Our current implementation of `GetValidationFailuresForBusinessObject` should already do what we want, so we can use it to replace the `GetUnhandledValidationFailures` method. Please check again if the logic matches up and we don't break anything by replacing the method. We'll also need a breaking change ticket to document the changed return type of the `GetUnhandledValdiationFailures` method.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx
a-ctor: discussed with mike: label is okay as is, we just need to add a header text that indicates that there should not be any unhandled validation failures and that this is probably a bug

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
a-ctor: resolved as the usage is visible two lines above ;)

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
a-ctor: discussed with mike: we want the old logic and, if enabled, the new logic so line 46 becomes unconditional and is done before reporting to the repo

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
a-ctor: Using a value tuple in a LINQ method is a bit annoying as it cannot detect the type correctly and the naming stuff won't work out of the box. In this case I would keep the anonymous type, these short-lived 15 allocations are nothing against other stuff.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
a-ctor: I think we should discuss this general problem as a team since the IDEs already add the inline tooltip already:

image

Adding an explicit parameter name then only has the benefit of being easier to read without an IDE (e.g. in Github reviews)

MichaelKetting: Exactly. Also, for those who disable this settng in the IDE because they find it distracting. Having this in case of multiple booleans is incredibly helpful when guarding against copy&paste errors, broken refactorings, and api upgrades. E.g. if there is an extra boolean parameter added in a method and you do not use the automated refactoring (or if you use a library which introduces a new parameter or changes a signature), the information which boolean we referred to is available right there in the code and will fail or do the correct thing. All these little guards go away if you have the IDE supply the information inline. 

I did anwser here because it popped up in my inbox, it's not meant to start a lengthy async discussion.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
Frazerus: I could not make this work. Found no method called prepareValidation on ISmartPage or anything similar to that.

MichaelKetting: Hmm...please check Form.aspx.cs, line 105. That's the one we want to call.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -178,18 +178,30 @@ private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObject
...
+    public IReadOnlyCollection<BusinessObjectValidationFailure> GetUnhandledValidationFailures (
+        [NotNull] IBusinessObject businessObject,
+        bool includePartiallyHandledFailures = false,
+        bool markAsHandled = false)
     {
-      ArgumentUtility.CheckNotNull("businessObject", businessObject);
+      ArgumentUtility.CheckNotNull(nameof(businessObject), businessObject);
 
-      return _unhandledValidationFailures
+      var validationFailures = _validationFailures
```
MichaelKetting, line 188: This one here changes the semantic. we need to keep using _unhandledValidationFailures to only get those that are still unhandled.

The intended semantic of the originel code was:
One ValidationFailure can result in multiple BusinessObjectValidationFailures.
include-partially-handled=true means we return BusinessObjectValidationFailures where any part of ValidationFailure was already handled.

With the semantic as it is written in the updated version, we use the complete set of known BusinessObjectValidationFailures as the jump-off point, including those already addressed by an earlier call.

Unless you already know the answer, we may need to discuss with @a-ctor the intended semantic to clear this up.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -178,18 +178,30 @@ private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObject
...
+
+      var result = new List<BusinessObjectValidationFailure>();
+      foreach (var (businessObjectValidationFailure, validationFailure) in validationFailures)
+      {
+        result.Add(businessObjectValidationFailure);
+
+        if (markAsHandled)
+        {
+          _unhandledValidationFailures.Remove((businessObjectValidationFailure, validationFailure));
+          _handledValidationFailures.Add(validationFailure);
```
MichaelKetting, line 200: Not sure, should we remove the code duplication that stems from these two lines (they are also used in GetValidationFailuers)

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
```diff
@@ -392,6 +401,189 @@ private void ChildrenListRemoveRowsButton_Click (object sender, EventArgs e)
...
+                    {
+                        new ValidatedProperty(lastChild, personLastNameProperty),
+                        new ValidatedProperty(lastChild, personFullNameProperty)
+                    },
+                    "Bad last child names",
+                    "Localized bad last child names"),
+            }));
+
+    PrepareValidation();
+    ((SmartPage)Page).PrepareValidation();
```
MichaelKetting, line 561:
```suggestion
    ((SmartPage)Page).PrepareValidation();
```

Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Web.UI.WebControls;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering;
+using Remotion.ServiceLocation;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls
+{
+  /// <summary>
+  /// A column definition for displaying a validation error indicator for a row.
+  /// </summary>
+  public class BocValidationErrorIndicatorColumnDefinition : BocColumnDefinition
```
MichaelKetting, line 27: this is part of a separate PR? I think you may have the wrong base right now?

MichaelKetting (changes requested):
didn't review the tests before clearing up the open questions.

Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
Frazerus: The proper implementation for this class will be/ is added in a separate issue, we need the knowledge of the type here already though. At least that's what I think this was about. @a-ctor?

MichaelKetting: @Frazerus okay, thanks.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -200,5 +210,12 @@ public IReadOnlyCollection<ValidationFailure> GetUnhandledValidationFailures (bo
           .Distinct()
           .ToArray();
     }
+
+    private void MoveFailureToHandled (BusinessObjectValidationFailure businessObjectValidationFailure, ValidationFailure validationFailure)
+    {
+
```
MichaelKetting, line 216:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
Frazerus: My last answer was not 100% correct. This does belong to this changeset, it does however require the BoValidationErrorIndicatorColumnRenderer from RM-8802. Will change this to a nullcolumnrenderer for the time being.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -170,26 +170,36 @@ private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObject
...
+      var result = new List<BusinessObjectValidationFailure>();
+      foreach (var (businessObjectValidationFailure, validationFailure) in validationFailures)
+      {
+        result.Add(businessObjectValidationFailure);
+
+        if (markAsHandled)
+        {
+          MoveFailureToHandled(businessObjectValidationFailure, validationFailure);
+        }
```
MichaelKetting, line 199:
```suggestion
        if (markAsHandled)
          MoveFailureToHandled(businessObjectValidationFailure, validationFailure);
```
Since we're touching it anyway... 
Suggestion, it just looks "cleaner" to me.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -170,26 +170,36 @@ private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObject
 
         if (markAsHandled)
         {
```
MichaelKetting, line 172:
```suggestion
        if (markAsHandled)
```

Since we're touching it anyway... 
Suggestion, it just looks "cleaner" to me.

Remotion/ObjectBinding/Validation/BusinessObjectValidationResult.cs
```diff
@@ -170,26 +170,36 @@ private static IEnumerable<BusinessObjectValidationFailure> CreateBusinessObject
...
-          _unhandledValidationFailures.Remove((businessObjectValidationFailure, validationFailure));
-          _handledValidationFailures.Add(validationFailure);
+          MoveFailureToHandled(businessObjectValidationFailure, validationFailure);
         }
```
MichaelKetting, line 174:
```suggestion
```

Since we're touching it anyway... 
Suggestion, it just looks "cleaner" to me.

Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
```diff
@@ -0,0 +1,43 @@
...
+  {
+    public BocValidationErrorIndicatorColumnDefinition ()
+    {
+      Width = new Unit(3, UnitType.Em);
```
MichaelKetting, line 31: @a-ctor Drop this in this PR?
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /// <inheritdoc />
+    protected override IBocColumnRenderer GetRendererInternal (IServiceLocator serviceLocator)
+    {
+      throw new NotImplementedException();
```
MichaelKetting: Recommendation: in cases such as this, I like to add a message of "TODO <ticket>" into the NotImplEx.

MichaelKetting (approved):
A couple of simple formatting things. I don't need to re-see it. I can click merge during the afternoon if you don't want to wait for TC.

PR #708, opened 2023-05-26
RM-8808 Implement tests for FallbackValidationMessageFactory

KuJo7:
https://re-motion.atlassian.net/browse/RM-8808


Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
     public void CreateValidationMessageForPropertyValidator ()
     {
+      var propertyValidatorMock = new Mock<IPropertyValidator>();
+      var validatedPropertyMock = new Mock<IPropertyInformation>();
+
+      var fallbackFactory = new FallbackValidationMessageFactory();
+      var result = fallbackFactory.CreateValidationMessageForPropertyValidator(propertyValidatorMock.Object, validatedPropertyMock.Object);
+
+      Assert.That(result, Is.Not.Null);
+      Assert.IsInstanceOf<InvariantValidationMessage>(result);
```
Frazerus, line 40: I think we should keep the typical assert structure here as well: 

`Assert.That(result, Is.InstanceOf<InvariantValidationMessage>()`

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
+      Assert.That(result.ToString(), Is.EqualTo($"{propertyValidatorMock.Object.GetType().Name}: Validation error."));
```
Frazerus, line 41: I think we typically want to write the full expected thing. This does not tell me what output the toString method returns without having to jump through some thinking hoops.

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
     public void CreateValidationMessageForObjectValidator ()
     {
+      var objectValidatorMock = new Mock<IObjectValidator>();
+      var validatedType = TypeAdapter.Create(typeof(Person));
+
+      var fallbackFactory = new FallbackValidationMessageFactory();
+      var result = fallbackFactory.CreateValidationMessageForObjectValidator(objectValidatorMock.Object, validatedType);
+
+      Assert.That(result, Is.Not.Null);
+      Assert.IsInstanceOf<InvariantValidationMessage>(result);
```
Frazerus, line 74: Same as above

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
+      Assert.That(result.ToString(), Is.EqualTo($"{objectValidatorMock.Object.GetType().Name}: Validation error."));
```
Frazerus, line 75: Same-same as above-above

Frazerus (approved):

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
KuJo7: Since `validationMessage` is private, I have to use `.ToString()` to access the string. I am not sure how it can be done in a different way.

Frazerus: I meant that the expected value should not use String.Format (which happens if you use `$"{someVariable}"`.

The inside of the assert should probably look like this:
`result.ToString(), Is.EqualTo("This is the actual string output without any formatting happening")`

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
+
+    [Test]
+    public void CreateValidationMessageForPropertyValidator_WithNullAsValidator_ThrowsArgumentNullException ()
+    {
+      var validatedPropertyMock = new Mock<IPropertyInformation>();
+
+      var fallbackFactory = new FallbackValidationMessageFactory();
+
+      Assert.That(() => fallbackFactory.CreateValidationMessageForPropertyValidator(null, validatedPropertyMock.Object), Throws.ArgumentNullException);
+    }
```
MichaelKetting, line 52: Typically, we don't test argument-null and argument-empty exceptions because our infrastructre provides the helpers, they are everywhere and it's just a matter of forgetting the call. Since we don't have them in other places, I would recommend dropping them. I would also place in APIs where the signature allows null values but we expcicitly disallow null. This would be done to document/enforce the API rules and not to unit test that we haven't made a mistake during implementation.

As for future stuff: there was a long time hoping that Microsoft would solve arg-null-guards via the compiler. Since they dropped this again, it could be a simple option to use roslyn to verify that each parameter has an argument null check.

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
     public void CreateValidationMessageForPropertyValidator ()
     {
+      var propertyValidatorMock = new Mock<IPropertyValidator>();
+      var validatedPropertyMock = new Mock<IPropertyInformation>();
+
+      var fallbackFactory = new FallbackValidationMessageFactory();
+      var result = fallbackFactory.CreateValidationMessageForPropertyValidator(propertyValidatorMock.Object, validatedPropertyMock.Object);
+
+      Assert.That(result, Is.Not.Null);
+      Assert.That(result, Is.InstanceOf<InvariantValidationMessage>());
```
MichaelKetting, line 40:
```suggestion
      Assert.That(result, Is.InstanceOf<InvariantValidationMessage>());
```
In this scenario, the null-check is redundant. The type check would complain with a meaningful message anyway.

Remotion/Validation/UnitTests/Implementation/FallbackValidationMessageFactoryTest.cs
```diff
@@ -15,23 +15,84 @@
...
     public void CreateValidationMessageForObjectValidator ()
     {
+      var objectValidatorMock = new Mock<IObjectValidator>();
+      var validatedType = TypeAdapter.Create(typeof(Person));
+
+      var fallbackFactory = new FallbackValidationMessageFactory();
+      var result = fallbackFactory.CreateValidationMessageForObjectValidator(objectValidatorMock.Object, validatedType);
+
+      Assert.That(result, Is.Not.Null);
+      Assert.That(result, Is.InstanceOf<InvariantValidationMessage>());
```
MichaelKetting, line 74:
```suggestion
      Assert.That(result, Is.InstanceOf<InvariantValidationMessage>());
```

MichaelKetting (approved):
Some small suggestions.

PR #710, opened 2023-05-31
RM-8822 Implement tests for ResourceManagerBasedValidationMessage

KuJo7:
https://re-motion.atlassian.net/browse/RM-8822

### DIFF
--- a/Remotion/Validation/Globalization.UnitTests/ResourceManagerBasedValidationMessageTest.cs
+++ b/Remotion/Validation/Globalization.UnitTests/ResourceManagerBasedValidationMessageTest.cs
@@ -15,35 +15,100 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Globalization;
+using Moq;
 using NUnit.Framework;
+using Remotion.Globalization;
 
 namespace Remotion.Validation.Globalization.UnitTests
 {
   [TestFixture]
   public class ResourceManagerBasedValidationMessageTest
   {
-    [Test]
-    [Ignore("RM-5906")]
-    public void Format ()
+    [ResourceIdentifiers]
+    private enum TestResourceIdentifiers
     {
+      One,
+      Two,
     }
 
     [Test]
-    [Ignore("RM-5906")]
+    [SetCulture("it-IT")]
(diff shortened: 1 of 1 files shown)


MichaelKetting (changes requested):

Remotion/Validation/Globalization.UnitTests/ResourceManagerBasedValidationMessageTest.cs
```diff
@@ -15,35 +15,100 @@
...
+            actualCulture = CultureInfo.CurrentUICulture;
+          })
+          .Returns(true);
+
+      var validationMessage  = new ResourceManagerBasedValidationMessage(resourceManagerMock.Object, TestResourceIdentifiers.Two);
+      var expectedCulture = CultureInfo.GetCultureInfo("fr-FR");
+
+      var result = validationMessage.ToString();
+
+      Assert.That(result, Is.EqualTo("Remotion.Validation.Globalization.UnitTests.ResourceManagerBasedValidationMessageTest.Two"));
```
MichaelKetting, line 110:
```suggestion
      string value = "Expected Value";
      CultureInfo actualCulture = null;

      var resourceManagerMock = new Mock<IResourceManager>();
      resourceManagerMock
          .Setup(_ => _.TryGetString("Remotion.Validation.Globalization.UnitTests.ResourceManagerBasedValidationMessageTest.Two", out value))
          .Callback(() =>
          {
            actualCulture = CultureInfo.CurrentUICulture;
          })
          .Returns(true);

      var validationMessage  = new ResourceManagerBasedValidationMessage(resourceManagerMock.Object, TestResourceIdentifiers.Two);
      var expectedCulture = CultureInfo.GetCultureInfo("fr-FR");

      var result = validationMessage.ToString();

      Assert.That(result, Is.EqualTo("Expected Value"));
```

It's good practice to use simple strings that are obviously fake values when setting up test doubles. That way, the test is a) shorter and b) there is no chance these values hide incorrect behavior. E.g. you could accidentally return the key instead of the result and then this error would remain undetected.

MichaelKetting (approved):

a-ctor (approved):

PR #711, opened 2023-06-02
RM-8678 Add localizations for de-CH

KuJo7:
https://re-motion.atlassian.net/browse/RM-8678

### DIFF
--- a/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.de-ch.resx
+++ b/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.de-ch.resx
@@ -1,0 +1,128 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
+    <resheader name="version">2.0</resheader>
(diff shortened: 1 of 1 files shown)


Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
```diff
@@ -736,6 +736,34 @@ public void CreateValidationMessageForPropertyValidator_WithDecimalValidator_Ret
...
+      _propertyStub.Setup(_ => _.PropertyType).Returns(typeof(object));
+      var validator = new InclusiveRangeValidator(5, 10, _validationMessageStub.Object);
+      var validationMessage = _factory.CreateValidationMessageForPropertyValidator(validator, _propertyStub.Object);
+
+      Assert.That(validationMessage, Is.Not.Null);
+      Assert.That(validationMessage, Is.InstanceOf<ResourceManagerBasedValidationMessage>());
+
+      Assert.That(validationMessage.ToString(), Is.EqualTo("Geben Sie einen Wert von \"{0}\" bis \"{1}\" ein."));
+    }
+
```
MichaelKetting, line 766: Please revert the tests. We typically do not test the different localizations via unit tests because it creates a maintainance dependency, in particular for non-English texts.

We do sometimes have unit tests that must work with the default culture texts because we a) want to test that the placeholders are filled correctly or b) the correct resource is resolved or c) it's easier for reason. 

For any other language, specific unit tests which test that the locatization has been created correctly does not scale. It's a) a copy&paste operation of potentially longer texts in a language we don't know so it's much harder to verify and b) localizations are updated on demand if there's typos found or user feedback. Needing to hunt down the correct test and fixing the string by also pasting it in the test does not add value.

So, instead we rely on tooling support (e.g. Rider/Resharper's resource editor) and simple review to make sure the changes are correct (no extra placeholders, the correct value was pasted, etc)

MichaelKetting (approved):

PR #713, opened 2023-06-06
RM-8832 Implement tests for AddingPropertyValidationRuleCollector

KuJo7:
https://re-motion.atlassian.net/browse/RM-8832

### DIFF
--- a/Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
+++ b/Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
@@ -15,6 +15,7 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -26,6 +27,7 @@ using Remotion.Utilities;
 using Remotion.Validation.Implementation;
 using Remotion.Validation.Merging;
 using Remotion.Validation.RuleCollectors;
+using Remotion.Validation.Rules;
 using Remotion.Validation.UnitTests.TestDomain;
 using Remotion.Validation.UnitTests.TestDomain.Collectors;
 using Remotion.Validation.UnitTests.TestHelpers;
@@ -37,8 +39,7 @@ namespace Remotion.Validation.UnitTests.RoleCollectors
   public class AddingPropertyValidationRuleCollector1Test
   {
     private Expression<Func<Customer, string>> _userNameExpression;
-    private Expression<Func<Customer, string>> _lastNameExpression;
-    private IAddingPropertyValidationRuleCollector _addingPropertyValidationRuleCollector;
+    private AddingPropertyValidationRuleCollector<Customer, string> _addingPropertyValidationRuleCollector;
     private IPropertyInformation _property;
     private Mock<IPropertyValidatorExtractor> _propertyValidatorExtractorMock;
     private StubPropertyValidator _stubPropertyValidator1;
@@ -51,7 +52,6 @@ namespace Remotion.Validation.UnitTests.RoleCollectors
       _property = PropertyInfoAdapter.Create(typeof(Customer).GetProperty("UserName"));
 
       _userNameExpression = ExpressionHelper.GetTypedMemberExpression<Customer, string>(c => c.UserName);
-      _lastNameExpression = ExpressionHelper.GetTypedMemberExpression<Customer, string>(c => c.LastName);
 
       _stubPropertyValidator1 = new StubPropertyValidator();
       _stubPropertyValidator2 = new NotEmptyOrWhitespaceValidator(new InvariantValidationMessage("Fake Message"));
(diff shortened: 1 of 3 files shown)


Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +123,151 @@ public void Create_PropertyDeclaredInBaseClass ()
...
+      validationMessageMock
+          .Setup(e => e.CreateValidationMessageForPropertyValidator(It.IsAny<NotNullValidator>(), _property))
+          .Returns(validationMessage);
+
+      Func<PropertyValidationRuleInitializationParameters, IPropertyValidator> validatorFactory = param => new NotNullValidator(param.ValidationMessage);
+
+      _addingPropertyValidationRuleCollector.RegisterValidator(validatorFactory);
+      var result = _addingPropertyValidationRuleCollector.CreateValidationRule(validationMessageMock.Object);
+
+      var propertyValidator = ((IPropertyValidationRule)result).Validators.ToArray()[0] as NotNullValidator;
```
Frazerus, line 148: Actually I think we might want to use First() here. We also probably don't want to use as, and rather add an extra Assert for which type of validator it is (so no casting in the variable assignment). Then we can safely cast the validator in the already existing assert. I think that might make this test more understandable.

Frazerus: Try it, and if it looks good, please also do on the other tests.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +123,151 @@ public void Create_PropertyDeclaredInBaseClass ()
...
+      _addingPropertyValidationRuleCollector.SetCondition(predicate);
+
+      var result = _addingPropertyValidationRuleCollector.CreateValidationRule(validationMessageMock.Object);
+
+      Assert.That(((PropertyValidationRule<Customer, string>)result).Condition, Is.SameAs(predicate));
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void SetCondition ()
```
Frazerus, line 241: Maybe SetCondition_SetsConditionToBeUsedByCreate? Vielleicht reicht der derzeitige Name aber eh auch.

Frazerus (changes requested):

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
KuJo7: I think Mike or Patrick once told me to not use `.First()` in the tests and its better to use `.ToArray()[0]`
The cast is actually necessary because otherwise I cannot access ValidationMessage

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -59,7 +61,8 @@ public void SetUp ()
 
       _propertyValidatorExtractorMock = new Mock<IPropertyValidatorExtractor>(MockBehavior.Strict);
 
-      _addingPropertyValidationRuleCollector = AddingPropertyValidationRuleCollector.Create(_userNameExpression, typeof(CustomerValidationRuleCollector1));
+      _addingPropertyValidationRuleCollectorType = AddingPropertyValidationRuleCollector.Create(_userNameExpression, typeof(CustomerValidationRuleCollector1));
```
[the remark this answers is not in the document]
KuJo7: It should be 180 and this line is not too long as I checked

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
KuJo7: Es wird eigentlich nur die Condition Property in der Klasse gesetzt, daher glaub ich passt einfach nur SetCondition (HappyPath)

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
KuJo7: I guess its even better to compare the whole list?

Frazerus: Yea, comparing the whole list would be optimal.

KuJo7: Since I want to assert the validationMessages, it doesn't make sense to compare the validator instances, because the test will also pass if no validationMessage was initialized. So, I have to do it as I did before

Frazerus (approved):

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -59,7 +62,8 @@ public void SetUp ()
...
+      _addingPropertyValidationRuleCollector = _addingPropertyValidationRuleCollectorType;
```
MichaelKetting, line 66: The naming doesn't name it clear what the field is for. Now that I understand why we have it:

the interface-version is used because we just want to test the interface api origianlly and the name is much shorter. It is okay to replace the interface with the type in the field type. In the one or two places where we need the interface due to expicit implementations, I think would work best if we do a cast inline. The assumtion is that this is the special case, not the default case.

 Separately, we can decide if we want to verify that the interface is implemented correctly. In the past, we did this in a few places. I don't think there's a benefit to this here so we can just switch the type and keep the rest simple.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -73,12 +77,27 @@ public void Initialization_PropertyDeclaredInSameClass ()
...
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void Initialization_CollectorTypeDoesNotImplementIValidationRuleCollector_ThrowsArgumentException ()
     {
+      Assert.That(
+          () => new AddingPropertyValidationRuleCollector<Customer, string>(_property, _ => "", typeof(Customer)),
+          Throws.ArgumentException
+              .With.ArgumentExceptionMessageEqualTo("Parameter 'collectorType' is a 'Remotion.Validation.UnitTests.TestDomain.Customer', which cannot be assigned to "
+                                                    + "type 'Remotion.Validation.IValidationRuleCollector'.", "collectorType"));
```
MichaelKetting, line 90: new line formatting of the message.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -73,12 +77,27 @@ public void Initialization_PropertyDeclaredInSameClass ()
...
+    }
+
+    [Test]
+    public void Create_CollectorTypeDoesNotImplementIValidationRuleCollector_ThrowsArgumentException ()
+    {
+      Assert.That(
+          () => AddingPropertyValidationRuleCollector.Create(_userNameExpression, typeof(Customer)),
+          Throws.ArgumentException
+              .With.ArgumentExceptionMessageEqualTo("Parameter 'collectorType' is a 'Remotion.Validation.UnitTests.TestDomain.Customer', which cannot be assigned to "
+                                                    + "type 'Remotion.Validation.IValidationRuleCollector'.", "collectorType"));
```
MichaelKetting, line 100: new line formatting of the message.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -73,12 +77,27 @@ public void Initialization_PropertyDeclaredInSameClass ()
...
     }
```
MichaelKetting, line 101: This factory method actually belongs to `AddingPropertyValidationRuleCollectorTest.cs`. We have separate test fixtures for separates classes, including generics. That's what the 1-index in the name implies. It's short for `\`1` before I knew I could actually write `\`1`, too in file names.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +124,155 @@ public void Create_PropertyDeclaredInBaseClass ()
...
+    [Test]
+    public void SetCondition_WithInvalidTypeOfPredicate_ThrowsArgumentException ()
+    {
+      Func<Person, bool> predicate = _ => true;
+
+      Assert.That(
+          () => _addingPropertyValidationRuleCollector.SetCondition(predicate),
+          Throws.ArgumentException
+              .With.Message.EqualTo("The type 'Remotion.Validation.UnitTests.TestDomain.Person' of the predicate does not match the "
+                                    + "type 'Remotion.Validation.UnitTests.TestDomain.Customer' of the validation rule."));
```
MichaelKetting, line 263: move message to new line. ALso, please use the argumentmessage-check API and also assert the parameter name for completeness' sake.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +124,155 @@ public void Create_PropertyDeclaredInBaseClass ()
...
+          Throws.InvalidOperationException
+              .With.Message.EqualTo("The IValidationMessageFactory did not return a result for NotNullValidator applied to property 'UserName' on "
+                                    + "type 'Remotion.Validation.UnitTests.TestDomain.Customer'."));
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CreateValidationRule_WithConditionNotNull_InitializesConditionForCreatedPropertyValidationRule ()
     {
+      var validationMessage = new DelegateBasedValidationMessage(() => "Test");
```
MichaelKetting, line 227: We also have the InvariantValidationMessage. My personal style is using that one because it's more straight forward. In the end, it doesn't really matter which version you use, it's just a case of simplifying the test and best practice / a typical pattern. In more advanced scenarios such simplfiications could become a code-style requirement instead of good practice in order to keep things readable.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +124,155 @@ public void Create_PropertyDeclaredInBaseClass ()
...
+
+      Func<PropertyValidationRuleInitializationParameters, IPropertyValidator> validatorFactory = param => new NotNullValidator(param.ValidationMessage);
+
+      _addingPropertyValidationRuleCollector.RegisterValidator(validatorFactory);
+
+      Assert.That(
+          () => _addingPropertyValidationRuleCollector.CreateValidationRule(validationMessageMock.Object),
+          Throws.InvalidOperationException
+              .With.Message.EqualTo("The IValidationMessageFactory did not return a result for NotNullValidator applied to property 'UserName' on "
+                                    + "type 'Remotion.Validation.UnitTests.TestDomain.Customer'."));
```
MichaelKetting, line 221: move message to new line

MichaelKetting, line 221: I found a probable message bug in prod code: we try to use single-quotes for parameter values. => `NotNullValidator` should have quotes. Please update the test and prod code. The `IValidationMessageFactory` string is an edge case. here, the sentence / readability works better if it is considered the subject and part of the sentence instead of information provided to the user.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +124,155 @@ public void Create_PropertyDeclaredInBaseClass ()
...
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CreateValidationRule_WhenCalledTwice_InitializesDeferredInitializationValidationMessagesOnlyForNewlyRegisteredValidators ()
     {
+      var validationMessage1 = new DelegateBasedValidationMessage(() => "Test");
+      var validationMessage2 = new DelegateBasedValidationMessage(() => "newTest");
+
+      var validationMessageMock1 = new Mock<IValidationMessageFactory>();
```
MichaelKetting, line 160: the name of the mock is incorrect. it should be a) called a `stub` and be this is `validationMessageFactory` not a `validationMessage`. Please check the other tests, too.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -105,51 +124,155 @@ public void Create_PropertyDeclaredInBaseClass ()
...
+      validationMessageMock
+          .Setup(e => e.CreateValidationMessageForPropertyValidator(It.IsAny<NotNullValidator>(), _property))
+          .Returns(validationMessage);
+
+      var result = _addingPropertyValidationRuleCollector.CreateValidationRule(validationMessageMock.Object);
+
+      var propertyValidators = ((IPropertyValidationRule)result).Validators;
+
+      Assert.That(propertyValidators, Is.Empty);
     }
```
MichaelKetting, line 203: This test misses part of the spec. we still need to register a validator. we simply go and do not use the `param.ValidationMessage` when instantiating the validator in the factory.

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
KuJo7: It was a bit confusing at first, then I realised that there is actually only one class `AddingPropertyValidationRuleCollector1Test.cs`. Should I create a class `AddingPropertyValidationRuleCollectorTest.cs` and move the factory method to it?

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
KuJo7: So, I guess this message bug is also in `AddingObjectValidationRuleCollector1.cs` ?

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
MichaelKetting: yes, please.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
MichaelKetting: I didn't verify it but it sounds reasonable to me :)

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -37,8 +39,8 @@ namespace Remotion.Validation.UnitTests.RoleCollectors
   public class AddingPropertyValidationRuleCollector1Test
   {
     private Expression<Func<Customer, string>> _userNameExpression;
-    private Expression<Func<Customer, string>> _lastNameExpression;
-    private IAddingPropertyValidationRuleCollector _addingPropertyValidationRuleCollector;
+    //private IAddingPropertyValidationRuleCollector _addingPropertyValidationRuleCollector;
+    private AddingPropertyValidationRuleCollector<Customer, string> _addingPropertyValidationRuleCollector;
```
MichaelKetting, line 43:
```suggestion
    private AddingPropertyValidationRuleCollector<Customer, string> _addingPropertyValidationRuleCollector;
```

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -73,83 +74,167 @@ public void Initialization_PropertyDeclaredInSameClass ()
...
+    [Test]
+    public void SetCondition_WithInvalidTypeOfPredicate_ThrowsArgumentException ()
+    {
+      Func<Person, bool> predicate = _ => true;
+
+      Assert.That(
+          () => _addingPropertyValidationRuleCollector.SetCondition(predicate),
+          Throws.ArgumentException
+              .With.ArgumentExceptionMessageEqualTo("The type 'Remotion.Validation.UnitTests.TestDomain.Person' of the predicate does not match the "
+                                    + "type 'Remotion.Validation.UnitTests.TestDomain.Customer' of the validation rule.", "predicate"));
```
MichaelKetting, line 225: Formatting

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -73,83 +74,167 @@ public void Initialization_PropertyDeclaredInSameClass ()
...
     public void CreateValidationRule_IgnoresValidationMessagesForValidatorsWithoutADeferredInitializationValidationMessage ()
     {
+      var validationMessage = new InvariantValidationMessage("Test");
+
+      var validationMessageFactoryStub = new Mock<IValidationMessageFactory>();
+      validationMessageFactoryStub
+          .Setup(e => e.CreateValidationMessageForPropertyValidator(It.IsAny<NotNullValidator>(), _property))
+          .Returns(validationMessage);
+
+      Func<PropertyValidationRuleInitializationParameters, IPropertyValidator> validatorFactory = _ => new NotNullValidator(new InvariantValidationMessage("newTest"));
```
MichaelKetting, line 156: Good test values: I would recommend changing this to "expectedMessage" or "originalMessage" since we do not change the value and expect it to remain the same. "newTest" implies something else. 

Probably good to change the other messages in a simarliar way. It's not mandatory, though, just best practice for selecting good test values.

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
```diff
@@ -73,83 +74,167 @@ public void Initialization_PropertyDeclaredInSameClass ()
...
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CreateValidationRule_IgnoresValidationMessagesForValidatorsWithoutADeferredInitializationValidationMessage ()
     {
+      var validationMessage = new InvariantValidationMessage("Test");
+
+      var validationMessageFactoryStub = new Mock<IValidationMessageFactory>();
+      validationMessageFactoryStub
+          .Setup(e => e.CreateValidationMessageForPropertyValidator(It.IsAny<NotNullValidator>(), _property))
+          .Returns(validationMessage);
```
MichaelKetting, line 154: We could actually rewrite this particular setup to use `.Throws(new InvalidOperationException("Should not be called."))` instead of the `.Returns(...)` since the affected code won't get exercised. There is however no actual contract that we are forbidden from doing the extra work. Given the entire setup, it might still be cleaner because it encourages future developers to not produce wasted CPU cycles.

If you keep it, you should update the message from `Test` to `Unexpected Message` to indicate that we don't expect this text to show up in the assertion.

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/RoleCollectors/AddingPropertyValidationRuleCollector1Test.cs
KuJo7:
```suggestion
      var validationMessageFactoryStub = new Mock<IValidationMessageFactory>();
      validationMessageFactoryStub
          .Setup(e => e.CreateValidationMessageForPropertyValidator(It.IsAny<NotNullValidator>(), _property))
          .Returns(new InvariantValidationMessage("unexpectedMessage"));

      Func<PropertyValidationRuleInitializationParameters, IPropertyValidator> validatorFactory = _ => new NotNullValidator(new InvariantValidationMessage("expectedMessage"));

      (_addingPropertyValidationRuleCollector).RegisterValidator(validatorFactory);
      var result = ((IAddingPropertyValidationRuleCollector)_addingPropertyValidationRuleCollector).CreateValidationRule(validationMessageFactoryStub.Object);

      var propertyValidators = ((IPropertyValidationRule)result).Validators.Cast<NotNullValidator>().ToArray();
      Assert.That(propertyValidators.Length, Is.EqualTo(1));
      Assert.That(propertyValidators[0].ValidationMessage.ToString(), Is.EqualTo("expectedMessage"));
```
I am a bit confused now, I changed the setup to use `.Throws()` instead of the `.Returns()`, but the code gets exercised on line 82 in `AddingPropertyValidationRuleCollector1.cs`.
If understood it right, only by registering the validator with a `DeferredInitializationValidationMessage`, the message will be initialized and a message without a `DeferredInitializationValidationMessage` should be ignored. So, I think that I have to leave it as I did before and update the message to `unexpectedMessage`

MichaelKetting: Ah, yes, that's correct. I thought that only Validators that actually use the DeferredMessage gets hooked up. So, your choice of using `.Returns()` is correct.

PR #728, opened 2023-06-14
RM-8829 Implement missing tests and assertions in LocalizedValidation…

KuJo7:
https://re-motion.atlassian.net/browse/RM-8829

### DIFF
--- a/Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
+++ b/Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
@@ -7,6 +7,8 @@ using NUnit.Framework;
 using Remotion.Globalization;
 using Remotion.Reflection;
 using Remotion.ServiceLocation;
+using Remotion.Validation.Globalization.UnitTests.TestDomain;
+using Remotion.Validation.Globalization.UnitTests.TestHelpers;
 using Remotion.Validation.Implementation;
 using Remotion.Validation.Validators;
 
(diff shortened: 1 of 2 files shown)


Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
```diff
@@ -734,12 +840,17 @@ public void CreateValidationMessageForPropertyValidator_WithDecimalValidator_Ret
...
-    [Ignore("TODO RM-5906")]
     public void CreateValidationMessageForObjectValidator_NoLocalizationsAvailableForObjectValidators_ReturnsNull ()
     {
+      var validator = new RealPersonValidator();
+      var validationMessage = _factory.CreateValidationMessageForObjectValidator(validator, TypeAdapter.Create(typeof(Person)));
+
+      Assert.That(validationMessage, Is.Null);
     }
 
     [Test]
```
KuJo7, line 856: @MichaelKetting Can you please check if this test might be wrong, because the method always returns null and I am not sure if we need to add logic to the implementation or not.

Remotion/Validation/Globalization.UnitTests/Validation.Globalization.UnitTests.csproj
```diff
@@ -15,6 +15,7 @@
     <ProjectReference Include="..\..\Globalization\Core\Globalization.Core.csproj" />
     <ProjectReference Include="..\Core\Validation.Core.csproj" />
     <ProjectReference Include="..\Globalization\Validation.Globalization.csproj" />
+    <ProjectReference Include="..\IntegrationTests\Validation.IntegrationTests.csproj" />
```
MichaelKetting, line 18: this dependency is incorrect. please remove.

Remotion/Validation/Globalization.UnitTests/LocalizedValidationMessageFactoryTest.cs
MichaelKetting: The null-value is correct. there is no good way to return a default (meaningful) validation message for generic object validation. So, we return null, which is a valid result for the case that the factory doesn't know what to do.

MichaelKetting: Yes.

Remotion/Validation/Globalization.UnitTests/Validation.Globalization.UnitTests.csproj
```diff
@@ -15,6 +15,7 @@
     <ProjectReference Include="..\..\Globalization\Core\Globalization.Core.csproj" />
     <ProjectReference Include="..\Core\Validation.Core.csproj" />
     <ProjectReference Include="..\Globalization\Validation.Globalization.csproj" />
+    <ProjectReference Include="..\UnitTests\Validation.UnitTests.csproj" />
```
MichaelKetting, line 18: Please drop this dependency.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #731, opened 2023-06-15
Feature/rm 8804 add validation summary row to boc list

Frazerus:
https://re-motion.atlassian.net/browse/RM-8804

The only thing still missing are possible improvements on how the cell is stickied, as we cant just make things slide around willy nilly (I think)

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
@@ -51,7 +51,7 @@ class BocList_RowBlock
 {
   public Row: HTMLElement;
   public SelectorControl: HTMLInputElement;
-
+  public ValidationRow: Nullable<HTMLElement>
   constructor(row: HTMLElement, selectorControl: HTMLInputElement)
   {
     if (row.nodeName !== 'TR')
@@ -59,6 +59,9 @@ class BocList_RowBlock
 
     this.Row = row;
     this.SelectorControl = selectorControl
+    this.ValidationRow = this.Row.classList.contains("hasValidationRow")
+        ? this.Row.nextSibling as HTMLElement
+        : null;
   }
 }
 
@@ -177,6 +180,13 @@ class BocList
         selectedRows.OnSelectionChanged (bocList, false);
       }
     });
+
+    if (row.classList.contains("hasValidationRow"))
+    {
+      let validationRow = row.nextSibling as HTMLElement;
+
+      validationRow.addEventListener('click', () => row.dispatchEvent(new Event("click")));
+    }
   
     selectorControl.addEventListener('click', function (evt)
     {
@@ -201,7 +211,7 @@ class BocList
     }
   }
 
-  //  Event handler for a table row in the BocList. 
+  //  Event handler for a table row in the BocList.
   //  Selects/unselects a row/all rows depending on its selection state,
   //      whether CTRL has been pressed and if _bocList_isSelectorControlClick is true.
   //  Aborts the execution if _bocList_isCommandClick or _bocList_isSelectorControlClick is true.
@@ -288,8 +298,10 @@ class BocList
     // Select currentRow
     rowBlock.SelectorControl.checked = true;
     if (isRowHighlightingEnabled)
+    {
       rowBlock.Row.classList.add(BocList.TrClassNameSelected);
-
+      rowBlock.ValidationRow?.classList.add(BocList.TrClassNameSelected);
+    }
     BocList.SetSelectAllRowsSelectorOnDemand (selectedRows);
   }
 
@@ -328,7 +340,10 @@ class BocList
     // Unselect currentRow
     rowBlock.SelectorControl.checked = false;
     if (isRowHighlightingEnabled)
+    {
       rowBlock.Row.classList.remove(BocList.TrClassNameSelected);
+      rowBlock.ValidationRow?.classList.remove(BocList.TrClassNameSelected);
+    }
 
     BocList.ClearSelectAllRowsSelector (selectedRows);
   }
@@ -503,12 +518,27 @@ class BocList
     {
       if (!PageUtility.Instance.IsInDom (scrollableContainer))
         return;
-  
+
       BocList.FixHeaderSize(scrollableContainer);
       BocList.FixHeaderPosition(tableContainer, scrollableContainer);
+      BocList.FixValidationErrorOverflowSize(scrollableContainer);
       setTimeout(resizeHandler, resizeInterval);
     };
     resizeHandler();
+  }
+
+  private static FixValidationErrorOverflowSize(scrollableContainer: HTMLElement)
+  {
+    // Sets the width of the validation error field to be as wide as the visible part of the bocList.
+    // This enables sticky behaviour while still allowing for a scrollbar if required.
+
+    const width = LayoutUtility.GetWidth(scrollableContainer);
+
+    const overflowContainer: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow div');
+    if (overflowContainer)
+    {
+      overflowContainer.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
+    }
   }
 
   private static CreateFakeTableHead(tableContainer: HTMLElement, scrollableContainer: HTMLElement, bocList: HTMLElement): void
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
@@ -15,14 +15,19 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Linq;
 using Moq;
 using NUnit.Framework;
 using Remotion.Development.Web.UnitTesting.Resources;
 using Remotion.Development.Web.UnitTesting.UI.Controls.Rendering;
+using Remotion.ObjectBinding.Validation;
 using Remotion.ObjectBinding.Web.Contracts.DiagnosticMetadata;
 using Remotion.ObjectBinding.Web.Services;
 using Remotion.ObjectBinding.Web.UI.Controls;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
 using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+using Remotion.Utilities;
 using Remotion.Web;
 using Remotion.Web.Contracts.DiagnosticMetadata;
 using Remotion.Web.UI.Controls.Rendering;
@@ -35,6 +40,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
     private BocListCssClassDefinition _bocListCssClassDefinition;
     private BocColumnRenderer[] _columnRenderers;
     private StubLabelReferenceRenderer _stubLabelReferenceRenderer;
+    private StubValidationSummaryRenderer _stubValidationSummaryRenderer;
 
     [SetUp]
     public void SetUp ()
@@ -60,6 +66,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
 
       _bocListCssClassDefinition = new BocListCssClassDefinition();
       _stubLabelReferenceRenderer = new StubLabelReferenceRenderer();
+      _stubValidationSummaryRenderer = new StubValidationSummaryRenderer();
     }
 
     [Test]
@@ -69,7 +76,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderTitlesRow(CreateRenderingContext());
 
       var document = Html.GetResultDocument();
@@ -91,7 +99,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderTitlesRow(CreateRenderingContext());
 
       var document = Html.GetResultDocument();
@@ -117,7 +126,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderTitlesRow(CreateRenderingContext());
 
       var document = Html.GetResultDocument();
@@ -137,7 +147,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderDataRow(
           CreateRenderingContext(),
           new BocListRowRenderingContext(new BocListRow(0, BusinessObject), 1, false),
@@ -161,7 +172,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       var columnRenderers = new[]
                              {
                                  CreateBocColumnRenderer(3, isRowHeader: false),
@@ -234,7 +246,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       var columnRenderers = new[]
                             {
                                 new BocColumnRenderer(
@@ -275,7 +288,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       var columnRenderers = new[]
                             {
                                 new BocColumnRenderer(
@@ -316,7 +330,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       var columnRenderers = new[]
                             {
                                 new BocColumnRenderer(
@@ -359,7 +374,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderDataRow(
           CreateRenderingContext(),
           new BocListRowRenderingContext(new BocListRow(0, BusinessObject), 1, true),
@@ -387,7 +403,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderEmptyListDataRow(CreateRenderingContext());
 
       var document = Html.GetResultDocument();
@@ -407,7 +424,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.Default);
+          RenderingFeatures.Default,
+          _stubValidationSummaryRenderer);
       renderer.RenderEmptyListDataRow(CreateRenderingContext());
 
       var document = Html.GetResultDocument();
@@ -422,7 +440,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
           _bocListCssClassDefinition,
           new BocIndexColumnRenderer(RenderingFeatures.WithDiagnosticMetadata, _bocListCssClassDefinition),
           new BocSelectorColumnRenderer(RenderingFeatures.WithDiagnosticMetadata, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
-          RenderingFeatures.WithDiagnosticMetadata);
+          RenderingFeatures.WithDiagnosticMetadata,
+          _stubValidationSummaryRenderer);
       renderer.RenderDataRow(
           CreateRenderingContext(),
           new BocListRowRenderingContext(new BocListRow(0, BusinessObject), 1, false),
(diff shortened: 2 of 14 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocListEvents.cs
```diff
@@ -150,6 +156,12 @@ public void SetRowReadOnly ()
     _isEditableRow = false;
   }
 
+  public void SetAdditionalCssClassForValidationRow (string cssClass)
```
a-ctor, line 159: To keep the symmetry, this should be below `SetAdditionalCssClassForDataRow`

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -181,6 +197,40 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+        BocListDataRowRenderEventArgs rowRenderEventArgs,
+        IReadOnlyCollection<BocListValidationFailureWithLocationInformation> validationFailuresWithLocationInfo)
+    {
+      var cssClassTableRow = GetCssClassValidationRow(renderingContext, rowRenderingContext.IsSelected, rowRenderEventArgs);
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, cssClassTableRow);
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Tr);
+
+      //The BocList also contains two columns not existent in the ColumnRenderers property (the number and selection column)
+      //Therefore we need to add these to the count.
```
a-ctor, line 212:
```suggestion
      // The BocList also contains two columns not existent in the ColumnRenderers property (the number and selection column)
      // Therefore we need to add these to the count.
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,38 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
+div.bocList div.bocListTableBlock tr.bocListValidationRow td.bocListValidationRowStickyCell
+{
+  position: sticky;
+  left: 0;
+  white-space: nowrap;
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow td.bocListValidationRowStickyCell ul
+{
+  max-width: 1px;
```
a-ctor, line 393: The current solution with the special table cell setup and the 1px column to get the sticky effect feels too complex for me. In addition, when we extend the solution to have a JS component that helps with sizing, we would need to rewrite a lot of it anyway.

I experimented a bit and came up with the following solution:
image

 - (This CSS rule and the rules above were disabled for this demo)
 - The `td` element now spans the whole row with a `colspan` (this should be set to the exact number of columns instead of `99` in the end)
 - The `ul` element is wrapped in two `div` elements
   - The first provides overflow+sticky logic and should always at most as big as the visible space
     - JS would set the size of this div to the visible width, providing the correct sticky behavior
     - For now I just added a `max-width` to show the sticky behavior
   - The second div is responsible for the sizing of the text. By using `min-inline-size` the text is forced to take at least that much space, overflowing in tight scenarios like we want

I think this setup provides a clean solution that we can extend later on by setting the width using JS

a-ctor (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,41 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
+{
+  margin: calc(var(--remotion-themed-spacing) / 2) var(--remotion-themed-spacing);
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow div.bocListValidationRowStickyContainer
+{
+  width: 300px;
```
a-ctor, line 386: I would choose a higher value for the `width` to display all the error text without a scrollbar for now. Once we integrate the width calculation we won't have to specify it but until then its better to have a no scrollbar.

If you want we can add the width detection right now since it should be much work. We already have a `resizeHandler` in line 512 of BocList.ts. We can then use `LayoutUtility.GetWidth(scrollableContainer)` to set the width of the overflow container. This would allow us to get rid of this `width` setting here entirely.

We have discuss with mike what a good cut-off point in terms of size would be. Right now, we have `400px` but I think that is a bit to small - a bigger start for the overflow is probably better.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -246,7 +246,7 @@ div.bocList table.bocListTable tr.bocListDataRow.selected td.bocListDataCell
   background-color: var(--color-highlight);
 }
 
-div.bocList table.bocListTable td.bocListDataCell
+div.bocList table.bocListTable tr:not(.hasValidationRow) td.bocListDataCell
```
a-ctor, line 249: We also have the `NovaViso` theme, which should probably? also get the validation functionality so we need to add CSS for that as well. To toggle the theme look at the `NovaViso`/`NovaGray` classes

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -181,6 +197,39 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+
+      // The BocList also contains two columns not existent in the ColumnRenderers property (the number and selection column)
+      // Therefore we need to add these to the count.
+      var totalColumnsCount = renderingContext.ColumnRenderers.Length + 2;
+
+      //QUESTION: Is there a better way to do this? saying int.toString() feels so wrong.
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Colspan, totalColumnsCount.ToString());
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Td);
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, "bocListValidationRowStickyContainer");
```
a-ctor, line 219: Classes are usually only applied to "top-level-like" elements and not used as marker for containers inside the control. This is something that needs to be discussed with @MichaelKetting

a-ctor (changes requested):

a-ctor:
I won't rebase these changes onto the new branch because it might mess with the existing review comments. We should do it once the review changes have been addressed

 - [ ] rebase onto the new base branch that is located on develop
 - [ ] change the base branch of this commit to develop

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
MichaelKetting: oh, okay. yeah, in that case it's not really needed or helpful.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
Frazerus: Nah, I don't think NovaGray needs validation. It's good as is, no need to change/add anything.

MichaelKetting: All choking (sic) aside, there's a couple of situation where we actually do skip something in NovaGray, so sarcasm about not doing NovaGray isn't always obvious (depends on just how much sarcasm/irony drips from the wording). In particular, for our primary usecase, the new validation will only get used in applications that use NovaViso. So, it could actually be a valid discussion if we should skip NovaGray. The basic rule is: is it cheap, then we cover NovaGray, too. Is it expensive, we need to discuss it.

MichaelKetting: (NovaGray is backwards compatiblity support only)

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -181,6 +197,37 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+      var cssClassValidationRow = GetCssClassValidationRow(rowRenderingContext.IsSelected, rowRenderEventArgs);
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, cssClassValidationRow);
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Tr);
+
+      // The BocList also contains two columns not existent in the ColumnRenderers property (the number and selection column)
+      // Therefore we need to add these to the count.
+      var totalColumnsCount = renderingContext.ColumnRenderers.Length + 2;
+
+      //QUESTION: Is there a better way to do this? saying int.toString() feels so wrong.
```
a-ctor, line 215: Remove leftover comment

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -40,19 +42,22 @@ public class BocRowRenderer : IBocRowRenderer
...
 
     public BocRowRenderer (
         BocListCssClassDefinition cssClasses,
         IBocIndexColumnRenderer indexColumnRenderer,
         IBocSelectorColumnRenderer selectorColumnRenderer,
-        IRenderingFeatures renderingFeatures)
+        IRenderingFeatures renderingFeatures,
+        IBocListValidationSummaryRenderer validationSummaryRenderer)
     {
       ArgumentUtility.CheckNotNull("cssClasses", cssClasses);
```
a-ctor, line 54: Missing arg check

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -503,14 +511,26 @@ class BocList
...
       setTimeout(resizeHandler, resizeInterval);
     };
     resizeHandler();
   }
 
+  private static FixValidationErrorOverflowSize(scrollableContainer: HTMLElement)
+  {
+    const width = LayoutUtility.GetWidth(scrollableContainer);
+
+    const overflowContainer: HTMLElement | null = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow div');
```
a-ctor, line 527: Please use `Nullable<HTMLElement>` instead. Same type but we use the `Nullable<T>` helper everywhere else so we should use it here as well.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -503,14 +511,26 @@ class BocList
...
+    if (overflowContainer != null)
```
a-ctor, line 528: While possibly not a problem, we should not use the type converting operators. So either we do `if (overflowContainer)` or `if (overflowContainer !=== null)`

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx
```diff
@@ -24,7 +24,7 @@
   <tr>
     <td></td>
     <td>
-      <remotion:boclist id=JobList runat="server" datasourcecontrol="CurrentObject" showallproperties="True" showavailableviewslist="False" propertyidentifier="Jobs" alwaysshowpageinfo="True" listmenulinebreaks="BetweenGroups" selection="SingleRadioButton" index="Disabled" PageSize="2" ShowSortingOrder="True" ListMenuHeading="Jobs Options" ListMenuHeadingLevel="H5" Width="100%" Height="15em">
+      <remotion:boclist id=JobList runat="server" datasourcecontrol="CurrentObject" showallproperties="True" showavailableviewslist="False" propertyidentifier="Jobs" alwaysshowpageinfo="True" listmenulinebreaks="BetweenGroups" selection="SingleRadioButton" index="Disabled" PageSize="2" ShowSortingOrder="True" ListMenuHeading="Jobs Options" ListMenuHeadingLevel="H5" Width="100%" Height="30em">
```
a-ctor, line 27: Probably test changes? If so we can revert them

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx
```diff
@@ -34,7 +34,7 @@
   </tr>
   <tr>
     <td colSpan=2>
-      <ros:TestBocList id=ChildrenList runat="server" datasourcecontrol="CurrentObject" propertyidentifier="Children" alwaysshowpageinfo="True" listmenulinebreaks="BetweenGroups" pagesize="4" indexoffset="100" RowMenuDisplay="Manual" ShowEmptyListMessage="True" Index="InitialOrder" Selection="Multiple" Width="100%" Height="15em" ControlServicePath="BocListWebService.asmx" ControlServiceArguments="ControlServiceArgs">
+      <ros:TestBocList id=ChildrenList runat="server" datasourcecontrol="CurrentObject" propertyidentifier="Children" alwaysshowpageinfo="True" listmenulinebreaks="BetweenGroups" pagesize="4" indexoffset="100" RowMenuDisplay="Manual" ShowEmptyListMessage="True" Index="InitialOrder" Selection="Multiple" Width="100%" Height="30em" ControlServicePath="BocListWebService.asmx" ControlServiceArguments="ControlServiceArgs">
```
a-ctor, line 37: Probably test changes? If so we can revert them

a-ctor (approved):

Remotion/ObjectBinding/Web/UI/Controls/BocListEvents.cs
```diff
@@ -155,6 +161,13 @@ public void SetAdditionalCssClassForDataRow (string cssClass)
     ArgumentUtility.CheckNotEmpty("cssClass", cssClass);
     _additionalCssClassForDataRow = cssClass;
   }
+
+  public void SetAdditionalCssClassForValidationRow (string cssClass)
+  {
+    ArgumentUtility.CheckNotEmpty("cssClass", cssClass);
+    _additionalCssClassForValidationRow = cssClass;
+  }
+
```
MichaelKetting, line 170: Hmm...I don't think we need `AdditionalCssClassForValidationRow`. Instead, the `AdditionalCssClassForDataRow` would also get applied to the validation row. The idea is to have both rows styles identically. If someone really did want special treatment, they can also use `tr.validationRow.myCustomRowCssClass` to slove this.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,56 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
 }
 /*endregion*/
 
+/*region bocListValidationRow*/
+div.bocList div.bocListTableBlock tr.bocListValidationRow
+{
+  border-bottom: var(--boclist-tabledata-border);
+  border-top: var(--boclist-tabledata-border);
+  border-top-style: dashed;
+  color: red;
```
MichaelKetting, line 371: there's a dedicated color property for this.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,56 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
 }
 /*endregion*/
 
+/*region bocListValidationRow*/
+div.bocList div.bocListTableBlock tr.bocListValidationRow
+{
+  border-bottom: var(--boclist-tabledata-border);
+  border-top: var(--boclist-tabledata-border);
+  border-top-style: dashed;
```
MichaelKetting, line 370: drop, please. we need to work within the existing parameters.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,56 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
+}
+
+div.bocList table.bocListTable tr.bocListValidationRow.selected td
+{
+  background-color: var(--color-highlight);
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
+{
+  margin: calc(var(--remotion-themed-spacing) / 2) var(--remotion-themed-spacing);
```
MichaelKetting, line 381: not sure where this is going, please demo

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,56 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow div
+{
+  width: 300px;
```
MichaelKetting, line 386: We shouldn't use PX. instead, we can go with EM and a reasonable number of characters that should be used, e.g. 20?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -362,6 +362,56 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
+div.bocList div.bocListTableBlock tr.bocListValidationRow div::-webkit-scrollbar-thumb
+{
+   background: var(--remotion-themed-scrollbar-color-thumb);
+   border-radius: var(--remotion-themed-scrollbar-border-radius);
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow div div
+{
+  inline-size: fit-content;
+  min-inline-size: 400px;
```
MichaelKetting, line 410: shoudl be related to the 300px size? maybe link via css property?

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -102,6 +102,47 @@ div.bocListNavigator a[href]:active img
   filter: progid:DXImageTransform.Microsoft.BasicImage(grayscale=1);
 }
 
+div.bocList div.bocListTableBlock tr.bocListValidationRow.even
```
MichaelKetting, line 105: I didn't review this in detail. the NovaGray is simpler and with a lot more css duplication.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
```diff
@@ -60,6 +61,7 @@ public void SetUp ()
 
       _bocListCssClassDefinition = new BocListCssClassDefinition();
       _stubLabelReferenceRenderer = new StubLabelReferenceRenderer();
+      _stubValidationSummaryRenderer = new StubValidationSummaryRenderer();
```
MichaelKetting, line 64: I'm missing the tests for rendereing a validation row in this Test fixture.

MichaelKetting (changes requested):
I think we should check the CSS for both NovaViso and NovaGray in a pair session. I think there's a couple of open questions / checks that need direct input about what stuff works based on our UI style guide :)

Frazerus:
The width calculation of the div is now broken due to changing it to only fill 14/16 columns. 

Will add tests later, think this is still a good point at which you can review.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -188,6 +205,48 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+      var cssClassValidationRow = GetCssClassValidationRow(rowRenderingContext.IsSelected, rowRenderEventArgs);
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, cssClassValidationRow);
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Tr);
+
+      var columnRenderers = renderingContext.ColumnRenderers;
+
+      var validationColumnRenderer = columnRenderers.First(r => r.ColumnDefinition is BocValidationErrorIndicatorColumnDefinition);
+
+      var amountColumnsLeftOfValidation = validationColumnRenderer.VisibleColumnIndex;
```
MichaelKetting, line 224: `numberOfColumnsBeforeValidationColumn`

'amount' is usually a sum result, not a count.
'LeftOfValidation' could also be read as some sort of leftover value. There's also the semantic detail that left/right are only helpful in LTR writing. If the BocList is displayed in RTL writing, the semantics of left/right are inverted so before/after are more precise.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -188,6 +205,48 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+        int rowIndex)
+    {
+      var cssClassValidationRow = GetCssClassValidationRow(rowRenderingContext.IsSelected, rowRenderEventArgs);
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, cssClassValidationRow);
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Tr);
+
+      var columnRenderers = renderingContext.ColumnRenderers;
+
+      var validationColumnRenderer = columnRenderers.First(r => r.ColumnDefinition is BocValidationErrorIndicatorColumnDefinition);
```
MichaelKetting, line 222: Is it guaranteed that there is always a BocValidationErrorIndicatorColumnDefinition when this code is reached? should we still be a bit more expressive and use FirstOrDefault and then add an `Assertion.IsNotNull(..., "BocValidationErrorIndicatorColumnDefinition is not proivded.")` to give better diagnostics / document our assumtions?

MichaelKetting, line 222: possible alternative. Not sure which is better.
```suggestion
      var validationColumnRenderer = columnRenderers.OfType<IBocValidationErrorIndicatorColumnRenderer>().First();
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -188,6 +205,48 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+
+      var amountColumnsLeftOfValidation = validationColumnRenderer.VisibleColumnIndex;
+
+      var totalColumnsCount = columnRenderers.Select(columnRenderer => columnRenderer.VisibleColumnIndex).Prepend(0).Max();
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Colspan, "" + amountColumnsLeftOfValidation);
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Td);
+      renderingContext.Writer.RenderEndTag();
```
MichaelKetting, line 230: I think you need a conditional here, if ColSpan is zero.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -177,6 +180,22 @@ class BocList
         selectedRows.OnSelectionChanged (bocList, false);
       }
     });
+
+    if (row.classList.contains("hasValidationRow")){
+      let validationRow = row.nextSibling as HTMLElement;
```
MichaelKetting, line 185:
```suggestion
    if (row.classList.contains("hasValidationRow"))
    {
      let validationRow = row.nextSibling as HTMLElement;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -177,6 +180,22 @@ class BocList
...
+      validationRow.addEventListener('click', (evt) =>
+      {
+        let hasSelectionChanged = BocList.OnRowClick(evt, bocList, validationRow, selectorControl);
+
+        if (hasSelectionChanged)
+        {
+          row.dispatchEvent(new Event("click"));
+          let selectedRows = BocList._selectedRows[bocList.id]!;
+          selectedRows.OnSelectionChanged(bocList, false);
+        }
```
MichaelKetting, line 196: Not sure if that's not duplicate logic. I see that you dispatch the click event on the "parent" row and then also do all the things yourself, too. It seems to me that all we need to do is perform an unconditional disapatch onto the parent row and we're done.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -201,7 +220,7 @@ class BocList
     }
   }
 
-  //  Event handler for a table row in the BocList. 
+//  Event handler for a table row in the BocList.
```
MichaelKetting, line 223: revert

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -503,14 +527,29 @@ class BocList
...
+  {
+    // Sets the width of the validation error field to be as wide as the visible part of the bocList.
+    // This enables sticky behaviour while still allowing for a scrollbar if required.
+
+    const width = LayoutUtility.GetWidth(scrollableContainer) - 1;
+
+    const overflowContainer: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow div');
+    if (overflowContainer)
+    {
+      //overflowContainer.style.width = width.toString() + "px";
```
MichaelKetting, line 549: that's commented out?

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Frazerus: Yeah, my comment above is about this: Due to the colspan rendering, we now cannot just apply the width of the visible part of the bocList, as this would cause overflowing. As this still needs fixing, I just commented it out.

Frazerus: Unless I have no clue how this works and it's really easy, I don't think this should happen in this task.

This will probably take significantly more time than I think would be reasonable to append onto this PR now (0.5 - 1pt I would guess, unless there's a really easy way to do it...which I do not know)

Frazerus:
Everything we wanted should be here now, no more "need to still do x"

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
```diff
@@ -435,6 +451,65 @@ public void TestDiagnosticMetadataRendering ()
...
+      var outerDiv = Html.GetAssertedChildElement(td, "div", 0);
+
+      var innerDiv = Html.GetAssertedChildElement(outerDiv, "div", 0);
+
+      var ul = Html.GetAssertedChildElement(innerDiv, "ul", 0);
+      var li = Html.GetAssertedChildElement(ul, "li", 0);
+
+      var span = Html.GetAssertedChildElement(li, "span", 0);
+
+      Assert.That(span.InnerText, Is.EqualTo("Big error"));
```
MichaelKetting, line 510: Please move to separate test for BocListValidationSummaryRenderer. There shold be more test code anway. It's possible that we need to move these tests into a separate task so we can finish up here.

We can use a stub-implementation here, that just renders a dummy tag with relevant information. It's also okay (we just didn't do this yet) that we capture the BocListValidationSummaryRenderArguments in the stub-object and then assert the BocListValidationSummaryRenderArguments after rendering has completed. This would depend on how many details we want to assert.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
```diff
@@ -435,6 +451,65 @@ public void TestDiagnosticMetadataRendering ()
...
+
+      var trRow = Html.GetAssertedChildElement(root, "tr", 0);
+      var trValidationRow = Html.GetAssertedChildElement(root, "tr", 1);
+
+      var td = Html.GetAssertedChildElement(trValidationRow, "td", 0);
+      td.AssertAttributeValueEquals("colspan", "6");
+
+      var outerDiv = Html.GetAssertedChildElement(td, "div", 0);
+
+      var innerDiv = Html.GetAssertedChildElement(outerDiv, "div", 0);
```
MichaelKetting, line 503: somewhere here we have CSS classes. please add assertions for those.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocRowRenderer.cs
```diff
@@ -188,6 +205,48 @@ private IBocIndexColumnRenderer GetIndexColumnRenderer ()
...
+
+      var lastColumnIndex = 0;
+      for (int i = columnRenderers.Length - 1; i >= 0; i--)
+      {
+        if (lastColumnIndex > columnRenderers[i].VisibleColumnIndex)
+          continue;
+
+        lastColumnIndex = columnRenderers[i].VisibleColumnIndex + 1;
+        break;
+      }
```
MichaelKetting, line 230: I try to never have `continue` in a loop and to avoid `break` to make the code easier to reason over. Having both `continue` and `break` in a single loop makes me queasy. I did not yet think how to improve this, am open for a pair session on this. One final break could be okay if we need to have an early exit condition.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
Frazerus: There are already tests for the SummaryRenderer, if there were too little, we should probably add them from the other ticket/ created a ticket during the other ticket.

MichaelKetting: ah, ok. didn't check this PR in detail. maybe quick look-see tomorrow (Wednesday)

MichaelKetting (approved):

PR #732, opened 2023-06-15
RM-6423 Unify implementation of CreateValidators()

KuJo7:
https://re-motion.atlassian.net/browse/RM-6423

### DIFF
--- a/Remotion/ObjectBinding/Web/UI/Controls/BusinessObjectBoundWebControl.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BusinessObjectBoundWebControl.cs
@@ -432,10 +432,8 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
       get { return false; }
     }
 
-    IEnumerable<BaseValidator> ISmartControl.CreateValidators ()
-    {
-      return Enumerable.Empty<BaseValidator>();
-    }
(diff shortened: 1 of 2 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BusinessObjectBoundWebControl.cs
```diff
@@ -432,10 +432,10 @@ bool ISmartControl.IsRequired
...
     }
 
-    IEnumerable<BaseValidator> ISmartControl.CreateValidators ()
-    {
-      return Enumerable.Empty<BaseValidator>();
-    }
+    IEnumerable<BaseValidator> ISmartControl.CreateValidators () => CreateValidators ();
+
+    /// <inheritdoc cref="ISmartControl.CreateValidators"/>
+    protected virtual IEnumerable<BaseValidator> CreateValidators () => Enumerable.Empty<BaseValidator>();
```
MichaelKetting, line 438: We need to keep CreateValidators as seen from BusinessObjectBoundEditableWebControl as a public member. Otherwise, we introduce a breaking change for projects that access the CreateValidators() method.

=> unify the explcit interface implementation and the protected virtual with a public implicit interface impelmetnation on BusinessObjectBoundWebControl.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #736, opened 2023-06-21
RM-8849 Implement tests for AddingObjectValidationRuleCollector

KuJo7:
https://re-motion.atlassian.net/browse/RM-8849

### DIFF
--- a/Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
+++ b/Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
@@ -18,9 +18,12 @@ using System;
 using System.Linq;
 using Moq;
 using NUnit.Framework;
+using Remotion.Development.NUnit.UnitTesting;
+using Remotion.Reflection;
 using Remotion.Validation.Implementation;
 using Remotion.Validation.Merging;
 using Remotion.Validation.RuleCollectors;
+using Remotion.Validation.Rules;
 using Remotion.Validation.UnitTests.TestDomain;
 using Remotion.Validation.UnitTests.TestDomain.Collectors;
 using Remotion.Validation.UnitTests.TestDomain.Validators;
@@ -32,11 +35,12 @@ namespace Remotion.Validation.UnitTests.RoleCollectors
   [TestFixture]
   public class AddingObjectValidationRuleCollector1Test
   {
-    private IAddingObjectValidationRuleCollector _addingObjectValidationRuleCollector;
+    private AddingObjectValidationRuleCollector<Customer> _addingObjectValidationRuleCollector;
     private Mock<IObjectValidatorExtractor> _objectValidatorExtractorMock;
     private IObjectValidator _stubObjectValidator1;
     private Mock<IObjectValidator> _stubObjectValidator2;
     private IObjectValidator _stubObjectValidator3;
+    private ITypeInformation _validatedType;
 
     [SetUp]
     public void SetUp ()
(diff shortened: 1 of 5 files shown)


Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
```diff
@@ -46,54 +50,159 @@ public void SetUp ()
...
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void Initialization_CollectorTypeDoesNotImplementIValidationRuleCollector_ThrowsArgumentException ()
     {
+      Assert.That(
+          () => new AddingObjectValidationRuleCollector<Customer>(typeof(Customer)),
+          Throws.ArgumentException
+              .With.ArgumentExceptionMessageEqualTo(
+                  "Parameter 'collectorType' is a 'Remotion.Validation.UnitTests.TestDomain.Customer', "
```
Frazerus, line 63:
```suggestion
                  "Parameter 'collectorType' is a 'Remotion.Validation.UnitTests.TestDomain.Customer' "
```

Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
```diff
@@ -46,54 +50,159 @@ public void SetUp ()
...
+      var fakeObjectValidators1 = ((ObjectValidationRule<Customer>)result1).Validators.Cast<FakeObjectValidator>().ToArray();
+      Assert.That(fakeObjectValidators1.Length, Is.EqualTo(1));
+      Assert.That(fakeObjectValidators1[0].ValidationMessage.ToString(), Is.EqualTo("expectedMessage1"));
+
+      _addingObjectValidationRuleCollector.RegisterValidator(validatorFactory);
+      var result2 = ((IAddingObjectValidationRuleCollector)_addingObjectValidationRuleCollector).CreateValidationRule(validationMessageFactoryStub2.Object);
+
+      var fakeObjectValidators2 = ((ObjectValidationRule<Customer>)result2).Validators.Cast<FakeObjectValidator>().ToArray();
+      Assert.That(fakeObjectValidators2.Length, Is.EqualTo(2));
+      Assert.That(fakeObjectValidators2[0].ValidationMessage.ToString(), Is.EqualTo("expectedMessage1"));
```
Frazerus, line 113: We may want to check for the whole object at once instead of checking each message separately:
```suggestion
      Assert.That(fakeObjectValidators2.Select(f => f.ValidationMessage.ToString()), Is.EquivalentTo(expectedMessages));
```
Not sure if you want to actually do it here, but I got feedback about checking things separately in multiple asserts when it would be more readable to have one assert with an expected list of objects.

Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
```diff
@@ -46,54 +50,159 @@ public void SetUp ()
...
+          .Setup(e => e.CreateValidationMessageForObjectValidator(It.IsAny<IObjectValidator>(), _validatedType))
+          .Returns(new InvariantValidationMessage("expectedMessage"));
+
+      Func<ObjectValidationRuleInitializationParameters, IObjectValidator> validatorFactory = param => new FakeObjectValidator(param.ValidationMessage);
+
+      _addingObjectValidationRuleCollector.RegisterValidator(validatorFactory);
+      var result = ((IAddingObjectValidationRuleCollector)_addingObjectValidationRuleCollector).CreateValidationRule(validationMessageFactoryStub.Object);
+
+      var fakeObjectValidators = ((ObjectValidationRule<Customer>)result).Validators.Cast<FakeObjectValidator>().ToArray();
+      Assert.That(fakeObjectValidators.Length, Is.EqualTo(1));
```
Frazerus, line 82: You might want to use `Single()` here, as it throws when there is more than one object contained. Using that, you can reduce these 2 asserts to a single one.

Frazerus: This can also be done with all the other ones where you have 2 asserts for a singular object. I don't know if this is actually good here, but I got feedback in my PRs that using Single would be a nice option.

Frazerus (approved):

Remotion/Validation/UnitTests/TestDomain/Validators/FakeObjectValidator.cs
```diff
@@ -0,0 +1,38 @@
...
+    public ValidationMessage ValidationMessage { get; }
+
+    public FakeObjectValidator (ValidationMessage validationMessage)
+    {
+      ValidationMessage = validationMessage;
+    }
+
+    public IEnumerable<ObjectValidationFailure> Validate (ObjectValidatorContext context)
+    {
+      throw new NotImplementedException();
```
MichaelKetting, line 35: It's okay to keep this as NotImplementedException. 

Suggestion:
I think I usually go to `return Enumreable,Empty<...>()` in cases such as this because it means the type won't break validation runs should it ever get used during such an operation.

Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
```diff
@@ -46,54 +50,159 @@ public void SetUp ()
...
     public void CreateValidationRule_IgnoresValidationMessagesForValidatorsWithoutADeferredInitializationValidationMessage ()
     {
+      var validationMessageFactoryStub = new Mock<IValidationMessageFactory>();
+      validationMessageFactoryStub
+          .Setup(e => e.CreateValidationMessageForObjectValidator(It.IsAny<IObjectValidator>(), _validatedType))
+          .Returns(new InvariantValidationMessage("unexpectedMessage"));
+
+      Func<ObjectValidationRuleInitializationParameters, IObjectValidator> validatorFactory = _ => new FakeObjectValidator(new InvariantValidationMessage("expectedMessage"));
+
+      (_addingObjectValidationRuleCollector).RegisterValidator(validatorFactory);
```
MichaelKetting, line 127:
```suggestion
      _addingObjectValidationRuleCollector.RegisterValidator(validatorFactory);
```

MichaelKetting (approved):

Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
KuJo7: I don't think that I should make this change. Then I would have to change it in `ArgumentUtility.cs` and also adapt all usages.

Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
KuJo7: Since I have it already this way in `AddingPropertyValidationRuleCollector1Test.cs` (merged), I think I would leave it as it is, because the tests should follow a similar logic

Remotion/Validation/UnitTests/RoleCollectors/AddingObjectValidationRuleCollector1Test.cs
KuJo7: Same as above. Thanks for the tip anyway, I will keep this in mind for my next PRs :)

PR #744, opened 2023-06-27
RM-8851 Update constraints of Length-Validators

KuJo7:
https://re-motion.atlassian.net/browse/RM-8851

### DIFF
--- a/Remotion/Validation/UnitTests/Validators/MinimumLengthValidatorTest.cs
+++ b/Remotion/Validation/UnitTests/Validators/MinimumLengthValidatorTest.cs
@@ -29,7 +29,7 @@ namespace Remotion.Validation.UnitTests.Validators
     public void Validate_WithPropertyValueNull_ReturnsNoValidationFailures ()
     {
       var propertyValidatorContext = CreatePropertyValidatorContext(null);
-      var validator = new MinimumLengthValidator(0, new InvariantValidationMessage("Fake Message"));
+      var validator = new MinimumLengthValidator(1, new InvariantValidationMessage("Fake Message"));
 
       var validationFailures = validator.Validate(propertyValidatorContext);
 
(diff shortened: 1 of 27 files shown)


MichaelKetting (commented):
@KuJo7 We have a lot of Length-Validators that previously only checked the MaxLength and used zero for MinLength. Could you do a quick check what you would need to update in the tests if you changed those LengthAttributes and .Lenth(...) extension methods to the MaxLengthAttribute and .MaxLength(...) extension method? Inside the test-domain, it's not really mandatory that everything is semantically sensible but if we can easily switch it around, it might be a good idea. Also, by changing the min-length, it can result in changed assertions.

Please timebox this. I have no idea just how many tests will fail if we do this. it could also be that we only change some of the validation rules.

MichaelKetting (commented):
We need to remember to add BreakingChange doc for this and the LengthAttribute tasks

Remotion/Validation/UnitTests/Validators/MinimumLengthValidatorTest.cs
```diff
@@ -40,22 +40,27 @@ public void Validate_WithPropertyValueNull_ReturnsNoValidationFailures ()
...
+      var validator = new MinimumLengthValidator(1, new InvariantValidationMessage("Fake Message"));
 
       var validationFailures = validator.Validate(propertyValidatorContext);
 
       Assert.That(validationFailures, Is.Empty);
     }
 
     [Test]
-    public void Validate_WithPropertyValueEqualsEmptyStringAndMinLengthZero_ReturnsNoValidationFailures ()
+    public void Validate_WithPropertyValueEqualsEmptyStringAndMinLengthOne_ReturnsSingleValidationFailure ()
```
MichaelKetting, line 51: Please check, not sure if this test isn't redundant now.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #746, opened 2023-06-28
RM-8854 Add tests for AddingObject/PropertyValidationRuleBuilder

KuJo7:
https://re-motion.atlassian.net/browse/RM-8854

### DIFF
--- a/Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
+++ b/Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
@@ -14,61 +14,156 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
+using System;
+using Moq;
 using NUnit.Framework;
+using Remotion.Reflection;
+using Remotion.Validation.MetaValidation;
+using Remotion.Validation.RuleBuilders;
+using Remotion.Validation.RuleCollectors;
+using Remotion.Validation.Rules;
+using Remotion.Validation.UnitTests.TestDomain;
+using Remotion.Validation.Validators;
 
 namespace Remotion.Validation.UnitTests.RuleBuilders
 {
   [TestFixture]
   public class AddingObjectValidationRuleBuilderTest
   {
+    private Mock<IAddingObjectValidationRuleCollector> _addingObjectValidationRuleCollectorMock;
+    private AddingObjectValidationRuleBuilder<Customer> _addingObjectValidationBuilder;
+    private Mock<IObjectValidator> _objectValidatorStub;
+    private Mock<IObjectMetaValidationRuleCollector> _objectMetaValidationRuleCollectorMock;
 
+    [SetUp]
+    public void SetUp ()
+    {
+      _addingObjectValidationRuleCollectorMock = new Mock<IAddingObjectValidationRuleCollector>(MockBehavior.Strict);
+      _addingObjectValidationRuleCollectorMock
+          .Setup(stub => stub.ValidatedType)
+          .Returns(TypeAdapter.Create(typeof(Customer)));
+
+      _objectMetaValidationRuleCollectorMock = new Mock<IObjectMetaValidationRuleCollector>(MockBehavior.Strict);
+      _addingObjectValidationRuleCollectorMock
+          .Setup(stub => stub.ValidatedType)
+          .Returns(TypeAdapter.Create(typeof(Customer)));
+
+      _addingObjectValidationBuilder = new AddingObjectValidationRuleBuilder<Customer>(
+          _addingObjectValidationRuleCollectorMock.Object,
+          _objectMetaValidationRuleCollectorMock.Object);
+
+      _objectValidatorStub = new Mock<IObjectValidator>();
+    }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void Initialization ()
     {
+      Assert.That(_addingObjectValidationBuilder.AddingObjectValidationRuleCollector, Is.SameAs(_addingObjectValidationRuleCollectorMock.Object));
+      Assert.That(_addingObjectValidationBuilder.ObjectMetaValidationRuleCollector, Is.SameAs(_objectMetaValidationRuleCollectorMock.Object));
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void SetValidator ()
     {
+      Func<ObjectValidationRuleInitializationParameters, IObjectValidator> validatorFactory = _ => _objectValidatorStub.Object;
+      _addingObjectValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterValidator(validatorFactory))
+          .Verifiable();
+
+      ((IAddingObjectValidationRuleBuilder<Customer>)_addingObjectValidationBuilder).SetValidator(validatorFactory);
+
+      _addingObjectValidationRuleCollectorMock.Verify();
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CanBeRemoved ()
     {
+      _addingObjectValidationRuleCollectorMock
+          .Setup(mock => mock.SetRemovable())
+          .Verifiable();
+
+      _addingObjectValidationBuilder.CanBeRemoved();
+
+      _addingObjectValidationRuleCollectorMock.Verify();
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void AddMetaValidationRule ()
     {
+      var metaValidationRuleStub = new Mock<IObjectMetaValidationRule>();
+
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(metaValidationRuleStub.Object))
+          .Verifiable();
+
+      _addingObjectValidationBuilder.AddMetaValidationRule(metaValidationRuleStub.Object);
+
+      _objectMetaValidationRuleCollectorMock.Verify();
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void AddMetaValidationRule_FuncOverload ()
     {
+      var fakeValidationResult = MetaValidationRuleValidationResult.CreateValidResult();
+
+      IObjectMetaValidationRule actualRule = null;
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => actualRule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule(v => fakeValidationResult);
+
+      Assert.That(actualRule, Is.TypeOf<DelegateObjectMetaValidationRule<IObjectValidator>>());
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Is.EquivalentTo(new [] { fakeValidationResult }));
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void AddMetaValidationRule_ExpressionOverload_IsValidIsTrue ()
     {
+      IObjectMetaValidationRule actualRule = null;
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => actualRule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule<IObjectValidator>(v => true);
+
+      Assert.That(actualRule, Is.TypeOf<DelegateObjectMetaValidationRule<IObjectValidator>>());
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Has.Exactly(1).With.Property(nameof(MetaValidationRuleValidationResult.IsValid)).True);
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void AddMetaValidationRule_ExpressionOverload_IsValidIsFalse ()
     {
+      IObjectMetaValidationRule actualRule = null;
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => actualRule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule<IObjectValidator>(v => false);
+
+      Assert.That(actualRule, Is.TypeOf<DelegateObjectMetaValidationRule<IObjectValidator>>());
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Has.Exactly(1).Items);
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Has.All.With.Property(nameof(MetaValidationRuleValidationResult.IsValid)).False);
+      Assert.That(
+          actualRule.Validate(new IObjectValidator[0]),
+          Has.All.With.Property(nameof(MetaValidationRuleValidationResult.Message))
+              .EqualTo(
+                  "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IObjectValidator' "
+                  + "on type 'Remotion.Validation.UnitTests.TestDomain.Customer'."));
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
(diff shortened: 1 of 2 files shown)


Remotion/Validation/UnitTests/RuleBuilders/AddingPropertyValidationRuleBuilderTest.cs
```diff
@@ -146,9 +147,15 @@ public void AddMetaValidationRule_ExpressionOverload_IsValidIsFalse ()
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void SetCondition ()
     {
+      Func<Customer, bool> predicate = _ => true;
+
+      _addingPropertyValidationRuleCollectorMock
+          .Setup(mock => mock.SetCondition(predicate)).Verifiable();
```
MichaelKetting, line 155:
```suggestion
      _addingPropertyValidationRuleCollectorMock
          .Setup(mock => mock.SetCondition(predicate))
          .Verifiable();
```
Fluent Moq syntax gets alined 

Note: there's another way of writing this by adding the verify for the call at the end.
```
      _addingPropertyValidationRuleCollectorMock
          .Verify(mock => mock.SetCondition(predicate));
```

This way we don't use the setup and the general purpose Verify. It only works if no setup is required (i.e. no return value is needed). I tend to mix and mingle based on readability.

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,148 @@
...
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void SetCondition ()
     {
+      Func<Customer, bool> predicate = _ => true;
+
+      _addingObjectValidationRuleCollectorMock
+          .Setup(mock => mock.SetCondition(predicate)).Verifiable();
```
MichaelKetting, line 154:
```suggestion
      _addingObjectValidationRuleCollectorMock
          .Setup(mock => mock.SetCondition(predicate))
          .Verifiable();
```

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,148 @@
...
+      _objectMetaValidationRuleCollectorMock
+          .Setup(
+              mock => mock.RegisterMetaValidationRule(
+                  It.Is<IObjectMetaValidationRule>(
+                      rule => rule.GetType() == typeof(DelegateObjectMetaValidationRule<IObjectValidator>)
+                              && !rule.Validate(new IObjectValidator[0]).First().IsValid
+                              && rule.Validate(new IObjectValidator[0]).First().Message
+                              == "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IObjectValidator' "
+                              + "on type 'Remotion.Validation.UnitTests.TestDomain.Customer'.")))
+          .Verifiable();
```
MichaelKetting, line 141: I would prefer we captur the rule via a callback and do the assertion a the end of the test. that way, such a complex assertion logic can provide helpful diagnostic. now, if any of the boolean conditions fails, the test is not matched and we have a really hard time figuring out what went wrong. This applies to the other tests, too.

MichaelKetting (changes requested):

Remotion/Validation/UnitTests/RuleBuilders/AddingPropertyValidationRuleBuilderTest.cs
MichaelKetting: okay, so we keep the existing syntax

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,153 @@
...
+      _objectMetaValidationRuleCollectorMock.Verify();
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void AddMetaValidationRule_FuncOverload ()
     {
+      var fakeValidationResult = MetaValidationRuleValidationResult.CreateValidResult();
+
+      IObjectMetaValidationRule rule = null;
```
MichaelKetting, line 110: naming: I prefer to use "actual" in the name, e.g. "actualRule" to identify the value we want to check. Or "result", when it's the return value of the exercised SUT.

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,153 @@
...
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => rule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule(v => fakeValidationResult);
+
+      Assert.That(rule.GetType(), Is.EqualTo(typeof(DelegateObjectMetaValidationRule<IObjectValidator>)));
+      Assert.That(rule.Validate(new IObjectValidator[0]).First(), Is.EqualTo(fakeValidationResult));
```
MichaelKetting, line 118:
```suggestion
      Assert.That(rule, Is.TypeOf<DelegateObjectMetaValidationRule<IObjectValidator>>));
      Assert.That(rule.Validate(new IObjectValidator[0]), Is.EquivalentTo(new[]{ fakeValidationResult }));
```
In both cases, we use native NUnit operations to do the assertions, getting a better error message in case of red tests. in the second case, we want to check a result collection. this can be done via EqualTo or EquivalentTo, the difference that EqualTo also asserts position, not just value. Since the order of the result items isn't relevant here, EquivalentTo is best. By using First, we only test that there's a first item, not that there could also be others. In total, the updated version is more robust and has better diagnostics.

Please check the rest of the PR for similiar cases.

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,153 @@
...
     {
+      IObjectMetaValidationRule rule = null;
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => rule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule<IObjectValidator>(v => true);
+
+      Assert.That(rule.GetType(), Is.EqualTo(typeof(DelegateObjectMetaValidationRule<IObjectValidator>)));
+      Assert.That(rule.Validate(new IObjectValidator[0]).First().IsValid, Is.True);
```
MichaelKetting, line 132: This one's a bit trickier. We fully expect that there's just one item, so we could at least use Single instead of First. Alternatively, and possibly nicer, is that we can just go an do the following:
```suggestion
      Assert.That(rule.Validate(new IObjectValidator[0]), Has.Exactly(1).With.Property(nameof(MetaValidationRuleValidationResult.IsValid)).True);
```

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,153 @@
...
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => rule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule<IObjectValidator>(v => false);
+
+      Assert.That(rule.GetType(), Is.EqualTo(typeof(DelegateObjectMetaValidationRule<IObjectValidator>)));
+      Assert.That(rule.Validate(new IObjectValidator[0]).First().IsValid, Is.False);
+      Assert.That(rule.Validate(new IObjectValidator[0]).First().Message, Is.EqualTo(
+          "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IObjectValidator' on type 'Remotion.Validation.UnitTests.TestDomain.Customer'."));
```
MichaelKetting, line 148:
```suggestion
      var result = rule.Validate(new IObjectValidator[0]);
      Assert.That(result , Has.Exactly(1).Items);
      Assert.That(result , Has.All.With.Property(nameof(MetaValidationRuleValidationResult.IsValid)).False);
            Assert.That(result , Has.All.With.Property(nameof(MetaValidationRuleValidationResult.Message)).EqualTo(
          "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IObjectValidator' on type 'Remotion.Validation.UnitTests.TestDomain.Customer'."));
```
We could also combine the two conditions using `And` and then then we don't need the `result` variable. It's a matter of style and formatting.  I went with the above approach because it was easier to prepare in the review.

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,153 @@
...
+      Func<Customer, bool> predicate = _ => true;
+
+      _addingObjectValidationRuleCollectorMock
+          .Setup(mock => mock.SetCondition(predicate))
+          .Verifiable();
+
+      _addingObjectValidationBuilder.SetCondition(predicate);
+
+      _addingObjectValidationRuleCollectorMock
+          .Verify(mock => mock.SetCondition(predicate));
```
MichaelKetting, line 163:
```suggestion
      _addingObjectValidationRuleCollectorMock.Verify();
```
Unless there's a good reason we need to do it this way, I try to avoid repeating myself. Either, we don't need the setup in line 156 or we do just a generic `Verify` in the assertion block of the test. Repeating the actual call is annoying when you need to update stuff in the future.

MichaelKetting (changes requested):
I found a bunch of things that apply to many of the tests in this PR. Please check them all :)

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
KuJo7: There seems to be an error. I think you missed `typeof()` in the first Assert:
`Assert.That(rule, Is.TypeOf(typeof(DelegateObjectMetaValidationRule<IObjectValidator>)));`

MichaelKetting: @KuJo7, actually, I was using Generics syntax and missed the method call parens at the end.

Remotion/Validation/UnitTests/RuleBuilders/AddingPropertyValidationRuleBuilderTest.cs
```diff
@@ -40,10 +39,14 @@ public class AddingPropertyValidationRuleBuilderTest
     public void SetUp ()
     {
       _addingPropertyValidationRuleCollectorMock = new Mock<IAddingPropertyValidationRuleCollector>(MockBehavior.Strict);
-      _addingPropertyValidationRuleCollectorMock.Setup(stub => stub.Property).Returns(PropertyInfoAdapter.Create(typeof(Customer).GetProperty("UserName")));
+      _addingPropertyValidationRuleCollectorMock
+          .Setup(stub => stub.Property)
+          .Returns(PropertyInfoAdapter.Create(typeof(Customer).GetProperty("UserName")));
```
MichaelKetting, line 44:
```suggestion
          .Returns(PropertyInfoAdapter.Create(MemberInfoFromExpressionUtility.GetProperty((Customer _) => _.UserName)));
```
```suggestion
          .Returns(PropertyInfoAdapter.Create(MemberInfoFromExpressionUtility.GetProperty((Customer _) => _.UserName)));
```
Info:
This syntax is typesafe. I prefer to use it with new code but we do have enough older code, too, so it doesn't stand out as "strange" when it is used.

Remotion/Validation/UnitTests/RuleBuilders/AddingPropertyValidationRuleBuilderTest.cs
```diff
@@ -97,58 +106,62 @@ public void AddMetaValidationRule_FuncOverload ()
...
 
       _addingPropertyValidationBuilder.AddMetaValidationRule<IPropertyValidator>(v => false);
 
-      _propertyMetaValidationRuleCollectorMock.Verify();
+      Assert.That(actualRule, Is.TypeOf<DelegatePropertyMetaValidationRule<IPropertyValidator>>());
+      Assert.That(actualRule.Validate(new IPropertyValidator[0]), Has.Exactly(1).Items);
+      Assert.That(actualRule.Validate(new IPropertyValidator[0]), Has.All.With.Property(nameof(MetaValidationRuleValidationResult.IsValid)).False);
+      Assert.That(actualRule.Validate(new IPropertyValidator[0]), Has.All.With.Property(nameof(MetaValidationRuleValidationResult.Message)).EqualTo(
+          "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IPropertyValidator' "
+          + "on property 'Remotion.Validation.UnitTests.TestDomain.Customer.UserName'."));
```
MichaelKetting, line 150:
```suggestion
      Assert.That(
          actualRule.Validate(new IPropertyValidator[0]),
          Has.All.With.Property(nameof(MetaValidationRuleValidationResult.Message))
              .EqualTo(
                  "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IPropertyValidator' "
                  + "on property 'Remotion.Validation.UnitTests.TestDomain.Customer.UserName'."));
```

Linebreaks: 
when a method call contains linebreaks, the arguments start in a new line.
For assertions, the Message-test (EqualTo etc) goes into a new line and depending on the length of the string, we can keep it in a signle line or need more line wraps.

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,152 @@
...
+
+      IObjectMetaValidationRule actualRule = null;
+      _objectMetaValidationRuleCollectorMock
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => actualRule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule(v => fakeValidationResult);
+
+      Assert.That(actualRule, Is.TypeOf<DelegateObjectMetaValidationRule<IObjectValidator>>());
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Is.EquivalentTo(new[]{ fakeValidationResult }));
```
MichaelKetting, line 117:
```suggestion
      Assert.That(actualRule.Validate(new IObjectValidator[0]), Is.EquivalentTo(new [] { fakeValidationResult }));
```

Remotion/Validation/UnitTests/RuleBuilders/AddingObjectValidationRuleBuilderTest.cs
```diff
@@ -14,61 +14,152 @@
...
+          .Setup(mock => mock.RegisterMetaValidationRule(It.IsAny<IObjectMetaValidationRule>()))
+          .Callback((IObjectMetaValidationRule r) => actualRule = r);
+
+      _addingObjectValidationBuilder.AddMetaValidationRule<IObjectValidator>(v => false);
+
+      Assert.That(actualRule, Is.TypeOf<DelegateObjectMetaValidationRule<IObjectValidator>>());
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Has.Exactly(1).Items);
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Has.All.With.Property(nameof(MetaValidationRuleValidationResult.IsValid)).False);
+      Assert.That(actualRule.Validate(new IObjectValidator[0]), Has.All.With.Property(nameof(MetaValidationRuleValidationResult.Message)).EqualTo(
+          "Meta validation rule 'v => False' failed for validator 'Remotion.Validation.Validators.IObjectValidator' on type 'Remotion.Validation.UnitTests.TestDomain.Customer'."));
```
MichaelKetting, line 148: See corresponding test in the other testfixture about formating.

MichaelKetting (changes requested):
formatting.

PR #747, opened 2023-06-28
RM-8855 Add tests for TypeBasedValidationMessageFactory

KuJo7:
https://re-motion.atlassian.net/browse/RM-8855

### DIFF
--- a/Remotion/Validation/Globalization.UnitTests/TypeBasedValidationMessageFactoryTest.cs
+++ b/Remotion/Validation/Globalization.UnitTests/TypeBasedValidationMessageFactoryTest.cs
@@ -1,32 +1,82 @@
-﻿using NUnit.Framework;
+﻿using Moq;
+using NUnit.Framework;
+using Remotion.Globalization;
+using Remotion.Reflection;
+using Remotion.Validation.Globalization.UnitTests.TestHelpers;
+using Remotion.Validation.Implementation;
 
 namespace Remotion.Validation.Globalization.UnitTests
 {
   [TestFixture]
   public class TypeBasedValidationMessageFactoryTest
   {
+    private TypeBasedValidationMessageFactory _factory;
+    private Mock<IPropertyInformation> _propertyStub;
+    private Mock<ITypeInformation> _objectStub;
+    private Mock<IMemberInformationGlobalizationService> _serviceStub;
+
+    [SetUp]
+    public void SetUp ()
+    {
+      _serviceStub = new Mock<IMemberInformationGlobalizationService>();
+      _factory = new TypeBasedValidationMessageFactory(_serviceStub.Object);
+
+      _propertyStub = new Mock<IPropertyInformation>();
+      _objectStub = new Mock<ITypeInformation>();
+    }
+
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CreateValidationMessageForPropertyValidator_WithTypeNotHavingLocalizations_ReturnsNull ()
     {
+      var validator = new StubPropertyValidator();
+      var result = _factory.CreateValidationMessageForPropertyValidator(validator, _propertyStub.Object);
+
+      Assert.That(result, Is.Null);
     }
 
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CreateValidationMessageForPropertyValidator_WithLocalizedType_ReturnsLocalizedValidationMessage ()
     {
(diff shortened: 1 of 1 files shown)


Remotion/Validation/Globalization.UnitTests/TypeBasedValidationMessageFactoryTest.cs
```diff
@@ -1,32 +1,87 @@
...
     [Test]
-    [Ignore("TODO RM-5906")]
     public void CreateValidationMessageForObjectValidator_WithLocalizedType_ReturnsLocalizedValidationMessage ()
     {
+      var validator = new StubObjectValidator();
+
+      var str = "LocalizedValidationMessage";
+      _serviceStub
+          .Setup(s => s.TryGetTypeDisplayName(
+              It.Is<ITypeInformation>(e => e is TypeAdapter && ((TypeAdapter)e).Type == typeof(StubObjectValidator)),
```
MichaelKetting, line 76: What does 'e' stand for? 

```suggestion
              It.Is<TypeAdapter>(e => e.Type == typeof(StubObjectValidator)),
```

I think this should work.  If so, please update the other tests, too.

Actually, there is an even better option: just use TypeAdapter.Get(typeof(StubObjectValidator)) for the first parameter and TypeAdapter.Get(_objectStub.Object.GetType()) for the second parameter. This way, we don';t need any `It.Is<..>` logic. You might need to capture those into variables before using them in the Setup call (not sure about this detail). The reason for the need to capture is that Moq might interpret the TypeAdapter.Get-call as part of the setup logic. So, first you can try if this works without using intermediary variables, and if not, introduce variables.

MichaelKetting (changes requested):

Remotion/Validation/Globalization.UnitTests/TypeBasedValidationMessageFactoryTest.cs
a-ctor: @KuJo7 Since we went over this yesterday. I made a wrong assumption about `TypeAdapter`. While the equality uses reference equality the `TypeAdapter.Create` method caches instances so you will get back the same instance and equality works as expected. As mike has said this allows us to drop the complicated `It.Is` setup.

Side note: the method is called create so it is `TypeAdapter.Create(typeof(StubObjectValidator))` for both parameters as the type is used for both parameters for some reason.

MichaelKetting: > Side note: the method is called create so it is TypeAdapter.Create(typeof(StubObjectValidator)) for both parameters as the type is used for both parameters for some reason.

This second bit, I want to post an update around noon.

Remotion/Validation/Globalization.UnitTests/TypeBasedValidationMessageFactoryTest.cs
```diff
@@ -1,32 +1,88 @@
...
+
+      var validator = new StubObjectValidator();
+
+      var actualMessage = "LocalizedValidationMessage";
+      _serviceStub
+          .Setup(s => s.TryGetTypeDisplayName(
+              typeInformation,
+              typeInformation,
+              out actualMessage))
+          .Returns(true);
```
MichaelKetting, line 80:
```suggestion
      _serviceStub
          .Setup(_ => _.TryGetTypeDisplayName(typeInformation, typeInformation, out actualMessage))
          .Returns(true);
```

given the call length, doing this in a single line feels like better use of space.

Remotion/Validation/Globalization.UnitTests/TypeBasedValidationMessageFactoryTest.cs
```diff
@@ -1,32 +1,88 @@
...
+
+      var validator = new StubPropertyValidator();
+
+      var actualMessage = "LocalizedValidationMessage";
+      _serviceStub
+          .Setup(s => s.TryGetTypeDisplayName(
+              typeInformation,
+              typeInformation,
+              out actualMessage))
+          .Returns(true);
```
MichaelKetting, line 50:
```suggestion
      _serviceStub
          .Setup(_ => _.TryGetTypeDisplayName(typeInformation, typeInformation, out actualMessage))
          .Returns(true);
```

We just go with `_` for this kind of parameter because it reduces the noise.

MichaelKetting (approved):
formatting

PR #748, opened 2023-06-30
RM-8863 Create specific type of length-validator based on min and max…

KuJo7:
https://re-motion.atlassian.net/browse/RM-8863

### DIFF
--- a/Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
+++ b/Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
@@ -29,29 +29,21 @@ namespace Remotion.Validation.Attributes.Validation
   /// </summary>
   public class LengthValidationAttribute : AddingValidationAttributeBase
   {
-    // TODO RM-5906: make max-length nullable and create specific type of length-validator based on min and max-length values
-    private readonly int _maxLength;
-    private readonly int _minLength;
+    /// <summary>
+    /// Gets or sets the minimum number of characters required.
+    /// </summary>
+    public int MinLength { get; set; }
+
+    /// <summary>
+    /// Gets or sets the maximum number of characters allowed.
+    /// </summary>
+    public int MaxLength { get; set; }
 
     /// <summary>
     /// Instantiates a new <see cref="LengthValidationAttribute"/>.
     /// </summary>
-    /// <param name="minLength">The minimum number of characters required.</param>
-    /// <param name="maxLength">The maximum number of characters allowed.</param>
-    public LengthValidationAttribute (int minLength, int maxLength)
+    public LengthValidationAttribute ()
     {
-      _minLength = minLength;
-      _maxLength = maxLength;
-    }
-
-    public int MinLength
-    {
-      get { return _minLength; }
-    }
-
-    public int MaxLength
-    {
-      get { return _maxLength; }
     }
 
     protected override IEnumerable<IPropertyValidator> GetValidators (IPropertyInformation property, IValidationMessageFactory validationMessageFactory)
@@ -59,20 +51,32 @@ namespace Remotion.Validation.Attributes.Validation
       ArgumentUtility.CheckNotNull("property", property);
       ArgumentUtility.CheckNotNull("validationMessageFactory", validationMessageFactory);
 
-      LengthValidator validator;
+      IPropertyValidator validator;
       if (string.IsNullOrEmpty(ErrorMessage))
       {
         validator = PropertyValidatorFactory.Create(
             property,
-            parameters => new LengthValidator(MinLength, MaxLength, parameters.ValidationMessage),
+            parameters => CreateValidator(parameters.ValidationMessage),
             validationMessageFactory);
       }
       else
       {
-        validator = new LengthValidator(MinLength, MaxLength, new InvariantValidationMessage(ErrorMessage));
+        validator = CreateValidator(new InvariantValidationMessage(ErrorMessage));
       }
 
       return EnumerableUtility.Singleton(validator);
+    }
+
+    private IPropertyValidator CreateValidator (ValidationMessage message)
+    {
+      if (MinLength == 0 && MaxLength > 0)
+        return new MaximumLengthValidator(MaxLength, message);
+      else if (MaxLength == 0 & MinLength > 0)
+        return new MinimumLengthValidator(MinLength, message);
+      else if (MinLength == MaxLength)
+        return new ExactLengthValidator(MinLength, message);
+      else
+        return new LengthValidator(MinLength, MaxLength, message);
     }
   }
 }
--- a/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.de.resx
+++ b/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.de.resx
@@ -206,8 +206,8 @@
     <value>Geben Sie maximal 1 Zeichen ein.</value>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Geben Sie maximal {1} Zeichen ein.</value>
-    <comment>{1} = expected length</comment>
+    <value>Geben Sie maximal {0} Zeichen ein.</value>
+    <comment>{0} = expected length</comment>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMinimumLengthOfOneValidationMessage" xml:space="preserve">
     <value>Geben Sie mindestens 1 Zeichen ein.</value>
--- a/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.fr.resx
+++ b/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.fr.resx
@@ -200,8 +200,8 @@
     <value>Ne saisissez pas plus de 1 caractère.</value>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Ne saisissez pas plus de {1} caractères.</value>
-    <comment>{1} = expected length</comment>
+    <value>Ne saisissez pas plus de {0} caractères.</value>
+    <comment>{0} = expected length</comment>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMinimumLengthOfOneValidationMessage" xml:space="preserve">
     <value>Saisissez au moins 1 caractère.</value>
--- a/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.it.resx
+++ b/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.it.resx
@@ -200,8 +200,8 @@
     <value>Non immettere più di 1 carattere.</value>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Non immettere più di {1} caratteri.</value>
-    <comment>{1} = expected length</comment>
+    <value>Non immettere più di {0} caratteri.</value>
+    <comment>{0} = expected length</comment>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMinimumLengthOfOneValidationMessage" xml:space="preserve">
     <value>Immetti almeno 1 carattere.</value>
--- a/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.resx
+++ b/Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.resx
@@ -173,8 +173,8 @@
     <comment>{0} = maximum number of integer digits; {1} = maximum number of decimal places</comment>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Enter no more than {1} characters.</value>
-    <comment>{1} = expected length</comment>
+    <value>Enter no more than {0} characters.</value>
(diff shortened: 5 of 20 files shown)


Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
```diff
@@ -40,39 +40,56 @@ public class LengthValidationAttribute : AddingValidationAttributeBase
     /// <param name="maxLength">The maximum number of characters allowed.</param>
     public LengthValidationAttribute (int minLength, int maxLength)
```
Frazerus, line 41: In the ticket it says that the constructor has to be parameterless. Is this ctor meant? Or the ones from the ExactLength & MinLength Validators?

Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
```diff
@@ -40,39 +40,56 @@ public class LengthValidationAttribute : AddingValidationAttributeBase
...
-        validator = new LengthValidator(MinLength, MaxLength, new InvariantValidationMessage(ErrorMessage));
+        validator = CreateValidator(new InvariantValidationMessage(ErrorMessage));
       }
 
       return EnumerableUtility.Singleton(validator);
     }
+
+    private IPropertyValidator CreateValidator (ValidationMessage message)
+    {
+      if (MinLength != 0 && MaxLength > MinLength)
```
Frazerus, line 70: This if else construct might be better suited to a switch expression/statement. Am not sure if it's more readable though, probably needs investigation.

Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs
```diff
@@ -65,16 +80,75 @@ public void GetPropertyValidator ()
...
+      var propertyInformation = PropertyInfoAdapter.Create(typeof(Customer).GetProperty("LastName"));
+      var validationMessageStub = new Mock<ValidationMessage>();
+      _validationMessageFactoryStub
+          .Setup(_ => _.CreateValidationMessageForPropertyValidator(It.IsAny<ExactLengthValidator>(), propertyInformation))
+          .Returns(validationMessageStub.Object);
 
+      var result = _exactLengthValidationAttribute.GetPropertyValidators(propertyInformation, _validationMessageFactoryStub.Object).ToArray();
+
+      Assert.That(result.Length, Is.EqualTo(1));
+      Assert.That(result[0], Is.TypeOf(typeof(ExactLengthValidator)));
```
Frazerus, line 97: You might want to use ``result.Single()`` here, I don't know if it's preferable. As we already used this implementation in the previous tests however, it's probably going to stay this way.

Frazerus (approved):

Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
KuJo7: Ah, the ticket has to be updated. It is not possible to make this ctor parameterless, because the idea was at first to set the properties using a setter, but this doesn't work with attributes. That's why, we have to keep it like this.

MichaelKetting: There was a misunderstanding: we want the parameterless ctor and use zero instead of null as the default for the properties since nullable integers are not supported on attributes.

Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs
KuJo7: @MichaelKetting Do we want to change it?

MichaelKetting: actually, no, it's fine. 
Before we go to result.Single, we could use 
`Assert.That(result, Has.Exacxly(1).All.TypeOf<ExactLegnthValidator>());`

(not 100% sure about the API, but I think it's close to what is possible)

Or for less compact
```
Assert.That(result.Length, Has.EqualTo(1));
Assert.That(result, Has.All.TypeOf<ExactLegnthValidator>());

Comment (MichaelKetting) [reply Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs:97]: Also, the `result.Single` gives you much less helpful information incase there's more than one item in the list, so less recommended.

Comment (Frazerus) [reply Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs:97]: Is there a similar implementation of `Is` using `Has` since you used `Has`? If so, what are the differences?

Comment (MichaelKetting) [reply Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs:97]: `Is` is for single.
`Has` is for collection contraints, i.e. stuff that is applied to each member of the collection.

Comment (Frazerus) [reply Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs:97]: Ok, thank you!

File: Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs [hunk 4]
```diff
@@ -29,33 +29,48 @@ namespace Remotion.Validation.UnitTests.Attributes.Validation
...
     private Mock<IValidationMessageFactory> _validationMessageFactoryStub;
 
     [SetUp]
     public void SetUp ()
     {
-      _attribute = new LengthValidationAttribute(10, 20);
+      _lengthValidationAttribute = new LengthValidationAttribute() { MinLength = 10, MaxLength = 20 };
+      _exactLengthValidationAttribute = new LengthValidationAttribute() { MinLength = 10, MaxLength = 10 };
+      _minimumLengthValidationAttribute = new LengthValidationAttribute() { MinLength = 10 };
+      _maximumLengthValidationAttribute = new LengthValidationAttribute() { MaxLength = 10 };
```
MichaelKetting, line 44: Best practice: originally we had the same setup for each test. it is best practice to keep individual setup in the test itself. In particular, here we create different attributes, each is used just once. This should be in a local variable in the individual tests. An added advantage of this is that you ensure you are using the correct value in the test and do not accidentally access use the wrong field. (yes, this can happen. quite easily, too, with copy&paste :) )

Remotion/Validation/UnitTests/Attributes/Validation/LengthValidationAttributeTest.cs
```diff
@@ -29,33 +29,48 @@ namespace Remotion.Validation.UnitTests.Attributes.Validation
...
+      Assert.That(_lengthValidationAttribute.MaxLength, Is.EqualTo(20));
+
+      Assert.That(_exactLengthValidationAttribute.MinLength, Is.EqualTo(10));
+      Assert.That(_exactLengthValidationAttribute.MaxLength, Is.EqualTo(10));
+
+      Assert.That(_minimumLengthValidationAttribute.MinLength, Is.EqualTo(10));
+      Assert.That(_minimumLengthValidationAttribute.MaxLength, Is.EqualTo(0));
+
+      Assert.That(_maximumLengthValidationAttribute.MinLength, Is.EqualTo(0));
+      Assert.That(_maximumLengthValidationAttribute.MaxLength, Is.EqualTo(10));
```
MichaelKetting, line 61: Here we're testing if the properties work. since we're excercising the properties anyway and they have been implemented as autoproperties, there is no benefit. originally, we tested if the properties returned the values from the ctor and where assigned in the correct "order" (min to min, max to max). 

What you could do is write two tests: GetAndSetMinimumLength and GetAndSetMaximumLength. There, you first assert if the property is zero, then assign it a value and then assert that it has the new value.

MichaelKetting (changes requested):

Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
```diff
@@ -29,50 +29,69 @@ namespace Remotion.Validation.Attributes.Validation
...
```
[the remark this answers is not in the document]
KuJo7:
```suggestion
          private IPropertyValidator CreateValidator (ValidationMessage message)
    {
      if (MinLength == 0 && MaxLength > 0)
        return new MaximumLengthValidator(MaxLength, message);
      else if (MaxLength == 0 & MinLength > 0)
        return new MinimumLengthValidator(MinLength, message);
      else if (MinLength > 0 && MaxLength > 0 && MinLength == MaxLength)
        return new ExactLengthValidator(MinLength, message);
      else
        return new LengthValidator(MinLength, MaxLength, message);
    }
```
I would prefer here to use the if/else instead of the switch expression because this way it is more readable. Now, it is possible to get rid of the exception since the lengthvalidator is the default. If an incorrect value (<1) gets assigned, it should throw the exception in LengthValidator

What do you think?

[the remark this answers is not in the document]
MichaelKetting: That's what I was looking for / expecting :)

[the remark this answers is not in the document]
KuJo7: Ah yeah, thanks :)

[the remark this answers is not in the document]
Frazerus: Just because I wanted to try how this might look:

image

It might be a bit unfitting in re-motion though, so I also agree with taking the if-else blocks 👍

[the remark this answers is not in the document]
MichaelKetting: Thanks @Frazerus, for the demo!

My thoughts: since this check relies only on properties of the switched-on value, not the actual value itself (which would be `this` in our case), it feels idiomatically wrong to use a switch expression. The main reason for using switch is, that it's more compact but the resulting syntax, in particular the fact that we start with empty braces, shows to me that we're misusing a language feature (it looks awkward).

That said, there is also pattern matching which is implemented by switch:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#positional-pattern
https://www.thomasclaudiushuber.com/2021/02/25/c-9-0-pattern-matching-in-switch-expressions/
```
return (MinLength, MaxLength) switch
{
    (0, > 0) => new MaxiumLengthValidator(MaxLength),
    (> 0, 0) => new MinimumLengthValidator(MinLength),
    (_, MaxLength) => new ExactLengthValidator(MaxLength), // not 100% sure on the syntax
    _ => new LengthValidator(MinLength, MaxLength)
}
```
It's a very fine difference in the use of syntax elements compared to the simple application of using empty braces and `when`. However, idiomatic code often is :)

Another possible way to write this could be
```
return this switch
{
    { MinLength: 0, MaxLength: > 0 } => new MaxiumLengthValidator(MaxLength),
    { MinLength: > 0, MaxLength: 0 } => new MinimumLengthValidator(MinLength),
    { MinLength: MaxLength } => new ExactLengthValidator(MaxLength), // not 100% sure on the syntax
    _ => new LengthValidator(MinLength, MaxLength)
}
```
Not sure which is best.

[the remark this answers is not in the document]
Frazerus: Yeah, I see how this might be problematic, I think we should however talk about switch expressions more in-depth at some point (dev-jour-fixe?). 

I think they can be really useful and I haven't seen much code written with them, making it harder to know what you can and cannot do.

Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.de.resx
```diff
@@ -206,7 +206,7 @@
     <value>Geben Sie maximal 1 Zeichen ein.</value>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Geben Sie maximal {1} Zeichen ein.</value>
+    <value>Geben Sie maximal {0} Zeichen ein.</value>
     <comment>{1} = expected length</comment>
```
MichaelKetting, line 210:
```suggestion
    <comment>{0} = expected length</comment>
```

Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.fr.resx
```diff
@@ -200,7 +200,7 @@
     <value>Ne saisissez pas plus de 1 caractère.</value>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Ne saisissez pas plus de {1} caractères.</value>
+    <value>Ne saisissez pas plus de {0} caractères.</value>
     <comment>{1} = expected length</comment>
```
MichaelKetting, line 204:
```suggestion
    <comment>{0} = expected length</comment>
```

Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.it.resx
```diff
@@ -200,7 +200,7 @@
     <value>Non immettere più di 1 carattere.</value>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Non immettere più di {1} caratteri.</value>
+    <value>Non immettere più di {0} caratteri.</value>
     <comment>{1} = expected length</comment>
```
MichaelKetting, line 204:
```suggestion
    <comment>{0} = expected length</comment>
```

Remotion/Validation/Globalization/Globalization/LocalizedValidationMessageFactory.resx
```diff
@@ -173,7 +173,7 @@
     <comment>{0} = maximum number of integer digits; {1} = maximum number of decimal places</comment>
   </data>
   <data name="Remotion.Validation.Globalization.LocalizedValidationMessageFactory.ValueMustHaveMaximumLengthValidationMessage" xml:space="preserve">
-    <value>Enter no more than {1} characters.</value>
+    <value>Enter no more than {0} characters.</value>
     <comment>{1} = expected length</comment>
```
MichaelKetting, line 177:
```suggestion
    <comment>{0} = expected length</comment>
```

MichaelKetting (approved):

PR #751, opened 2023-07-06
RM-8821 Render validation failure icon in cells

a-ctor:
https://re-motion.atlassian.net/browse/RM-8821

### DIFF
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocListAsGridControlObjectTest.cs
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocListAsGridControlObjectTest.cs
@@ -640,7 +640,7 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
     }
 
     [Test]
-    public void GetValidationErrors_CellValidationFailure ()
+    public void GetValidationErrors_CellValidationFailureOnRow ()
     {
       var home = Start();
       var bocList = home.ListAsGrids().GetByLocalID("JobList_Validation");
@@ -658,6 +658,28 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
                                        };
 
       var bocListValidationFailures = bocList.GetRow(1).GetValidationErrors();
+      Assert.That(bocListValidationFailures, Is.EqualTo(expectedValidationFailures));
+    }
+
+    [Test]
+    public void GetValidationErrors_CellValidationFailure ()
+    {
+      var home = Start();
+      var bocList = home.ListAsGrids().GetByLocalID("JobList_Validation");
+
+      home.WebButtons().GetByLocalID("ValidationTestCaseCellButton").Click();
+
+      var expectedValidationFailures = new []
+                                       {
+                                           new BocListValidationError(
+                                               "Localized cell validation failure message",
+                                               "89dc8cd2-30e0-4bb3-92a0-4587f32492f5",
+                                               "DisplayName",
+                                               "89dc8cd2-30e0-4bb3-92a0-4587f32492f5",
+                                               "DisplayName")
+                                       };
+
+      var bocListValidationFailures = bocList.GetRow(1).GetCell().WithColumnTitle("DisplayName").GetValidationErrors();
       Assert.That(bocListValidationFailures, Is.EqualTo(expectedValidationFailures));
     }
 
@@ -696,7 +718,7 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
     }
 
     [Test]
-    public void GetValidationErrors_EditableRowAndCellValidationFailure ()
+    public void GetValidationErrors_EditableRowAndCellValidationFailureOnRow ()
     {
       var home = Start();
       var bocList = home.ListAsGrids().GetByLocalID("JobList_Validation");
@@ -718,6 +740,28 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests
       Assert.That(row.GetValidationErrors(), Is.EqualTo(expectedValidationFailures));
     }
 
+    [Test]
+    public void GetValidationErrors_EditableRowAndCellValidationFailure ()
+    {
+      var home = Start();
+      var bocList = home.ListAsGrids().GetByLocalID("JobList_Validation");
+
+      var row = bocList.GetRow(1);
+
+      home.WebButtons().GetByLocalID("ValidationTestCaseCellButton").Click();
+
+      var expectedValidationFailures = new []
+                                       {
+                                           new BocListValidationError(
+                                               "Localized cell validation failure message",
+                                               "89dc8cd2-30e0-4bb3-92a0-4587f32492f5",
+                                               "DisplayName",
+                                               "89dc8cd2-30e0-4bb3-92a0-4587f32492f5",
+                                               "DisplayName")
+                                       };
+
+      Assert.That(row.GetCell().WithColumnTitle("DisplayName").GetValidationErrors(), Is.EqualTo(expectedValidationFailures));
+    }
 
     private WxePageObject Start ()
     {
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRenderer.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRenderer.cs
@@ -129,10 +129,14 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering
         int rowIndex,
         string? cellID,
         IReadOnlyCollection<string> headerIDs,
+        IReadOnlyList<bool> columnsWithValidationFailures,
         BocListDataRowRenderEventArgs dataRowRenderEventArgs)
     {
       ArgumentUtility.CheckNotNull("renderingContext", renderingContext);
       ArgumentUtility.CheckNotNull("dataRowRenderEventArgs", dataRowRenderEventArgs);
+      ArgumentUtility.CheckNotNull("columnsWithValidationFailures", columnsWithValidationFailures);
+      if (renderingContext.ColumnRenderers.Length != columnsWithValidationFailures.Count)
+        throw new ArgumentException("The number of validation failures flags must match the number of column renderers.", nameof(columnsWithValidationFailures));
 
       var columnRenderingContext = CreateBocColumnRenderingContext(renderingContext);
 
@@ -143,7 +147,8 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering
               rowIndex: rowIndex,
               showIcon: _showIcon,
               cellID: cellID,
-              headerIDs: headerIDs));
+              headerIDs: headerIDs,
+              columnsWithValidationFailures: columnsWithValidationFailures));
     }
 
     private BocColumnRenderingContext CreateBocColumnRenderingContext (BocListRenderingContext renderingContext)
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
@@ -16,6 +16,7 @@
 // 
 using System;
 using System.Linq;
+using System.Text;
 using System.Web.UI;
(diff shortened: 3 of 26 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -45,6 +47,12 @@ public abstract class BocColumnRendererBase<TBocColumnDefinition> : IBocColumnRe
     /// <summary>Filename of the image used to indicate an descending sort order of the column in its title cell.</summary>
     protected const string c_sortDescendingIcon = "sprite.svg#SortDescending";
 
+    /// <summary>Filename of the image used to indicate that there are validation failures in a cell.</summary>
+    protected string c_validationIndicatorIcon = "sprite.svg#ValidationError";
+
+    /// <summary>Filename of the image used to indicate that there are validation failures in a cell.</summary>
```
MichaelKetting, line 53: doc

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -45,6 +47,12 @@ public abstract class BocColumnRendererBase<TBocColumnDefinition> : IBocColumnRe
...
+    protected string c_validationIndicatorPlaceholderIcon = "sprite.svg#ValidationErrorPlaceholder";
```
MichaelKetting, line 54: We have a simpler option. if we actually do need to render an image instead of just a span with a size, we always use the "Spacer.gif" for this. See IconInfo.CreateSpacer

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -355,8 +363,80 @@ private void RenderDataCell (BocColumnRenderingContext<TBocColumnDefinition> ren
...
+      if (validationFailures.Count > 0)
+      {
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, _cssClasses.CssClassScreenReaderText);
+        renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Span);
+
+        renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Span);
+        renderingContext.Writer.WriteEncodedText(validationInfoText);
+        renderingContext.Writer.RenderEndTag();
+
+        renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Ul);
```
MichaelKetting, line 414: We need to use DIV here because UL is a blocklevel element which may only be nested in other blocklevel elements.

MichaelKetting, line 414: We need to verify this: it is possible the best way to handle this is by using aria-label instead of printing a separate span:
https://www.scottohara.me/blog/2020/05/02/labelled-lists.html

gut feeling: we should use aria-label. This result in the same or better UX for screenreader users.

alternative (more effort and no payoff in our scenario): annodate the span as aria-hidden, add an ID and link it to the UL via aria-labeledby

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -355,8 +363,80 @@ private void RenderDataCell (BocColumnRenderingContext<TBocColumnDefinition> ren
...
+            var rowObjectWithIdentity = validationFailure.RowObject as IBusinessObjectWithIdentity;
+
+            renderingContext.Writer.AddAttribute(DiagnosticMetadataAttributesForObjectBinding.BocListValidationFailureSourceRow, rowObjectWithIdentity?.UniqueIdentifier ?? string.Empty);
+            renderingContext.Writer.AddAttribute(DiagnosticMetadataAttributesForObjectBinding.BocListValidationFailureSourceColumn, validationFailure.ColumnDefinition?.ItemID ?? string.Empty);
+            renderingContext.Writer.AddAttribute(DiagnosticMetadataAttributesForObjectBinding.ValidationFailureSourceBusinessObject, validatedObjectWithIdentity?.UniqueIdentifier ?? string.Empty);
+            renderingContext.Writer.AddAttribute(DiagnosticMetadataAttributesForObjectBinding.ValidationFailureSourceProperty, validationFailure.Failure.ValidatedProperty?.Identifier ?? string.Empty);
+          }
+
+          renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Li);
+          renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Span);
```
MichaelKetting, line 429: do we need the SPAN inside the LI?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -355,8 +363,80 @@ private void RenderDataCell (BocColumnRenderingContext<TBocColumnDefinition> ren
...
+          renderingContext.Writer.WriteEncodedText(validationFailure.Failure.ErrorMessage);
+          renderingContext.Writer.RenderEndTag();
+          renderingContext.Writer.RenderEndTag();
+        }
+
+        renderingContext.Writer.RenderEndTag();
+
+        renderingContext.Writer.RenderEndTag();
```
MichaelKetting, line 437: I usually add comments on which element gets closed when there's more than one item in the method. makes for easier troubleshooting / review / maintenance.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -355,8 +363,80 @@ private void RenderDataCell (BocColumnRenderingContext<TBocColumnDefinition> ren
...
+          renderingContext.ColumnDefinition,
+          false);
+
+      if (validationFailures.Count > 0)
+      {
+        var tooltipStringBuilder = new StringBuilder();
+        tooltipStringBuilder.AppendLine(validationInfoText);
+        foreach (var validationFailure in validationFailures)
+        {
+          tooltipStringBuilder.Append(" • ");
```
MichaelKetting, line 378:
```suggestion
          tooltipStringBuilder.Append("• ");
```

gut feeling: we don't need the left space.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -355,8 +363,80 @@ private void RenderDataCell (BocColumnRenderingContext<TBocColumnDefinition> ren
...
+          tooltipStringBuilder.AppendLine(validationFailure.Failure.ErrorMessage);
+        }
+
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Id, BocRowRenderer.GetCellIDForValidationMarker(renderingContext.Control, arguments.RowIndex, renderingContext.VisibleColumnIndex));
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Title, tooltipStringBuilder.ToString());
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClasses.ValidationErrorMarker);
+        renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Span);
```
MichaelKetting, line 385: we need to annotate with aria-hidden (there's readymade const for this on HtmlAttriute2 and HtmlAriaHiddenAttributeValue

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRenderer.cs
```diff
@@ -129,10 +129,14 @@ public void RenderDataColumnDeclaration (BocListRenderingContext renderingContex
         int rowIndex,
         string? cellID,
         IReadOnlyCollection<string> headerIDs,
+        bool[] columnsWithValidationFailures,
```
MichaelKetting, line 132: We used IReadOnlyCollection instead of array because of the immutatility. I think using IReadOnlyList<bool> would be more consistent?

(everything in me screams to use BitArray, but that's not really "better" of course)

MichaelKetting (commented):
partial reivew, the annotated files are "done"

MichaelKetting (commented):
partial review: testfixtures are still left to review

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocListAsGridControlObjectTest.cs
```diff
@@ -661,6 +661,28 @@ public void GetValidationErrors_CellValidationFailure ()
...
+                                           new BocListValidationError(
+                                               "Localized cell validation failure message",
+                                               "89dc8cd2-30e0-4bb3-92a0-4587f32492f5",
+                                               "DisplayName",
+                                               "89dc8cd2-30e0-4bb3-92a0-4587f32492f5",
+                                               "DisplayName")
+                                       };
+
+      var bocListValidationFailures = bocList.GetRow(1).GetCell().WithColumnTitle("DisplayName").GetValidationErrors();
+      Assert.That(bocListValidationFailures, Is.EqualTo(expectedValidationFailures));
```
MichaelKetting, line 683: I do prefer this style to inling the variable into the Assert call due to the length / compexity of the actual call. Do you want to update the other usages in this changeset (BocListAsGridControlObjectTest.cs BocListControlObjectTest.cs

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
a-ctor: The only reason for the span here is that we have this structure of UL already for the row/validation column and it allows us to reuse the error parsing logic without adding a special case.

I am unsure what exactly you are getting at with the aria-labels. We would have the same problem in the previously rendered lists for the row/validation column wouldn't we?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
a-ctor: As mentioned in the previous comment. This is to have the same structure as in the validation row/column, allowing the error parsing logic to be reused

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
MichaelKetting: As discussed: `aria-label`

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererBase.cs
```diff
@@ -355,9 +360,81 @@ private void RenderDataCell (BocColumnRenderingContext<TBocColumnDefinition> ren
...
+        foreach (var validationFailure in validationFailures)
+        {
+          tooltipStringBuilder.Append("• ");
+          tooltipStringBuilder.AppendLine(validationFailure.Failure.ErrorMessage);
+        }
+
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Id, BocRowRenderer.GetCellIDForValidationMarker(renderingContext.Control, arguments.RowIndex, renderingContext.VisibleColumnIndex));
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Title, tooltipStringBuilder.ToString());
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClasses.ValidationErrorMarker);
+        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, "true");
```
MichaelKetting, line 382:
```suggestion
        renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
```

MichaelKetting (approved):

PR #754, opened 2023-07-10
RM-8820 Integrate BocList edit mode into the new BocList validation

a-ctor:
https://re-motion.atlassian.net/browse/RM-8820

### DIFF
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListAsGridUserControl.ascx.cs
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListAsGridUserControl.ascx.cs
@@ -32,6 +32,8 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite.Shared.Cont
 {
   public partial class BocListAsGridUserControl : DataEditUserControl
   {
+    private IBusinessObjectValidationResult _validationResult = BusinessObjectValidationResult.Create(new ValidationResult());
+
     protected string SampleIconUrl;
 
     public override IBusinessObjectDataSourceControl DataSource
@@ -72,15 +74,24 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite.Shared.Cont
     {
       base.OnLoad(e);
 
-      JobList_Normal.SwitchListIntoEditMode();
-      JobList_Empty.SwitchListIntoEditMode();
-      JobList_Validation.SwitchListIntoEditMode();
+      if (!IsPostBack)
+      {
+        JobList_Normal.SwitchListIntoEditMode();
+        JobList_Empty.SwitchListIntoEditMode();
+        JobList_Validation.SwitchListIntoEditMode();
+      }
     }
 
     protected override void OnPreRender (EventArgs e)
     {
       base.OnPreRender(e);
       SetTestOutput();
+
+      // We do the validation on every request since it is easier that covering it in multiple event handlers.
+      // While we use domain validation, it should only affect the last control that is intended for testing validation.
+      ((SmartPage)Page)!.PrepareValidation();
+      CurrentObjectValidationResultDispatchingValidator.DispatchValidationFailures(_validationResult);
+      CurrentObjectValidationResultDispatchingValidator.Validate();
     }
 
     private void MenuItemClickHandler (object sender, WebMenuItemClickEventArgs e)
@@ -163,11 +174,7 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite.Shared.Cont
 
     private void AddValidationFailures (params ValidationFailure[] validationFailures)
     {
-      var validationResult = BusinessObjectValidationResult.Create(new ValidationResult(validationFailures));
-
-      ((SmartPage)Page)!.PrepareValidation();
-      CurrentObjectValidationResultDispatchingValidator.DispatchValidationFailures(validationResult);
-      CurrentObjectValidationResultDispatchingValidator.Validate();
+      _validationResult = BusinessObjectValidationResult.Create(new ValidationResult(validationFailures));
     }
   }
 }
--- a/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListUserControl.ascx.cs
+++ b/Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListUserControl.ascx.cs
@@ -32,6 +32,8 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite.Shared.Cont
 {
   public partial class BocListUserControl : DataEditUserControl
   {
+    private IBusinessObjectValidationResult _validationResult = BusinessObjectValidationResult.Create(new ValidationResult());
+
     protected string SampleIconUrl;
 
     public override IBusinessObjectDataSourceControl DataSource
@@ -95,6 +97,12 @@ namespace Remotion.ObjectBinding.Web.Development.WebTesting.TestSite.Shared.Cont
     {
       base.OnPreRender(e);
       SetTestOutput();
+
+      // We do the validation on every request since it is easier that covering it in multiple event handlers.
+      // While we use domain validation, it should only affect the last control that is intended for testing validation.
+      ((SmartPage)Page)!.PrepareValidation();
+      CurrentObjectValidationResultDispatchingValidator.DispatchValidationFailures(_validationResult);
+      CurrentObjectValidationResultDispatchingValidator.Validate();
     }
 
     private void MenuItemClickHandler (object sender, WebMenuItemClickEventArgs e)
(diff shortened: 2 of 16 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -3436,6 +3439,15 @@ protected bool IsSelectionEnabled
       get { return _selection != RowSelection.Undefined && _selection != RowSelection.Disabled; }
     }
 
+    public bool IsInlineValidationDisplayEnabled
```
MichaelKetting, line 3442: Am thinking: should we make this explicit iface? it's a gut feeling thing.

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
```diff
@@ -50,7 +50,7 @@ public class BocListUserControl : BaseUserControl
   protected Button ChildrenListSetPageButton;
   protected Button ChildrenListAddRowButton;
   protected Button ChildrenListRemoveRowsButton;
-  protected Button ChildrenAddValidationErrorsButton;
+  protected CheckBox EnableValidationErrorsCheckbox;
```
MichaelKetting, line 53: this change is an x-over with one of @Frazerus changesets? Changes look okay, btw.

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListUserControl.ascx.cs
```diff
@@ -95,6 +97,10 @@ protected override void OnPreRender (EventArgs e)
     {
       base.OnPreRender(e);
       SetTestOutput();
+
+      ((SmartPage)Page)!.PrepareValidation();
+      CurrentObjectValidationResultDispatchingValidator.DispatchValidationFailures(_validationResult);
+      CurrentObjectValidationResultDispatchingValidator.Validate();
```
MichaelKetting, line 103: Shouldn't this be the same conditional logic as in IndividualControlTests or similiarily only get triggered when performing validation?

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListAsGridUserControl.ascx.cs
```diff
@@ -72,15 +74,22 @@ protected override void OnLoad (EventArgs e)
...
     }
 
     protected override void OnPreRender (EventArgs e)
     {
       base.OnPreRender(e);
       SetTestOutput();
+
+      ((SmartPage)Page)!.PrepareValidation();
+      CurrentObjectValidationResultDispatchingValidator.DispatchValidationFailures(_validationResult);
+      CurrentObjectValidationResultDispatchingValidator.Validate();
```
MichaelKetting, line 92: See BocListUserControl.ascx

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -95,6 +95,24 @@ private void DispatchToValidationFailureRepository (IBusinessObjectValidationRes
           if (otherRowFailures.Count > 0)
             validationFailureRepository.AddValidationFailuresForDataRow(rowObject, otherRowFailures);
         }
+
+        var editModeController = ((IBocList)bocListControl).EditModeController;
+        if (editModeController.IsRowEditModeActive)
+        {
+          var bocListRow = editModeController.GetEditedRow();
+          var editableRow = editModeController.GetEditableRow(bocListRow.Index);
```
MichaelKetting, line 103:
```suggestion
          var editableRow = editModeController.GetEditedRow();
```
Unless I'm missing something in the API?

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -144,6 +162,32 @@ protected override bool ControlPropertiesValid ()
         return NamingContainer.FindControl(controlToValidate) != null;
     }
 
+    private void DispatchEditableRowValidationFailures (
+        IEditableRow editableRow,
+        IBusinessObject rowObject,
+        IReadOnlyList<BocColumnDefinition> columnDefinitions,
+        IBocListValidationFailureRepository validationFailureRepository)
+    {
+      var castedEditableRow = (EditableRow)editableRow;
```
MichaelKetting, line 171: We can move the relevant APIs to the interface. Perhaps not just GetValidators but also the related HasValidators etc methods?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/EditableRowSupport/EditableRow.cs
```diff
@@ -490,30 +490,33 @@ public IBusinessObjectBoundEditableWebControl[] GetEditControlsAsArray ()
...
-          else
-            writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassEditModeValidationMessage);
           writer.RenderBeginTag(HtmlTextWriterTag.Div);
-          PlainTextString.CreateFromText(validator.ErrorMessage).WriteTo(writer);
+          validator.RenderControl(writer);
           writer.RenderEndTag();
+
+          if (   ! validator.IsValid
+                 && validator.Display == ValidatorDisplay.None
+                 && ! _editModeHost.DisableEditModeValidationMessages)
```
MichaelKetting, line 506: Whitespace/Indent.

MichaelKetting (approved):
I think there's just some small stuff that doesn't need a re-review?

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
a-ctor: `GetEditedRow` does not return an `IEditableRow` for some reason so this hop is required

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
a-ctor: `EditableRow` has a lot of members that are not part of the interface. Right here we only need the `GetValidators` method. If you want to include other members as well we should talk about what members you want to move.

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListUserControl.ascx.cs
a-ctor: The main difference here is that we have our own BocList control that has the validation column added. The checkbox stuff in the IndividualControlTests was mainly because this is only for manual clicking and introducing a new BocList seems unnecessary. In the case of the web tests it is kind of strange to enable the functionality via checkbox even if we have our own control for the validation scenario. The other BocList controls should ignore the validation anyways.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
MichaelKetting: Ah, yes. no, that's actually correct and sensible after looking into the API. The naming might not be perfect and stems from the name on BocList itself. so, no change. GetEditedRow is just returning metadata.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
MichaelKetting: For symmerty with GetEditControls:
* HasValidators()
* HasValidator(int)
* GetValidators(int)

that's it for now.

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocListUserControl.ascx.cs
MichaelKetting: Hmm...so we're using the CurrentObjectValidationResultDispatchingValidator which applies to all BO-Controls on the page, knowing that it's only relevant to one of them so we don't care that we do it in each roundtrip? Wouldn't it then be best to call this only when Validate is clicked in the webtests?

Remotion/ObjectBinding/Web.Test/IndividualControlTests/BocListUserControl.ascx.cs
MichaelKetting: >  if there was a reason for storing the boolean in the session instead of using the checkbox directly. 
I have also mentioned this in my review with the recommendation to use just the checkbox, not the session.

PR #757, opened 2023-07-10
RM-8866 Add icon for validator column

a-ctor:
https://re-motion.atlassian.net/browse/RM-8866

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
@@ -361,6 +361,13 @@ th.bocListTitleCell > a[href]:active
   background-color: #C5E0FC;
 }
 
+th.bocListTitleCell > a > img,
+th.bocListTitleCell > span > img
+{
+  width: 16px;
+  height: 16px;
+}
+
 th.bocListTitleCell .bocListSortingOrder img
 {
   height: 16px;
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
@@ -125,6 +125,20 @@ div.bocList table.bocListTable td.bocListTitleCell > span
   display: inline-flex;
   align-items: center;
   padding: calc(var(--remotion-themed-spacing) / 2) var(--remotion-themed-spacing);
+  gap: var(--remotion-themed-spacing-small);
+}
+
+div.bocList table.bocListTable th.bocListTitleCell > a > img,
+div.bocList table.bocListTable th.bocListTitleCell > span > img
+{
+  width: var(--remotion-themed-img-width);
+  height: var(--remotion-themed-img-width);
+}
+
+div.bocList table.bocListTable th.bocListTitleCell > span
+{
+  /* Prevent layout changes due to a missing border for anchor when switching between sortable/non-sortable. */
+  border: var(--remotion-themed-border-placeholder);
 }
 
 div.bocList table.bocListTable td.bocListDataCell > a:not([hidden]),
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocColumnDefinition.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocColumnDefinition.cs
@@ -16,6 +16,7 @@
 // 
 using System;
 using System.ComponentModel;
+using System.Drawing;
 using System.Web.UI;
 using System.Web.UI.WebControls;
 using Remotion.Globalization;
@@ -24,6 +25,7 @@ using Remotion.ServiceLocation;
 using Remotion.Utilities;
 using Remotion.Web;
 using Remotion.Web.Globalization;
+using Remotion.Web.UI.Controls;
 using Remotion.Web.UI.Globalization;
 
 namespace Remotion.ObjectBinding.Web.UI.Controls
@@ -32,8 +34,9 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
   public abstract class BocColumnDefinition : BusinessObjectControlItem
   {
     private string _itemID = string.Empty;
-    private bool _showColumnTitle = true;
     private WebString _columnTitle = WebString.Empty;
+    private IconInfo _columnTitleIcon = new IconInfo();
+    private BocColumnTitleStyle _columnTitleStyle = BocColumnTitleStyle.Text;
     private Unit _width = Unit.Empty;
     private string _cssClass = string.Empty;
 
@@ -88,17 +91,14 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
       get { return ColumnTitle; }
     }
 
-    /// <summary> Gets or sets a flag to show/hide the columns's title. </summary>
-    /// <value> <see langword="false" /> to hide the columns's title. </value>
-    [PersistenceMode(PersistenceMode.Attribute)]
-    [Category("Appearance")]
-    [Description("Set false to hide the columns's title")]
-    [DefaultValue(true)]
-    [NotifyParentProperty(true)]
-    public bool ShowColumnTitle
+    /// <summary> Gets the displayed image of the column title. </summary>
+    /// <remarks> Override this property to change the way the column title icon is generated. </remarks>
+    /// <value> A <see cref="IconInfo"/> representing this column's title icon. </value>
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+    [Browsable(false)]
+    public virtual IconInfo ColumnTitleIconDisplayValue
     {
-      get { return _showColumnTitle; }
-      set { _showColumnTitle = value; }
+      get { return ColumnTitleIcon; }
     }
 
     /// <summary> Gets or sets the text displayed in the column title. </summary>
@@ -115,6 +115,39 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     {
       get { return _columnTitle; }
       set { _columnTitle = value; }
+    }
+
+    /// <summary>
+    ///   Gets or sets the image displayed in the title cell of the column. Must not be <see langword="null"/>.
+    ///   Only displayed if <see cref="BocColumnTitleStyle"/>.<see cref="BocColumnTitleStyle.Icon"/> is set in <see cref="ColumnTitleStyle"/>.
+    /// </summary>
+    [PersistenceMode(PersistenceMode.Attribute)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Content)]
+    [Category("Appearance")]
+    [Description("The image displayed in the column title in the rendered page.")]
+    [NotifyParentProperty(true)]
+    public IconInfo ColumnTitleIcon
+    {
+      get { return _columnTitleIcon; }
+      set
+      {
+        ArgumentUtility.CheckNotNull("Icon", value);
+        _columnTitleIcon = value;
+      }
+    }
+
+    /// <summary>
+    /// Gets or sets the <see cref="BocColumnTitleStyle"/> that is used to determine how the column title cell is displayed.
+    /// </summary>
+    [PersistenceMode(PersistenceMode.Attribute)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Content)]
+    [Category("Appearance")]
+    [Description("The style determining how the column title is displayed in the rendered page.")]
+    [NotifyParentProperty(true)]
+    public BocColumnTitleStyle ColumnTitleStyle
+    {
+      get { return _columnTitleStyle; }
+      set { _columnTitleStyle = value; }
     }
 
     /// <summary> Gets or sets the width of the column definition. </summary>
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
@@ -18,10 +18,15 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Web.UI;
+using System.Web.UI.WebControls;
 using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
 using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
+using Remotion.Web;
+using Remotion.Web.Globalization;
+using Remotion.Web.Infrastructure;
+using Remotion.Web.UI.Controls;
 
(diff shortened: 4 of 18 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocValidationErrorIndicatorColumnRenderer.cs
```diff
@@ -65,6 +65,33 @@ protected override string GetAdditionalCssClassForDataCell (BocCellAttributeRend
...
+      if (columnTitle.IsEmpty)
+        columnTitle = WebString.CreateFromText("Validation errors");
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClasses.ValidationErrorMarker);
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Title, columnTitle.ToString(WebStringEncoding.Html));
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Span);
+
+      var iconUrl = renderingContext.Control.ValidationFailureRepository.HasValidationFailuresForDataRows()
+          ? c_validationIndicatorIcon
+          : c_validationIndicatorDisabledIcon;
```
MichaelKetting, line 80: I looked into the represenative samples:
We have used
```
img {
    opacity: 0.5;
    filter: grayscale(100%);
}
```
for disabled icons in NovaViso.

Alternative might be the gray-shade from the null-Checkbox value `#979797` in NovaViso

Disclaimer: need to see it live for final word.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocValidationErrorIndicatorColumnRenderer.cs
```diff
@@ -65,6 +65,33 @@ protected override string GetAdditionalCssClassForDataCell (BocCellAttributeRend
       return CssClasses.DataCellValidationFailureIndicator;
     }
 
+    protected override void RenderTitleCellContents (BocColumnRenderingContext<BocValidationErrorIndicatorColumnDefinition> renderingContext)
```
MichaelKetting, line 68: I want to try and avoid this extension point if we can do this via CSS. There's an idea/comment about replacing the `nbsp;` with a `span` and dummy width. This can then serve as the icon-target. 

Reason for avoiding the extension point is that making sure we cover "all" that is needed would be more effort than we should invest right now and doing it partially invites non-optimal usages downstream and added costs when we rework the extension point.

Having an icon for the title could perhaps work, doing it the same way as we do it with the title. (similiar properties, etc). 

TBD: do we perhaps want to a tooltip for the icon anyway? should the tooltip simply be the otherwise invisible column title?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocValidationErrorIndicatorColumnRenderer.cs
```diff
@@ -65,6 +65,33 @@ protected override string GetAdditionalCssClassForDataCell (BocCellAttributeRend
...
+    {
+      var columnTitle = renderingContext.ColumnDefinition.ColumnTitle;
+      if (columnTitle.IsEmpty)
+        columnTitle = WebString.CreateFromText("Validation errors");
+
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClasses.ValidationErrorMarker);
+      renderingContext.Writer.AddAttribute(HtmlTextWriterAttribute.Title, columnTitle.ToString(WebStringEncoding.Html));
+      renderingContext.Writer.RenderBeginTag(HtmlTextWriterTag.Span);
```
MichaelKetting, line 76: Am not sure about the semantics. Will become moot point anyway.

MichaelKetting (changes requested):
I think we could most easily do this in pair.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -127,6 +127,25 @@ div.bocList table.bocListTable td.bocListTitleCell > span
   padding: calc(var(--remotion-themed-spacing) / 2) var(--remotion-themed-spacing);
 }
 
+div.bocList table.bocListTable th.bocListTitleCell > a > img,
+div.bocList table.bocListTable th.bocListTitleCell > span > img
+{
+  width: 16px;
+  height: 16px;
+  margin-right: var(--remotion-themed-spacing); /* Provides a spacing between icon and text, which is reset below if only an icon is displayed */
```
MichaelKetting, line 135:
```suggestion
  margin-right: var(--remotion-themed-spacing-small); /* Provides a spacing between icon and text, which is reset below if only an icon is displayed */
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -127,6 +127,25 @@ div.bocList table.bocListTable td.bocListTitleCell > span
...
+}
+
+div.bocList table.bocListTable th.bocListTitleCell > a > img:has(+ span.screenReaderText),
+div.bocList table.bocListTable th.bocListTitleCell > span > img:has(+ span.screenReaderText)
+{
+  margin-right: 0;
+}
```
MichaelKetting, line 142: Simple CSS:
if we do the following, we should be able to get the same effect with more straight-forward markup:
```
div.bocList table.bocListTable th.bocListTitleCell > a > img+span:not(.screenReaderText),
div.bocList table.bocListTable th.bocListTitleCell > span > img+span:not(.screenReaderText)
{
  padding-left: var(--remotion-themed-spacing-small); 
}
```

Actually, we don't even need to test for `:not(.screenReaderText)` on the span since the padding would be invisible anyway.

I used padding instead of margin. It should work. If not, we can stick with margin. (I use padding everywhere else in these cases)

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -127,6 +127,25 @@ div.bocList table.bocListTable td.bocListTitleCell > span
...
+
+div.bocList table.bocListTable th.bocListTitleCell > span > img
+{
+  margin: 1px; /* Prevent layout changes due to a missing 1px border when switching between sortable/non-sortable. */
```
MichaelKetting, line 146: not sure I understand this. Yes, the span doesn't have a border-placeholder applied. Is this a gneeral layout bug and we need to apply the missing placeholder-border to the span?

either way, please update to use `--remotion-themed-border-width` instead of magic number.

For cases such as this, I actually have a border placeholder defined:
`border: var(--remotion-themed-border-placeholder)`
This is applied when no border is visible. Unless there's a good reason, we should go with this option.

So, to sumarize:
```
div.bocList table.bocListTable th.bocListTitleCell > span,
div.bocList table.bocListTable td.bocListTitleCell > span
{
  border:  var(--remotion-themed-border-placeholder);
}
```
(we need to do both `td` and `th`.)

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -127,6 +127,25 @@ div.bocList table.bocListTable td.bocListTitleCell > span
...
+div.bocList table.bocListTable th.bocListTitleCell > a > img,
+div.bocList table.bocListTable th.bocListTitleCell > span > img
+{
+  width: 16px;
+  height: 16px;
+  margin-right: var(--remotion-themed-spacing); /* Provides a spacing between icon and text, which is reset below if only an icon is displayed */
+}
+
+div.bocList table.bocListTable th.bocListTitleCell > a > img:has(+ span.screenReaderText),
+div.bocList table.bocListTable th.bocListTitleCell > span > img:has(+ span.screenReaderText)
```
MichaelKetting, line 139:
```suggestion
div.bocList table.bocListTable th.bocListTitleCell > a > img:has(+ span.screenReaderText),
div.bocList table.bocListTable td.bocListTitleCell > a > img:has(+ span.screenReaderText),
div.bocList table.bocListTable th.bocListTitleCell > span > img:has(+ span.screenReaderText),
div.bocList table.bocListTable td.bocListTitleCell > span > img:has(+ span.screenReaderText)
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -127,6 +127,25 @@ div.bocList table.bocListTable td.bocListTitleCell > span
   padding: calc(var(--remotion-themed-spacing) / 2) var(--remotion-themed-spacing);
 }
 
+div.bocList table.bocListTable th.bocListTitleCell > a > img,
+div.bocList table.bocListTable th.bocListTitleCell > span > img
```
MichaelKetting, line 131:
```suggestion
div.bocList table.bocListTable th.bocListTitleCell > a > img,
div.bocList table.bocListTable td.bocListTitleCell > a > img,
div.bocList table.bocListTable th.bocListTitleCell > span > img,
div.bocList table.bocListTable td.bocListTitleCell > span > img
```

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -361,6 +361,13 @@ th.bocListTitleCell > a[href]:active
   background-color: #C5E0FC;
 }
 
+th.bocListTitleCell > a > img,
+th.bocListTitleCell > span > img
```
MichaelKetting, line 365: Oh no, I found a css bug in NovaGray: we need to duplicate the title-rules for TD elements. please create a ticket, thanks! This was changed when we introduced column title referneces for screenreaders.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -361,6 +361,13 @@ th.bocListTitleCell > a[href]:active
...
+{
+  width: 16px;
+  height: 16px;
+}
```
MichaelKetting, line 369: I think we shoudl also add a padding to the follow-up span, like in NovaViso.

MichaelKetting (changes requested):
Reminder: breaking change ticket and feature ticket for the new / changes APIs.

Remotion/ObjectBinding/Web/UI/Controls/BocValidationErrorIndicatorColumnDefinition.cs
```diff
@@ -39,5 +71,12 @@ protected override IBocColumnRenderer GetRendererInternal (IServiceLocator servi
 
     /// <inheritdoc />
     protected override string DisplayedTypeName => "ValidationErrorIndicatorColumnDefinition";
+
+    public bool IsSortable => true;
+
+    public IComparer<BocListRow> CreateCellValueComparer ()
+    {
+      return new NullComparer<BocListRow>();
+    }
```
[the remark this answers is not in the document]
a-ctor: Its just in there to have a better test experience. The relevant commit is not merged onto develop

Remotion/ObjectBinding/Web/UI/Controls/BocColumnDefinition.cs
```diff
@@ -117,6 +117,42 @@ public virtual WebString ColumnTitle
...
+    [Description("The image displayed in the column title in the rendered page.")]
+    [NotifyParentProperty(true)]
+    public IconInfo ColumnTitleIcon
+    {
+      get { return _columnTitleIcon; }
+      set
+      {
+        ArgumentUtility.CheckNotNull("Icon", value);
+        _columnTitleIcon = value;
+      }
```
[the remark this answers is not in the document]
a-ctor: does not work yet :(

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
a-ctor: We don't need this style on the `td`s? This is the spacing between column title icon and column title text.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
a-ctor: Same here. This is for the column title icons/text so there is no reason to apply the style to `td`.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
a-ctor: I don't think I have enough information to create a bug ticket from your comment.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
a-ctor: I have now solved this problem using a `gap` in the flex box definition above :)

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
a-ctor: The reason I did not introduce it in NovaGray is that the icons there don't have a spacing (wrong icon in column to illustrate):
image

I changed the icon from 12x17 to 16x16 and used it here with a small shift upwards, leaving us with this:
image

Because the icon has some whitespace left and right we don't need to introduce additional margin, which fits in with the existing formatting.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
MichaelKetting: please re-read: that's TD with class "bocListTitleCell" => we have both TH and TD elements in the header row of the BocList. It's for reasons of how screenreaders interpret column headers.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -125,6 +125,20 @@ div.bocList table.bocListTable td.bocListTitleCell > span
...
+{
+  width: var(--remotion-themed-img-width);
+  height: var(--remotion-themed-img-width);
+}
+
+div.bocList table.bocListTable th.bocListTitleCell > span
+{
+  /* Prevent layout changes due to a missing 1px border when switching between sortable/non-sortable. */
```
MichaelKetting, line 140: If you want, keep the comment. personally, the variable name in combination with the border makes it clear that we compensate stuff.

If you keep it, I recommend changing the "missing 1px border" to "missing border for anchor".

MichaelKetting (approved):
Seen the gap. went all 👀. continued with review.

PR #758, opened 2023-07-11
Feature/rm 8867 add tests for evaluate is valid on boc list validation result dispatching validator cs

Frazerus:
https://re-motion.atlassian.net/browse/RM-8867

### DIFF
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
@@ -175,5 +175,70 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
 
       Assert.That(validator.ErrorMessage, Is.EqualTo("Errors on BocList\r\nErrors on rows in BocList\r\n"));
     }
+
+    [Test]
+    public void EvaluateIsValid_WithoutFailures_EmptyErrorMessageAndValid ()
+    {
+      var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
+      validationFailureRepoStub
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(false))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+      var error = "BocList has validation errors.";
+      var resourceManagerStub = new Mock<IResourceManager>();
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.RemainingValidationFailureText", out error)).Returns(true);
+
+      var bocListMock = new Mock<Control> { CallBase = true };
+      bocListMock.As<IBocList>().Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepoStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
+      bocListMock.Object.ID = "someBocList";
+
+      var validator = new BocListValidationResultDispatchingValidator();
+      validator.ErrorMessage = "This should not be output.";
+      validator.ControlToValidate = "someBocList";
+
+      NamingContainer.Controls.Add(bocListMock.Object);
+      NamingContainer.Controls.Add(validator);
+
+      validator.Validate();
+
+      Assert.That(validator.ErrorMessage, Is.EqualTo(""));
+      Assert.That(validator.IsValid, Is.True);
+    }
+
+    [Test]
+    public void EvaluateIsValid_WithFailures_ErrorMessageFromResourceManagerAndInvalid ()
+    {
+      var validationFailuresForBocListRows = new[]
+                                             {
+                                               BocListValidationFailureWithLocationInformation.CreateFailure(BusinessObjectValidationFailure.Create("Errors on BocList rows."))
+                                             };
+
+      var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
+      validationFailureRepoStub
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(false))
+          .Returns(validationFailuresForBocListRows);
+
+      var error = "BocList has validation errors.";
+      var resourceManagerStub = new Mock<IResourceManager>();
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.ValidationFailuresFoundInListErrorMessage", out error)).Returns(true);
+
+      var bocListMock = new Mock<Control> { CallBase = true };
+      bocListMock.As<IBocList>().Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepoStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
+      bocListMock.Object.ID = "someBocList";
+
+      var validator = new BocListValidationResultDispatchingValidator();
+      validator.ErrorMessage = "This should not be output.";
+      validator.ControlToValidate = "someBocList";
+
+      NamingContainer.Controls.Add(bocListMock.Object);
+      NamingContainer.Controls.Add(validator);
+
+      validator.Validate();
+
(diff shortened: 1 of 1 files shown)


Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -145,5 +143,125 @@ public void LazyValidate_WithInternalAndRowFailures_ReturnsAppendedMessageText (
...
+      validator.ErrorMessage = "This should not be output.";
+      validator.ControlToValidate = "someBocList";
+
+      NamingContainer.Controls.Add(bocListMock.Object);
+      NamingContainer.Controls.Add(validator);
+
+      validator.Validate();
+
+      Assert.That(validator.ErrorMessage, Is.EqualTo(""));
+      Assert.That(validator.IsValid, Is.EqualTo(true));
```
a-ctor, line 172: Use `Is.True`/`Is.False` instead

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -145,5 +143,125 @@ public void LazyValidate_WithInternalAndRowFailures_ReturnsAppendedMessageText (
 
       Assert.That(validator.ErrorMessage, Is.EqualTo("Errors on BocList\r\nErrors on rows in BocList"));
     }
+
+    [Test]
+    public void EvaluateIsValid_WithoutFailures_EmptyErrormessageAndValid ()
+    {
+      var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
+      validationFailureRepoStub.Setup(_ => _.HasValidationFailuresForDataRow(It.IsAny<IBusinessObject>())).Returns(false);
```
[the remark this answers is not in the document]
MichaelKetting: The simple rule that was employed in the past  is the following: if the value is "important" to setting up the test, we should still stub it so the behavior/usage is documented. if it's just guard condition satisfaction that's not relevant for the test, we can leave it out. 

So, if we test: 
* do we get "validation failures found" when a validator returns "false" when calling `Validate()`, we should have two tests, one where `Validate()` returns true and one where it returns false. 
* Same goes for a test that actually is intended to verify that an exit condtion is triggered upon false. 

All other cases where the default-return-value case is already covered and noone cares about this check, we can skip stubbing it. 

As for this particular case: If that's the only / primary setup of the test, we should stub it. From the name of the test but without looking at the rest of the code, it might really do feel like it's relevant information that this method causes the emtpy result. Similiar situation would exist for instance when testing a reader-api on a stream and we want to test when the stream is empty. we would still stub reader to return 0 when calling the read method. Otherwise, we have a mock, do nothing with it and then get a result. that just feels...weird when reading the test.

Also, casing-typo in the Test name ("Errormessage" -> "ErrorMessage")

@a-ctor

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -170,5 +170,68 @@ public void RefreshErrorMessage_WithInternalAndRowFailures_ReturnsAppendedMessag
...
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+      var error = "BocList has validation errors.";
+      var resourceManagerStub = new Mock<IResourceManager>();
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.RemainingValidationFailureText", out error)).Returns(true);
+
+      var bocListMock = new Mock<Control> { CallBase = true };
+      bocListMock.As<IBocList>().Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepoStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
+      bocListMock.Object.ID = "someBocList";
```
a-ctor, line 188: We don't need the resource manage setups for this test if I am not mistaken?

a-ctor (approved):

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
Frazerus: We do, implementation changed. 

We don't need it per se, but I would prefer no crashes from the resource manager being null in an extension method if the test somehow doesn't go the right way.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -170,5 +170,68 @@ public void RefreshErrorMessage_WithInternalAndRowFailures_ReturnsAppendedMessag
...
+    public void EvaluateIsValid_WithFailures_ErrorMessageFromResourceManagerAndInvalid ()
+    {
+      var validationFailuresForBocListRows = new[]
+                                             {
+                                               BocListValidationFailureWithLocationInformation.CreateFailure(BusinessObjectValidationFailure.Create("Errors on BocList rows"))
+                                             };
+
+      var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
+      validationFailureRepoStub.Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(false))
+          .Returns(validationFailuresForBocListRows);
```
MichaelKetting, line 213:
```suggestion
      validationFailureRepoStub
          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(false))
          .Returns(validationFailuresForBocListRows);
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -170,5 +170,68 @@ public void RefreshErrorMessage_WithInternalAndRowFailures_ReturnsAppendedMessag
...
+      validationFailureRepoStub.Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(false))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+      var error = "BocList has validation errors.";
+      var resourceManagerStub = new Mock<IResourceManager>();
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.RemainingValidationFailureText", out error)).Returns(true);
+
+      var bocListMock = new Mock<Control> { CallBase = true };
+      bocListMock.As<IBocList>().Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepoStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
```
MichaelKetting, line 187: Depending on the production code, you could also set this up with a `Throws(new  InvalidOperationException("GetResourceManager() should be called in this test.")` to avoid needing a complete stubbing.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -170,5 +170,68 @@ public void RefreshErrorMessage_WithInternalAndRowFailures_ReturnsAppendedMessag
...
+      Assert.That(validator.ErrorMessage, Is.EqualTo(""));
+      Assert.That(validator.IsValid, Is.True);
+    }
+
+    [Test]
+    public void EvaluateIsValid_WithFailures_ErrorMessageFromResourceManagerAndInvalid ()
+    {
+      var validationFailuresForBocListRows = new[]
+                                             {
+                                               BocListValidationFailureWithLocationInformation.CreateFailure(BusinessObjectValidationFailure.Create("Errors on BocList rows"))
```
MichaelKetting, line 208:
```suggestion
                                               BocListValidationFailureWithLocationInformation.CreateFailure(BusinessObjectValidationFailure.Create("Errors on BocList rows."))
```

Since you use complete sentences elsewhere, too.

MichaelKetting (approved):

PR #763, opened 2023-07-13
RM-8869 Align ErrorMessage in BocValidationRow of BocList

Frazerus:
https://re-motion.atlassian.net/browse/RM-8869

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
@@ -118,6 +118,8 @@ class BocList
       BocList.FixUpScrolling(bocList);
     }
 
+    BocList.FixupValidationErrorOverflowSize(bocList);
+
     var selectedRows = new BocList_SelectedRows (selection);
     if (   selectedRows.Selection != BocList._rowSelectionUndefined
         && selectedRows.Selection != BocList._rowSelectionDisabled)
@@ -150,7 +152,7 @@ class BocList
 
         if (hasClickSensitiveRows)
           BocList.BindRowClickEventHandler (bocList, tableRow, tableCell, checkBox);
-    
+
         if (checkBox.checked)
         {
           var rowBlock = new BocList_RowBlock(tableRow, checkBox);
@@ -521,6 +523,24 @@ class BocList
 
       BocList.FixHeaderSize(scrollableContainer);
       BocList.FixHeaderPosition(tableContainer, scrollableContainer);
+      setTimeout(resizeHandler, resizeInterval);
+    };
+    resizeHandler();
+  }
+
+  private static FixupValidationErrorOverflowSize(bocList: HTMLElement): void
+  {
+    const tableBlock = bocList.querySelector(':scope > div.bocListTableBlock')!;
+    const tableContainer = tableBlock.querySelector<HTMLElement>(':scope > div.bocListTableContainer')!;
+    const scrollableContainer = tableContainer.querySelector<HTMLElement>(':scope > div.bocListTableScrollContainer')!;
+
+    const resizeInterval = 50;
+
+    let resizeHandler = function ()
+    {
+      if (!PageUtility.Instance.IsInDom (scrollableContainer))
+        return;
+
       BocList.FixValidationErrorOverflowSize(scrollableContainer);
       setTimeout(resizeHandler, resizeInterval);
     };
@@ -529,14 +549,30 @@ class BocList
 
   private static FixValidationErrorOverflowSize(scrollableContainer: HTMLElement)
   {
-    // Sets the width of the validation error field to be as wide as the visible part of the bocList.
+    // Sets the width of the validation error field to be as wide as the visible part of the bocList
+    // minus how much space the td on the left requires to align with the error indicator column.
     // This enables sticky behaviour while still allowing for a scrollbar if required.
 
-    const width = LayoutUtility.GetWidth(scrollableContainer);
+    const availableWidthInContainer = scrollableContainer.clientWidth;
 
-    const overflowContainers = [...scrollableContainer.querySelectorAll<HTMLElement>(':scope tr.bocListValidationRow > td > div')];
-    overflowContainers.forEach(e => {
-      e.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
+    const tdBeforeValidationFailureColumn: Nullable<HTMLElement> = scrollableContainer.querySelector('tr.bocListValidationRow td:first-child:not(.bocListValidationFailureCell)');
+
+    let widthBeforeValidationFailureColumn = 0;
+    if (tdBeforeValidationFailureColumn != null)
+    {
+      widthBeforeValidationFailureColumn = Math.max(tdBeforeValidationFailureColumn.clientWidth - scrollableContainer.scrollLeft, 0);
+    }
+
+    const overflowContainers = [...scrollableContainer.querySelectorAll<HTMLElement>(':scope tr.bocListValidationRow > td.bocListValidationFailureCell > div')];
+
+    overflowContainers.forEach(container =>
+    {
+      let parentStyle = window.getComputedStyle(<HTMLElement>container.parentNode);
+
+      let padding = parseFloat(parentStyle.paddingLeft) + parseFloat(parentStyle.paddingRight);
+      let border = parseFloat(parentStyle.borderLeftWidth) + parseFloat(parentStyle.borderRightWidth);
+        
+      container.style.width = (availableWidthInContainer - widthBeforeValidationFailureColumn - padding - border).toString() + "px";
     });
   }
 
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
@@ -127,7 +127,14 @@ div.bocList table.bocListTable tr.bocListValidationRow.selected td
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
 {
-  margin: 1em 0;
+  margin: 0.5em 0;
+  padding-left: 0;
+  list-style: none;
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow td li:not(:last-child)
+{
+  margin-bottom: 0.5em;
 }
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow div
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
@@ -393,7 +393,14 @@ div.bocList table.bocListTable tr.bocListValidationRow.selected td
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
 {
+  padding-left: 0;
   margin: var(--remotion-themed-spacing) 0;
+  list-style: none;
+}
+
+div.bocList div.bocListTableBlock tr.bocListValidationRow td li:not(:last-child)
+{
+  margin-bottom: calc(var(--remotion-themed-spacing-small));
 }
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow div
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocRowRendererTest.cs
@@ -521,6 +521,85 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
       Assert.That(ctx.ColumnRenderers, Is.EqualTo(columnRenderers));
     }
 
+    [Test]
+    public void RenderValidationRow_WithValidationFailureColumn_AddsTDBeforeErrorMessageContainer ()
+    {
+      var validationSummaryRendererStub = new StubValidationSummaryRenderer();
+      BocListValidationSummaryRenderArguments? args = null;
+      BocListRenderingContext ctx = null;
+      validationSummaryRendererStub.Callback = (BocRenderingContext<IBocList> context, in BocListValidationSummaryRenderArguments arguments) =>
+      {
+        ctx = context as BocListRenderingContext;
+        args = arguments;
+      };
+
+      var validationFailureRepository = new BocListValidationFailureRepository();
+      validationFailureRepository.AddValidationFailuresForDataRow(BusinessObject, new []{BusinessObjectValidationFailure.Create("Big error")});
+
+      List.Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepository);
+
+      IBocRowRenderer renderer = new BocRowRenderer(
+          _bocListCssClassDefinition,
+          new BocIndexColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition),
+          new BocSelectorColumnRenderer(RenderingFeatures.Default, _bocListCssClassDefinition, _stubLabelReferenceRenderer),
+          RenderingFeatures.Default,
+          validationSummaryRendererStub);
+      var columnRenderers = new[]
+                            {
+                                new BocColumnRenderer(
+                                    new StubColumnRenderer(new FakeResourceUrlFactory()),
+                                    new StubColumnDefinition(),
+                                    columnIndex: 0,
+                                    visibleColumnIndex: 2,
+                                    isRowHeader: true,
+                                    showIcon: false,
+                                    SortingDirection.None,
+                                    orderIndex: -1),
+                                new BocColumnRenderer(
+                                    new StubColumnRenderer(new FakeResourceUrlFactory()),
+                                    new BocValidationErrorIndicatorColumnDefinition(),
+                                    columnIndex: 1,
+                                    visibleColumnIndex: 3,
+                                    isRowHeader: true,
+                                    showIcon: false,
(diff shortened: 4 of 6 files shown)


KuJo7 (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,16 +529,27 @@ class BocList
...
+
+    let widthBeforeValidationFailureColumn = 0;
+    if (tdBeforeValidationErrorColumn)
+        // +1 for rounding errors
+      widthBeforeValidationFailureColumn = LayoutUtility.GetWidth(tdBeforeValidationErrorColumn) + 1;
+
+    // -1 for rounding errors
+    const availableWidthInContainer = LayoutUtility.GetWidth(scrollableContainer) - 1;
+
+    overflowContainer.style.width = "calc(" + (availableWidthInContainer - widthBeforeValidationFailureColumn).toString() + "px - var(--remotion-themed-scrollbar-width))";
```
MichaelKetting, line 552: We missed this bug in the earlier PR: must not use css-variables like this in Javascript. They are theme-specific, not a contract.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,16 +529,27 @@ class BocList
...
-      overflowContainer.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
+      return;
     }
+
+    const tdBeforeValidationErrorColumn: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow td');
+
+    let widthBeforeValidationFailureColumn = 0;
+    if (tdBeforeValidationErrorColumn)
+        // +1 for rounding errors
+      widthBeforeValidationFailureColumn = LayoutUtility.GetWidth(tdBeforeValidationErrorColumn) + 1;
```
MichaelKetting, line 547: please add braces. multiline blocks always have braces.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -377,6 +378,7 @@ div.bocList table.bocListTable tr.bocListValidationRow.selected td
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
 {
+  padding-left: var(--bocList-validationrow-ul-padding);
```
MichaelKetting, line 381: would like to do a quick live demo once the changes are done to check how things look with the various column configurations.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -521,22 +524,54 @@ class BocList
...
     // This enables sticky behaviour while still allowing for a scrollbar if required.
 
-    const width = LayoutUtility.GetWidth(scrollableContainer);
+    // -1 for rounding errors
+    const availableWidthInContainer = LayoutUtility.GetWidth(scrollableContainer) - 1;
+
+    // +1 for rounding errors
+    const scrollbarWidth = scrollableContainer.offsetWidth - scrollableContainer.clientWidth + 1;
+
+    const tdBeforeValidationErrorColumn: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow td');
```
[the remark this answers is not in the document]
Frazerus: Well we would just have `widthBeforeValidationFailurecolumn` as 0. We do not care about scrolling in that case, as this also means that we always have  the full length of the row.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -521,22 +524,54 @@ class BocList
...
+
+    let widthBeforeValidationFailureColumn = 0;
+    if (tdBeforeValidationErrorColumn){
+      // +1 for rounding errors
+      widthBeforeValidationFailureColumn = Math.max(LayoutUtility.GetWidth(tdBeforeValidationErrorColumn) + 1 - scrollableContainer.scrollLeft + 1, 0);
```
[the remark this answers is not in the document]
Frazerus: This is so incredibly imprecise that without that extra +1 it simply wouldn't work. At this point, i might just have a `-5px` at the end just to be sure that we have enough space. Thoughts? (Maybe demo, would probaby be good here, maybe there is an error here I'm overlooking.)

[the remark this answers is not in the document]
Frazerus: There is also a scrollbar appearing, which might not be what we want -> We should just talk about it.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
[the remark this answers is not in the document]
MichaelKetting: Check. Just for completeness sake: my assumption was, that "width" would be incorrect, subtract the whole width from the width (something like that). Anyway, not just a "it's weird" but rather an "it would be incorrect results".

Am guessing your edit also refers to "found out that there really is an actual problem"

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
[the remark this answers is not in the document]
MichaelKetting: yes, please. I don't mind dirty hacks in CSS/JS if all else fails. Right now it just looks like double-compensation of the same issue. Looking forward to learning more tomorrow. Will not think about it tonight so I can sleep peacefully ;)

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -9,6 +9,7 @@ div.bocList
   --boclist-menublock-minimum-width: 10em;
   --boclist-menublock-maximum-width: auto;
   --boclist-validationrow-minimum-width: 20em;
+  --bocList-validationrow-ul-padding: 1em;
```
MichaelKetting, line 12:
```suggestion
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -393,6 +394,7 @@ div.bocList table.bocListTable tr.bocListValidationRow.selected td
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
 {
+  padding-left: var(--bocList-validationrow-ul-padding);
   margin: var(--remotion-themed-spacing) 0;
 }
```
MichaelKetting, line 399:
```suggestion
div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
{
  margin: var(--remotion-themed-spacing) 0;
  padding-left: 0;
  list-style: none;
}

div.bocList div.bocListTableBlock tr.bocListValidationRow td li:not(:last-child)
{
  margin-bottom: calc(var(--remotion-themed-spacing-small));
}
```

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -127,6 +127,7 @@ div.bocList table.bocListTable tr.bocListValidationRow.selected td
 
 div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
 {
+  padding-left: 1.3em;
   margin: 1em 0;
 }
```
MichaelKetting, line 132:
```suggestion
div.bocList div.bocListTableBlock tr.bocListValidationRow td ul
{
  margin: 0.5em 0;
  padding-left: 0;
  list-style: none;
}

div.bocList div.bocListTableBlock tr.bocListValidationRow td li:not(:last-child)
{
  margin-bottom: 0.5em;
}
```

MichaelKetting (commented):
I considered the CSS a bit more and optimized :)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,33 @@ class BocList
...
+    const td: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow td:first-child');
+
+    let widthBeforeValidationFailureColumn = 0;
+    // width does not matter if the first td is already the one with the error messages (and therefore has child nodes)
+    if (td != null && !td.hasChildNodes())
+    {
+      widthBeforeValidationFailureColumn = Math.max(td.clientWidth + 1 - scrollableContainer.scrollLeft, 0);
+    }
 
     const overflowContainers = [...scrollableContainer.querySelectorAll<HTMLElement>(':scope tr.bocListValidationRow > td > div')];
```
MichaelKetting, line 568: if we do introduce the css class, we can also update this rule here.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,33 @@ class BocList
...
+
+    overflowContainers.forEach(container => {
+
+      let style = window.getComputedStyle(container);
+
+      let margin = parseFloat(style.marginLeft) + parseFloat(style.marginRight);
+      let padding = parseFloat(style.paddingLeft) + parseFloat(style.paddingRight);
+      let border = parseFloat(style.borderLeftWidth) + parseFloat(style.borderRightWidth);
+
+      container.style.width = (availableWidthInContainer - widthBeforeValidationFailureColumn - margin - padding - border).toString() + "px";
```
MichaelKetting, line 578: nitpick: this depends on using the correct `box-sizing` model. If you use `content-box`, the width is without the padding, border, and marging, and with `border-box`, the padding and border are part of the width. 

I'm not 100% sure I get the intention of the width calculation just from reading the code. what I see is that we get the margin+padding+border for the container element and then set a width minus all those. if we want to be 100% correct, we need to test which box-sizing is used and adjust accordingly. On the other hand, if we define that margin, padding and border of the root div must always be zero, we don't need to account for anything. 

I checked and it seems like container doesn't have any padding, border, or margin anyway. We could perhaps simply ignore this? or perhaps add an entry in CSS with a border, margin and padding on this element being `0 !important` plus a note. That way, it's not (easily) overrideable. The alternative is to also account for box-sizing changes here in the code (probably cleaenst?) or force those properties to zero in javascript (not preferred)

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Frazerus: This is supposed to fix the padding on the parent td, which is 1 by default, but I used the div instead of the td. Therefore I think we don't care about the box sizing as it should always add the padding to the inner width of a child element.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,39 @@ class BocList
...
-      e.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
+      let margin = 0;
+      let padding = 0;
+      let border = 0;
+
+      let parentTd = container.parentNode as HTMLElement;
+      if(parentTd != null){
+        let style = window.getComputedStyle(parentTd);
+
+        margin = parseFloat(style.marginLeft) + parseFloat(style.marginRight);
```
MichaelKetting, line 580: waiting for the push of the final update with `margin`

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,36 @@ class BocList
...
 
-    const overflowContainers = [...scrollableContainer.querySelectorAll<HTMLElement>(':scope tr.bocListValidationRow > td > div')];
-    overflowContainers.forEach(e => {
-      e.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
+      let padding = 0;
+      let border = 0;
+
+      let parentTd = container.parentNode as HTMLElement;
+      if(parentTd != null){
+        let style = window.getComputedStyle(parentTd);
```
MichaelKetting, line 576:
```suggestion
      if (parentTd != null)
      {
        let style = window.getComputedStyle(parentTd);
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,36 @@ class BocList
...
     // This enables sticky behaviour while still allowing for a scrollbar if required.
 
-    const width = LayoutUtility.GetWidth(scrollableContainer);
+    const availableWidthInContainer = scrollableContainer.clientWidth;
+
+    const td: Nullable<HTMLElement> = scrollableContainer.querySelector('tr.bocListValidationRow td:first-child:not(.bocListValidationFailureCell)');
+
+    let widthBeforeValidationFailureColumn = 0;
+    // width does not matter if the first td is already the one with the error messages (and therefore has child nodes)
+    if (td != null && !td.hasChildNodes())
```
MichaelKetting, line 562:
```suggestion
    if (tdBeforeValidationFailureColumn != null)
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,36 @@ class BocList
...
   {
-    // Sets the width of the validation error field to be as wide as the visible part of the bocList.
+    // Sets the width of the validation error field to be as wide as the visible part of the bocList
+    // minus how much space the td on the left requires to align with the error indicator column.
     // This enables sticky behaviour while still allowing for a scrollbar if required.
 
-    const width = LayoutUtility.GetWidth(scrollableContainer);
+    const availableWidthInContainer = scrollableContainer.clientWidth;
+
+    const td: Nullable<HTMLElement> = scrollableContainer.querySelector('tr.bocListValidationRow td:first-child:not(.bocListValidationFailureCell)');
```
MichaelKetting, line 558:
```suggestion
    const tdBeforeValidationFailureColumn: Nullable<HTMLElement> = scrollableContainer.querySelector('tr.bocListValidationRow td:first-child:not(.bocListValidationFailureCell)');
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,36 @@ class BocList
...
+
+    let widthBeforeValidationFailureColumn = 0;
+    // width does not matter if the first td is already the one with the error messages (and therefore has child nodes)
```
MichaelKetting, line 561:
```suggestion
```
the part in the parenthesis is obsolete now and I think we can drop the entire comment because the code says it all.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,36 @@ class BocList
...
+      let padding = 0;
+      let border = 0;
+
+      let parentTd = container.parentNode as HTMLElement;
+      if(parentTd != null){
+        let style = window.getComputedStyle(parentTd);
+
+        padding = parseFloat(style.paddingLeft) + parseFloat(style.paddingRight);
+        border = parseFloat(style.borderLeftWidth) + parseFloat(style.borderRightWidth);
+      }
```
MichaelKetting, line 580:
```suggestion
      let style = window.getComputedStyle(container.parentNode!);
      padding = parseFloat(style.paddingLeft) + parseFloat(style.paddingRight);
      border = parseFloat(style.borderLeftWidth) + parseFloat(style.borderRightWidth);
```

The container.parentNode can't be null at this point because we just got it from the page. You may need to add a cast back in, just not a soft-cast if possible. That way, we can simplify a lot and don't need to invent names like `parentTd`.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -529,14 +549,36 @@ class BocList
...
-    const width = LayoutUtility.GetWidth(scrollableContainer);
+    const availableWidthInContainer = scrollableContainer.clientWidth;
+
+    const td: Nullable<HTMLElement> = scrollableContainer.querySelector('tr.bocListValidationRow td:first-child:not(.bocListValidationFailureCell)');
+
+    let widthBeforeValidationFailureColumn = 0;
+    // width does not matter if the first td is already the one with the error messages (and therefore has child nodes)
+    if (td != null && !td.hasChildNodes())
+    {
+      widthBeforeValidationFailureColumn = Math.max(td.clientWidth - scrollableContainer.scrollLeft, 0);
```
MichaelKetting, line 564:
```suggestion
      widthBeforeValidationFailureColumn = Math.max(tdBeforeValidationFailureColumn.clientWidth - scrollableContainer.scrollLeft, 0);
```

MichaelKetting (approved):
I think I was able to streamline a bit. pelase verify if my suggestions make sense and if so, we're done :)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Frazerus: I dont understand which part in the parenthesis youre talking about here. The "hasChildNodes()" part?

MichaelKetting: I was referrring to the selected comment line and this text part of the comment: `(and therefore has child nodes)`
Unless we drop the entire comment.

PR #764, opened 2023-07-14
Feature/rm 8870 enable dispatching validation failures for a custom set of boc list rows

Frazerus:
https://re-motion.atlassian.net/browse/RM-8870

### DIFF
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListListValidationFailureHandlerTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListListValidationFailureHandlerTest.cs
@@ -1,0 +1,78 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System.Text;
+using Moq;
+using NUnit.Framework;
+using Remotion.ObjectBinding.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class BocListListValidationFailureHandlerTest
+  {
+    [Test]
+    public void HandleValidationFailures_WithValidationFailures_ContextContainsConcatenatedMessages ()
+    {
+      var bocListStub = new Mock<IBocList>();
+      var repository = new BocListValidationFailureRepository();
+
+      bocListStub.Setup(_ => _.ValidationFailureRepository).Returns(repository);
+
+      var validationFailures = new[]
+                               {
+                                     BusinessObjectValidationFailure.Create("Error #1"),
+                                     BusinessObjectValidationFailure.Create("Error #2")
+                               };
+
+      repository.AddValidationFailuresForBocList(validationFailures);
+
+      var context = new ValidationFailureHandlingContext(bocListStub.Object);
+
+      var handler = new BocListListValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("Error #1\r\nError #2\r\n"));
+    }
+
+    [Test]
+    public void HandleValidationFailures_WithoutValidationFailures_ContextContainsEmptyMessage ()
+    {
+      var bocListStub = new Mock<IBocList>();
+
+      var repository = new BocListValidationFailureRepository();
+
+      bocListStub.Setup(_ => _.ValidationFailureRepository).Returns(repository);
+
+      var context = new ValidationFailureHandlingContext(bocListStub.Object);
+
+      var handler = new BocListListValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.Empty);
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListRowAndCellValidationFailureHandlerTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListRowAndCellValidationFailureHandlerTest.cs
@@ -1,0 +1,90 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System.Text;
+using Moq;
+using NUnit.Framework;
+using Remotion.Globalization;
+using Remotion.ObjectBinding.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class BocListRowAndCellValidationFailureHandlerTest
+  {
+    private Mock<IBocList> _bocListStub;
+    private BocListValidationFailureRepository _repository;
+
+    [SetUp]
+    public void Setup ()
+    {
+      _bocListStub = new Mock<IBocList>();
+      _repository = new BocListValidationFailureRepository();
+    }
+
+    [Test]
+    public void HandleValidationFailures_WithoutValidationFailures_ContextContainsEmptyMessage ()
+    {
+      _bocListStub.Setup(_ => _.ValidationFailureRepository).Returns(_repository);
+
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      var handler = new BocListRowAndCellValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.Empty);
+    }
+
+    [Test]
+    public void HandleValidationFailures_WithValidationFailures_ContextContainsConcatenatedMessages ()
+    {
+      var resourceManagerStub = new Mock<IResourceManager>();
+
+      _bocListStub.Setup(_ => _.ValidationFailureRepository).Returns(_repository);
+      _bocListStub.Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
+
+      var validationFailures = new[]
+                               {
+                                     BusinessObjectValidationFailure.Create("Oh no such an error."),
+                                     BusinessObjectValidationFailure.Create("What, another one?")
+                               };
+
+      _repository.AddValidationFailuresForBocList(validationFailures);
+
+      var errorMessage = "Error message from resource manager.";
+      resourceManagerStub
+          .Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.ValidationFailuresFoundInOtherListPagesErrorMessage", out errorMessage))
+          .Returns(true);
+
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      var handler = new BocListRowAndCellValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("Error message from resource manager.\r\n"));
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
@@ -37,9 +37,17 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
       validationFailureRepoStub
           .Setup(_ => _.GetUnhandledValidationFailuresForBocList(true))
           .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+
+      var validationFailuresForBocList = new[]
+                                         {
+                                           BocListValidationFailureWithLocationInformation.CreateFailure(BusinessObjectValidationFailure.Create("Errors on BocList"))
+                                         };
+
       validationFailureRepoStub
-          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
-          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+          .SetupSequence(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>())
+          .Returns(validationFailuresForBocList);
 
       var error = "New error just arrived";
       var resourceManagerStub = new Mock<IResourceManager>();
@@ -66,9 +74,11 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
     public void RefreshErrorMessage_WithBocListValidationFailures_ReturnsFailuresOfBocList ()
     {
       var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
+      var resourceManagerStub = new Mock<IResourceManager>();
 
       var bocListMock = new Mock<Control> { CallBase = true };
       bocListMock.As<IBocList>().Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepoStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
       bocListMock.Object.ID = "someBocList";
 
       var validationFailuresForBocList = new[]
@@ -107,9 +117,12 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
       validationFailureRepoStub
           .Setup(_ => _.GetUnhandledValidationFailuresForBocList(true))
           .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
       validationFailureRepoStub
-          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
-          .Returns(validationFailuresForBocListRows);
+          .SetupSequence(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(validationFailuresForBocListRows)
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
       var error = "Errors on rows in BocList";
 
       var resourceManagerStub = new Mock<IResourceManager>();
@@ -136,6 +149,44 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
     }
 
     [Test]
+    public void RefreshErrorMessage_WithoutErrors_ReturnsLocalizedInfo ()
+    {
+      var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
+      validationFailureRepoStub
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocList(true))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+      validationFailureRepoStub
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+
+      var error = "Errors on rows in BocList";
+
+      var resourceManagerStub = new Mock<IResourceManager>();
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.ValidationFailuresFoundInListErrorMessage", out error)).Returns(true);
+
+      var businessObjectStub = new Mock<IBusinessObject>();
+
+      var bocListMock = new Mock<Control> { CallBase = true };
+      bocListMock.As<IBocList>().Setup(_ => _.ValidationFailureRepository).Returns(validationFailureRepoStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.GetResourceManager()).Returns(resourceManagerStub.Object);
+      bocListMock.As<IBocList>().Setup(_ => _.Value).Returns(new[] { businessObjectStub.Object });
+      bocListMock.Object.ID = "someBocList";
+
+      var validator = new BocListValidationResultDispatchingValidator();
+      validator.ErrorMessage = "This should not be output.";
+      validator.ControlToValidate = "someBocList";
+
+      NamingContainer.Controls.Add(bocListMock.Object);
+      NamingContainer.Controls.Add(validator);
+
+      ((IValidatorWithDynamicErrorMessage)validator).RefreshErrorMessage();
+
+      Assert.That(validator.ErrorMessage, Is.EqualTo("Errors on rows in BocList"));
+    }
+
+    [Test]
     public void RefreshErrorMessage_WithValidationFailuresForBoundPropertyAndRowFailures_ReturnsAppendedMessageText ()
     {
       var validationFailuresForBocList = new[]
@@ -149,8 +200,16 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
                                              };
 
       var validationFailureRepoStub = new Mock<IBocListValidationFailureRepository>();
-      validationFailureRepoStub.Setup(_ => _.GetUnhandledValidationFailuresForBocList(true)).Returns(validationFailuresForBocList);
-      validationFailureRepoStub.Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true)).Returns(validationFailuresForBocListRows);
+      validationFailureRepoStub
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocList(true))
+          .Returns(validationFailuresForBocList);
+
+      validationFailureRepoStub
+          .SetupSequence(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(validationFailuresForBocListRows)
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+
       var error = "Errors on rows in BocList";
 
       var resourceManagerStub = new Mock<IResourceManager>();
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/CompoundBocListValidationFailureHandlerTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/CompoundBocListValidationFailureHandlerTest.cs
@@ -1,0 +1,37 @@
+﻿using Moq;
+using NUnit.Framework;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class CompoundBocListValidationFailureHandlerTest
+  {
+    [Test]
+    public void HandleValidationFailures_CallsContainedHandlers ()
+    {
+      var handler1Mock = new Mock<IBocListValidationFailureHandler>();
+      var handler2Mock = new Mock<IBocListValidationFailureHandler>();
+
+      var validationFailureHandler = new CompoundBocListValidationFailureHandler(new[] { handler1Mock.Object, handler2Mock.Object });
+      var context = new ValidationFailureHandlingContext(new Mock<IBocList>().Object);
+
+      var sequence = new VerifiableSequence();
+      handler1Mock
+          .InVerifiableSequence(sequence)
+          .Setup(_ => _.HandleValidationFailures(context))
+          .Verifiable();
+      handler2Mock
+          .InVerifiableSequence(sequence)
+          .Setup(_ => _.HandleValidationFailures(context))
+          .Verifiable();
+
+      validationFailureHandler.HandleValidationFailures(context);
+
+      handler1Mock.Verify();
+      handler2Mock.Verify();
+      sequence.Verify();
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
@@ -1,0 +1,127 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System;
+using System.Text;
+using Moq;
+using NUnit.Framework;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class ValidationFailureHandlingContextTest
+  {
+    private Mock<IBocList> _bocListStub;
+
+    [SetUp]
+    public void Setup ()
+    {
+      _bocListStub = new Mock<IBocList>();
+    }
+
+    [Test]
+    public void AppendErrorMessage_AppendsReportedMessagesToStringBuilder ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      context.ReportErrorMessage("First error message.");
+      context.ReportErrorMessage("Second error message.");
+
+      var stringBuilder = new StringBuilder();
+
+      context.AppendErrorMessages(stringBuilder);
+
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("First error message.\r\nSecond error message.\r\n"));
+    }
+
+    [Test]
+    public void AppendErrorMessages_WithExistingContentInStringBuilder_AppendsNewLineAndReportedMessagesToStringBuilder ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      context.ReportErrorMessage("First error message.");
+      context.ReportErrorMessage("Second error message.");
+
+      var stringBuilder = new StringBuilder();
+      stringBuilder.Append("Existing message without newline.");
+
+      context.AppendErrorMessages(stringBuilder);
+
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("Existing message without newline.\r\nFirst error message.\r\nSecond error message.\r\n"));
+    }
+
+    [Test]
+    public void AppendErrorMessages_WithTrailingNewLineInStringBuilder_AppendsReportedMessagesToStringBuilder ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      context.ReportErrorMessage("First error message.");
+      context.ReportErrorMessage("Second error message.");
+
+      var stringBuilder = new StringBuilder();
+      stringBuilder.AppendLine("Existing message with newline.");
+
+      context.AppendErrorMessages(stringBuilder);
+
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("Existing message with newline.\r\nFirst error message.\r\nSecond error message.\r\n"));
+    }
+
+    [Test]
+    public void AppendErrorMessage_WithMessagesAfterReporting_AppendsReportedMessagesToStringBuilder ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      context.ReportErrorMessage("First error message.");
+      context.ReportErrorMessage("Second error message.");
+
+      var stringBuilder = new StringBuilder();
+
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("First error message.\r\nSecond error message.\r\n"));
+
+      context.ReportErrorMessage("Third error message");
+
+      stringBuilder = new StringBuilder();
+
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("First error message.\r\nSecond error message.\r\nThird error message\r\n"));
+    }
+
+    [Test]
+    public void BocList_ReturnsBocListSuppliedInCtor ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      Assert.That(context.BocList, Is.SameAs(_bocListStub.Object));
+    }
+
+    [Test]
+    public void ValidationFailureRepository_ReturnsValidationFailureRepositoryOfBocList ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      var repository = new BocListValidationFailureRepository();
+
+      _bocListStub.Setup(_ => _.ValidationFailureRepository).Returns(repository);
+
+      Assert.That(context.ValidationFailureRepository, Is.SameAs(repository));
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
@@ -203,6 +203,10 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation
     /// Gets the list of validation errors for this control.
     /// </summary>
     IEnumerable<PlainTextString> GetValidationErrors ();
+
+    /// <summary> Gets the <see cref="IBusinessObject"/>s relevant for validation.</summary>
+    IReadOnlyList<IBusinessObject>? GetBusinessObjectsForValidation ();
+
     string? ControlServicePath { get; }
   }
 }
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
@@ -204,7 +204,7 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation
     /// </summary>
     IEnumerable<PlainTextString> GetValidationErrors ();
 
-    /// <summary> Gets the <see cref="IBusinessObject"/>s relevant for validation.</summary>
+    /// <summary> Gets the <see cref="IBusinessObject"/>s relevant for validation. This list can be a superset of the <see cref="Value" /> property.</summary>
     IReadOnlyList<IBusinessObject>? GetBusinessObjectsForValidation ();
 
     string? ControlServicePath { get; }
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListListValidationFailureHandler.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListListValidationFailureHandler.cs
@@ -1,0 +1,54 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation
+{
+  /// <summary>
+  /// Handles all validation failures for the <see cref="IBusinessObjectProperty"/> of the
+  /// <see cref="IBocList"/>. Should always be the first handler to run.
+  /// </summary>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = Position)]
+  public class BocListListValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    public const int Position = -1_000_000_000;
+
+    public BocListListValidationFailureHandler ()
+    {
+    }
+
+    /// <summary>
+    ///   Handles all unhandled validation failures on the <see cref="IBusinessObjectProperty"/>
+    ///   of the <see cref="IBocList"/>.
+    ///   Reports all found failures as separate messages to the <see cref="ValidationFailureHandlingContext"/>.
+    /// </summary>
+    public void HandleValidationFailures (ValidationFailureHandlingContext context)
+    {
+      ArgumentUtility.CheckNotNull("context", context);
+
+      var validationFailures = context.ValidationFailureRepository.GetUnhandledValidationFailuresForBocList(true);
+
+      foreach (var validationFailure in validationFailures)
+      {
+        context.ReportErrorMessage(validationFailure.Failure.ErrorMessage);
+      }
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListRowAndCellValidationFailureHandler.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListRowAndCellValidationFailureHandler.cs
@@ -1,0 +1,61 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System;
+using System.Linq;
+using Remotion.Globalization;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation
+{
+  /// <summary>
+  /// Handles all still unhandled validation failures and should always be the last handler to run.
+  /// </summary>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = Position)]
+  public class BocListRowAndCellValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    public const int Position = 1_000_000_000;
+
+    public BocListRowAndCellValidationFailureHandler ()
+    {
+    }
+
+    /// <summary>
+    /// Handles all still unhandled validation failures by reporting a single message to the <see cref="ValidationFailureHandlingContext"/>.
+    /// </summary>
+    /// <remarks>
+    ///   It is expected that all unhandled validation failures at this point are unhandled due to them being on a different page of the
+    ///   <see cref="IBocList"/>.
+    /// </remarks>
+    public void HandleValidationFailures (ValidationFailureHandlingContext context)
+    {
+      ArgumentUtility.CheckNotNull("context", context);
+
+      var hasRowsWithUnhandledValidationFailures = context.ValidationFailureRepository.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true).Any();
+
+      if (!hasRowsWithUnhandledValidationFailures)
+        return;
+
+      var remainingValidationFailuresText = context.BocList
+          .GetResourceManager()
+          .GetString(BocList.ResourceIdentifier.ValidationFailuresFoundInOtherListPagesErrorMessage);
+
+      context.ReportErrorMessage(remainingValidationFailuresText);
+    }
+  }
+}
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/CompoundBocListValidationFailureHandler.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/CompoundBocListValidationFailureHandler.cs
@@ -1,0 +1,51 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation
+{
+  /// <summary>
+  /// Aggregates a set of <see cref="IBocListValidationFailureHandler"/> instances. When performing validation failure handling all handlers are executed.
+  /// </summary>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Compound)]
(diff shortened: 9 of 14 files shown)


Frazerus:
Unit tests still missing, will add on Monday.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  public class ValidationFailureHandlingContext
+  {
+    public IBocList BocList { get; }
+
+    public IBocListValidationFailureRepository ValidationFailureRepository { get; }
+
+    public string CompoundedErrorMessage => _sb.ToString();
```
a-ctor, line 12: I personally would use a method for this functionality. This makes it clearer that there is some kind of computation behind it and it also hides it a bit more from the user.

a-ctor (commented):
Looks good. Please don't forget the license header and the arg checks. I am going to do a full review once the unit tests are also here.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    private readonly StringBuilder _sb;
+
+    public ValidationFailureHandlingContext (IBocList bocList)
+    {
+      BocList = bocList;
+      ValidationFailureRepository = bocList.ValidationFailureRepository;
```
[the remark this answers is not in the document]
Frazerus: I don't quite understand which one you mean and what benefits there would be? We could also just have something like `public IBocListValidationFailureRepository ValidationFailureRepository => BocList.ValidationFailureRepository;`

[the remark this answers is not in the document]
a-ctor: > `public IBocListValidationFailureRepository ValidationFailureRepository => BocList.ValidationFailureRepository;`

That is exactly what I mean by getter implementation :D

Previously, it was created as auto-property which includes a backing field. So we are storing the `ValidationFailureRepository` in addition to the `BocList` in our context. While not wrong in this case, it a) adds another field for nothing to the context and b) if the repo would change on `BocList` we still use the old value.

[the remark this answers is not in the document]
Frazerus: Huh, somehow I thought Getter implementation, hmm that means method with getter in name.... Java & C++ baggage :(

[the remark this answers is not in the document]
a-ctor: image

pls don't tell the java guys we are so much cooler than them

Remotion/Web/Core/UI/Controls/IValidatorWithDynamicErrorMessage.cs
```diff
@@ -0,0 +1,33 @@
...
+//
+using System;
+using System.Web.UI;
+
+namespace Remotion.Web.UI.Controls
+{
+  /// <summary>
+  /// This interface defines features for a validator which needs to have its <see cref="IValidator.ErrorMessage"/> updated at a late point.
+  /// </summary>
+  public interface IValidatorWithDynamicErrorMessage : IValidator
```
a-ctor, line 26: I don't know how much of the naming is already discussed but there are some inconsistencies. Here we are talking about a "dynamic error message" which is "refreshed". In the custom control we added, we are talking about "lazy evaluation". These two descriptions don't fit well together IMO. I think we should change one side to fit the other.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Multiple, Position = 100)]
+  public class BocListRowAndCellValidationFailureHandler : IBocListValidationFailureHandler
+  {
+
+    public void HandleValidationFailures (ValidationFailureHandlingContext context)
```
a-ctor, line 29:
```suggestion
  {
    public void HandleValidationFailures (ValidationFailureHandlingContext context)
```

a-ctor (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/IBocList.cs
```diff
@@ -203,6 +203,10 @@ public interface IBocList
     /// Gets the list of validation errors for this control.
     /// </summary>
     IEnumerable<PlainTextString> GetValidationErrors ();
+
+    /// <summary> Gets the <see cref="IBusinessObject"/>s relevant for validation.</summary>
```
MichaelKetting, line 207:
```suggestion
    /// <summary> Gets the <see cref="IBusinessObject"/>s relevant for validation. This list can be a superset of the <see cref="Value" /> property.</summary>
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+//
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Multiple, Position = -100)]
+  public class BocListListValidationFailureHandler : IBocListValidationFailureHandler
+  {
+
```
MichaelKetting, line 26: I try to always add a ctor, even if it's empty. that way, there is a visual hook when reading the class about what is required for instantiation.

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  public interface IBocListValidationFailureHandler
```
MichaelKetting, line 20: please add xml doc for the interface type and member. Since this is a required extension point, we may need to add how-to docs about what can be done / how it is done.

see other interfaces, I usually add a `seealso` for implementors to make this easier discoverable when reading the docs.

Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+// using NUnit.Framework;
+//
+using System.Collections.Generic;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Compound)]
+  public class CompoundBocListValidationFailureHandler : IBocListValidationFailureHandler
```
MichaelKetting, line 25: class doc, see other Compound types.

Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System.Collections.Generic;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Compound)]
```
MichaelKetting, line 24: please add lifetimekind. unless there's a reason to be stateful, it's usally singleton for all implementations.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/CompoundBocListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,26 @@
...
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class CompoundBocListValidationFailureHandlerTest
+  {
+    [Test]
+    public void HandleValidationFailures_CallsContainedHandlers ()
```
MichaelKetting, line 12: Please add VerifiableSequence to the test. The order is relevant during execution.
see CompoundExtensibleEnumGlobalizationServiceTest

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
MichaelKetting, line 20: please add integeration test for IoC, see other interfaces composed via ImplementFor attribute
e.g. `IExtensibleEnumGlobalizationServiceTest`

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Linq;
+using Remotion.Globalization;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Multiple, Position = 100)]
+  public class BocListRowAndCellValidationFailureHandler : IBocListValidationFailureHandler
+  {
```
MichaelKetting, line 27: please add empty default ctor. Implementationkind=signleton.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+// using NUnit.Framework;
+//
+using System.Linq;
+using Remotion.Globalization;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Multiple, Position = 100)]
```
MichaelKetting, line 25: please move the value of Positon into a const (see other ImplemntationFor usages).

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  public class ValidationFailureHandlingContext
+  {
+    public IBocList BocList { get; }
+
+    public IBocListValidationFailureRepository ValidationFailureRepository => BocList.ValidationFailureRepository;
+
+    private readonly StringBuilder _sb;
```
MichaelKetting, line 30:
```suggestion
    private readonly StringBuilder _errorMessage;
```
or `_errorMessageBuilder`

`sb` is only okay in very limited scenarios (mostly only places where we would use other abbreviations, too. `stringBuilder` is okay when it's in the very local scope and obvious what it is (e.g. the only purpose of the method is building that string). And even then, it's probably more of a legacy thing than good practice. For fields it should always be named, same as when we have a `string` or `int` field, which also have semantic names.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,52 @@
...
+
+      BocList = bocList;
+      _sb = new StringBuilder();
+    }
+
+    public string GetCompoundedErrorMessage () => _sb.ToString();
+
+    public void ReportErrorMessage (string errorMessage)
+    {
+      ArgumentUtility.CheckNotNull("errorMessage", errorMessage);
```
MichaelKetting, line 44:
```suggestion
      ArgumentUtility.CheckNotNullOrEmpty("errorMessage", errorMessage);
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+      ArgumentUtility.CheckNotNull("bocList", bocList);
+
+      BocList = bocList;
+      _sb = new StringBuilder();
+    }
+
+    public string GetCompoundedErrorMessage () => _sb.ToString();
+
+    public void ReportErrorMessage (string errorMessage)
```
MichaelKetting, line 42: Since this is a user api, we should add doc and parameter doc incl. not-null-or-empty.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,52 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System.Text;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  public class ValidationFailureHandlingContext
```
MichaelKetting, line 24: clss doc

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
```diff
@@ -0,0 +1,57 @@
...
+  public class ValidationFailureHandlingContextTest
+  {
+    private Mock<IBocList> _bocListStub;
+
+    [SetUp]
+    public void Setup ()
+    {
+      _bocListStub = new Mock<IBocList>();
+    }
+
```
MichaelKetting, line 35: test for the repository-property please. actually, also for the boclist property.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Test]
+    public void ReportErrorMessage_WithEmptyMessage_EmptyMessageSaved ()
```
MichaelKetting, line 37: Ah, I see you're passing empty as an expected case. see other comment on Context itself. Not sure if Empty is a good idea / helpful? my goto would be "no".

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    public ValidationFailureHandlingContext (IBocList bocList)
+    {
+      ArgumentUtility.CheckNotNull("bocList", bocList);
+
+      BocList = bocList;
+      _sb = new StringBuilder();
+    }
+
+    public string GetCompoundedErrorMessage () => _sb.ToString();
```
MichaelKetting, line 40: `GetErrorMessage` or `GetAggregatedErrorMessage`

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    [Test]
+    public void ReportErrorMessage_WithMessages_ConcatenatesMessages ()
+    {
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      context.ReportErrorMessage("First error message.");
+      context.ReportErrorMessage("Second error message.");
+
+      Assert.That(context.GetCompoundedErrorMessage(), Is.EqualTo("First error message.\r\nSecond error message.\r\n"));
```
MichaelKetting, line 54: please add a test where another message is added after Get is called the first time.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), RegistrationType = RegistrationType.Multiple, Position = -100)]
```
MichaelKetting, line 23: please move the value of Positon into a const (see other ImplemntationFor usages).

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,72 @@
...
+
+      var handler = new BocListListValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      Assert.That(context.GetCompoundedErrorMessage(), Is.EqualTo("Error #1\r\nError #2\r\n"));
+    }
+
+    [Test]
+    public void HandleValidationFailures_WithoutValidationFailures_ContextContainsEmptyMessage ()
```
MichaelKetting, line 55: It's okay to keep it. Might be overkill given the implementation.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,72 @@
...
+  [TestFixture]
+  public class BocListListValidationFailureHandlerTest
+  {
+    [Test]
+    public void HandleValidationFailures_WithValidationFailures_ContextContainsConcatenatedMessages ()
+    {
+      var bocListStub = new Mock<IBocList>();
+      var repository = new BocListValidationFailureRepository();
+
+      bocListStub.Setup(_ => _.ValidationFailureRepository).Returns(repository);
```
MichaelKetting, line 35: you could move that one into the Setup since it's needed every time and is just test overhead.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,72 @@
...
+
+      Assert.That(context.GetCompoundedErrorMessage(), Is.EqualTo("Error #1\r\nError #2\r\n"));
+    }
+
+    [Test]
+    public void HandleValidationFailures_WithoutValidationFailures_ContextContainsEmptyMessage ()
+    {
+      var bocListStub = new Mock<IBocList>();
+
+      var repository = new BocListValidationFailureRepository();
```
MichaelKetting, line 59:
```suggestion
      var bocListStub = new Mock<IBocList>();
      var repository = new BocListValidationFailureRepository();
```

Symmetry with other test.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListRowAndCellValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,78 @@
...
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class BocListRowAndCellValidationFailureHandlerTest
+  {
+    [Test]
+    public void HandleValidationFailures_WithoutValidationFailures_ContextContainsEmptyMessage ()
```
MichaelKetting, line 31: here the empty test is indeed relevant.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListRowAndCellValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,78 @@
...
+      var validationFailures = new[]
+                               {
+                                     BusinessObjectValidationFailure.Create("Oh no such an error."),
+                                     BusinessObjectValidationFailure.Create("What, another one?")
+                               };
+
+      repository.AddValidationFailuresForBocList(validationFailures);
+
+      var errorMessage = "Error message from resource manager.";
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.ValidationFailuresFoundInOtherListPagesErrorMessage", out errorMessage)).Returns(true);
```
MichaelKetting, line 67: is this still within the line length limit?

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -37,10 +37,22 @@ public void RefreshErrorMessage_WithoutValidationFailures_ReturnsMessageFromReso
       validationFailureRepoStub
           .Setup(_ => _.GetUnhandledValidationFailuresForBocList(true))
           .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
+      var sequence = new MockSequence();
```
MichaelKetting, line 41: We have completely removed MockSequence because it is hard to work with (it hides calls etc).

For testing ordering, we use VerifiableSequence. In case when there's multiple calls with the same arguments, we use 
`mock.SetupSequence(mock => mock.Read())
          .Returns(true)
          .Returns(false);
`

Here, you can't assert ordering between calls, though.

From reading the tests, I think the SetupSequence/Returns semantic is what you want.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
```diff
@@ -231,30 +242,20 @@ private IBocList GetControlToValidate ()
...
-
-        errorMessageBuilder.AppendLine(remainingValidationFailuresText);
+        errorMessageBuilder.Append(bocList.GetResourceManager().GetString(BocList.ResourceIdentifier.ValidationFailuresFoundInListErrorMessage));
       }
 
-      if (errorMessageBuilder.Length > 0)
-        ErrorMessage = errorMessageBuilder.ToString();
-      else
-        ErrorMessage = GetControlToValidate().GetResourceManager().GetString(BocList.ResourceIdentifier.ValidationFailuresFoundInListErrorMessage);
+      ErrorMessage = errorMessageBuilder.ToString();
```
MichaelKetting, line 258: I think you have changed the semantic here? Do you already have tests for the Refresh error message method? 

Before, `ValidationFailuresFoundInListErrorMessage` was only appended if there where no other messages in the result.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
MichaelKetting, line 40: After reading the rest of the PR, in particular the usage pattern of this method:
I think storing the messages in a List<string>, passing the StringBuilder into an `AppendErrorMessages(StringBuilder)` method would be better. That way we do not need to copy the strings twice and also have less memory allocations.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListValidationResultDispatchingValidator.cs
Frazerus: @a-ctor: https://github.com/re-motion/Framework/pull/764#discussion_r1266491136

I think both versions can be correct, I honestly don't know which one we should go for.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+  /// <seealso cref="IBocListValidationFailureHandler"/>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = c_position)]
+  public class BocListListValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    private const int c_position = -100;
+
+    public BocListListValidationFailureHandler ()
+    {
+    }
+    /// <summary>
```
MichaelKetting, line 36:
```suggestion
    public BocListListValidationFailureHandler ()
    {
    }

    /// <summary>
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  /// <summary>
+  /// Handles all validation failures for the <see cref="IBusinessObjectProperty"/> of the <see cref="IBocList"/>. Should always be the first handler to run.
+  /// </summary>
+  /// <seealso cref="IBocListValidationFailureHandler"/>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = c_position)]
+  public class BocListListValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    private const int c_position = -100;
```
MichaelKetting, line 31:
```suggestion
  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = Position)]
  public class BocListListValidationFailureHandler : IBocListValidationFailureHandler
  {
    public const int Position = -100;
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public BocListListValidationFailureHandler ()
+    {
+    }
+    /// <summary>
+    ///   Handles all unhandled validation failures on the <see cref="IBusinessObjectProperty"/> of the <see cref="IBocList"/>.
+    ///   Reports all found failures as separate messages to the <see cref="ValidationFailureHandlingContext"/>.
+    /// </summary>
+    /// <seealso cref="IBocListValidationFailureHandler.HandleValidationFailures"/>
```
MichaelKetting, line 40:
```suggestion
```
That's usally just the implementation types we should mention when declaring an interface so it's easier to navigate

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  /// <summary>
+  /// Handles all validation failures for the <see cref="IBusinessObjectProperty"/> of the <see cref="IBocList"/>. Should always be the first handler to run.
+  /// </summary>
+  /// <seealso cref="IBocListValidationFailureHandler"/>
```
MichaelKetting, line 27:
```suggestion
```
Just for derived types that are important, not for basetypes.

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
```
MichaelKetting, line 18:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System.Linq;
+using Remotion.Globalization;
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
```
MichaelKetting, line 20:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  /// <summary>
+  /// Handles all still unhandled validation failures and should always be the last handler to run.
+  /// </summary>
+  /// <seealso cref="IBocListValidationFailureHandler"/>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = c_position)]
```
MichaelKetting, line 30:
```suggestion
  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = Position)]
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  /// <summary>
+  /// Handles all still unhandled validation failures and should always be the last handler to run.
+  /// </summary>
+  /// <seealso cref="IBocListValidationFailureHandler"/>
```
MichaelKetting, line 29:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Multiple, Position = c_position)]
+  public class BocListRowAndCellValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    private const int c_position = 100;
```
MichaelKetting, line 33:
```suggestion
    public const int Position = 100;
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Handles all still unhandled validation failures by reporting a single message to the <see cref="ValidationFailureHandlingContext"/>.
+    /// </summary>
+    /// <remarks>
+    ///   It is expected that all unhandled validation failures at this point are unhandled due to them being on a different page of the <see cref="IBocList"/>.
+    /// </remarks>
+    /// <seealso cref="IBocListValidationFailureHandler.HandleValidationFailures"/>
```
MichaelKetting, line 45:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+  /// <summary>
+  /// The <see cref="IBocListValidationFailureHandler"/> interface provides functionality to handle validation failures.
+  /// </summary>
+  /// <remarks>
+  ///   <note type="inotes">
+  ///     New inheritors of the <see cref="IBocListValidationFailureHandler"/> interface must be registered as seen in <see cref="BocListListValidationFailureHandler"/>.
+  ///     The position should be between -100 and 100.
+  ///   </note>
+  /// </remarks>
+  public interface IBocListValidationFailureHandler
```
MichaelKetting, line 31: Here we need the `seealso` xml docs

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  /// <summary>
+  /// The <see cref="IBocListValidationFailureHandler"/> interface provides functionality to handle validation failures.
+  /// </summary>
+  /// <remarks>
+  ///   <note type="inotes">
+  ///     New inheritors of the <see cref="IBocListValidationFailureHandler"/> interface must be registered as seen in <see cref="BocListListValidationFailureHandler"/>.
+  ///     The position should be between -100 and 100.
+  ///   </note>
```
MichaelKetting, line 29: When going for "this must be in between, we need to leave more room, i.e. go to +/- 10e9. If we don't really specify it in detail, we can also leave it for now.

With my doc-request I was actually not referring to documenting how to register a service, but what sort of semantic should be impelmented when using the extension point (= what should I do when I implement the method?)

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
MichaelKetting, line 31: Since we already have a BocListImplementation/Validation namespace, I think we should move everything but the BocListDispatchingValidator over there.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,70 @@
...
+  /// </summary>
+  public class ValidationFailureHandlingContext
+  {
+    /// <summary>
+    /// The <see cref="IBocList"/> to be handled in this context.
+    /// </summary>
+    public IBocList BocList { get; }
+
+    /// <summary>
+    /// The <see cref="IBocListValidationFailureRepository"/> to be used in this context. Stems from the <see cref="IBocList"/>.
```
MichaelKetting, line 37:
```suggestion
    /// The <see cref="IBocListValidationFailureRepository"/> to be used in this context.
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,70 @@
...
+      ArgumentUtility.CheckNotNull("bocList", bocList);
+
+      BocList = bocList;
+      _errorMessages = new List<string>();
+    }
+
+    /// <summary>
+    /// Appends all reported error messages to the <see cref="StringBuilder"/>
+    /// </summary>
+    public void AppendErrorMessages (StringBuilder stringBuilder) => _errorMessages.ForEach(e => stringBuilder.AppendLine(e));
```
MichaelKetting, line 54: argument null check.

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// Reports an error message to this context. Will trim the whitespace of the message.
+    /// </summary>
+    /// <param name="errorMessage">Must not be <see langword="null"/> or empty.</param>
+    public void ReportErrorMessage (string errorMessage)
+    {
+      ArgumentUtility.CheckNotNullOrEmpty("errorMessage", errorMessage);
+
+      if (string.IsNullOrWhiteSpace(errorMessage))
+        return;
+
```
MichaelKetting, line 66:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,70 @@
...
+      _errorMessages.Add(errorMessage.Trim());
```
MichaelKetting, line 67:
```suggestion
      _errorMessages.Add(errorMessage);
```
There is no good way to both disallow empty and at the same time, trim the message. this way is at least consistent.

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+  /// </remarks>
+  public interface IBocListValidationFailureHandler
+  {
+    /// <summary>
+    /// Handles validation failures.
+    /// </summary>
+    /// <param name="context">
+    ///   Contains the <see cref="IBocList"/> to handle validation failures for.
+    ///   Should contain the finished error output after failure handling.
+    /// </param>
```
MichaelKetting, line 39:
```suggestion
```suggestion
    /// <remarks>
    ///   Use <paramref name="context"> to access the <see cref="IBocList"/> and <see cref="IBocListValidationFailureRepository" /> and for reporting the relevant validation failures.
    /// </remarks>
```
```

File: Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundBocListValidationFailureHandler.cs [hunk 46]
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.Validation
+{
+  /// <summary>
+  /// Aggregates a set of <see cref="IBocListValidationFailureHandler"/> instances. When performing validation failure handling all handlers are executed.
+  /// </summary>
+  /// <seealso cref="IBocListValidationFailureHandler"/>
```
MichaelKetting, line 27:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/IBocListValidationFailureHandler.cs
MichaelKetting, line 31:
```suggestion
  /// <seealso cref="BocListListValidationFailureHandler"/>
  /// <seealso cref="BocListRowAndCellValidationFailureHandler"/>
  public interface IBocListValidationFailureHandler
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+    public IEnumerable<IBocListValidationFailureHandler> ValidationFailureHandlers { get; }
+
+    public CompoundBocListValidationFailureHandler (IEnumerable<IBocListValidationFailureHandler> validationFailureHandlers)
+    {
+      ArgumentUtility.CheckNotNull("validationFailureHandlers", validationFailureHandlers);
+
+      ValidationFailureHandlers = validationFailureHandlers;
+    }
+
+    /// <seealso cref="IBocListValidationFailureHandler.HandleValidationFailures"/>
```
MichaelKetting, line 40:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/CompoundBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Compound)]
+  public class CompoundBocListValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    public IEnumerable<IBocListValidationFailureHandler> ValidationFailureHandlers { get; }
+
+    public CompoundBocListValidationFailureHandler (IEnumerable<IBocListValidationFailureHandler> validationFailureHandlers)
+    {
+      ArgumentUtility.CheckNotNull("validationFailureHandlers", validationFailureHandlers);
+
+      ValidationFailureHandlers = validationFailureHandlers;
```
MichaelKetting, line 37: You need to expand the sequence. Otherwise you'd evaluate during each invocation.
```suggestion
      ValidationFailureHandlers = validationFailureHandlers.ToArray();
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
```diff
@@ -0,0 +1,90 @@
...
+    private Mock<IBocList> _bocListStub;
+
+    [SetUp]
+    public void Setup ()
+    {
+      _bocListStub = new Mock<IBocList>();
+    }
+
+    [Test]
+    public void ReportErrorMessage_WithMessages_ConcatenatesMessages ()
```
MichaelKetting, line 39:
```suggestion
    public void AppendErrorMessages_AppendsReportedMessagesToStringBuilder ()
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ValidationFailureHandlingContextTest.cs
```diff
@@ -0,0 +1,90 @@
...
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      context.ReportErrorMessage("First error message.");
+      context.ReportErrorMessage("Second error message.");
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo("First error message.\r\nSecond error message.\r\n"));
+    }
+
```
MichaelKetting, line 50:
```suggestion

    [Test]
    public void AppendErrorMessages_WithExistingContentInStringBuilder_AppendsNewLineAndReportedMessagesToStringBuilder()
    {
      var context = new ValidationFailureHandlingContext(_bocListStub.Object);

      context.ReportErrorMessage("First error message.");
      context.ReportErrorMessage("Second error message.");

      var stringBuilder = new StringBuilder();
      stringBuilder.Append("Existing message without newline.");
      context.AppendErrorMessages(stringBuilder);
      Assert.That(stringBuilder.ToString(), Is.EqualTo("Existing message without newline.\r\nFirst error message.\r\nSecond error message.\r\n"));
    }

    [Test]
    public void AppendErrorMessages_WithTrailingNewLineInStringBuilder_AppendsReportedMessagesToStringBuilder()
    {
      var context = new ValidationFailureHandlingContext(_bocListStub.Object);

      context.ReportErrorMessage("First error message.");
      context.ReportErrorMessage("Second error message.");

      var stringBuilder = new StringBuilder();
      stringBuilder.AppendLine("Existing message with newline.");
      context.AppendErrorMessages(stringBuilder);
      Assert.That(stringBuilder.ToString(), Is.EqualTo("Existing message with newline.\r\nFirst error message.\r\nSecond error message.\r\n"));
    }
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/CompoundBocListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,37 @@
...
+          .Verifiable();
+      handler2
+          .InVerifiableSequence(sequence)
+          .Setup(_ => _.HandleValidationFailures(context))
+          .Verifiable();
+
+      validationFailureHandler.HandleValidationFailures(context);
+
+      handler1.Verify(_ => _.HandleValidationFailures(context), Times.Once());
+      handler2.Verify(_ => _.HandleValidationFailures(context), Times.Once());
```
MichaelKetting, line 33:
```suggestion
      handler1.Verify();
      handler2.Verify();
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/CompoundBocListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class CompoundBocListValidationFailureHandlerTest
+  {
+    [Test]
+    public void HandleValidationFailures_CallsContainedHandlers ()
+    {
+      var handler1 = new Mock<IBocListValidationFailureHandler>();
+      var handler2 = new Mock<IBocListValidationFailureHandler>();
```
MichaelKetting, line 15:
```suggestion
      var handler1Mock = new Mock<IBocListValidationFailureHandler>();
      var handler2Mock = new Mock<IBocListValidationFailureHandler>();
```
(sadly we don't already suffix "mock" or "stub", but we should)

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -149,8 +205,21 @@ public void RefreshErrorMessage_WithValidationFailuresForBoundPropertyAndRowFail
...
+      validationFailureRepoStub
+          .InSequence(sequence)
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(validationFailuresForBocListRows);
+      validationFailureRepoStub
+          .InSequence(sequence)
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
```
MichaelKetting, line 221: Please change to use SetupSequence and multiple returns isntead. For ease of maintenance, I want to avoid `MockSquence` in the codebase.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -107,9 +117,17 @@ public void RefreshErrorMessage_WithValidationFailuresForBoundProperty_ReturnsLo
...
+      var sequence = new MockSequence();
       validationFailureRepoStub
+          .InSequence(sequence)
           .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
           .Returns(validationFailuresForBocListRows);
+      validationFailureRepoStub
+          .InSequence(sequence)
+          .Setup(_ => _.GetUnhandledValidationFailuresForBocListAndContainingDataRowsAndDataCells(true))
+          .Returns(Array.Empty<BocListValidationFailureWithLocationInformation>());
+
```
MichaelKetting, line 130: See other comment on MockSequence

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListRowAndCellValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,89 @@
...
+
+      var context = new ValidationFailureHandlingContext(_bocListStub.Object);
+
+      var handler = new BocListRowAndCellValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo(""));
```
MichaelKetting, line 54:
```suggestion
      Assert.That(stringBuilder.ToString(), Is.Empty);
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListListValidationFailureHandlerTest.cs
```diff
@@ -0,0 +1,77 @@
...
+
+      var context = new ValidationFailureHandlingContext(bocListStub.Object);
+
+      var handler = new BocListListValidationFailureHandler();
+
+      handler.HandleValidationFailures(context);
+
+      var stringBuilder = new StringBuilder();
+      context.AppendErrorMessages(stringBuilder);
+      Assert.That(stringBuilder.ToString(), Is.EqualTo(""));
```
MichaelKetting, line 74:
```suggestion
      Assert.That(stringBuilder.ToString(), Is.Empty);
```

Remotion/ObjectBinding/Web/UI/Controls/Validation/BocListListValidationFailureHandler.cs
Frazerus: Why do we need them public?

MichaelKetting: So other users can define their position relative to ours.

MichaelKetting: OtherImpl{ const int Position = BaseImpl.Position + 10; }

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System;
+using Remotion.ServiceLocation;
```
MichaelKetting, line 19:
```suggestion
using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
using Remotion.ServiceLocation;
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+using Remotion.Utilities;
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation
+{
+  /// <summary>
+  /// Handles all validation failures for the <see cref="IBusinessObjectProperty"/> of the
+  /// <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>. Should always be the first handler to run.
```
MichaelKetting, line 26:
```suggestion
  /// <see cref="IBocList"/>. Should always be the first handler to run.
```
The T-syntax doesn't have compiler checks and is longer

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+  {
+    public const int Position = -1_000_000_000;
+
+    public BocListListValidationFailureHandler ()
+    {
+    }
+
+    /// <summary>
+    ///   Handles all unhandled validation failures on the <see cref="IBusinessObjectProperty"/>
+    ///   of the <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>.
```
MichaelKetting, line 39:
```suggestion
    ///   of the <see cref="IBocList"/>.
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+    public BocListRowAndCellValidationFailureHandler ()
+    {
+    }
+
+    /// <summary>
+    /// Handles all still unhandled validation failures by reporting a single message to the <see cref="ValidationFailureHandlingContext"/>.
+    /// </summary>
+    /// <remarks>
+    ///   It is expected that all unhandled validation failures at this point are unhandled due to them being on a different page of the
+    ///   <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>.
```
MichaelKetting, line 43:
```suggestion
    ///   <see cref="IBocList"/>.
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListRowAndCellValidationFailureHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+using System;
+using System.Linq;
+using Remotion.Globalization;
+using Remotion.ServiceLocation;
```
MichaelKetting, line 21:
```suggestion
using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
using Remotion.ServiceLocation;
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/CompoundBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Validation
+{
+  /// <summary>
+  /// Aggregates a set of <see cref="IBocListValidationFailureHandler"/> instances. When performing validation failure handling all handlers are executed.
+  /// </summary>
+  [ImplementationFor(typeof(IBocListValidationFailureHandler), Lifetime = LifetimeKind.Singleton, RegistrationType = RegistrationType.Compound)]
+  public class CompoundBocListValidationFailureHandler : IBocListValidationFailureHandler
+  {
+    public IBocListValidationFailureHandler[] ValidationFailureHandlers { get; }
```
MichaelKetting, line 32:
```suggestion
    public IReadOnlyCollection<IBocListValidationFailureHandler> ValidationFailureHandlers { get; }
```
We try to never expose an actually array type these days. It is not imutable. Yes, you can still hard-cast, but that's already pretty low-level and not worth the effort to prevent for most cases.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /// <summary>
+    /// Appends all reported error messages to the <see cref="StringBuilder"/>
+    /// </summary>
+    public void AppendErrorMessages (StringBuilder stringBuilder)
+    {
+      ArgumentUtility.CheckNotNull("stringBuilder", stringBuilder);
+
+      if (stringBuilder.Length > 0 && !stringBuilder.ToString().EndsWith("\r\n"))
```
MichaelKetting, line 59:
```suggestion
      if (stringBuilder.Length > 0 && !stringBuilder.ToString().EndsWith("\n"))
```
Newline Linux vs Windows. Detecting "\r\n" could result in weird special cases.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/ValidationFailureHandlingContext.cs
```diff
@@ -0,0 +1,76 @@
...
+        stringBuilder.AppendLine();
+
+      _errorMessages.ForEach(e => stringBuilder.AppendLine(e));
+    }
+
+    /// <summary>
+    /// Reports an error message to this context. Will trim the whitespace of the message.
```
MichaelKetting, line 66:
```suggestion
    /// Reports an error message to this context.
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+  /// <seealso cref="BocListListValidationFailureHandler"/>
+  /// <seealso cref="BocListRowAndCellValidationFailureHandler"/>
+  public interface IBocListValidationFailureHandler
+  {
+    /// <summary>
+    /// Handles validation failures.
+    /// </summary>
+    /// <remarks>
+    ///   Implementations of this method should handle some part of the validation failures belonging to the
+    ///   <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/> found in its
```
MichaelKetting, line 39:
```suggestion
    ///   <see cref="IBocList"/> found in its
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// using NUnit.Framework;
+//
+
+using System;
+using Remotion.ObjectBinding.Web.UI.Controls.Validation;
```
MichaelKetting, line 20:
```suggestion
using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
using Remotion.ObjectBinding.Web.UI.Controls.Validation;
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+    ///   Implementations of this method should handle some part of the validation failures belonging to the
+    ///   <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/> found in its
+    ///   <see cref="IBocListValidationFailureRepository"/>.
+    ///
+    ///   Use <paramref name="context"/> to access the <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>
+    ///   and <see cref="IBocListValidationFailureRepository" /> and for reporting the relevant validation failures.
+    ///
+    ///   All unhandled failures will be handled by the <see cref="BocListRowAndCellValidationFailureHandler"/>,
+    ///   unless it has been removed.
+    ///   There will be no <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>
```
MichaelKetting, line 47:
```suggestion
    ///   There will be no <see cref="IBocList"/>
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+  {
+    /// <summary>
+    /// Handles validation failures.
+    /// </summary>
+    /// <remarks>
+    ///   Implementations of this method should handle some part of the validation failures belonging to the
+    ///   <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/> found in its
+    ///   <see cref="IBocListValidationFailureRepository"/>.
+    ///
+    ///   Use <paramref name="context"/> to access the <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>
```
MichaelKetting, line 42:
```suggestion
    ///   Use <paramref name="context"/> to access the <see cref="IBocList"/>
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+  public interface IBocListValidationFailureHandler
+  {
+    /// <summary>
+    /// Handles validation failures.
+    /// </summary>
+    /// <remarks>
+    ///   Implementations of this method should handle some part of the validation failures belonging to the
+    ///   <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/> found in its
+    ///   <see cref="IBocListValidationFailureRepository"/>.
+    ///
```
MichaelKetting, line 41: you need to add `<para>...</para>` around paragraphs or insert a `<br/>` in between if you want linebreaks.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+    ///   Use <paramref name="context"/> to access the <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>
+    ///   and <see cref="IBocListValidationFailureRepository" /> and for reporting the relevant validation failures.
+    ///
+    ///   All unhandled failures will be handled by the <see cref="BocListRowAndCellValidationFailureHandler"/>,
+    ///   unless it has been removed.
+    ///   There will be no <see cref="T:Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.IBocList"/>
+    ///   business object property validation failures in the <see cref="IBocListValidationFailureRepository"/>
+    ///   when calling this method on your implementation,
+    ///   unless the <see cref="BocListListValidationFailureHandler"/> has been removed or your implementation is out of order in the dependency container.
```
MichaelKetting, line 50: Not sure about the meaning of this last sentence.

still incorrect, but not the reason I did not understand it:
"out of order" does not mean "incorrect sequence" but means "the machine does not work"

Maybe you can say it in a posible way: For there to be validation failures, you need to...?

MichaelKetting (changes requested):
Some minor tweaks and one doc-line that needs a quick session

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/BocListRowAndCellValidationFailureHandler.cs
Frazerus: This isn't required anymore as we moved these classes.

MichaelKetting: ah, right. yes, parent namespace.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Validation/IBocListValidationFailureHandler.cs
Frazerus: For the <see cref="IBocListValidationFailureRepository"/> to contain any <see cref="IBocList"/> business object property validation failures your implementation has to be called before the <see cref="BocListListValidationFailureHandler"/> (provided the service has not been removed).

MichaelKetting (approved):

a-ctor (approved):

PR #765, opened 2023-07-14
RM-8886 Fix incorrect sizing in BocList validation rows

a-ctor:
https://re-motion.atlassian.net/browse/RM-8886

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
@@ -534,11 +534,10 @@ class BocList
 
     const width = LayoutUtility.GetWidth(scrollableContainer);
 
-    const overflowContainer: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow div');
-    if (overflowContainer)
-    {
-      overflowContainer.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
-    }
+    const overflowContainers = [...scrollableContainer.querySelectorAll<HTMLElement>(':scope tr.bocListValidationRow > td > div')];
+    overflowContainers.forEach(e => {
+      e.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
+    });
   }
 
   private static CreateFakeTableHead(tableContainer: HTMLElement, scrollableContainer: HTMLElement, bocList: HTMLElement): void


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -534,11 +534,10 @@ class BocList
 
     const width = LayoutUtility.GetWidth(scrollableContainer);
 
-    const overflowContainer: Nullable<HTMLElement> = scrollableContainer.querySelector('div.bocListTableBlock tr.bocListValidationRow div');
-    if (overflowContainer)
-    {
-      overflowContainer.style.width = "calc(" + width.toString() + "px - var(--remotion-themed-scrollbar-width))";
-    }
+    const overflowContainers = [...scrollableContainer.querySelectorAll<HTMLElement>('div.bocListTableBlock tr.bocListValidationRow div')];
```
MichaelKetting, line 537: I realized the selector is written incorrectly. Right now, it selects "some" element on the page (`div.bocListTableBlock` is actually a parent element of `scrollableContainer`).

Second, right now, both `div` are selected which I believe is also incorrect?

I believe you need to use the following (please verify solution)
```
scrollableContainer.querySelectorAll<HTMLElement>(':scope tr.bocListValidationRow > td > div')
```

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: Fun fact: the css selector in `querySelector`/`querySelectorAll` need not be relative to the scope. The results are filtered to descendants anyway so `tr.bocListValidationRow > td > div` would also work as the `:scope ` is implied by the method. I left it in for clarity but it would not be necessary :D

MichaelKetting: Thanks, which now also poses the question how this could have ever worked because in my mental image, the div.bocListTableBlock is not nested in scrollable Container.

MichaelKetting (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: Ok I did not get my point across :D The specified selector is executed in the global scope but the results of this selection are then filtered to the elements that are descendants of the scrollable container.

MichaelKetting: Ah, okay. follow-up question: with ":scope: does it still execute in global scope or in a reduced part of the DOM? (asking for my friend, Intel Celereron 50)

a-ctor: Sadly I am not a CSS execution engine so I can't really answer that. In theory, it should not make a difference if the filtering is done in the CSS selector or afterwards

MichaelKetting: Sadly, I am. The selectors work by going from right to left. So, the more generic the right end, the more CPU it takes. My hope (unconfirmed) is that it would scope the damage to the relevant control tree when starting from a specific part in the DOM.

PR #767, opened 2023-07-18
RM-8885 Fix bug when no min/max values are provided

KuJo7:
https://re-motion.atlassian.net/browse/RM-8885


Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
```diff
@@ -51,6 +52,9 @@ protected override IEnumerable<IPropertyValidator> GetValidators (IPropertyInfor
       ArgumentUtility.CheckNotNull("property", property);
       ArgumentUtility.CheckNotNull("validationMessageFactory", validationMessageFactory);
 
+      if (MinLength == 0 && MaxLength == 0)
+        return Enumerable.Empty<IPropertyValidator>();
```
MichaelKetting, line 56: unit test please.
also, we should integrate this into the IF/ELSE tree. it's part of the remaining rule set and if we had thought of it before, we wouldn't have killed the original ELSE.

MichaelKetting (changes requested):

Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
KuJo7: I hope you mean the IF/ELSE tree in `GetValidators()` as I made the changes now, because if you meant the IF/ELSE tree in `CreateValidator()`, I am not sure if it is necessary to do a null-check and then return empty array in the outer method

MichaelKetting: Yeah, I actually meant `CreateValidator()` but I didn't realize that this would lead to furthor problems. I'll do the actual review tomorrow. Thanks for the feedback on what works and what doesn't!

Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
```diff
@@ -52,7 +53,11 @@ protected override IEnumerable<IPropertyValidator> GetValidators (IPropertyInfor
...
       IPropertyValidator validator;
-      if (string.IsNullOrEmpty(ErrorMessage))
+      if (MinLength == 0 && MaxLength == 0)
+      {
+        return Enumerable.Empty<IPropertyValidator>();
+      }
+      else if (string.IsNullOrEmpty(ErrorMessage))
       {
         validator = PropertyValidatorFactory.Create(
             property,
```
MichaelKetting, line 63: okay, there is no good way to do this inside `CreateValidator` apprently, atleast in the existing code structure. Because of the way PropertyValidatorFactory works, there is also no way to change this without introducing a NullValidator object that is then filtered. 

Basically, this would look like this:
CreateValidator returns a NullValidator (i.e. a null object) for Min==Max==0 and in GetValidators we check if the result is the NullValidator and return Enumerable.empty or Enumerable.Singleton based on this result check.

Advantage of that solution:
We only have one place where we decide how to handle Min/Max Length. Is it worth the back and forth? Perhaps. I think we should take a moment and do a separate version of the code with the alternative implementation. that way, we can see how it feels like.

Should we ever decide to also gracefully handle other incorrect configurations, this might actually be necessary.


Code style for the current solution:
please make the IF as an exit condition because it has nothing to do with the original IF, but rather cuts out before things get interesting.

MichaelKetting (changes requested):

Remotion/Validation/Core/Attributes/Validation/LengthValidationAttribute.cs
KuJo7: I guess for now we want to revert it to my first apporach d92963a0fae07f887e622052d8e6e7429f831db3. Do we want to implement the other version of the code in this pr or should I create a ticket for this?

MichaelKetting: > I guess for now we want to revert it to my first apporach https://github.com/re-motion/Framework/commit/d92963a0fae07f887e622052d8e6e7429f831db3.
Yep, sounds good. checked the old code and yeah, I though it was part of the helper method not the "outer" method.

> Do we want to implement the other version of the code in this pr or should I create a ticket for this?
just do a quick demo in the current PR based on a copy of the current file. it's just so we can check out the code/discuss and then decide which to keep.

MichaelKetting (approved):

PR #768, opened 2023-07-19
RM-8875 Correct BocList validation localization

Frazerus:
https://re-motion.atlassian.net/browse/RM-8875

### DIFF
--- a/Remotion/ObjectBinding/Web/Globalization/BocList.de.resx
+++ b/Remotion/ObjectBinding/Web/Globalization/BocList.de.resx
@@ -124,7 +124,7 @@
     <value>Ansicht</value>
   </data>
   <data name="Remotion.ObjectBinding.Web.UI.Controls.BocList.RowEditModeErrorMessage" xml:space="preserve">
-    <value>In der aktuellen Zeile sind Fehler aufgetreten.</value>
+    <value>Ungültige Eingabe in der aktuellen Zeile.</value>
   </data>
   <data name="Remotion.ObjectBinding.Web.UI.Controls.BocList.RequiredFieldTitle" xml:space="preserve">
     <value>Eingabe erforderlich in der aktuellen Zeile</value>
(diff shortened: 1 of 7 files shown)


Remotion/ObjectBinding/Web/Globalization/BocList.de.resx
```diff
@@ -124,7 +124,7 @@
     <value>Ansicht</value>
   </data>
   <data name="Remotion.ObjectBinding.Web.UI.Controls.BocList.RowEditModeErrorMessage" xml:space="preserve">
-    <value>In der aktuellen Zeile sind Fehler aufgetreten.</value>
+      <value>Ungültige Eingabe in der aktuellen Zeile</value>
```
MichaelKetting, line 127:
```suggestion
    <value>Ungültige Eingabe in der aktuellen Zeile</value>
```

pleas echeck your resx. soemthing changed the indent on some of the entries. since I can't say if that's a setting on your system, just because the changes where made via Rider-Localization UI or due to some other unknown reason, it's best to fix them since there is no expectation that other developers wouldn't revert them to the original two-whitespace indent.

Remotion/ObjectBinding/Web/Globalization/BocList.fr.resx
```diff
@@ -174,13 +174,13 @@
   <data name="Remotion.ObjectBinding.Web.UI.Controls.BocList.ValidationFailuresFoundInOtherListPagesErrorMessage" xml:space="preserve">
     <value>Des erreurs sont survenues sur au moins une autre page.</value>
```
MichaelKetting, line 175: please check the wording for FR and IT regarding error/invalid input. It seems like the correct wording is "Saisie non valable" / "Input non valido" (correctly adjusted for grammar). Doku should do a final check here before we're done.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #769, opened 2023-07-19
Feature/rm 8716 dispatching a validation failure to an invisible control should not mark it as handled

Frazerus:
https://re-motion.atlassian.net/browse/RM-8716

### DIFF
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
@@ -21,8 +21,10 @@ using Moq;
 using NUnit.Framework;
 using Remotion.Globalization;
 using Remotion.ObjectBinding.Validation;
+using Remotion.ObjectBinding.Web.UI.Controls;
 using Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation;
 using Remotion.ObjectBinding.Web.UI.Controls.Validation;
+using Remotion.ObjectBinding.Web.UnitTests.Domain;
 using Remotion.Web.UI.Controls;
 
 namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
@@ -30,6 +32,36 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
   [TestFixture]
   public class BocListValidationResultDispatchingValidatorTest : BocTest
   {
+    [Test]
+    public void DispatchValidationFailures_WithInvisibleControl_UsesReadOnlyValidationResultForDispatching ()
+    {
+      IBusinessObject businessObject = TypeWithReference.Create();
+      var businessObjectProperty = businessObject.BusinessObjectClass.GetPropertyDefinition("ReferenceList");
+
+      var bocList = new BocListMock();
+      bocList.ID = "someBocList";
+      bocList.DataSource = new StubDataSource(businessObject.BusinessObjectClass);
+      bocList.DataSource.BusinessObject = businessObject;
+      bocList.Property = (IBusinessObjectReferenceProperty)businessObjectProperty;
+      bocList.Visible = false;
+      bocList.FixedColumns.Add(new BocValidationErrorIndicatorColumnDefinition());
+
+      var validationResultMock = new Mock<IBusinessObjectValidationResult>(MockBehavior.Strict);
+      validationResultMock.Setup(e => e.GetValidationFailures(businessObject, businessObjectProperty, false))
+          .Returns(Array.Empty<BusinessObjectValidationFailure>())
+          .Verifiable();
+
+      var validator = new BocListValidationResultDispatchingValidator();
+      validator.ControlToValidate = "someBocList";
+
+      NamingContainer.Controls.Add(bocList);
+      NamingContainer.Controls.Add(validator);
+
+      validator.DispatchValidationFailures(validationResultMock.Object);
+
+      validationResultMock.Verify();
+    }
+
     [Test]
     public void RefreshErrorMessage_WithoutValidationFailures_ReturnsMessageFromResourceManager ()
     {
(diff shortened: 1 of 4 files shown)


Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -240,5 +242,167 @@ public void EvaluateIsValid_WithFailures_ErrorMessageFromResourceManagerAndInval
       Assert.That(validator.ErrorMessage, Is.EqualTo("BocList has validation errors."));
       Assert.That(validator.IsValid, Is.False);
     }
+
+    [Test]
+    public void DispatchValidationFailures_WithVisibleControl_DispatchesValidationFailuresToRepositoryAndMarkAsHandled ()
+    {
+      var validationResultMock = new Mock<IBusinessObjectValidationResult>();
```
a-ctor, line 249: For me this test is too complex to understand if it does what it should. There are so mocks and verifications that make the test hard to read. I am unsure on how to best proceed. Maybe @MichaelKetting can chime in here as well.

We want to test that the dispatching uses `control.Visible` in three places:
1. When using `IValidationFailureMatcher`
2. When querying validation failures for the row object
3. When querying validation failures for the list object

The only difference between the `true` and the `false` test is the value so it might be a good idea to use a test parameter instead of duplicating the test. To reduce the amount of mocks I would introduce stub classes or use existing stub classes.

```c#
[Test]
public void METHOD ([Values(true, false)] bool markAsHandled)
{
  var mock = new Mock<IBusinessObjectValidationResult>(MockBehavior.Strict);
  mock.Setup(e => e.GetValidationFailures(It.IsAny<IBusinessObject>(), It.IsAny<IBusinessObjectProperty>(), markAsHandled));
  mock.Setup(e => e.GetUnhandledValidationFailures(It.IsAny<IBusinessObject>(), It.IsAny<bool>(), markAsHandled));

  var validationFailureMatcherMock = new Mock<IValidationFailureMatcher>(MockBehavior.Strict);
  validationFailureMatcherMock.Setup(e => e.GetMatchingValidationFailures(It.IsAny<IBusinessObject>(), mock.Object, markAsHandled));

  var columnDefinitionMock = new Mock<BocColumnDefinition>();
  columnDefinitionMock.As<IBocColumnDefinitionWithValidationSupport>().Setup(e => e.GetValidationFailureMatcher()).Returns(validationFailureMatcherMock.Object);

  var bocListMock = new BocListMock();
  bocListMock.ID = "someBocList";
  bocListMock.Visible = false;
  bocListMock.FixedColumns.Add(columnDefinitionMock.Object);
  // todo setup Business Object and row objects (i would use a similar setup to ManualBusinessObject)

  var validator = new BocListValidationResultDispatchingValidator();
  validator.ControlToValidate = "someBocList";

  NamingContainer.Controls.Add(bocListMock);
  NamingContainer.Controls.Add(validator);

  validator.DispatchValidationFailures(mock.Object);
  
  // we probably want to assert the three scenarios here to ensure that there were any calls since we only check that there are no wrong calls
}
```

By extracting logic into stub I think we can reduce the complexity of the test and make it more readable.

a-ctor (changes requested):

a-ctor:
As discussed: we use a decorator solution instead to prevent the need to introduce `markAsHandled` everywhere. In `BocListValidationResultDispatchingValidator.DispatchValidationFailures` we wrap the `validationResult` using a "read-only" decorator and use this decorated instances instead. The decorator always provides `markedAsHandled: false` to the inner instance.

a-ctor (approved):

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ReadOnlyBusinessObjectValidationResultDecoratorTest.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
+{
+  [TestFixture]
+  public class ReadOnlyBusinessObjectValidationResultDecoratorTest
+  {
+    [Test]
+    public void GetValidationFailures_AlwaysPassesMarkAsHandledFalse ()
+    {
+      var businessObject = Mock.Of<IBusinessObject>();
+      var businessObjectProperty = Mock.Of<IBusinessObjectProperty>();
```
MichaelKetting, line 32: naming, suffix with `Stub`

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ReadOnlyBusinessObjectValidationResultDecoratorTest.cs
```diff
@@ -0,0 +1,61 @@
...
+
+      var validationResultMock = new Mock<IBusinessObjectValidationResult>(MockBehavior.Strict);
+      validationResultMock.Setup(e => e.GetValidationFailures(businessObject, businessObjectProperty, false))
+          .Returns(Array.Empty<BusinessObjectValidationFailure>())
+          .Verifiable();
+
+      var readOnlyDecorator = new ReadOnlyBusinessObjectValidationResultDecorator(validationResultMock.Object);
+      readOnlyDecorator.GetValidationFailures(businessObject, businessObjectProperty, true);
```
MichaelKetting, line 40: I think we should test that the inner result is passed to the callsite (return collection instance and use `Is.SameAs`). Otherwise, we could return anything from `null` to `Array.Emtpy` and no one would know.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/ReadOnlyBusinessObjectValidationResultDecoratorTest.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+      var businessObject = Mock.Of<IBusinessObject>();
+
+      var validationResultMock = new Mock<IBusinessObjectValidationResult>(MockBehavior.Strict);
+      validationResultMock.Setup(e => e.GetUnhandledValidationFailures(businessObject, true, false))
+          .Returns(Array.Empty<BusinessObjectValidationFailure>())
+          .Verifiable();
+
+      var readOnlyDecorator = new ReadOnlyBusinessObjectValidationResultDecorator(validationResultMock.Object);
+      readOnlyDecorator.GetUnhandledValidationFailures(businessObject, true, true);
```
MichaelKetting, line 56: Same review comments as in `GetValidationFailures`

MichaelKetting (changes requested):

PR #776, opened 2023-07-26
RM-8406 Create a test infrastructure for interface mapping tests

a-ctor:
https://re-motion.atlassian.net/browse/RM-8406


a-ctor:
Stale. We can reopen in the future.

PR #779, opened 2023-07-31
RM-8765 Make Domain-based maximum length validators non-removable

Frazerus:
https://re-motion.atlassian.net/browse/RM-8765


Remotion/Data/DomainObjects.Validation.UnitTests/DomainObjectAttributesBasedValidationPropertyRuleReflectorTests/DomainObject_DOABCPRRTest.cs
```diff
@@ -363,9 +362,7 @@ public void GetMetaValidationRules_NullableStringPropertyAttribute ()
     {
       var result = _propertyWithNullableStringPropertyAttributeReflector.GetMetaValidationRules().ToArray();
 
-      Assert.That(result.Count(), Is.EqualTo(1));
-      Assert.That(result[0], Is.TypeOf(typeof(RemotionMaxLengthPropertyMetaValidationRule)));
-      Assert.That(((RemotionMaxLengthPropertyMetaValidationRule)result[0]).MaxLength, Is.EqualTo(10));
+      Assert.That(result.Any, Is.False);
```
MichaelKetting, line 365:
```suggestion
      Assert.That(result, Is.Empty);
```

Remotion/Data/DomainObjects.Validation.UnitTests/DomainObjectAttributesBasedValidationPropertyRuleReflectorTests/DomainObjectMixin_DOABCPRRTest.cs
```diff
@@ -371,21 +370,9 @@ public void GetRemovablePropertyValidators_MandatoryAttributeAndValidationMessag
...
-          .Returns(validationMessageStub.Object);
-
       var result = _propertyWithNullableStringPropertyAttributeReflector.GetRemovablePropertyValidators().ToArray();
 
-      validationMessageStub.Setup(_ => _.ToString()).Returns("Stub Message");
-      Assert.That(result.Count(), Is.EqualTo(1));
-      Assert.That(result[0], Is.TypeOf(typeof(MaximumLengthValidator)));
-      Assert.That(((MaximumLengthValidator)result[0]).Max, Is.EqualTo(10));
-      Assert.That(((MaximumLengthValidator)result[0]).ValidationMessage.ToString, Is.EqualTo("Stub Message"));
+      Assert.That(result.Any, Is.False);
```
MichaelKetting, line 375:
```suggestion
      Assert.That(result, Is.Empty);
```

Remotion/Data/DomainObjects.Validation.UnitTests/DomainObjectAttributesBasedValidationPropertyRuleReflectorTests/DomainObjectMixin_DOABCPRRTest.cs
```diff
@@ -435,28 +417,51 @@ public void GetRemovablePropertyValidators_MandatoryStringPropertyAndValidationM
...
+      _validationMessageFactoryStub
+          .Setup(
+              _ => _.CreateValidationMessageForPropertyValidator(
+                  It.IsAny<MaximumLengthValidator>(),
+                  PropertyInfoAdapter.Create(_mixinPropertyWithNullableStringPropertyAttribute)))
+          .Returns(lengthValidationMessageStub.Object);
+
       var result = _propertyWithNullableStringPropertyAttributeReflector.GetNonRemovablePropertyValidators().ToArray();
 
-      Assert.That(result.Any(), Is.False);
```
MichaelKetting, line 440: No, I can't tell you why it was written that way. Either because it wasn't possible back than or because it wasn't noticed during review.

Remotion/Data/DomainObjects.Validation.UnitTests/DomainObjectAttributesBasedValidationPropertyRuleReflectorTests/DomainObjectMixin_DOABCPRRTest.cs
```diff
@@ -480,9 +485,7 @@ public void GetMetaValidationRules_NullableStringPropertyAttribute ()
     {
       var result = _propertyWithNullableStringPropertyAttributeReflector.GetMetaValidationRules().ToArray();
 
-      Assert.That(result.Count(), Is.EqualTo(1));
-      Assert.That(result[0], Is.TypeOf(typeof(RemotionMaxLengthPropertyMetaValidationRule)));
-      Assert.That(((RemotionMaxLengthPropertyMetaValidationRule)result[0]).MaxLength, Is.EqualTo(10));
+      Assert.That(result.Any(), Is.False);
```
MichaelKetting, line 488:
```suggestion
      Assert.That(result, Is.Empty);
```

MichaelKetting (approved):
some minor adjustments.

Remotion/Data/DomainObjects.Validation.UnitTests/DomainObjectAttributesBasedValidationPropertyRuleReflectorTests/DomainObject_DOABCPRRTest.cs
Frazerus: All tests in this file use the any -> is.false syntax. I'm not sure that changes anything but I'd like you to know.

MichaelKetting: Ah, I see, there's a couple more in this and the other file. Hmm... no, we can still udpate the current tests. I would also be okay with changing the remaining couple of assertions to Is.Empty in these two files.

Remotion/Data/DomainObjects.Validation.UnitTests/DomainObjectAttributesBasedValidationPropertyRuleReflectorTests/DomainObjectMixin_DOABCPRRTest.cs
MichaelKetting: Just mentioning this: `result.Any` passes a delagate to the Assertion method. Am surprised that the test works and isn't complaining about the mismatch in semantics. Normally, this syntax is only used for exception-semantics and the result isn't relevant. so, what does `Is.False` assert in this instance. So, even if we wouldn't change it to `Is.Empty`, we would still need to update to `result.Any()`

MichaelKetting (approved):
I had to revert a couple of changes in the test domain. The deletions went a bit too far and dropped test examples, too, which just happend to also use "maxlength" as the demonstrated concept. The implemenetations where unrelated, though.

PR #781, opened 2023-08-01
RM-8904 Support for aggregating multiple IPropertyValidator instances…

Frazerus:
… to a single IBusinessObjectPropertyConstraint
https://re-motion.atlassian.net/browse/RM-8904

### DIFF
--- a/Remotion/ObjectBinding/Validation.UnitTests/PropertyValidatorToBusinessObjectPropertyConstraintConverterTest.cs
+++ b/Remotion/ObjectBinding/Validation.UnitTests/PropertyValidatorToBusinessObjectPropertyConstraintConverterTest.cs
@@ -15,7 +15,11 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Linq;
 using NUnit.Framework;
+using Remotion.ObjectBinding.BusinessObjectPropertyConstraints;
+using Remotion.Validation.Implementation;
+using Remotion.Validation.Validators;
 
 namespace Remotion.ObjectBinding.Validation.UnitTests
 {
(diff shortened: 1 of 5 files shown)


Remotion/ObjectBinding/Validation.UnitTests/PropertyValidatorToBusinessObjectPropertyConstraintConverterTest.cs
```diff
@@ -15,17 +15,84 @@
...
+      var converter = new PropertyValidatorToBusinessObjectPropertyConstraintConverter();
+
+      var constraints = converter.Convert(validators).ToArray();
+
+      // I'm not sure how I would go about testing this in a different manner. Any suggestions?
+      // The problem I'm seeing here is that the objects are created elsewhere, therefore I can't use
+      // ... constraints, Is.equivalentTo ...
+      Assert.That(constraints.Length, Is.EqualTo(2));
+      Assert.That(constraints[1], Is.InstanceOf<BusinessObjectPropertyValueLengthConstraint>());
+      Assert.That(constraints[0], Is.InstanceOf<BusinessObjectPropertyValueRequiredConstraint>());
```
KuJo7, line 95:
```suggestion
            Assert.That(constraints, Has.Some.InstanceOf<BusinessObjectPropertyValueLengthConstraint>().And.Some.InstanceOf<BusinessObjectPropertyValueRequiredConstraint>());
```

Remotion/ObjectBinding/Validation.UnitTests/PropertyValidatorToBusinessObjectPropertyConstraintConverterTest.cs
```diff
@@ -15,17 +15,82 @@
...
+                       };
+
+      var converter = new PropertyValidatorToBusinessObjectPropertyConstraintConverter();
+
+      var constraints = converter.Convert(validators).ToArray();
+
+      Assert.That(constraints.Length, Is.EqualTo(2));
+      Assert.That(constraints,
+          Has.Some.InstanceOf<BusinessObjectPropertyValueLengthConstraint>()
+              .And.Some.InstanceOf<BusinessObjectPropertyValueRequiredConstraint>());
```
MichaelKetting, line 93:
```suggestion
      Assert.That(
          constraints,
          Has.Some.InstanceOf<BusinessObjectPropertyValueLengthConstraint>()
              .And.Some.InstanceOf<BusinessObjectPropertyValueRequiredConstraint>());
```

MichaelKetting, line 93: I do like this. Never seen it before. But I like it :)

Remotion/ObjectBinding/Validation.UnitTests/PropertyValidatorToBusinessObjectPropertyConstraintConverterTest.cs
```diff
@@ -15,17 +15,82 @@
...
   {
     [Test]
-    [Ignore("TODO RM-5906")]
-    public void Convert ()
+    public void Convert_WithDifferentMaxLengthValidators_ContainsLowerMaxLength ()
     {
+      var validators = new []
+                       {
+                         new MaximumLengthValidator(7, new InvariantValidationMessage("Max length is 7")),
+                         new MaximumLengthValidator(5, new InvariantValidationMessage("Max length is 5"))
```
MichaelKetting, line 35:
```suggestion
                         new MaximumLengthValidator(7, new InvariantValidationMessage("Max length is 7")),
                         new MaximumLengthValidator(5, new InvariantValidationMessage("Max length is 5")),
                         new MaximumLengthValidator(7, new InvariantValidationMessage("Max length is 6")),
```

For cases such as this, I like to put the expected value in the middle to prevent code changes that simply fall back to the first / last entry when selecting the value. 

A similiar scenario is when I look for multiple values that also have a skip-condition. There, I make sure the skip-condition applies to a middle item and then assert that the first and last item are part of the result. That way, I make sure the skip condition results in a skip and not a loop-break.

MichaelKetting (approved):
Small change, then done.

Remotion/ObjectBinding/Validation.UnitTests/PropertyValidatorToBusinessObjectPropertyConstraintConverterTest.cs
Frazerus: Was done & researched by @KuJo7, I simply put a comment there on how to do it better, he implemented the suggestion.

PR #783, opened 2023-08-02
Feature/rm 8637 setting control required true while control enable optional validators false has no effect beyond rendering the required marker

Frazerus:
https://re-motion.atlassian.net/browse/RM-8637


Remotion/ObjectBinding/Web/UI/Controls/BusinessObjectBoundEditableWebControl.cs
KuJo7, line 1: remove unused `using Remotion.Web.UI.Controls;`

KuJo7 (approved):
Looks fine for me :)

MichaelKetting (changes requested):
There's a couple of simplifications we need to do.

MichaelKetting (approved):
I had some last minute tweaks. Now we just need to split the commits. I think I'll do that quickly.

MichaelKetting (approved):
Rebase is also done now.

PR #786, opened 2023-08-07
RM-8229 Hide overflowing content from BocBooleanValue

a-ctor:
https://re-motion.atlassian.net/browse/RM-8229


MichaelKetting (approved):
I found a more layout compatible solution. please verify. I also included this fix in the other elements that are affected from similiar issues.

a-ctor:
@MichaelKetting That solution does not work as the whole control now overflows the cell:

image

`min-content` also prefers the maximally soft-wrapped version of the children so it might not be correct in some scenarios. It apparently works like we want if the text is wrapped in child elements but if the text were to be a direct child instead of in a span this would be the result for the text `No No`:
image

PR #787, opened 2023-08-07
feature/RM-6055-Support-for-automatic-validation-of-ClientTransaction

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-6055


a-ctor (approved):
Only took 7 years for two LOC :p

PR #788, opened 2023-08-09
RM-8890 Create a script to generate a SBOM

a-ctor:
https://re-motion.atlassian.net/browse/RM-8890

Open questions/todos:
- Do we want to fix the tool versions?
- Integrate into nightly/full build on TC

### DIFF
--- a/Build/create-sbom.ps1
+++ b/Build/create-sbom.ps1
@@ -1,0 +1,102 @@
+﻿param(
+    [string] $TempFolder = "Build\BuildOutput\temp\sbom",
+    [Parameter(mandatory=$true)] [string] $NpmProjectJson,
+    [Parameter(mandatory=$true)] [string] $Solution,
+    [Parameter(mandatory=$true)] [string] $GithubUsername,
+    [Parameter(mandatory=$true)] [string] $GithubAccessToken,
+    [Parameter(mandatory=$true)] [string] $OutputPath
+)
+# temp folder: Use BuildOutput\temp as default -> set on TC
+# check if tools already exist in the temp folder before downloading again
+
+$ErrorActionPreference = "Stop"
+$originalDirectory = Get-Location
+
+trap {
+    Write-Host "Unexpected script exception: $($_.InvocationInfo.ScriptName)@$($_.InvocationInfo.ScriptLineNumber): '$($_.InvocationInfo.Line)'"
+    Write-Host $_
+    cd $originalDirectory
+    exit 1
+}
+
+function Fail-On-Error {
+    param(
+        $description
+    )
+
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Executing '$description' failed with exit code $LASTEXITCODE."
+        exit 1
+    }
+}
+
+# Set .NET current directory to enable correct path resolution
+[System.Environment]::CurrentDirectory = Get-Location
+
+$TempFolder = [System.IO.Path]::GetFullPath($TempFolder)
+[System.IO.Directory]::CreateDirectory($TempFolder)
+Write-Host "Using temp folder: $TempFolder"
+
+$NpmProjectJson = [System.IO.Path]::GetFullPath($NpmProjectJson)
+Write-Host "Using NPM project: $NpmProjectJson"
+
+$Solution = [System.IO.Path]::GetFullPath($Solution)
+Write-Host "Using solution file: $Solution"
+
+$OutputPath = [System.IO.Path]::GetFullPath($OutputPath)
+[System.IO.Directory]::CreateDirectory([System.IO.Path]::GetDirectoryName($OutputPath))
+Write-Host "Using output path: $OutputPath"
+
+cd $TempFolder
+
+Write-Host "# Installing dependencies"
+
+# Download cyclonedx-npm
+Write-Host "Installing cyclonedx-npm"
+npm install --save-dev @cyclonedx/cyclonedx-npm
+
+# Download CycloneDX
+if (Test-Path "dotnet-CycloneDX.exe") {
+    Write-Host "CycloneDX already installed."
+} else {
+    Write-Host "Installing CycloneDX"
+    dotnet tool install CycloneDX --tool-path .
+}
+
+# Download cyclonedx-cli
+if (Test-Path "cyclonedx-cli.exe") {
+    Write-Host "cyclonedx-cli already installed."
+} else {
+    Write-Host "Installing cyclonedx-cli"
+    $releases = Invoke-RestMethod -Uri "https://api.github.com/repos/CycloneDX/cyclonedx-cli/releases" -Headers @{ Authorization="Bearer $GithubAccessToken" }
+    $asset = ($releases | Select-Object -First 1).assets | ? { $_.name -eq "cyclonedx-win-x86.exe" }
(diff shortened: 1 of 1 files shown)


Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
...
+
+Write-Host "Analyzing NPM dependencies"
+npx "@cyclonedx/cyclonedx-npm" $NpmProjectJson --omit dev --output-format xml --output-file npm-dep.xml
+Fail-On-Error "cyclonedx-npm"
+
+Write-Host "Analyzing Nuget dependencies"
+.\dotnet-CycloneDX.exe -o . -gu $GithubUsername -gt $GithubAccessToken -f dotnet-dep.xml $Solution
+Fail-On-Error "dotnet-CycloneDX.exe"
+
+Write-Host "Analyzing Nuget dependencies"
```
Frazerus, line 93: Probably more like "Merging Nuget and NPM dependencies"

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
...
+    Write-Host "Installing cyclonedx-cli"
+    $releases = Invoke-RestMethod -Uri "https://api.github.com/repos/CycloneDX/cyclonedx-cli/releases" -Headers @{ Authorization="Bearer $GithubAccessToken" }
+    $asset = ($releases | Select-Object -First 1).assets | ? { $_.name -eq "cyclonedx-win-x86.exe" }
+    Invoke-WebRequest -Uri ($asset.browser_download_url) -OutFile "cyclonedx-cli.exe" -Headers @{ Authorization="Bearer $GithubAccessToken" }
+}
+
+# Run the analysis
+Write-Host "# Starting analysis"
+
+Write-Host "Ensuring NPM dependencies are up-to-date"
```
Frazerus, line 79: Do they actually need to be up-to-date?

Frazerus (approved):

Build/create-sbom.ps1
a-ctor: The tool requires a package.lock to be there so either that is checked in or created using `npm install`. Doing this again on an already restored application should be fast anyways since we only have our dependency packages

Build/create-sbom.ps1
a-ctor: Copy paste is evil :D

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
...
+Write-Host "Analyzing NPM dependencies"
+npx "@cyclonedx/cyclonedx-npm" $NpmProjectJson --omit dev --output-format xml --output-file npm-dep.xml
+Fail-On-Error "cyclonedx-npm"
+
+Write-Host "Analyzing Nuget dependencies"
+.\dotnet-CycloneDX.exe -o . -gu $GithubUsername -gt $GithubAccessToken -f dotnet-dep.xml $Solution
+Fail-On-Error "dotnet-CycloneDX.exe"
+
+Write-Host "Analyzing Nuget dependencies"
+.\cyclonedx-cli.exe merge --input-files "npm-dep.xml" "dotnet-dep.xml" --output-file $OutputPath
```
MichaelKetting, line 94: JSON output as second/separate file is still missing.

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
+﻿param(
```
MichaelKetting, line 1: >    Do we want to fix the tool versions?

do you mean "fix" as in "lock the version" or "fix" as in "repair"?

If it's "lock", I am inclined do say keep it floating until there is an issue? For most cases I would indeed go with "lock", here it feels like it's overkill? Unless we have a problem with different SDK versions etc.

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
...
+    [string] $TempFolder = "Build\BuildOutput\temp\sbom",
```
MichaelKetting, line 2: > Integrate into nightly/full build on TC

Intgrate in both builds so we always know when there's an issue.

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
...
+    [Parameter(mandatory=$true)] [string] $NpmProjectJson,
+    [Parameter(mandatory=$true)] [string] $Solution,
+    [Parameter(mandatory=$true)] [string] $GithubUsername,
+    [Parameter(mandatory=$true)] [string] $GithubAccessToken,
+    [Parameter(mandatory=$true)] [string] $OutputPath
```
MichaelKetting, line 7: Normally, the file name including version would be provided by build script. now with the tacked-on implementation, we can pass the version and the outputDirectory.

for localbuilds we use 0.0.0 as version and probably THIS as the directory like with the other paths and append "Build\BuildOutput"

In both cases we then append "SBOM" and the filename.

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
+﻿param(
+    [string] $TempFolder = "Build\BuildOutput\temp\sbom",
+    [Parameter(mandatory=$true)] [string] $NpmProjectJson,
```
MichaelKetting, line 3: hardcode it, based on the externally supplied working directory. Working Directory defaults to THIS.

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
...
+}
+
+# Run the analysis
+Write-Host "# Starting analysis"
+
+Write-Host "Ensuring NPM dependencies are up-to-date"
+cd ([System.IO.Path]::GetDirectoryName($NpmProjectJson))
+npm install
+Fail-On-Error "npm install"
+cd $TempFolder
```
MichaelKetting, line 83: I guess it's okay to change the directory, it just feels strange if the tool can be provided with paths based on the directory to be used, e.g. "{tempdirectoy}\npm-dep.xml"

MichaelKetting (changes requested):

Build/create-sbom.ps1
MichaelKetting: Edit/Update: just reread the spec. okay, we only said we use XML, not JSON. could you please check if the dependencytrack can also process XML, if so, there is no need for json

a-ctor: Examples in their documentation do include xml data as payload so I assume that they do. I didn't find a concrete statement though.

Build/create-sbom.ps1
a-ctor: I tried :(

Build/create-sbom.ps1
a-ctor: Why do we need to include the version info here? Right now, this is a problem that the caller has to solve and I think it is better this way instead of concatenating in the script. On TC it should be easy to put the version into the output file name. The script itself is not integrated into the local build so in the rare case we need to do a local built you need to pass the values yoursef anyway.

Build/create-sbom.ps1
```diff
@@ -0,0 +1,97 @@
+﻿param(
+    [string] $TempFolder = "Build\BuildOutput\temp\sbom",
+    [Parameter(mandatory=$true)] [string] $NpmProjectJson,
+    [Parameter(mandatory=$true)] [string] $Solution,
```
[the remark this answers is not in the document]
a-ctor: What do you mean by working directory defaults to this?

MichaelKetting (approved):

PR #789, opened 2023-08-09
feature/RM-8889-create-npm-package-inside-re-motion

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-8889

### DIFF
--- a/Build/Customizations/Extensions.targets
+++ b/Build/Customizations/Extensions.targets
@@ -79,4 +79,17 @@
                                                   AssemblyProduct="$(ProductNameForSecurityManager)"
                                                   AssemblyCopyright="$(CopyrightForSecurityManager)" />
   </Target>
+
+  <Target Name="CreateNpmDependencyPackage" AfterTargets="CreateNuGetPackagesWithDebugSymbols">
+    <PropertyGroup>
+      <_npmDependenciesFilePath>$(SolutionDirectory)Remotion\Web\Dependencies.JavaScript\package.json</_npmDependenciesFilePath>
+      <_outputPath>$(OutputDirectory)Npm\remotion.dependencies\package.json</_outputPath>
+    </PropertyGroup>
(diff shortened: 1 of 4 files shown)


a-ctor:
@MichaelKetting can't request your review for some reason but pls review ;)

MichaelKetting:
> @MichaelKetting can't request your review for some reason but pls review ;)

That's because I created it, even though you are assigned. Let's agree not to talk about it ;)

Build/Customizations/Extensions.targets
```diff
@@ -79,4 +79,26 @@
...
+  <Target Name="AddNpmDependencyProjectGenerationToBuild" AfterTargets="PrepareBuildTargets">
+    <PropertyGroup>
+      <BuildTargets>
+        $(BuildTargets);
+        GenerateNpmDependencyProjectInOutput;
+      </BuildTargets>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="GenerateNpmDependencyProjectInOutput">
```
MichaelKetting, line 92: Please move this to after-CreateNuGetPackagesWithDebugSymbols was because a) it should be inside the build and before the build-cleanup happens and b) it will then be controlled by the SkipNuGet flag. Right now, we would create it in each build, which isn't nice.

```suggestion
  <Target Name="CreateNpmDependencyPackage" AfterTargets="CreateNuGetPackagesWithDebugSymbols">
```
(better name for "NpmDependencyPackage" ?

MichaelKetting (commented):
(note, I also can't actually approve - Github :( )

a-ctor (approved):

a-ctor:
@MichaelKetting do you have any additional change requests or can I consider the PR approved? :D

MichaelKetting:
> @MichaelKetting do you have any additional change requests or can I consider the PR approved? :D

approved. thanks

PR #793, opened 2023-08-16
RM-8897 WebTreeView expander link has no discernable text

a-ctor:
https://re-motion.atlassian.net/browse/RM-8897

@MichaelKetting the ticket did not specify what the expected solution was so I used `aria-hidden` on the expander node. I assume that the expander node is not necessary for screen reader at all since keyboard navigation of the tree is supported. If this assumption is incorrect please tell me the desired solution :D

### DIFF
--- a/Remotion/Web/Core/UI/Controls/WebTreeView.cs
+++ b/Remotion/Web/Core/UI/Controls/WebTreeView.cs
@@ -772,7 +772,7 @@ namespace Remotion.Web.UI.Controls
           writer.AddAttribute(DiagnosticMetadataAttributes.TriggersPostBack, "true");
       }
       writer.AddAttribute("tabindex", "-1");
-      writer.AddAttribute(HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.None);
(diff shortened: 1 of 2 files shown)


Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -772,7 +772,7 @@ private void RenderNodes (HtmlTextWriter writer, WebTreeNodeCollection webTreeNo
           writer.AddAttribute(DiagnosticMetadataAttributes.TriggersPostBack, "true");
       }
       writer.AddAttribute("tabindex", "-1");
-      writer.AddAttribute(HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.None);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
```
MichaelKetting, line 775: Ah, yes. The assumtion was that role=none would be sufficient. for some reason, it wasn't. Or, actually, the problem is that axe core interprets role=none as still aria-visible. either way, aria-hidden takes care of it.

MichaelKetting (approved):

PR #795, opened 2023-08-16
RM-8917-upgrade-moq-to-4.18.4

Frazerus:
https://re-motion.atlassian.net/browse/RM-8917

### DIFF
--- a/Build/ProjectTypes/Test.props
+++ b/Build/ProjectTypes/Test.props
@@ -13,7 +13,7 @@
   <ItemGroup>
     <PackageReference Include="JetBrains.Annotations" Version="2022.1.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.10.0" />
-    <PackageReference Include="Moq" Version="4.16.1" PrivateAssets="all" />
+    <PackageReference Include="Moq" Version="4.17.2" PrivateAssets="all" />
     <PackageReference Include="Moq.VerifiableSequence" Version="4.16.1.2" PrivateAssets="all" />
     <PackageReference Include="NUnit" Version="3.13.3" PrivateAssets="all" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.0.0" />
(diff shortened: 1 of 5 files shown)


a-ctor (approved):

Build/ProjectTypes/Test.props
```diff
@@ -13,7 +13,7 @@
   <ItemGroup>
     <PackageReference Include="JetBrains.Annotations" Version="2022.1.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.10.0" />
-    <PackageReference Include="Moq" Version="4.16.1" PrivateAssets="all" />
+    <PackageReference Include="Moq" Version="4.18.4" PrivateAssets="all" />
     <PackageReference Include="Moq.VerifiableSequence" Version="4.16.1.2" PrivateAssets="all" />
```
MichaelKetting, line 17: There isn't a hard and fast rule yet. For consistency, having our Moq.VerifiableSequence build match Moq would be a nice touch, knowing the builds are officially compatible by also getting a `v4.18.4.1` version. please check with @a-ctor if that's a good practice to establish.

MichaelKetting (approved):

MichaelKetting (changes requested):
Requires solution for castle dependency.

Build/ProjectTypes/Test.props
a-ctor: I don't think is is necessary to keep your package version up-to-date if nothing has changed. We started with a matching version which kind of implies that we update the version but I don't think we need to. Its just unnecessary work as long as we don't have to fix anything.

If a higher version of Moq is installed it will be used for `Moq.VerifiableSequence` and in the case that something is broken our CI tests fail. If that happens we can update the package and bump the version to the one we are using right now. I see the chance of something silently breaking as very small since we mostly wrap Moq methods in the package.

Frazerus:
> Requires solution for castle dependency.

I guess the best option would be to only upgrade Moq to v4.17.2 then, the last version where we do not require a castle.core upgrade to any v >= 5.x

Frazerus:
The upgrade would include having to update moq.Verifiable Sequence also in the case of 4.17.2. There are many errors which need to be fixed before being able to upgrade in VerifiableSeq.

I therefore put this task to be a whole task where we also need to upgrade Moq.VerifiableSeq. It might not make sense to do this now, as when we upgrade again early next release, we might have to do a very similar upgrade.

PR #808, opened 2023-08-22
RM-8893 Add tests for dispatching without edit mode

Frazerus:
https://re-motion.atlassian.net/browse/RM-8893

### DIFF
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
@@ -36,57 +36,6 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
   public class BocListValidationResultDispatchingValidatorTest : BocTest
   {
 
-    public (BocListMock boclist, BocSimpleColumnDefinition simpleColumnDefinition) InitiateBocListForEditModeWithStringClass (TypeWithString[] typeWithStrings)
-    {
-      var typeWithStringClass = BindableObjectProviderTestHelper.GetBindableObjectClass(typeof(TypeWithString));
-
-      var typeWithStringFirstValuePath = BusinessObjectPropertyPath.CreateStatic(typeWithStringClass, "FirstValue");
-
-      var typeWithStringFirstValueSimpleColumn = new BocSimpleColumnDefinition();
-      typeWithStringFirstValueSimpleColumn.SetPropertyPath(typeWithStringFirstValuePath);
-
-
-      var hostTypeForTypeWithString = HostTypeForTypeWithString.Create();
-      hostTypeForTypeWithString.Strings = typeWithStrings;
-      var businessObject = (IBusinessObject)hostTypeForTypeWithString;
-      var businessObjectProperty = businessObject.BusinessObjectClass.GetPropertyDefinition("Strings");
-
-      var bocList = new BocListMock();
-      bocList.DataSource = new StubDataSource(typeWithStringClass);
-      bocList.DataSource.BusinessObject = businessObject;
-      bocList.DataSource.Mode = DataSourceMode.Edit;
-      bocList.Property = (IBusinessObjectReferenceProperty)businessObjectProperty;
-      bocList.FixedColumns.Add(typeWithStringFirstValueSimpleColumn);
-      bocList.FixedColumns.Add(new BocValidationErrorIndicatorColumnDefinition());
-      bocList.EnableAutoFocusOnSwitchToEditMode = false;
-
-      Assert.That(bocList.IsReadOnly, Is.False);
-
-      bocList.LoadValue(false);
-      bocList.SwitchRowIntoEditMode(0);
-
-      Assert.That(bocList.IsRowEditModeActive, Is.True);
-
-      var editableRow = ((IBocList)bocList).EditModeController.GetEditableRow(0);
-      ((EditableRow)editableRow).EnsureValidatorsRestored();
-
-      Assert.That(editableRow.HasValidators(), Is.True);
-
-      Assert.That(editableRow.HasValidators(0), Is.True);
-      ControlCollection editableRowValidators = editableRow.GetValidators(0);
-      Assert.That(editableRowValidators.Cast<BaseValidator>().Select(v => v.GetType()),
-          Is.EqualTo(
-              new[]
-              {
-                typeof(ControlCharactersCharactersValidator),
-                typeof(BusinessObjectBoundEditableWebControlValidationResultDispatchingValidator)
-              }));
-
-      ((BaseValidator)editableRowValidators[0]).Validate();
-
-      return (bocList, typeWithStringFirstValueSimpleColumn);
-    }
-
     [Test]
     public void DispatchValidationFailures_WithInvisibleControl_UsesReadOnlyValidationResultForDispatching ()
     {
@@ -130,6 +79,27 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
       var validator = new BocListValidationResultDispatchingValidator();
       validator.ControlToValidate = "someBocList";
       bocList.ID = "someBocList";
+      bocList.SwitchRowIntoEditMode(0);
+
+      Assert.That(bocList.IsRowEditModeActive, Is.True);
+
+      var editableRow = ((IBocList)bocList).EditModeController.GetEditableRow(0);
+      ((EditableRow)editableRow).EnsureValidatorsRestored();
+
+      Assert.That(editableRow.HasValidators(), Is.True);
+
+      Assert.That(editableRow.HasValidators(0), Is.True);
+      ControlCollection editableRowValidators = editableRow.GetValidators(0);
+      Assert.That(
+          editableRowValidators.Cast<BaseValidator>().Select(v => v.GetType()),
+          Is.EqualTo(
+              new[]
+              {
+                typeof(ControlCharactersCharactersValidator),
+                typeof(BusinessObjectBoundEditableWebControlValidationResultDispatchingValidator)
+              }));
+
+      ((BaseValidator)editableRowValidators[0]).Validate();
 
       NamingContainer.Controls.Add(bocList);
       NamingContainer.Controls.Add(validator);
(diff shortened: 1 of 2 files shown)


Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +161,169 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
...
+          .Returns(Array.Empty<BusinessObjectValidationFailure>())
+          .Verifiable();
+      validationResultMock
+          .Setup(_ => _.GetUnhandledValidationFailures(It.IsAny<IBusinessObject>(), It.IsAny<bool>(), It.IsAny<bool>()))
+          .Returns(Array.Empty<BusinessObjectValidationFailure>());
+
+      validator.DispatchValidationFailures(validationResultMock.Object);
+
+      var bocListValidationFailureRepository = bocList.ValidationFailureRepository;
+      var editedRow = ((IBocList)bocList).EditModeController.GetEditableRow(0);
```
a-ctor, line 225: Is the assert with the `editedRow` (that is now `values[0]`) no longer valid? if so we can remove this line as well

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +161,169 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
...
+      IBusinessObject businessObject = TypeWithReference.Create();
+      var businessObjectProperty = businessObject.BusinessObjectClass.GetPropertyDefinition("ReferenceList");
+
+      var bocList = new BocListMock();
+      bocList.ID = "someBocList";
+      bocList.DataSource = new StubDataSource(businessObject.BusinessObjectClass);
+      bocList.DataSource.BusinessObject = businessObject;
+      bocList.Property = (IBusinessObjectReferenceProperty)businessObjectProperty;
+
+      var validationResultMock = new Mock<IBusinessObjectValidationResult>(MockBehavior.Strict);
```
a-ctor, line 252: Nothing wrong with the test, but I think we should discuss as a team how we want to handle such "no calls" mock objects. I think it might be good to have a special name and also to add a `.VerifyNoOtherCalls()` at the end of the test. this makes it clearer that we assert that nothing is called on this mock. Otherwise, this fact is only visible when investigating the test in detail and realizing that this strict mock with no setups is here to ensure nothing was called.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +161,169 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
...
+
+      var editModeControllerStub = new Mock<IEditModeController>();
+      var dataSourceStub = new Mock<IBusinessObjectDataSource>();
+      dataSourceStub.Setup(_ => _.BusinessObject).Returns(businessObject);
+
+      var columnDefinitionMock = new Mock<BocColumnDefinition>();
+      var validationFailureMatcherMock = new Mock<IValidationFailureMatcher>();
+      columnDefinitionMock.As<IBocColumnDefinitionWithValidationSupport>().Setup(_ => _.GetValidationFailureMatcher()).Returns(validationFailureMatcherMock.Object);
+
+      var bocListMock = new Mock<Control> { CallBase = true };
```
a-ctor, line 285: Can we simplify this setup here by using `BocListStubMock` instead? Haven't thought it through but it might do a lot of things for us that we now have an explicit setup for.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
Frazerus: Yeah, I agree with having the verify no other calls. 
As for names, do you suggest having something like a stub/fake/etc. postfix in such cases? 

In that case I have 3 suggestions:
* requisite 
* dependency
* extra

I think extra would be the most appropriate as it suggests that it is not needed in this case.

@a-ctor, @MichaelKetting thoughts?

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
Frazerus: Im not sure what a BocListStubMock is, but taking a guess that you mean the BocListMock class: That does not work. As I did this some time ago I'm not quite sure why exactly, but there were problems with setting some specific things on that object.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
a-ctor: Added a discussion item for our next weekly meeting :)

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -35,8 +35,7 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.Validation
   [TestFixture]
   public class BocListValidationResultDispatchingValidatorTest : BocTest
   {
-
-    public (BocListMock boclist, BocSimpleColumnDefinition simpleColumnDefinition) InitiateBocListForEditModeWithStringClass (TypeWithString[] typeWithStrings)
+    private (BocListMock boclist, BocSimpleColumnDefinition simpleColumnDefinition) InitiateBocListForEditModeWithStringClass (TypeWithString[] typeWithStrings)
```
MichaelKetting, line 38: Noticed this the first time. This helper should be at the bottom of the file :)

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +160,170 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
...
+    }
+
+    [Test]
+    public void DispatchValidationFailures_WithErrorIndicatorColumn_DispatchesToRepository ()
+    {
+      var validationFailureRepo = new BocListValidationFailureRepository();
+
+      var error = "BocList has validation errors.";
+      var resourceManagerStub = new Mock<IResourceManager>();
+      resourceManagerStub.Setup(_ => _.TryGetString("Remotion.ObjectBinding.Web.UI.Controls.BocList.ValidationFailuresFoundInListErrorMessage", out error)).Returns(true);
```
MichaelKetting, line 272: based on the length, having like breaks here with Setup and Returns would be nice. Same in line 283. It's a looking-good gutfeeling.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +160,170 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
...
+      NamingContainer.Controls.Add(bocListMock.Object);
+      NamingContainer.Controls.Add(validator);
+
+      validator.DispatchValidationFailures(validationResultMock.Object);
+
+      validationResultMock.Verify();
+      Assert.That(validator.IsValid, Is.False);
+      Assert.That(validationFailureRepo.GetUnhandledValidationFailuresForBocList(true).Single().Failure.ErrorMessage, Is.EqualTo("ListErrorMessage"));
+      Assert.That(validationFailureRepo.GetUnhandledValidationFailuresForDataRow(businessObject, true).Single().Failure.ErrorMessage, Is.EqualTo("RowErrorMessage"));
+      Assert.That(validationFailureRepo.GetUnhandledValidationFailuresForDataCell(businessObject, columnDefinitionMock.Object, true).Single().Failure.ErrorMessage, Is.EqualTo("ColumnErrorMessage"));
```
MichaelKetting, line 326: Is this still 180chars?

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
MichaelKetting: Okay, I see there's "many" calls. With "many" calls, adding an excpetion for each would be more involved. If it's just the three methods I can see in the next test, it's probably a cointoss. if it's five or more, it's no fun to use the exception trick.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +160,170 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
...
+      Assert.That(failuresInRepository.Count, Is.EqualTo(1));
+      Assert.That(failuresInRepository.Single().ColumnDefinition, Is.EqualTo(simpleColumnDefinition));
+      Assert.That(failuresInRepository.Single().RowObject, Is.EqualTo(values[0]));
+      Assert.That(failuresInRepository.Single().Failure.ErrorMessage, Is.EqualTo("The value contains unsupported characters: \"\u0001\" (position: 1)."));
+
+      validationResultMock.Verify();
+
+      // Validates that dispatch was called to dispatching validator
+      var bocTextValue = ((Control)editableRow).FindControl("0") as BusinessObjectBoundEditableWebControl;
+      validationResultMock.Verify(_ => _.GetValidationFailures(bocTextValue.DataSource.BusinessObject, bocTextValue.Property, true));
```
MichaelKetting, line 235: Does Moq implictly reset the recordings after a verify? Otherwise, we test the same thing we do in line 231 with the validationResultMock.Verify call, correct?

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
```diff
@@ -160,6 +160,170 @@ public void DispatchValidationFailures_WithRowEditModeControlHavingInvalidValida
       Assert.That(failuresInRepository.Single().Failure.ErrorMessage, Is.EqualTo("The value contains unsupported characters: \"\u0001\" (position: 1)."));
 
       validationResultMock.Verify();
+
+      // Validates that dispatch was called to dispatching validator
+      var bocTextValue = ((Control)editableRow).FindControl("0") as BusinessObjectBoundEditableWebControl;
+      validationResultMock.Verify(_ => _.GetValidationFailures(bocTextValue.DataSource.BusinessObject, bocTextValue.Property, true));
```
MichaelKetting, line 166: see comment in line 235

MichaelKetting (commented):
I got a question or two and some formatting suggestions.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
Frazerus: These calles were already handled in the above Setup that uses `simpleColumnDefinition.GetPRopertyPath()...`, this was therefore an unrequired verify.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/Validation/BocListValidationResultDispatchingValidatorTest.cs
Frazerus: These calles were already handled in the above Setup that uses simpleColumnDefinition.GetPRopertyPath()..., this was therefore an unrequired verify.

MichaelKetting (approved):

PR #823, opened 2023-09-01
feature/RM-8945-BocList-content-width-does-not-match-column-width-for-inline-editing

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-8945


a-ctor (approved):

a-ctor:
Making the window smaller has the following effect:

image

We probably also want to add an `overflow: hidden`

MichaelKetting:
> Making the window smaller has the following effect:
> 
> image
> 
> We probably also want to add an `overflow: hidden`

Thanks. yes, this is a separate issue that existed previously. please create a bug ticket for it and we check how if/how this should be solved. probalby with overflow:hidden, but I can't say definitivly right now.

PR #831, opened 2023-09-08
Feature/rm 8939 boc list validation marker column should only be shown if there are validation errors

Frazerus:
https://re-motion.atlassian.net/browse/RM-8939

Tested Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocListControlObjectTest.cs, did work with alternative standard of not showing the validation column.

### DIFF
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
@@ -16,7 +16,6 @@
 // 
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using Remotion.Collections;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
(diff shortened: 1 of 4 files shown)


Remotion/ObjectBinding/Web/UI/Controls/ErrorColumnWithoutValidationFailureBehaviour.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Remotion.ObjectBinding.Web.UI.Controls
+{
+  public enum ErrorColumnWithoutValidationFailureBehaviour
+  {
+    HideUnlessAnyValidationFailures,
+    HideUnlessRowOrCellValidationFailures,
+    Show
```
a-ctor, line 7: The double negative/inversion (`Hide` + `Unless`) makes it hard to understand the names without docs. But it is kind of hard to find good names here... Maybe we discuss this separately and try to find good names together.

a-ctor (changes requested):
Haven't looked at the tests yet

Remotion/ObjectBinding/Web/UI/Controls/ErrorColumnWithoutValidationFailureBehaviour.cs
Frazerus: Possible dev jour fixe thing? As we have @MichaelKetting there as well?

Frazerus: Just renamed them from `HideUnless` to `DisplayIf` -> IMO easier to understand

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -164,7 +163,34 @@ private bool IsColumnVisible (BocColumnDefinition column)
...
+
+      if (columnAsValidationErrorIndicatorColumn.BehaviourWithoutValidationFailures == ErrorColumnWithoutValidationFailureBehaviour.HideUnlessAnyValidationFailures)
+      {
+        return repository.GetListFailureCount() + repository.GetRowAndCellFailureCount() != 0;
+      }
+
+      if (columnAsValidationErrorIndicatorColumn.BehaviourWithoutValidationFailures == ErrorColumnWithoutValidationFailureBehaviour.HideUnlessRowOrCellValidationFailures)
+      {
+        return repository.GetRowAndCellFailureCount() != 0;
+      }
```
MichaelKetting: I would ammend @a-ctor comment on the switch statement with:
1. move the guard condition  (repostiroy == null) to the top. 
2. make the remainding three cases into an if/else or a switch.

Note: to be very precise, having no OwnerControl would actually be a Assertion. This should never happen. I think we should add this guard (`Assertion.IsNotNull(OwnerControl)`). Disclaimer: I don't know how the other places usually handle OwnerControl==null. We should be reasonably cosinstent.

a-ctor (approved):

MichaelKetting (approved):

PR #862, opened 2023-09-27
Feature/rm 8576 replace common service locator with re motion implementation v4.10

Frazerus:
https://re-motion.atlassian.net/browse/RM-8576


Frazerus:
Issue not correct understood

PR #867, opened 2023-10-03
RM-8966 Fix WebTestHelperTest where processes hang when a test fails

a-ctor:
https://re-motion.atlassian.net/browse/RM-8966

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
@@ -21,6 +21,7 @@ using System.IO;
 using System.Linq;
 using NUnit.Framework;
 using Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure;
+using Remotion.Web.Development.WebTesting.Utilities;
 using Remotion.Web.Development.WebTesting.WebDriver;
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
```diff
@@ -31,6 +32,38 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
...
+                                                                         "edge",
+                                                                         "geckodriver",
+                                                                         "firefox"
+                                                                     };
+
+    // Filter for the driver process names that might be left open if something went wrong
+    private static readonly Func<Process, bool> s_processFilter = static p =>
+    {
+      return s_relevantProcessNames.Contains(p.ProcessName);
+    };
```
MichaelKetting, line 49: Feels weird to me. Wouldn't a static method be easier and have the same effect?
```suggestion
    private static bool FilterBrowserProcesses (Process process)
    {
      return s_relevantProcessNames.Contains(process.ProcessName);
    };
```

And add the comment as xmlcomment instead of code comment.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #868, opened 2023-10-03
RM-8951 Add test to ensure the available resources are set correctly on re-motion assemblies

a-ctor:
https://re-motion.atlassian.net/browse/RM-8951

Draft PR to review the current implementation. Once the design is approved I will add the individual test for all the assemblies.

### DIFF
--- a/Remotion/ObjectBinding/UnitTests/AvailableResourcesTest.cs
+++ b/Remotion/ObjectBinding/UnitTests/AvailableResourcesTest.cs
@@ -1,0 +1,32 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using NUnit.Framework;
+using Remotion.Development.NUnit.UnitTesting.Globalization;
+using Remotion.Globalization;
+
+namespace Remotion.ObjectBinding.UnitTests
+{
+  [TestFixture]
+  public class AvailableResourcesTest
+  {
+    [Test]
+    public void AllAvailableResourcesAreDeclaredCorrectlyAndHaveTheirResources ()
+    {
+      AvailableResourcesTester.CheckAvailableResources<AvailableResourcesLanguagesAttribute>(typeof(BusinessObject).Assembly);
+    }
+  }
+}
--- a/Remotion/ObjectBinding/UnitTests/ObjectBinding.UnitTests.csproj
+++ b/Remotion/ObjectBinding/UnitTests/ObjectBinding.UnitTests.csproj
@@ -11,7 +11,7 @@
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\..\..\SharedSource\Development\NUnit\UnitTesting\*.cs" Link="SharedSource\NUnit\UnitTesting\%(Filename)%(Extension)" />
+    <Compile Include="..\..\..\SharedSource\Development\NUnit\UnitTesting\**\*.cs" Link="SharedSource\NUnit\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)" />
     <ProjectReference Include="..\..\Core\Collections.DataStore\Core.Collections.DataStore.csproj" />
     <ProjectReference Include="..\..\Core\Core\Core.Core.csproj" />
     <ProjectReference Include="..\..\Core\ExtensibleEnums\Core.ExtensibleEnums.csproj" />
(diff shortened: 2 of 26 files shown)


Remotion/ObjectBinding/UnitTests/AvailableResourcesTest.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Remotion.ObjectBinding.UnitTests
+{
+  [TestFixture]
+  public class AvailableResourcesTest
+  {
+    [Test]
+    public void AllAvailableResourcesAreDeclaredCorrectlyAndHaveTheirResources ()
+    {
+      AvailableResourcesTester.CheckAvailableResources<AvailableResourcesLanguagesAttribute>(typeof(BusinessObject).Assembly);
```
MichaelKetting, line 29: why not hardcode the attribute?

Remotion/ObjectBinding/UnitTests/ObjectBinding.UnitTests.csproj
```diff
@@ -12,6 +12,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="..\..\..\SharedSource\Development\NUnit\UnitTesting\*.cs" Link="SharedSource\NUnit\UnitTesting\%(Filename)%(Extension)" />
+    <Compile Include="..\..\..\SharedSource\Development\Globalization\*.cs" Link="SharedSource\NUnit\Globalization\%(Filename)%(Extension)" />
```
MichaelKetting, line 15: THere's a different include pattern in line 14. I think this needs chaing in line 15?
-> `SharedSource\NUnit\UnitTesting\Globalization` would be the location for the shared file, also in the sharedsource location.

SharedSource/Development/Globalization/AvailableResourcesTester.cs
```diff
@@ -0,0 +1,41 @@
...
+// ReSharper disable once CheckNamespace
+namespace Remotion.Development.Globalization
+{
+  public static class AvailableResourcesTester
+  {
+    /// <summary>
+    /// Tests that the available resources declared by <typeparamref name="TAttribute"/> on <paramref name="targetAssembly"/>
+    /// match the resource files in the output directory.
+    /// </summary>
+    public static void CheckAvailableResources<TAttribute> (Assembly targetAssembly)
```
MichaelKetting, line 16: not yet reviewed

MichaelKetting (changes requested):

Remotion/ObjectBinding/UnitTests/AvailableResourcesTest.cs
a-ctor: Because I did not want to introduce a globalization reference in the shared source project and passing the attribute seems like reasonable compromise that won't break instantly on rename/move of the attribute

MichaelKetting: Ah, I see. hmm... got an idea; we could link the Attribute over into the shared files project. That way, we can simplify the code :)

Remotion/ObjectBinding/UnitTests/ObjectBinding.UnitTests.csproj
```diff
@@ -11,7 +11,7 @@
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\..\..\SharedSource\Development\NUnit\UnitTesting\*.cs" Link="SharedSource\NUnit\UnitTesting\%(Filename)%(Extension)" />
+    <Compile Include="..\..\..\SharedSource\Development\NUnit\UnitTesting\**\*.cs" Link="SharedSource\NUnit\UnitTesting\%(RecursiveDir)%(Filename)%(Extension)" />
```
MichaelKetting, line 14: check. ok.

SharedSource/Development/NUnit/UnitTesting/Globalization/AvailableResourcesTester.cs
```diff
@@ -0,0 +1,41 @@
...
+  public static class AvailableResourcesTester
+  {
+    /// <summary>
+    /// Tests that the available resources declared by <typeparamref name="TAttribute"/> on <paramref name="targetAssembly"/>
+    /// match the resource files in the output directory.
+    /// </summary>
+    public static void CheckAvailableResources<TAttribute> (Assembly targetAssembly)
+      where TAttribute : Attribute
+    {
+      // We don't have a reference to globalization so the call has to provide the right attribute
```
MichaelKetting, line 19: waiting on the final decision if linking the attribute into the dev project would make sense or not.

MichaelKetting (changes requested):

SharedSource/Development/NUnit/UnitTesting/Globalization/AvailableResourcesTester.cs
a-ctor: Linking does not work since a reference in the XML docs cannot be found. Including this file as well causes more XML doc issues and other compile errors because stuff could not be found.

MichaelKetting (changes requested):
Existing stuff looks okay, now we need to put it into all the unit test for the projects that have resources.

a-ctor:
@MichaelKetting we can do them now as it doesn't seem worth to me to open separate issues. There are 20 usages of the `AvailableResourcesLanguages`, i would need a list of what of those usages need tests and what do not.

MichaelKetting:
> @MichaelKetting we can do them now as it doesn't seem worth to me to open separate issues. There are 20 usages of the `AvailableResourcesLanguages`, i would need a list of what of those usages need tests and what do not.

SecurityManagerClientsWeb (needs also a test project)
objectbindingwebcompatiblity 
objectbindingwebvalidation
objectbindingwebpreview (needs also a test project)

a-ctor:
@MichaelKetting And what about the ones that have empty `AvailableResourceLanguages`? Which of those should I add the test to?

MichaelKetting (approved):

PR #876, opened 2023-10-11
RM-8777 BocList: Support for BusinessObject icons to be rendered in a…

Frazerus:
… specific column
https://re-motion.atlassian.net/browse/RM-8777

### DIFF
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
@@ -3297,7 +3297,7 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     }
 
     /// <summary>
-    ///   Gets or sets a flag that indicates whether the control automatically generates a column 
+    ///   Gets or sets a flag that indicates whether the control automatically generates a column
     ///   for each property of the bound object.
     /// </summary>
     /// <value> <see langword="true"/> show all properties of the bound business object. </value>
@@ -3311,12 +3311,14 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     }
 
     /// <summary>
-    ///   Gets or sets a flag that indicates whether to display an icon in front of the first value 
-    ///   column.
+    ///   Gets or sets a flag that indicates whether to display an icon for the row's business object.
     /// </summary>
+    /// <remarks>
+    ///   Customize which <see cref="BocValueColumnDefinition"/> displays the icon via the <see cref="IBocColumnDefinitionWithRowIconSupport.RowIconMode"/> flag.
+    /// </remarks>
     /// <value> <see langword="true"/> to enable the icon. </value>
     [Category("Appearance")]
-    [Description("Enables the icon in front of the first value column.")]
+    [Description("Enables the icon for the row's business object.")]
     [DefaultValue(true)]
     public virtual bool EnableIcon
     {
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
@@ -16,6 +16,7 @@
 // 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Remotion.Collections;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
@@ -63,14 +64,21 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering
 
       PrepareSorting(sortingDirections, sortingOrder);
 
-      var firstValueColumnRendered = false;
       var bocColumnRenderers = new List<BocColumnRenderer>(_columnDefinitions.Count);
       var visibleColumnIndex = GetInitialVisibleColumnIndex();
+
+      BocColumnDefinition? iconColumn = null;
+      if (EnableIcon)
+      {
+        iconColumn = _columnDefinitions.FirstOrDefault(cd => cd is IBocColumnDefinitionWithRowIconSupport { RowIconMode: RowIconMode.Preferred })
+                     ?? _columnDefinitions.FirstOrDefault(cd => cd is IBocColumnDefinitionWithRowIconSupport { RowIconMode: RowIconMode.Automatic });
+      }
+
       for (int columnIndex = 0; columnIndex < _columnDefinitions.Count; columnIndex++)
       {
         var columnDefinition = _columnDefinitions[columnIndex];
         var isRowHeader = (columnDefinition as IBocColumnDefinitionWithRowHeaderSupport)?.IsRowHeader ?? false;
-        var showIcon = !firstValueColumnRendered && columnDefinition is BocValueColumnDefinition && EnableIcon;
+        var showIcon = ReferenceEquals(columnDefinition, iconColumn);
 
         var sortingDirection = SortingDirection.None;
         if (sortingDirections.ContainsKey(columnIndex))
@@ -106,9 +114,6 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocListImplementation.Rendering
                   sortingDirection,
                   orderIndex));
         }
-
-        if (columnDefinition is BocValueColumnDefinition)
-          firstValueColumnRendered = true;
       }
       return bocColumnRenderers.ToArray();
     }
--- a/Remotion/ObjectBinding/Web/UI/Controls/IBocColumnDefinitionWithRowIconSupport.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/IBocColumnDefinitionWithRowIconSupport.cs
@@ -1,0 +1,27 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+namespace Remotion.ObjectBinding.Web.UI.Controls
+{
+  /// <summary>
+  /// Defines the required API for configuring if a <see cref="BocColumnDefinition"/> should show the row icon.
+  /// </summary>
+  public interface IBocColumnDefinitionWithRowIconSupport
+  {
+    RowIconMode RowIconMode { get; }
+  }
+
+}
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilderTest.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilderTest.cs
@@ -426,6 +426,132 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls.BocListImplementation
       Assert.That(bocColumnRenderers[2].ShowIcon, Is.False);
     }
 
+    [Test]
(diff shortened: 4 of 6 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -3311,12 +3311,15 @@ public virtual bool ShowAllProperties
     }
 
     /// <summary>
-    ///   Gets or sets a flag that indicates whether to display an icon in front of the first value 
-    ///   column.
+    ///   Gets or sets a flag that indicates whether to display an icon in a <see cref="IBocColumnDefinitionWithRowIconSupport"/>.
     /// </summary>
+    /// <remarks>
+    ///   Which column the icon is displayed in is decided in the <see cref="BocColumnRendererArrayBuilder.CreateColumnRenderers"/>
```
a-ctor, line 3317: The implementation detail of `CreateColumnRenderers` is probably not helpful for a consumer. I would cut this information an just mention that the location is controlled by the `RowIconEnabled` flag on a column definition.

Usually when we reference some property of a type, for example, `a.b` we construct the XML docs in a similar fashion using `<see cref="a"/>.<see cref="a.b"/>` to have same "dot syntax" in the docs.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -63,14 +64,34 @@ public BocColumnRenderer[] CreateColumnRenderers ()
 
       PrepareSorting(sortingDirections, sortingOrder);
 
-      var firstValueColumnRendered = false;
+      var renderedRowIcon = false;
       var bocColumnRenderers = new List<BocColumnRenderer>(_columnDefinitions.Count);
       var visibleColumnIndex = GetInitialVisibleColumnIndex();
+
+      var enableIconOnFirstPossible = _columnDefinitions.All(
```
a-ctor, line 71: `...FirstPossibleColumn`?

Remotion/ObjectBinding/Web/UI/Controls/IBocColumnDefinitionWithRowIconSupport.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+  /// <summary>
+  /// Defines the required API for allowing a <see cref="BocColumnDefinition"/> to show the row icon.
+  /// </summary>
+  public interface IBocColumnDefinitionWithRowIconSupport
+  {
+    RowIcon RowIconEnabled { get; }
+  }
+
+  public enum RowIcon
```
a-ctor, line 27: Move to its own file? Otherwise I would make it a nested type instead if we want to keep it close to the interface

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -63,14 +64,34 @@ public BocColumnRenderer[] CreateColumnRenderers ()
...
+          cd =>
+          {
+            if (cd is IBocColumnDefinitionWithRowIconSupport columnDefinitionWithRowIconSupport)
+              return columnDefinitionWithRowIconSupport.RowIconEnabled != RowIcon.Preferred;
+            return true;
```
a-ctor, line 76: Maybe we could simplify this to something like:

```c#
return cd is not IBocColumnDefinitionWithRowIconSupport { RowIconEnabled: RowIcon.Preferred };

```

a-ctor (approved):

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -3311,12 +3311,16 @@ public virtual bool ShowAllProperties
...
 
     /// <summary>
-    ///   Gets or sets a flag that indicates whether to display an icon in front of the first value 
-    ///   column.
+    ///   Gets or sets a flag that indicates whether to display an icon in a <see cref="IBocColumnDefinitionWithRowIconSupport"/>.
     /// </summary>
+    /// <remarks>
+    ///   Which column the icon is displayed in is controlled by the values of
+    ///   <see cref="IBocColumnDefinitionWithRowIconSupport"/>.<see cref="IBocColumnDefinitionWithRowIconSupport.RowIconEnabled"/>
+    ///   on the columns.
```
MichaelKetting, line 3319:
```suggestion
    ///   Customize which <see cref="BocValueColumnDefinition"/> displayes the icon via the <see cref="IBocColumnDefinitionWithRowIconSupport.RowIconEnabled"/> flag.
```

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -63,14 +64,32 @@ public BocColumnRenderer[] CreateColumnRenderers ()
...
-      var firstValueColumnRendered = false;
+      var renderedRowIcon = false;
       var bocColumnRenderers = new List<BocColumnRenderer>(_columnDefinitions.Count);
       var visibleColumnIndex = GetInitialVisibleColumnIndex();
+
+      var enableIconOnFirstPossibleColumn = _columnDefinitions.All(
+          cd =>
+          {
+            return cd is not IBocColumnDefinitionWithRowIconSupport { RowIconEnabled: RowIcon.Preferred };
+          });
```
MichaelKetting, line 75: It would be better to use Any to test for preferred and then use a `not` at the "correct" spot. That way, you only need to iterate until you find the first Preferred column, not check all the columns.

MichaelKetting, line 75: There's an alternative option which may be easier:

You could select min-value column from the list and then simply do a reference-equals in the for-loop. Or perhaps `order-by + first` might be even easier to understand. Yes, this will cause an extra loop, but the code is easier to grasp (I think, your thoughts?) and the 20 or so columns that are typically a maximum number used in the UI don't cost us that much CPU cycles that it benefits from more complex code.
```
var iconColumn;
if (EnableIcon)
  iconColumn = _columnDefinitions.OfType<IBocColumnDefinitionWithRowIconSupport>().Where(cd=> cd.RowIcon != Disabled).OrderByDescending(cd => cd.RowIcon).FirstOrDefault();
else 
  iconColumn = null;
```

and below:
`var showIcon = object.ReferenceEuqls(columnDefinition, iconColumn)`

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
Frazerus: Yea the logic in here is quite complex. I don't know if it's better to keep the complexity and add a comment for how it works, to have the "better" performance or just bin it and go for the less complex one either. 

I like the current implementation but I can see how it's going to take double the time to understand compared to the easier implementation.

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -63,14 +64,28 @@ public BocColumnRenderer[] CreateColumnRenderers ()
 
       PrepareSorting(sortingDirections, sortingOrder);
 
-      var firstValueColumnRendered = false;
+      var renderedRowIcon = false;
       var bocColumnRenderers = new List<BocColumnRenderer>(_columnDefinitions.Count);
       var visibleColumnIndex = GetInitialVisibleColumnIndex();
+
+      var enableIconOnFirstPossibleColumn = !_columnDefinitions.Any(cd => cd is IBocColumnDefinitionWithRowIconSupport { RowIconMode: RowIconMode.Preferred });
```
MichaelKetting, line 71: Question: my order-by suggestion, did it not work or is it not as readable/understandable as the present code? Am asking to understand how my suggestion helps/doesn't help.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
MichaelKetting: NOTE: I just noticed the comments above. Sorry! Need to read.

MichaelKetting (commented):
dicussing pro/con of alternative implementation

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
MichaelKetting: I would like to go through the pro and cons as you see them to have a shared understanding of which solution we should keep. From your desciption, I see you favor the performance-optimized version over the less-complex one. Is my summary correct?

a-ctor: Since I suggested the inital `is not ... {}` stuff. I generally don't think that its that bad to read. Yes you need to know about the feature and what what it does but after that it should be pretty clear.

If my memory of boolean logic is correct we can also convert the `All` into an `Any` like this:

```
!_columnDefinitions.Any(cd => cd is IBocColumnDefinitionWithRowIconSupport { RowIconEnabled: RowIcon.Preferred });
```

I think this is pretty readable as "There exists no `IBocColumnDefinitionWithRowIconSupport` where `RowIconEnabled` is set to `RowIcon.Preferred`. But that just another option :)

MichaelKetting: @a-ctor to avoid misunderstandings: my suggestion refers to removing about 15 lines of logic from the implementation, not simply rewriting the line with the ALL-statement. Am not sure if you also meant the whole part or just this one line.

a-ctor: I was referring to just the 5 lines you used as context for the comment. I mistakenly thought that the discussion is about the complexity of those lines and since I was the one who suggested the `is not` stuff I wanted to comment on that. Since its about something else - carry on without me :p

Frazerus:
I added the method that I wanted to propose last week. I think it's the most readable version :)

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -63,14 +64,40 @@ public BocColumnRenderer[] CreateColumnRenderers ()
...
+              }
+            });
+      }
+
+      var lookup = _columnDefinitions
+          .OfType<IBocColumnDefinitionWithRowIconSupport>()
+          .Select((cd, i) => (cd, i))
+          .ToLookup(x => x.cd.RowIconMode, x => x.i);
+
+      var resultIndex = lookup[RowIconMode.Preferred].OrderBy(i => i).Take(1).Concat(lookup[RowIconMode.Automatic].OrderBy(i => i).Take(1)).FirstOrDefault();
```
MichaelKetting, line 93: Did a quick check now (next appointment is looming)

I don't see where `enableIconOnFirstPossibleColumn` is used in the code? Is that an artifact? I would drop line 73-86

line 88 - 93
That's the one I know and looks okay. I'm wondering if it would be easier / have equivalent perofrmance if we did this instead:
var iconCOlumn  = _cols.OfType<IBocColumnDefinitionWithRowIconSupport>().FirstOrDefault(Automatic) ?? _cols.OfType<IBocColumnDefinitionWithRowIconSupport>().FirstOrDefault(Preferred)

I'm fine with either version, it's just a gut feeling about the option I just pasted so I want to make sure it's part of the concious choice :)

MichaelKetting (changes requested):
I think there's more code in there than we use? Is it for side-by-side compare as you mentioned?

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
Frazerus: And yes, this is a duplicate to compare

MichaelKetting (approved):

PR #878, opened 2023-10-16
RM-9881 Fix incorrect size calculation for BocList validation row

a-ctor:
https://re-motion.atlassian.net/browse/RM-8981


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -595,7 +595,7 @@ class BocList
       let padding = parseFloat(parentStyle.paddingLeft) + parseFloat(parentStyle.paddingRight);
       let border = parseFloat(parentStyle.borderLeftWidth) + parseFloat(parentStyle.borderRightWidth);
         
-      container.style.width = (availableWidthInContainer - widthBeforeValidationFailureColumn - padding - border).toString() + "px";
+      container.style.width = (availableWidthInContainer - widthBeforeValidationFailureColumn - padding - border - 10).toString() + "px";
```
MichaelKetting, line 598: Could you please add a short comment on the 10px offset. perhaps a variable to avoid the magic number and a comment line.

Don't think I need to see the final version :)

PR #891, opened 2023-10-18
RM-8872 Add retries to the downloads in the BinaryProvider classes

a-ctor:
https://re-motion.atlassian.net/browse/RM-8872


MichaelKetting (approved):
One rename, please.

PR #892, opened 2023-10-19
RM-8971 Change usages of overflow hidden to overflow clip

Frazerus:
https://re-motion.atlassian.net/browse/RM-8971

We can have a sit together to check on the changes :)


a-ctor (commented):
redirecting to mike

MichaelKetting (approved):

PR #895, opened 2023-10-24
Feature/rm 8980 remove wcag helper and associated code

Frazerus:
https://re-motion.atlassian.net/browse/RM-8980

Have fun!


Remotion/ObjectBinding/Web/Doc/include/UI/Controls/BocReferenceValue.xml
```diff
@@ -44,10 +44,6 @@
         <item>
           <term>WAI WCAG 1.0</term>
           <description>
-            <para> 
-              Level A conform if advanced features are turned off. This can be enforced by the 
-              <see cref="Remotion.Web.Configuration.WebConfiguration.Wcag"/> configuration entry.
-            </para>
```
a-ctor, line 50: I wouldn't delete the whole paragraph here. I would leave `Level A conform if advanced features are turned off.` since we only removed the enforcement and add this spot to the list of TODOs just like the other two places that have open XML doc questions

Remotion/ObjectBinding/Web/UI/Controls/BocListImplementation/Rendering/BocColumnRendererArrayBuilder.cs
```diff
@@ -201,17 +198,11 @@ private bool IsColumnVisibleForBocValidationErrorIndicatorColumnDefinition (BocV
 
     private bool IsColumnVisibleForBocCommandColumnDefinition (BocCommandColumnDefinition columnAsCommandColumn)
```
a-ctor, line 199: As this method now returns a constant and it is private, we can inline the value and remove a bit of code at the usage.

a-ctor (changes requested):

Remotion/ObjectBinding/Web/Doc/include/UI/Controls/BocReferenceValue.xml
```diff
@@ -44,10 +44,10 @@
         <item>
           <term>WAI WCAG 1.0</term>
```
a-ctor, line 45: We can remove all of the WCAG 1.0 stuff from the docs, which should clear up the last two unresolved issues

a-ctor (changes requested):

a-ctor (approved):

MichaelKetting (changes requested):
Some more doc cleanups

For the Breaking Change note:
We only removed special handling for WCAG 1.0, not WCAG support. Support for WCAG 2.0 was added in https://re-motion.atlassian.net/browse/RM-6809 subsequent issues. The breaking change ticket only needs to document the removal of the spcial APIs for completeness' sake.

MichaelKetting (approved):

PR #897, opened 2023-10-25
RM-8990/RM-8992/RM-8993 Replace QueryConfiguration with IoC system

a-ctor:
Multi-issue PR as all of them build on top of each other so I think it makes sense to provide them together:

https://re-motion.atlassian.net/browse/RM-8990
https://re-motion.atlassian.net/browse/RM-8992
https://re-motion.atlassian.net/browse/RM-8993


Remotion/Data/DomainObjects.UnitTests/Queries/ConfigurationLoader/BaseDirectoryBasedQueryFileFinderTest.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    [Test]
+    public void GetQueryFilePaths_WithNonExistentRelativePathButOptional_ReturnsNothing ()
+    {
+      var appContextProviderMock = new Mock<IAppContextProvider>();
+      appContextProviderMock.SetupGet(e => e.BaseDirectory).Returns(@"C:\base");
+
+      var queryFileFinder = new QueryFileFinderStub(appContextProviderMock.Object, "query.txt", true);
+      Assert.That(queryFileFinder.GetQueryFilePaths(), Is.Empty);
+    }
```
martin-herwich, line 105: See comment above, Suggestion:
GetQueryFilePaths_RelativePath_RequiredFile_NotFound_ThrowsException
GetQueryFilePaths_RelativePath_OptionalFile_NotFound_ReturnsNothing (or "Empty")

Remotion/Data/DomainObjects.UnitTests/Queries/ConfigurationLoader/BaseDirectoryBasedQueryFileFinderTest.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    [Test]
+    public void GetQueryFilePaths_WithNonExistentFullPathButOptional_ReturnsNothing ()
+    {
+      var appContextProviderMock = new Mock<IAppContextProvider>();
+      appContextProviderMock.SetupGet(e => e.BaseDirectory).Returns(@"C:\base");
+
+      var queryFileFinder = new QueryFileFinderStub(appContextProviderMock.Object, @"C:\fullpath\query.txt", true);
+      Assert.That(queryFileFinder.GetQueryFilePaths(), Is.Empty);
+    }
```
martin-herwich, line 64: The difference between these two tests is that the query file is required in the first but optional in the second - this distinction is not that obvious in the names. Also, the distinction between these two tests and those for relative paths should be made more obvious.
Suggestion:
GetQueryFilePaths_FullPath_RequiredFile_NotFound_ThrowsException
GetQueryFilePaths_FullPath_OptionalFile_NotFound_ReturnsNothing (or "Empty")

martin-herwich (changes requested):

Remotion/Data/DomainObjects.UnitTests/Serialization/QueriesTest.cs
```diff
@@ -81,13 +83,14 @@ public void QueryDefinitionInQueryConfiguration ()
     public void UnknownQueryDefinitionInQueryConfiguration ()
     {
       QueryDefinition unknownQueryDefinition = new QueryDefinition("UnknownQuery", TestDomainStorageProviderDefinition, "select 42", QueryType.Scalar);
```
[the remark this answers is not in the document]
a-ctor: Yes, `QueryDefinition` has `ISerializable` members that ensure that a serialized `QueryDefinition` is loaded using `repo.GetMandadtory(...)` if `IsPartOfQueryConfiguration` is set.

I am adding a comment to explain this complexity :)

Remotion/Data/DomainObjects.UnitTests/StandardMappingTest.cs
```diff
@@ -52,13 +55,17 @@ public override void SetUp ()
     {
       base.SetUp();
 
+      _serviceLocatorScope = new ServiceLocatorScope(QueriesForStandardMappingQueryFileFinder.CreateServiceConfigurationEntry());
```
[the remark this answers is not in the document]
a-ctor: The additions for `Queries1.xml` and `Queries2.xml` were done in a test project where the IoC registration is necessary for the app to run correctly.

Here we are in the unit tests and I want the tests to be as isolated as possible. As such, adding a global IoC registration that affects all tests feels wrong even though there probably is no impact. If you want I can change it to a global registration instead

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Remotion.Data.DomainObjects.Queries
+{
+  /// <summary>
+  /// Default implementation of the <see cref="IQueryDefinitionRepository"/> interface.
+  /// </summary>
+  [ImplementationFor(typeof(IQueryDefinitionRepository), Lifetime = LifetimeKind.Singleton)]
+  public class QueryDefinitionRepository : IQueryDefinitionRepository
+  {
+    public static QueryDefinitionRepository FromQueryDefinitions (IEnumerable<QueryDefinition> queryDefinitions)
```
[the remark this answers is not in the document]
a-ctor: I would suggest changing the private constructor to take an `IEnumerable<T>` and do the dictionary construction there. Then we don't have two `.ToDictionary` calls any more. 

I would then solve the "item is null" and "duplicated ID" issues by wrapping the exceptions thrown by `.ToDictionary`.

What do you think?

Remotion/Data/DomainObjects/Queries/ConfigurationLoader/QueryDefinitionLoader.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+      ArgumentUtility.CheckNotNull(nameof(queryFileFinder), queryFileFinder);
+
+      _queryFileFinder = queryFileFinder;
+
+      _queries = new DoubleCheckedLockingContainer<ReadOnlyCollection<QueryDefinition>>(LoadAllQueryDefinitionsImplementation);
+    }
+
+    /// <inheritdoc />
+    public ReadOnlyCollection<QueryDefinition> LoadAllQueryDefinitions () => _queries.Value!;
```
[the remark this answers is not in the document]
a-ctor: If we want to use an `IReadOnlyXXX` here instead of `ReadOnlyCollection<T>` I would suggest using `IReadOnlyList<T>` instead of the collection one. I feel like the collection variant does not add much over `IEnumerable<T>` which could then be used instead.

I personally, don't think it is necessary to wrap a created list with a `ReadOnlyCollection<QueryDefinition>` if we return an `IReadOnlyList<T>` but I know mike likes it so I would keep the creation and only change the return type.

Remotion/Data/DomainObjects/Queries/ConfigurationLoader/BaseDirectoryBasedQueryFileFinder.cs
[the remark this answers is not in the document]
a-ctor: We can introduce an abstract property `Required` if you want. What exactly does not feel right for you here? I introduced it as a parameter as implementation already need to call the constructor and won't use the value anywhere else anyways as they shouldn't override `GetQueryFilePaths` again. The implementations are mostly there for IoC registration just like the `Query1.xml` and `Query2.xml` registrations.

But looking at the implementation it might be a good idea to allow multiple query files to be passed at the same time? It seems like of annoying to register multiple classes in such cases. What do you think?

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Queries/CustomQueryDatabaseBehaviorTest.cs
```diff
@@ -45,6 +47,7 @@ public override void SetUp ()
 
       var locator = DefaultServiceLocator.Create();
       locator.RegisterSingle<IPersistenceExtensionFactory>(() => persistenceExtensionFactoryStub.Object);
+      locator.RegisterSingle<IQueryFileFinder>(() => new QueriesForStandardMappingQueryFileFinder(new AppContextProvider()));
```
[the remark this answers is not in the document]
a-ctor: I don't like adding service locator dependencies to the test so I would keep it as is.

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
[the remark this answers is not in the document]
martin-herwich: Sounds good, make it so!

Remotion/Data/DomainObjects/Queries/ConfigurationLoader/QueryDefinitionLoader.cs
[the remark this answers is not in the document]
martin-herwich: I have nothing against `IReadOnlyList<T>`, but I think that `IReadOnlyCollection<T>` adds a significant advantage in that it is already enumerated and can thus be checked by `ArgumentUtility.CheckNotNullOrItemsNull()`. In my opinion, `IEnumerable<T>` should only be used when lazy enumeration is a required feature.

With regards to wrapping, I had that discussion. If we don't keep the list, we can just return the `List<T>` because even if the caller casts from `IReadOnlyCollection<T>` and modifies it, it does not corrupt our internal state. In this case however, we keep the list and return it in all future calls, so we want to guard against that sort of caller misbehavior.

Remotion/Data/DomainObjects/Queries/ConfigurationLoader/BaseDirectoryBasedQueryFileFinder.cs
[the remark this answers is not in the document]
martin-herwich: I concur with Mike that there is no significant benefit to avoiding code duplication here that would outweigh the additional complexity. We should just have two side-by-side implementations, one for optional and one for required, without a base class.

I don't think multiple allowing query files adds value because a query file can already have multiple queries. Each query file has a role, wherefore we want to configure required/optional for each query file separately.

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -34,21 +34,33 @@ public static QueryDefinitionRepository FromQueryDefinitions (IEnumerable<QueryD
...
-      _queryLookup = queryLookup;
+      ArgumentUtility.CheckNotNull(nameof(queryDefinitions), queryDefinitions);
+
+      try
+      {
+        _queryLookup = queryDefinitions.ToDictionary(e => e.ID, e => e);
+      }
+      catch (NullReferenceException)
+      {
+        // Only way a null ref is causes is in the key selector of the .ToDictionary(...) call, indicating that an item is null
```
martin-herwich, line 57: These messages do not tell the caller which query is the problem, which makes fixing it harder. We should provide the ID of the QueryDefinition in the message, so that text search for the source of the problem is possible.
This may necessitate manually building the dictionary.

We could even strive to name not only the first ID with an error, but all IDs, so that fixing the problem does not require multiple attempts.

martin-herwich (changes requested):

martin-herwich (approved):

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private readonly Dictionary<string, QueryDefinition> _queryLookup;
+
+    public QueryDefinitionRepository (IQueryDefinitionLoader queryDefinitionLoader)
+        : this(ArgumentUtility.CheckNotNull(nameof(queryDefinitionLoader), queryDefinitionLoader).LoadAllQueryDefinitions())
+    {
+    }
+
+    private QueryDefinitionRepository (IEnumerable<QueryDefinition> queryDefinitions)
```
MichaelKetting, line 47: if we switch to IReadOnlyCollection, we can use the ArgUtility for the null-check. It would remove some noise from the code and the performance impact is negligible.

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -0,0 +1,96 @@
...
+          throw new ArgumentException($"Item {index} of parameter '{nameof(queryDefinitions)}' is null.", nameof(queryDefinitions));
+
+        if (queryLookup.ContainsKey(queryDefinition.ID))
+        {
+          duplicateQueryIDs ??= new HashSet<string>();
+          duplicateQueryIDs.Add(queryDefinition.ID);
+          continue;
+        }
+
+        queryLookup.Add(queryDefinition.ID, queryDefinition);
```
MichaelKetting, line 67: This is right on the edge between an exit condition and an if/else. since `continue` is something we try to avoid in the code base, this tips the scale to the if/else solution.

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -0,0 +1,96 @@
...
+  /// </summary>
+  [ImplementationFor(typeof(IQueryDefinitionRepository), Lifetime = LifetimeKind.Singleton)]
+  public class QueryDefinitionRepository : IQueryDefinitionRepository
+  {
+    public static QueryDefinitionRepository FromQueryDefinitions (IEnumerable<QueryDefinition> queryDefinitions)
+    {
+      ArgumentUtility.CheckNotNull(nameof(queryDefinitions), queryDefinitions);
+
+      return new QueryDefinitionRepository(queryDefinitions);
+    }
```
MichaelKetting, line 38: We should move this to the UnitTests / simply use a ctor. I don't see a benefit in having a static factory for an api that's normally resolved via service.

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    private readonly Dictionary<string, QueryDefinition> _queryLookup;
+
+    public QueryDefinitionRepository (IQueryDefinitionLoader queryDefinitionLoader)
+        : this(ArgumentUtility.CheckNotNull(nameof(queryDefinitionLoader), queryDefinitionLoader).LoadAllQueryDefinitions())
```
MichaelKetting, line 43: TODO: Test if DeferredQueryDefinitionRepository is something we need (see RM-8992)

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+    }
+
+    private QueryDefinitionRepository (IEnumerable<QueryDefinition> queryDefinitions)
+    {
+      ArgumentUtility.CheckNotNull(nameof(queryDefinitions), queryDefinitions);
+
+      var queryLookup = new Dictionary<string, QueryDefinition>();
+      HashSet<string>? duplicateQueryIDs = null;
```
MichaelKetting, line 52: I don't think the lazy initialization of the HashSet is worth the allocation we safe doing so since we instead now have a nullable field with the added overhead during execution. Plus a marked-as-not-const variable.

MichaelKetting (commented):
Review RM-8990

Remotion/Data/DomainObjects/Queries/QueryDefinitionRepository.cs
MichaelKetting: Can't change this because the ctor is private.

Remotion/Data/DomainObjects.UnitTests/UnitTesting/TemporaryTestFile.cs
```diff
@@ -0,0 +1,45 @@
...
+//
+using System;
+using System.IO;
+
+namespace Remotion.Data.DomainObjects.UnitTests.UnitTesting
+{
+  /// <summary>
+  /// Creates a temporary test file in the temp directory that is cleaned up when disposed.
+  /// </summary>
+  public class TemporaryTestFile : IDisposable
```
MichaelKetting, line 25: We already have TempFile with this semantics.

Remotion/Data/DomainObjects.UnitTests/QueriesForStandardMappingQueryFileFinder.cs
```diff
@@ -0,0 +1,36 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using Remotion.Data.DomainObjects.Queries.ConfigurationLoader;
+using Remotion.ServiceLocation;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.UnitTests
+{
+  public class QueriesForStandardMappingQueryFileFinder : BaseDirectoryBasedRequiredQueryFileFinder
```
MichaelKetting, line 24: This is solved via cached instance isntead of creating new scopes on demand. also helps with performance.

Remotion/Data/DomainObjects/Configuration/DomainObjectsConfiguration.cs
```diff
@@ -60,10 +58,8 @@ public StorageConfiguration Storage
     }
 
     [ConfigurationProperty(QueryPropertyName)]
-    public QueryConfiguration Query
-    {
-      get { return _queryConfiguration.Value; }
-    }
+    [Obsolete("QueryConfiguration is no longer supported. (Version 6.0.0)", true)]
```
MichaelKetting, line 61: did you create breaking change notes for this? (haven't checked)

PR #898, opened 2023-10-25
Feature/rm 8987 update security manager principal to use security client transaction factory

Frazerus:
https://re-motion.atlassian.net/browse/RM-8987

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/ClientTransactionFactoryTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/ClientTransactionFactoryTest.cs
@@ -54,7 +54,7 @@ namespace Remotion.Data.DomainObjects.UnitTests
           .Callback((ClientTransaction transaction) => transaction.Extensions.Add(extensionStub.Object))
           .Verifiable();
 
-      ITransaction transaction = factory.Object.CreateRootTransaction();
+      ITransaction transaction = ((ITransactionFactory)factory.Object).CreateRootTransaction();
 
       var clientTransaction = transaction.To<ClientTransaction>();
       Assert.That(clientTransaction.Extensions, Has.Member(extensionStub.Object));
@@ -74,7 +74,7 @@ namespace Remotion.Data.DomainObjects.UnitTests
           .Callback((ClientTransaction transaction) => transaction.Extensions.Add(extensionStub.Object))
           .Verifiable();
 
-      ITransaction rootTransaction = factory.Object.CreateRootTransaction();
+      ITransaction rootTransaction = ((ITransactionFactory)factory.Object).CreateRootTransaction();
       ITransaction childTransaction = rootTransaction.CreateChild();
 
       Assert.That(rootTransaction.To<ClientTransaction>().Extensions, Has.Member(extensionStub.Object));
--- a/Remotion/Data/DomainObjects/ClientTransactionFactory.cs
+++ b/Remotion/Data/DomainObjects/ClientTransactionFactory.cs
@@ -19,6 +19,7 @@ using System;
 namespace Remotion.Data.DomainObjects
 {
   /// <summary>The <see cref="ClientTransactionFactory"/> is the default implementation of the <see cref="ITransactionFactory"/> interface.</summary>
+  /// <threadsafety static="true" instance="true"/>
   [Serializable]
   public class ClientTransactionFactory : ITransactionFactory
   {
@@ -26,13 +27,17 @@ namespace Remotion.Data.DomainObjects
     /// Creates a new root transaction instance. This instance is not yet managed by a scope.
     /// </summary>
     /// <returns>A new root transaction.</returns>
-    public ITransaction CreateRootTransaction ()
+    public ClientTransaction CreateRootTransaction ()
     {
       var transaction = ClientTransaction.CreateRootTransaction();
 
       OnTransactionCreated(transaction);
+      return transaction;
+    }
 
-      return transaction.ToITransaction();
+    ITransaction ITransactionFactory.CreateRootTransaction ()
+    {
+      return CreateRootTransaction().ToITransaction();
     }
 
     protected virtual void OnTransactionCreated (ClientTransaction transaction)
--- a/SecurityManager/Core.UnitTests/Domain/SecurityManagerPrincipalTests/Common.cs
+++ b/SecurityManager/Core.UnitTests/Domain/SecurityManagerPrincipalTests/Common.cs
@@ -139,6 +139,12 @@ namespace Remotion.SecurityManager.UnitTests.Domain.SecurityManagerPrincipalTest
 
       Assert.That(deserializedPrincipal.Substitution.ID, Is.EqualTo(principal.Substitution.ID));
       Assert.That(deserializedPrincipal.Substitution, Is.Not.SameAs(principal.Substitution));
+
+      Assert.That(() => deserializedPrincipal.GetTenants(true), Throws.Nothing);
+      Assert.That(() => deserializedPrincipal.GetTenants(false), Throws.Nothing);
+      Assert.That(() => deserializedPrincipal.GetTenants(false), Is.Not.Empty);
+      Assert.That(() => deserializedPrincipal.GetActiveSubstitutions(), Throws.Nothing);
+      Assert.That(() => deserializedPrincipal.GetActiveSubstitutions(), Is.Not.Empty);
     }
 
     [Test]
--- a/SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
+++ b/SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using JetBrains.Annotations;
 using Remotion.Context;
 using Remotion.Data.DomainObjects;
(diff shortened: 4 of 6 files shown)


a-ctor (approved):

Remotion/Data/DomainObjects.UnitTests/ClientTransactionFactoryTest.cs
```diff
@@ -54,10 +54,9 @@ public void CreateRootTransaction_WithExtension ()
           .Callback((ClientTransaction transaction) => transaction.Extensions.Add(extensionStub.Object))
           .Verifiable();
 
-      ITransaction transaction = factory.Object.CreateRootTransaction();
+      var transaction = factory.Object.CreateRootTransaction();
```
MichaelKetting, line 57: This test verifies that the interface-version of the API works correctly. Therefor, we should keep the original test and instead do a cast:
((IClientTransactionFactory)factory.Object)

Remotion/Data/DomainObjects.UnitTests/ClientTransactionFactoryTest.cs
```diff
@@ -74,11 +73,11 @@ public void CreateChildTransaction_WithExtension ()
           .Callback((ClientTransaction transaction) => transaction.Extensions.Add(extensionStub.Object))
           .Verifiable();
 
-      ITransaction rootTransaction = factory.Object.CreateRootTransaction();
-      ITransaction childTransaction = rootTransaction.CreateChild();
+      var rootTransaction = factory.Object.CreateRootTransaction();
```
MichaelKetting, line 76: Same change as CreateRootTransaction_WithExtension

Remotion/Data/DomainObjects/ClientTransactionFactory.cs
```diff
@@ -19,20 +19,25 @@
...
   /// <summary>The <see cref="ClientTransactionFactory"/> is the default implementation of the <see cref="ITransactionFactory"/> interface.</summary>
+  /// <threadsafety static="true" instance="true"/>
   [Serializable]
   public class ClientTransactionFactory : ITransactionFactory
   {
+    ITransaction ITransactionFactory.CreateRootTransaction ()
+    {
+      return CreateRootTransaction().ToITransaction();
+    }
+
```
MichaelKetting, line 30: I think I usually put the explicit interface method after the public API (but before any protected virtual implementations). I can live with both.

SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
```diff
@@ -120,6 +127,8 @@ public static ISecurityManagerPrincipal Current
 
       _domainRevision = GetDomainRevision();
       _userRevision = GetUserRevision(_securityPrincipal.User);
+
+
```
MichaelKetting, line 131:
```suggestion
```

I found some whitespace you might be looking for somewhere :)

SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
```diff
@@ -109,6 +112,10 @@ public static ISecurityManagerPrincipal Current
       _substitutedUserHandle = substitutedUserHandle;
       _substitutedRoleHandles = substitutedRoleHandles;
 
+      _securityTransactionFactory =  new Lazy<SecurityClientTransactionFactory>(
```
MichaelKetting, line 115:
```suggestion
      _securityTransactionFactory = new Lazy<SecurityClientTransactionFactory>(
```

SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
```diff
@@ -85,6 +85,9 @@ public static ISecurityManagerPrincipal Current
     private readonly SubstitutionProxy? _substitutionProxy;
     private readonly SecurityPrincipal _securityPrincipal;
 
+    [NonSerialized]
+    private Lazy<SecurityClientTransactionFactory> _securityTransactionFactory;
+
```
MichaelKetting, line 90: This design doesn't satisfy the non-serialized + lazy-init concept. 

NonSerialized fields are NULL when the type is deserialized. One approach (probably the easiest, not sure if it's the only one) is to use a doublechecked lock. 

field: 
```
[NonSerialized]
private SecurityClientTransactionFactory _securityTransactionFactory;
```

in `private ClientTransaction CreateClientTransaction ()` 
```
  if (_securityTransactionFactory == null)
  {
    lock (_lockObject)
    {
      _securityTransactionFactory = new SecurityClientTransactionFactory();
     }
  }
  _securityTransactionFactory.Create();
```

Since we don't want to carry with us a lockobject, we can also do an Interlocked instead in `private ClientTransaction CreateClientTransaction ()` 
```
  if (_securityTransactionFactory == null)
  {
      var securityTransactionFactory = new SecurityClientTransactionFactory();
      Interlocked.CompareExchange(ref _securityTransactionFactory, securityTransactionFactory, null);
  }
  _securityTransactionFactory.Create();
```
This version will on occasion create an extra Factory instance, but that's okay. We don't care about this racy overhead.

In theory we could also use a MemoryBarrier but it is more explicit this way.

MichaelKetting: Note: in .NET 8, they make Interlocked.CompareExchange faster:
https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/#regex

MichaelKetting (changes requested):
I moved this over to develop. We shouldn't have an issue, there's just one ticket that depends on this and RM-8987 should be done soon, anyway.

SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
MichaelKetting: @a-ctor FYI missed a serialization bug.

Remotion/Data/DomainObjects.UnitTests/ClientTransactionFactoryTest.cs
```diff
@@ -54,7 +54,7 @@ public void CreateRootTransaction_WithExtension ()
           .Callback((ClientTransaction transaction) => transaction.Extensions.Add(extensionStub.Object))
           .Verifiable();
 
-      ITransaction transaction = factory.Object.CreateRootTransaction();
+      var transaction = ((ITransactionFactory)factory.Object).CreateRootTransaction();
```
MichaelKetting, line 57:
```suggestion
      ITransaction transaction = ((ITransactionFactory)factory.Object).CreateRootTransaction();
```

Remotion/Data/DomainObjects.UnitTests/ClientTransactionFactoryTest.cs
```diff
@@ -74,8 +74,8 @@ public void CreateChildTransaction_WithExtension ()
           .Callback((ClientTransaction transaction) => transaction.Extensions.Add(extensionStub.Object))
           .Verifiable();
 
-      ITransaction rootTransaction = factory.Object.CreateRootTransaction();
-      ITransaction childTransaction = rootTransaction.CreateChild();
+      var rootTransaction = ((ITransactionFactory)factory.Object).CreateRootTransaction();
+      var childTransaction = rootTransaction.CreateChild();
```
MichaelKetting, line 78:
```suggestion
      ITransaction rootTransaction = ((ITransactionFactory)factory.Object).CreateRootTransaction();
      ITransaction childTransaction = rootTransaction.CreateChild();
```
The interface is a relevant part of the tested semantics.

SecurityManager/Core/Domain/SecurityManagerPrincipal.cs
MichaelKetting: @Frazerus just FYI: I managed to mangle to double-checked lock pattern in the first example. It goes `if value == null` then `lock` followed by `if value == null` then `value = new`
That's why it's "double-checked" lock.

MichaelKetting (changes requested):
I managed to forget one bit in the last review: we need to add a test for the serialization of SecurityManagerPrincipal. Given there was a null-refernece exception that was missed should have been the dead giveaway.

A test for serializating is in Remotion.SecurityManager.UnitTests.Domain.SecurityManagerPrincipalTests/Common.

We should amend it by calling GetActiveSubstitutions and/or GetTenants on the deserialized object to document that there's no null-reference after deserialization. Alternatively (maybe better, don't know) copy the Serialization test over to the testfiles for these two methods and exercise the relevant APIs there on the deserialized object.

SecurityManager/Core.UnitTests/Domain/SecurityManagerPrincipalTests/Common.cs
```diff
@@ -139,6 +139,10 @@ public void Serialization ()
 
       Assert.That(deserializedPrincipal.Substitution.ID, Is.EqualTo(principal.Substitution.ID));
       Assert.That(deserializedPrincipal.Substitution, Is.Not.SameAs(principal.Substitution));
+
+      Assert.That(() => deserializedPrincipal.GetTenants(true), Throws.Nothing);
+      Assert.That(() => deserializedPrincipal.GetTenants(false), Throws.Nothing);
+      Assert.That(() => deserializedPrincipal.GetActiveSubstitutions(), Throws.Nothing);
```
MichaelKetting, line 145:
```suggestion
      Assert.That(() => deserializedPrincipal.GetTenants(true), Throws.Nothing);
      Assert.That(() => deserializedPrincipal.GetTenants(false), Throws.Nothing);
      Assert.That(() => deserializedPrincipal.GetActiveSubstitutions(), Is.Not.Empty);
```

There's active substitutions based on the line 126. It would be nice to also check for Not-Empty on GetTenants. I would assume atleast one of those methods also returns a non-empty value. Didn't check, though.

MichaelKetting (approved):

PR #899, opened 2023-10-25
RM-9005 Fix BocList client script error on empty table

a-ctor:
https://re-motion.atlassian.net/browse/RM-9005

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
@@ -514,7 +514,10 @@ class BocList
     var tableBlock = bocList.querySelector(':scope > div.bocListTableBlock')!;
 
     var scrollTimer: Nullable<number> = null;
-    var tableContainer = tableBlock.querySelector<HTMLElement>(':scope > div.bocListTableContainer')!;
+    const tableContainer = tableBlock.querySelector<HTMLElement>(':scope > div.bocListTableContainer');
+    if (!tableContainer) // In the case of an empty BocList we don't have a table container
+      return;
+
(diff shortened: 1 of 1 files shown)


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -514,7 +514,10 @@ class BocList
     var tableBlock = bocList.querySelector(':scope > div.bocListTableBlock')!;
 
     var scrollTimer: Nullable<number> = null;
-    var tableContainer = tableBlock.querySelector<HTMLElement>(':scope > div.bocListTableContainer')!;
+    const tableContainer = tableBlock.querySelector<HTMLElement>(':scope > div.bocListTableContainer');
+    if (!tableContainer) // In the case of an empty BocList we don't have a table container
+      return;
+
```
martin-herwich, line 520: tableContainer is now of type `HTMLElement?`, but `BocList.FixHeaderPosition` expects the non-nullable `HTMLElement` - see Build Log.

martin-herwich (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
a-ctor: Yes this was an initial problem that should be fixed now with the change from `var` to `const`. Since the variable is constant, its type gets "changed" to `HTMLElement` after the null check in the next line. I am re-running the build as it failed because of an external issue

martin-herwich: Ah, I see.

martin-herwich (approved):

PR #913, opened 2023-11-09
RM-8988 Remove SecurityConfiguration

Frazerus:
https://re-motion.atlassian.net/browse/RM-8988


Remotion/Security/Core/Configuration/SecurityConfiguration.cs
```diff
@@ -1,69 +0,0 @@
-// This file is part of the re-motion Core Framework (www.re-motion.org)
```
MichaelKetting, line 1: schema-file removal is missing

MichaelKetting (approved):

PR #933, opened 2023-11-14
RM-8975 Enable headless mode for the screenshot tests

Frazerus:
https://re-motion.atlassian.net/browse/RM-8975


Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
```diff
@@ -52,22 +52,29 @@ protected IDriver Driver
     public void IntegrationTestOneTimeSetUp ()
```
MichaelKetting, line 52: Not sure if this is done as intended. 

Before, the OneTimeSetup decided based on the fixture declaration. Now, we must decide based on the individual test declaration. 

For now, the idea is that we run each test assembly once, and only run those tests in head-mode that do not support headless. In the future, we can run the test assemblies twice. Once with headless and once with head, each executing only their own subset of tests.

The check as it is implemented now does not work with the require-head attribute applied to tests. The onetime-setup is not executed per test, only per testfixture.

With IntegrationTestSetup, this is different, here we can do it per test. The problem is, that the webtesthelper is already configured for headless. We may need to take a step back and updated the design concept in this type.

Frazerus: As for how I understood the ticket: We want to ignore the tests marked with `[RequiresUserInterface]` if we are in headless mode.
But we also want functionality to make the tests in a non-headless mode so we can only run the test which do require it?

MichaelKetting: I believe this this is meant to describe a future desired state. The current step is only meant to make sure only indivdual tests that require a UI are run with head. Before, we could only switch this based on the TestFixture-level but not on the level of individual tests.

The thing is, our build right now runs with headless-by-default and individual testfixtures opt out. if we implement this task as ignoring tests, we would ignore all tests that require head in the build without complensating measures. Therefor, this cannot be the intent :)

MichaelKetting:
> As of the current changes, the two integrationtest baseclasses are quite similar, it might be beneficial to have a baseclass for both? Unless there's a reason for not doing that in the first place...

@Frazerus i.e. move it into the IntegrationTests.Infrastructure project. Perhaps. Depends on whether there's enough similiarity and that it's not just accidental similiarity but should be maintained identically. If this is decideded, it would be a separate task. I don't have a preference as of now. @a-ctor thoughts?

a-ctor (commented):
We discussed some stuff on Friday that I think is not yet in this PR? Can you re-request me please once the changes are worked in? :)

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/App.config
```diff
@@ -0,0 +1,70 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+
+<configuration>
```
a-ctor, line 3: We need to add this new project to `Project.props` for it to run on CI. We also need to configure it so that it does not run in a docker images as that would prevent it from going non-headless. We should be able to use the configurations define at the top of `Project.props` to do this, although I am not sure how it works exactly :D

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/App.config
```diff
@@ -0,0 +1,70 @@
...
+  <remotion.webTesting
+      xmlns="http://www.re-motion.org/WebTesting/Configuration/2.0"
+      browser="Chrome"
+      searchTimeout="00:00:30"
+      downloadStartedTimeout="00:00:10"
+      downloadUpdatedTimeout="00:00:10"
+      verifyWebApplicationStartedTimeout="00:00:10"
+      retryInterval="00:00:00.025"
+      asyncJavaScriptTimeout="00:00:10"
+      headless="true"
```
a-ctor, line 26:
```suggestion
      headless="false"
```

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/DiagnosticInformationCollectioningRequestErrorDetectionStrategy.cs
```diff
@@ -0,0 +1,83 @@
...
+using System;
+using Coypu;
+using Remotion.Web.Development.WebTesting.RequestErrorDetectionStrategies;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.RequireUI
+{
+  /// <summary>
+  /// Testclass to be able to seperate Tests for <see cref="IRequestErrorDetectionStrategy"/> implementation and calls.
+  /// </summary>
+  public class DiagnosticInformationCollectioningRequestErrorDetectionStrategy : IRequestErrorDetectionStrategy
```
a-ctor, line 26: Duplicating this is not ideal since it is quite a bit of code. We do have an infrastructure project into which we could move this class. This way, we don't have to duplicate the strategy. Calling @MichaelKetting to confirm that this is a good approach :D

Remotion/Web/Development.WebTesting.IntegrationTests/MouseTest.cs
```diff
@@ -31,7 +30,6 @@
 namespace Remotion.Web.Development.WebTesting.IntegrationTests
 {
   [TestFixture]
-  [RequiresUserInterface]
   public class MouseTest : IntegrationTest
```
a-ctor, line 33: Since we added this in the other project we probably want to remove it from this project?

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/Web.Development.WebTesting.IntegrationTests.RequireUI.csproj
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ProjectType>Test</ProjectType>
+  </PropertyGroup>
+  <Import Project="..\..\..\Build\Shared.build.props" />
+  <PropertyGroup>
+    <AssemblyTitle>Remotion WebTesting Framework Integration Tests</AssemblyTitle>
+    <AssemblyName>Remotion.Web.Development.WebTesting.IntegrationTests</AssemblyName>
```
a-ctor, line 9: The details here need to be adapted

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/MouseTest.cs
```diff
@@ -0,0 +1,165 @@
...
+      Thread.Sleep(250);
+      Helper.BrowserConfiguration.MouseHelper.Hover(hoverElement);
+      Thread.Sleep(250);
+
+      Assert.That(GetColor(hoverElement).ToArgb(), Is.EqualTo(Color.Red.ToArgb()));
+    }
+
+    [Explicit("Testing the tooltip is flaky as long as there is no good implementation of .ShowTooltip")]
+    [Test]
+    public void Test_Tooltip ()
```
a-ctor, line 126: This test requires saved screenshots so we need to move the folder over as well.

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/ScreenshotTest.cs
```diff
@@ -0,0 +1,103 @@
...
+using Remotion.Web.Development.WebTesting.PageObjects;
+using Remotion.Web.Development.WebTesting.ScreenshotCreation;
+using Remotion.Web.Development.WebTesting.ScreenshotCreation.Annotations;
+using Remotion.Web.Development.WebTesting.ScreenshotCreation.Fluent;
+using Remotion.Web.Development.WebTesting.WebDriver;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.RequireUI
+{
+  [TestFixture]
+  public class ScreenshotTest : IntegrationTest
```
a-ctor, line 31: These tests require saves screenshots so we should move the relevant files over to this project

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/ScreenshotTest.cs
```diff
@@ -0,0 +1,103 @@
...
+        Helper.BrowserConfiguration.BrowserAnnotateHelper.DrawTooltip(builder, target, GetTooltipStyleForCurrentBrowser());
+
+        builder.Crop(target.ForElementScopeScreenshot(), new WebPadding(30, 60, 30, 60));
+      };
+
+      var home = Start();
+      var element = home.Scope.FindId("tooltipTarget");
+      Helper.RunScreenshotTestExact<ElementScope, ScreenshotTest>(element, ScreenshotTestingType.Both, test);
+    }
+
```
a-ctor, line 51: I think there are two tests missing here? We have 4 deleted methods in the other project but only two of them show up here.

a-ctor (changes requested):

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/App.config
Frazerus: Yeah I'll definitely need help with that.

Frazerus:
Tests can't be started yet, but fixed some things. However, when running locally, IIS doesn't get started for the tests. Any ideas why that might happen?

a-ctor (approved):
I am not sure how the "don't run in docker" is supposed to work and if the current implementation is okay, so I pass the torch over to mike :p

Frazerus:
> Tests can't be started yet, but fixed some things. However, when running locally, IIS doesn't get started for the tests. Any ideas why that might happen?

I still need help with that. @MichaelKetting?

MichaelKetting:
> > Tests can't be started yet, but fixed some things. However, when running locally, IIS doesn't get started for the tests. Any ideas why that might happen?
> 
> I still need help with that. @MichaelKetting?

@a-ctor ?

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/IntegrationTest.cs
```diff
@@ -0,0 +1,106 @@
...
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.RequireUI
+{
+  /// <summary>
+  /// Base class for all integration tests.
+  /// </summary>
+  public abstract class IntegrationTest
```
MichaelKetting, line 28: I'm still missing the rename to IntegrationTestBase, probably not pushed yet.

Remotion/Web/Web.Development.WebTesting.IntegrationTests.RequireUI/BrowserWindowTest.cs
```diff
@@ -0,0 +1,75 @@
...
+      get { return false; }
+    }
+
+    public BrowserHelper BrowserHelper
+    {
+      get { return Helper.BrowserConfiguration.BrowserHelper; }
+    }
+
+    [Test]
+    public void Test_ContentPosition ()
```
MichaelKetting, line 39: I think we should copy the other two tests, too sicne it makes a difference whether we're using UI or headless when moving the window / resizing the window.

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -175,105 +173,6 @@ public void BadgeTest ()
       Helper.RunScreenshotTestExact<IFluentScreenshotElement<ElementScope>, ScreenshotTest>(element.ForElementScopeScreenshot(), ScreenshotTestingType.Both, test);
     }
 
-    /// <summary>
-    /// Tests that the <see cref="ScreenshotTooltipAnnotation"/> is drawn correctly.
-    /// </summary>
-    [Category("Screenshot")]
-    [Test]
-    public void DrawTooltip ()
```
MichaelKetting, line 183: What is DrawTooltip and DrawTooltip_WithFollowingBoxAnnotation doing that require a UI? Don't we also have tooltips manually drawn for headless tests?

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -175,105 +173,6 @@ public void BadgeTest ()
...
-      mouseHelper.Hover(element);
-
-      Helper.RunScreenshotTestExact<ElementScope, ScreenshotTest>(element, ScreenshotTestingType.Both, test);
-
-      // Reset the cursor to the initial position (0, 0).
-      mouseHelper.Hover(Point.Empty);
-    }
-
     [Test]
     public void TooltipBounds_NotDrawnYet_ThrowsInvalidOperationException ()
```
MichaelKetting, line 177: Depending on the outcome of the question of DrawTooltip, this test might also belong to UI-tests.

Remotion/Web/Development.WebTesting.IntegrationTests/TestExecutionScreenshotRecorderTest.cs
```diff
@@ -93,43 +91,6 @@ public void TestExecutionScreenshotRecorderTest_TakeDesktopScreenshot_ReplacesIn
       Assert.That(File.Exists(fullPathWitCharReplaced), Is.True);
     }
 
-    [Test]
-    public void TestExecutionScreenshotRecorderTest_TakeBrowserScreenshot_SavesToCorrectPath ()
```
MichaelKetting, line 97: Why is are some of those tests moved to UI-tests. The ScreenshotRecorder basically works, doesn't it? so the moved tests must to something special that's not supported unless there's UI. From the test name it doesn't look like it's UI dependent / different from the other tests. please explain.

MichaelKetting: Note: it's possible that we wanna duplicate the tests (the entire testfixture) because we want to ensure the tests work in both UI and headless mode.

MichaelKetting (changes requested):
Some questions

Build/Customizations/Projects.props
```diff
@@ -285,7 +285,7 @@
       <TestSetupBuildFile>$(TestSetupBuildFileForWebTest)</TestSetupBuildFile>
     </IntegrationTestProjectFiles>
 
-    <IntegrationTestProjectFiles Include="$(SolutionDirectory)\Remotion\Web\IntegrationTests\Web.IntegrationTests.csproj">
+    <IntegrationTestProjectFiles Include="$(SolutionDirectory)\Remotion\Web\Development.WebTesting.IntegrationTests.RequireUI\Web.Developmnet.WebTesting.IntegrationTests.RequireUI.csproj">
```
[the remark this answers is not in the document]
Frazerus: Huh, seems like in changing this twice or thrice I messed up somewhere in between.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/DiagnosticInformationCollectioningRequestErrorDetectionStrategy.cs
```diff
@@ -18,7 +18,7 @@
 using Coypu;
 using Remotion.Web.Development.WebTesting.RequestErrorDetectionStrategies;
 
-namespace Remotion.Web.Development.WebTesting.IntegrationTests
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure
```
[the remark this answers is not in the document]
Frazerus: The first one is alright, but I think I need some input for the second one. I don't understand where the error detection strategy comes from and (without the current implementation) constantly get errors that the strategy could not be loaded.

Remotion/Web/Development.WebTesting.IntegrationTests/TestExecutionScreenshotRecorderTest.cs
Frazerus: I merely moved all tests which didn't work in headless but did work in headful to the new project (hopefully) without changing implementations. I currently can't say what the reason for this is, will look into it tomorrow.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/DiagnosticInformationCollectioningRequestErrorDetectionStrategy.cs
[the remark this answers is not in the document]
MichaelKetting: error detection strategy:
Short Version:
we should be able to remove `_aspNetRequestErrorDetectionScope` in `IntegrationTest` of the `RequireUI` project since this is not set in ObjectBinding.Web either.

Long version: need to see the errors live.

Remotion/Web/Development.WebTesting.IntegrationTests/TestExecutionScreenshotRecorderTest.cs
MichaelKetting: Yep, it's stuff noticed with a bit of distance :)

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.Shared.cs
```diff
@@ -242,46 +192,31 @@ public void DrawTooltip_WithFollowingBoxAnnotation ()
...
+    {
+      var tooltipAnnotation = new ScreenshotTooltipAnnotation("title", ScreenshotTooltipStyle.Chrome, WebPadding.None);
+
+      Assert.That(
+          () => tooltipAnnotation.TooltipBounds,
+          Throws.InstanceOf<InvalidOperationException>().With.Message.EqualTo("The tooltip needs to be drawn before its bounds can be accessed."));
+    }
     /// <summary>
-    /// Tests that the <see cref="ScreenshotTooltipAnnotation"/> is annotated correctly.
+    /// Tests that the <see cref="ScreenshotBadgeAnnotation"/> is drawn correctly.
```
MichaelKetting, line 205: Since there's a missing blank line anway, please move to original position. makes blaming / DIFF-check eaiser.

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.Shared.cs
```diff
@@ -706,6 +373,8 @@ public void ScreenshotTransformationTest ()
       Helper.RunScreenshotTestExact<ScreenshotTest>(PrepareTest(), ScreenshotTestingType.Browser, test);
     }
 
+
+
```
MichaelKetting, line 377:
```suggestion
```

Remotion/Web/Development.WebTesting.IntegrationTests/TestExecutionScreenshotRecorderTest.cs
```diff
@@ -94,40 +92,42 @@ public void TestExecutionScreenshotRecorderTest_TakeDesktopScreenshot_ReplacesIn
     }
 
     [Test]
```
MichaelKetting, line 94: it looks like the method order got changed. could you please sort it in the original order, thanks!

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/DiagnosticInformationCollectingRequestErrorDetectionStrategy.cs
```diff
@@ -18,20 +18,20 @@
...
 using Remotion.Web.Development.WebTesting.RequestErrorDetectionStrategies;
 
-namespace Remotion.Web.Development.WebTesting.IntegrationTests
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure
 {
   /// <summary>
   /// Testclass to be able to seperate Tests for <see cref="IRequestErrorDetectionStrategy"/> implementation and calls.
   /// </summary>
-  public class DiagnosticInformationCollectioningRequestErrorDetectionStrategy : IRequestErrorDetectionStrategy
+  public class DiagnosticInformationCollectingRequestErrorDetectionStrategy : IRequestErrorDetectionStrategy
```
MichaelKetting, line 26: to make the fture reading of the Feature Branch easer: we should move the changes from the rename to a separate commit. Also, do we still need this in the Infrastructure project now?

MichaelKetting (changes requested):
Some formatting/structuring stuff.

MichaelKetting (commented):
Note: please pick all the changes belonging to the renamed/name-fixed DiagnosticCollection-type into the separate commit. RIght now, only the type itself is isolated, not the code changes in the other files.

MichaelKetting (changes requested):
Marked for changes because of the two left open screenshot questions. the treeview-one we can check ourselves. the BorderElementB test I would like to get @a-ctor's input if this issomething we need to repair or if it's okay.

Remotion/Web/Development.WebTesting.IntegrationTests/SavedTestScreenshots/ScreenshotTest/ResolveBorderElementB.png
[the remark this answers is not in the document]
a-ctor: The new image is 2x2 compared to the previous 3x2 image. This is actually correct as the 1px element is located at the bottom right corner and, as such, should also be in the bottom right corner of the cropped image. Previously, this worked because of the scrollbar and the result was a 3x2 image with slightly gray pixes on the right. In headless we don't have a scrollbar, which removes this inconsistency.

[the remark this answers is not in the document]
MichaelKetting: Ah, thanks, I read this the other way around.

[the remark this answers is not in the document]
Frazerus: image
I don't think this means that the old one was 3x2, this seems very much the other way around. Or is githubs UI just bad there?

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/SavedTestScreenshots/BocTreeViewControlObjectTest/Chrome.any.WebTreeViewScreenshot_WithMarkedBadge0.png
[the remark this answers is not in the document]
MichaelKetting: screen got smaller. in the future (separate task) we should probably set a fixed width for the tree-area so it doesn't always change depending on the widnow size.

Remotion/Web/Development.WebTesting.IntegrationTests/SavedTestScreenshots/ScreenshotTest/ResolveBorderElementB.png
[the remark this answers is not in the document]
MichaelKetting: @Frazerus  I can confirm your finding.
@a-ctor did you check the correct file? there's two of them, one in RequiredUI and one in the non-ui screenshot tests

Frazerus:
All tests work now. By resizing the headless content size to the size we had before going headless, all screenshot test pictures could be reverted. DropDownMenuControlObjectTest required extra size setting due to how it previously worked.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -207,8 +207,9 @@ public void OnSetUp ([NotNull] string testName)
     /// </summary>
     /// <param name="maximizeWindow">Specified whether the new browser session's window should be maximized.</param>
     /// <param name="configurationOverride">Specifies additional options applied when creating the browser.</param>
+    /// <param name="headlessSize">If not set, a value of 1050x868 is set as the browser size for headless windows.</param>
```
MichaelKetting, line 210: value doesn't match the size in line 225. 

The size used in line 225 is specifically for our tests, correct? then we should set it there. 1296x868 does not match a standard screen size such as 1280x1024 because it's wider than 1280. 

Looking further, the maximizeWindow paramerter and the headlessSize parameter are mutually exclusive, aren't they? Not sure how to best handle this, given we have the DriverConfigOverride which controls the headlessmode.

I also noticed that the headless tests still have a maxmizeWindow flag. This is no longer sensible with headless, isn't it? I think a bit more cleanup would be a good thing for splitting the code into UI and headless. We should discuss what is needed and if it should be in here or a separate task.

MichaelKetting (changes requested):
Request Change only because we need to discuss how to handle maximize window.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
Frazerus: Do we want the size to actually match a reasonable screensize? If yes, I would create a ticket for that.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -162,9 +162,9 @@ public IBrowserSession MainBrowserSession
     /// <summary>
     /// SetUp method for each web test fixture.
     /// </summary>
-    /// <param name="maximizeWindow">Specifies whether the main browser session's window should be maximized.</param>
+    /// <param name="windowSize">Specifies the main browser session's window size and whether it should be maximized.</param>
```
MichaelKetting, line 165:
```suggestion
    /// <param name="windowSize">Specifies the main browser session's window size. Maximizes the window if left <see langword="null"/>.</param>
```

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -205,10 +205,10 @@ public void OnSetUp ([NotNull] string testName)
     /// <summary>
     /// Creates a new browser session using the configured settings from <see cref="WebTestConfigurationFactory"/>.
     /// </summary>
-    /// <param name="maximizeWindow">Specified whether the new browser session's window should be maximized.</param>
+    /// <param name="windowSize">Specifies the new browser session's size and whether it should by maximized.</param>
```
MichaelKetting, line 208:
```suggestion
    /// <param name="windowSize">Specifies the main browser session's window size. Maximizes the window if left <see langword="null"/>.</param>
```

Remotion/Web/Development.WebTesting/WindowSize.cs
```diff
@@ -0,0 +1,28 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+namespace Remotion.Web.Development.WebTesting
+{
+  public record WindowSize(int Width, int Height, bool IsMaximized = false)
```
MichaelKetting, line 19: xml doc please. "Describes the size of the browser window."

Remotion/Web/Development.WebTesting/WindowSize.cs
```diff
@@ -0,0 +1,28 @@
...
+  {
+    public static readonly WindowSize Maximized = new WindowSize(-1, -1, true );
+
+    public int Width { get; } = Width;
+    public int Height { get; } = Height;
+
+    public bool IsMaximized { get; } = IsMaximized;
```
MichaelKetting, line 26: probably should also get property comments since this is a web test api

MichaelKetting (approved):
xml doc details.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
MichaelKetting: need to read this comment.

MichaelKetting: This means, to have a good default for headless instead of something the project developer must configure?

Ah, idea! and, yes, can be separate ticket if you want:

in WebTestHelper.CreateNewBrowserSession, we already use maximize if the size is null. We could simply use a fixed size as the null fallback value if it's headless:
line 220: isntead of `windowSize ??= WindowSize.Maximized;` we use `windowSize ??= configuration.Headless ? new WindowSize(1280, 900): WindowSize.Maximize`

PR #950, opened 2023-11-20
RM-9019 Fix incorrect resource URL in BocBooleanValueResourceSet

a-ctor:
https://re-motion.atlassian.net/browse/RM-9019


MichaelKetting (approved):
👀🙃

PR #951, opened 2023-11-21
RM-9031 Add focus highlight to validation error marker in BocList

a-ctor:
https://re-motion.atlassian.net/browse/RM-9031

NovaGray and NovaViso screenshot:

image

NovaGray layouting is already scuffed so the outline itself looks bad.


Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -373,15 +373,21 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
 /*region validation marker*/
-div.bocList .bocListCellStructureElement > .validationErrorMarker
-{
-  margin-top: calc(var(--remotion-themed-spacing) / 2);
-}
-
 div.bocList .bocListCellStructureElement > .validationErrorMarker,
 div.bocList .bocListContent > .validationErrorMarker
 {
+  padding: calc(var(--remotion-themed-spacing) / 2);
```
MichaelKetting, line 379:
```suggestion
  padding: calc(var(--remotion-themed-spacing-small) - var(--remotion-themed-border-width));
```
Also, both versions result in the ValidationColumn Icon to go below the baseline. The selector might be too generic or there is another sideeffect.

image

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -373,15 +373,21 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
   border: var(--remotion-themed-border-placeholder);
+  border-radius: var(--remotion-themed-border-radius);
+}
+
+div.bocList .bocListCellStructureElement > .validationErrorMarker:focus,
+div.bocList .bocListContent > .validationErrorMarker:focus
+{
+  outline: none;
```
MichaelKetting, line 387: TODO: need to review if we use to "correct" optiosn. Probably but needs review.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
a-ctor: A bit unsure of what you mean but I copied this from the following `Common.css` rule:

```css
input:not(:disabled).remotion-themed:focus,
textarea:not(:disabled).remotion-themed:focus,
select:not(:disabled).remotion-themed:focus,
button:not(:disabled).remotion-themed:focus,
.remotion-themed input:not(:disabled):focus,
.remotion-themed textarea:not(:disabled):focus,
.remotion-themed select:not(:disabled):focus,
.remotion-themed button:not(:disabled):focus
{
  outline: none;
  background-color: var(--remotion-themed-background-color-focus);
  border: var(--remotion-themed-border-focus);
  box-shadow: var(--remotion-themed-box-shadow-focus);
}
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
a-ctor: How did I miss that 😅 The first marker was in a `.bocListContent` which already applies a padding so it was added twice.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -373,15 +373,26 @@ div.bocList div.bocListTableBlock div.bocListNavigator a:not([href]) img
...
+{
+  border: var(--remotion-themed-border-placeholder);
+  border-radius: var(--remotion-themed-border-radius);
+}
+
 div.bocList .bocListCellStructureElement > .validationErrorMarker
 {
-  margin-top: calc(var(--remotion-themed-spacing) / 2);
+  /* .bocListContent already comes with padding so we only need this for .bocListCellStructureElement */
+  padding: calc(var(--remotion-themed-spacing) / 2);
```
MichaelKetting, line 386: `padding: calc(var(--remotion-themed-spacing-small) - var(--remotion-themed-border-width));`

I believe this version creates a more precise alignment. Only, it doesn't applies to the validation column icon itself. Could you please check if/wat options we have?

(Without this, the validation icosn are 1px below the selection radio button and the top border of the selection checkbox.

a-ctor: Even with your suggestion the initial problem still exists. The problem is unrelated to this issue but rather a rounding problem that causes the checkbox to move slightly depending on where it is positioned on the screen. If you remove the error messages from the BocList above (see screenshot below) then the positioning is correct again.

image

MichaelKetting: I'm not convinced. we're adding a 1px border and a padding. in other places I believe add a padding minus the border plus a separate border space. This results in a 1px offset.

a-ctor: The diff makes it seem like the border is new but it isn't. If you look at the bottom left of the diff you can see that the "placeholder" border always existed. The border isn't new and the added padding replaces the `margin-top` so there should not be a change in the positioning of the markers. Even on develop the checkboxes and markers were misaligned by one pixel in the body.

I think that we should look at this problem in a separate issue as there is a new interaction that makes it clearer that this is a rounding problem that causes the slight misalignment. Without validation markers we get the following alignment:

image

The top of the checkbox and marker align perfectly. Now here is the screenshot with validation errors enabled:

image

We can see that the all the markers aligned correctly to each other and to the text baseline. But the checkbox is now a pixel above the marker. In fact, the checkbox did not move but rather the text and marker moved down a pixel. To show that this is a rounding error here is a screenshot with the form grid validation messages from the top BocList removed (see my other comment above):

image

Without any CSS changes the alignment is fixed again, just because the location of the bottom BocList has changed. I think that the changes here are good as proposed but that we have to look into the positioning in general if we want to fix the 1px offset in the checkbox. But I would do that as part of a separate issue.

MichaelKetting (approved):
still not convinced but approving this. please create a follow=up ticket or check, we already have a ticket for rounding errors.

my main issue is that the radio button, not the checkbox was off, too.

a-ctor:
@MichaelKetting Both radio and checkbox are off by 1px in regard to the validation marker. But this is already the case on develop. On Firefox I even have a 2px distance between checkbox and marker...

PR #953, opened 2023-11-27
RM-9035 Clenaup for missed details in WebConfiguration tasks

a-ctor:
https://re-motion.atlassian.net/browse/RM-9035

I did not update the `WebConfiguration` to use the .NET Framework 2.0 syntax as I felt that there would be no benefit to it. We are replacing the `ConfigurationManager` implementation anyway so if it works now I think we should just leave it as is.

### DIFF
--- a/Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
+++ b/Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
@@ -81,7 +81,28 @@ namespace Remotion.Web.ExecutionEngine
     {
       UrlMappingFile = urlMappingFile;
       MaximumUrlLength = maximumUrlLength;
-      DefaultWxeHandler = defaultWxeHandler;
+      DefaultWxeHandler = CheckDefaultWxeHandler(defaultWxeHandler);
+    }
+
+    private string? CheckDefaultWxeHandler (string? defaultWxeHandler)
+    {
+      if (string.IsNullOrEmpty(defaultWxeHandler))
+      {
+        return null;
+      }
+      else
+      {
+        defaultWxeHandler = defaultWxeHandler.Trim();
+        ArgumentUtility.CheckNotNullOrEmpty(nameof(defaultWxeHandler), defaultWxeHandler);
+
+        if (defaultWxeHandler.StartsWith("/") || defaultWxeHandler.IndexOf(":") != -1)
+          throw new ArgumentException($"No absolute paths are allowed. Resource: '{defaultWxeHandler}'", nameof(defaultWxeHandler));
+
+        if (!defaultWxeHandler.StartsWith("~/"))
+          defaultWxeHandler = "~/" + defaultWxeHandler;
+
+        return defaultWxeHandler;
+      }
     }
   }
 }
--- a/Remotion/Web/Core/Schemas/WebConfiguration.xsd
+++ b/Remotion/Web/Core/Schemas/WebConfiguration.xsd
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <xs:schema 
     xmlns:xs="http://www.w3.org/2001/XMLSchema"
-    xmlns:c="http://www.re-motion.org/web/configuration" 
+    xmlns:c="http://www.re-motion.org/web/configuration/v2"
     elementFormDefault="qualified" 
     attributeFormDefault="unqualified" 
-    targetNamespace="http://www.re-motion.org/web/configuration">
+    targetNamespace="http://www.re-motion.org/web/configuration/v2">
   <xs:element name="remotion.web">
     <xs:complexType>
       <xs:all>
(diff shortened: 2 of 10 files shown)


Remotion/Web/Core/Schemas/WebConfiguration.xsd
```diff
@@ -4,7 +4,8 @@
     xmlns:c="http://www.re-motion.org/web/configuration" 
     elementFormDefault="qualified" 
     attributeFormDefault="unqualified" 
-    targetNamespace="http://www.re-motion.org/web/configuration">
+    targetNamespace="http://www.re-motion.org/web/configuration"
+    version="2.0">
```
MichaelKetting, line 8: I don't know about this. We usually did 
```suggestion
    targetNamespace="http://www.re-motion.org/web/configuration/v2">
```

which does require updating the config files

Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
```diff
@@ -58,7 +58,7 @@ public static WxeUrlSettings Create (string? urlMappingFile = null, int? maximum
 
     /// <summary> Gets the path to the file holding the URL mapping configuration. </summary>
     /// <value> A string. Defaults to 'UrlMapping.xml'. </value>
-    public string? UrlMappingFile { get; }
+    public string UrlMappingFile { get; }
```
MichaelKetting, line 61: This should remain nullable. null means not here. Empty is not supported. Also, a follow-up task will remove this property completely so no need to change this here.

Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
```diff
@@ -70,7 +70,7 @@ public static WxeUrlSettings Create (string? urlMappingFile = null, int? maximum
     ///   A virtual path, relative to the application root. Will always start with <c>~/</c>.
     ///   Defaults to an empty string if no handler is assigned.
     /// </value>
-    public string? DefaultWxeHandler { get; }
+    public string DefaultWxeHandler { get; }
```
MichaelKetting, line 73: remains nullable.

Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
```diff
@@ -79,9 +79,30 @@ public WxeUrlSettings ()
...
       MaximumUrlLength = maximumUrlLength;
-      DefaultWxeHandler = defaultWxeHandler;
+      DefaultWxeHandler = ValidateDefaultWxeHandler(defaultWxeHandler);
+    }
+
+    private string ValidateDefaultWxeHandler (string? defaultWxeHandler)
+    {
+      if (string.IsNullOrEmpty(defaultWxeHandler))
+      {
+        return string.Empty;
```
MichaelKetting, line 91:
```suggestion
        return null;
```

Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
```diff
@@ -79,9 +79,30 @@ public WxeUrlSettings ()
 
     private WxeUrlSettings (string? urlMappingFile, int maximumUrlLength, string? defaultWxeHandler)
```
MichaelKetting, line 80: unit tests please

Remotion/Web/Core/ExecutionEngine/WxeUrlSettings.cs
```diff
@@ -79,9 +79,30 @@ public WxeUrlSettings ()
...
     {
-      UrlMappingFile = urlMappingFile;
+      UrlMappingFile = urlMappingFile ?? string.Empty;
       MaximumUrlLength = maximumUrlLength;
-      DefaultWxeHandler = defaultWxeHandler;
+      DefaultWxeHandler = ValidateDefaultWxeHandler(defaultWxeHandler);
+    }
+
+    private string ValidateDefaultWxeHandler (string? defaultWxeHandler)
```
MichaelKetting, line 87: not sure about the method naming. wuldn't expect "Validate" to return a modified value.

MichaelKetting (changes requested):
about the config implelemetation: having it without setter would be nice. no chance there, right?

Remotion/Web/Core/Schemas/WebConfiguration.xsd
MichaelKetting: Thanks. And yeah...

MichaelKetting (commented):
Github doesn't yet have the new commit set.

MichaelKetting (approved):
Please also update the breaking change ticket for the config section.

Remotion/Web/Core/Schemas/WebConfiguration.xsd
a-ctor: Just because I saw it with the WXE Url mapping XML schema: There we are using `/1.0` as suffix. Do you want to leave it here with `/v2` or use `/2.0` to keep it in sync?

MichaelKetting: It's "as you like."

The major.minor syntax was meant to convey if there's compatible or incompatible changes to the config. Since Namespace is always a breaking change, the major.minor syntx is kind of useless. It might be easier to read, though.

MichaelKetting (approved):

PR #956, opened 2023-11-29
RM-9000 Replace public access to the storage configuration with IoC

Frazerus:
https://re-motion.atlassian.net/browse/RM-9000
Sorry for only one commit, there were some...rocks. 

Some tests still don't work, I think I need some input for those.


Remotion/Data/DomainObjects/Persistence/Configuration/IStorageSettings.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Gets the default storage provider definition or null if it doesn't exist.
+    /// </summary>
+    StorageProviderDefinition? DefaultStorageProviderDefinition { get; }
+
+    /// <summary>
+    /// Gets all storage provider definitions
+    /// </summary>
+    IReadOnlyCollection<StorageProviderDefinition> StorageProviderDefinitions { get; }
```
a-ctor, line 56: In mikes design those were methods. I am fine with them being properties but we should move them above the methods to keep the ordering.

Remotion/Data/DomainObjects/Persistence/Configuration/StorageProviderDefinition.cs
```diff
@@ -28,27 +27,20 @@ namespace Remotion.Data.DomainObjects.Persistence.Configuration
...
-
-    protected StorageProviderDefinition (string name, IStorageObjectFactory factory)
-        : base(name, new NameValueCollection())
+    protected StorageProviderDefinition (string name, IStorageObjectFactory factory, IReadOnlyCollection<Type>? assignedStorageGroups = null)
     {
       ArgumentUtility.CheckNotNull("factory", factory);
-
+      Name = name;
       _factory = factory;
+      AssignedStorageGroups = assignedStorageGroups ?? new Type[]{};
```
a-ctor, line 43:
```suggestion
      AssignedStorageGroups = assignedStorageGroups ?? Array.Empty<Type>();
```

Remotion/Data/DomainObjects/Persistence/Configuration/StorageProviderDefinition.cs
```diff
@@ -28,27 +27,20 @@ namespace Remotion.Data.DomainObjects.Persistence.Configuration
...
-      var factoryTypeName = GetAndRemoveNonEmptyStringAttribute(config, "factoryType", name, required: true)!;
-      var configuredFactoryType = TypeUtility.GetType(factoryTypeName, throwOnError: true)!;
-      _factory = CreateStorageObjectFactory(configuredFactoryType);
-    }
-
-    protected StorageProviderDefinition (string name, IStorageObjectFactory factory)
-        : base(name, new NameValueCollection())
+    protected StorageProviderDefinition (string name, IStorageObjectFactory factory, IReadOnlyCollection<Type>? assignedStorageGroups = null)
     {
       ArgumentUtility.CheckNotNull("factory", factory);
```
a-ctor, line 40:
```suggestion
      ArgumentUtility.CheckNotNullOrEmpty("name", name);
      ArgumentUtility.CheckNotNull("factory", factory);
```

Remotion/Data/DomainObjects/Persistence/Configuration/StorageProviderDefinition.cs
```diff
@@ -28,27 +27,20 @@ namespace Remotion.Data.DomainObjects.Persistence.Configuration
   /// Defines the configuration for a specific <see cref="StorageProvider"/>. Subclasses of <see cref="StorageProviderDefinition"/> can be 
   /// instantiated from a config file entry.
   /// </summary>
-  public abstract class StorageProviderDefinition: ExtendedProviderBase
+  public abstract class StorageProviderDefinition
   {
+    public string Name { get; }
+
```
a-ctor, line 33: Order private before public. I know that it is kind of strange in property-field mixes if you want to retain the constructor order.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -186,7 +186,7 @@ public virtual IRdbmsPersistenceModelProvider CreateRdbmsPersistenceModelProvide
       var storageTypeInformationProvider = CreateStorageTypeInformationProvider(storageProviderDefinition);
       var storageNameProvider = CreateStorageNameProvider(storageProviderDefinition);
       var persistenceModelProvider = CreateRdbmsPersistenceModelProvider(storageProviderDefinition);
-      var storageProviderDefinitionFinder = new StorageEntityBasedStorageProviderDefinitionFinder();
+      var storageProviderDefinitionFinder = SafeServiceLocator.Current.GetInstance<IStorageSettings>();
```
a-ctor, line 189: I am not sure if we want to replace the `StorageEntityBasedStorageProviderDefinitionFinder` as well as it has nothing to do with the config but rather uses the class definition to get to the storage provider definition. This is something we'll have to check with @MichaelKetting

Remotion/Data/DomainObjects/Persistence/Configuration/IStorageObjectFactoryFactoryExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+
+namespace Remotion.Data.DomainObjects.Persistence.Configuration
+{
+  public static class IStorageObjectFactoryFactoryExtensions
+  {
+    /// <summary>
+    /// Resolves or creates an <see cref="IStorageObjectFactory"/> object based on the <see cref="Type"/> of the generic parameter.
+    /// </summary>
+    public static T Create<T> (this IStorageObjectFactoryFactory storageObjectFactoryFactory)
```
a-ctor, line 26: We usually also add arg checks for extension methods. Even for the `this` parameter as it can be null due to the translation. e.g. `value.DoSomething(1)` is translated to `MyExtensions.DoSomething(value, 1)` so if values is null no null ref exception is thrown and the null values is passed to the extension method

Remotion/Data/DomainObjects/Persistence/Configuration/StorageObjectFactoryFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+  {
+    /// <summary>
+    /// Resolves or creates an <see cref="IStorageObjectFactory"/> object based on the supplied <see cref="Type"/>.
+    /// If the <see cref="IStorageObjectFactory"/> of the requested type cannot be resolved by the service locator, a
+    /// new instance is created.
+    /// </summary>
+    public IStorageObjectFactory Create (Type storageObjectFactoryType)
+    {
+
+      ArgumentUtility.CheckNotNullAndTypeIsAssignableFrom("storageObjectFactoryType", storageObjectFactoryType, typeof(IStorageObjectFactory));
```
a-ctor, line 42:
```suggestion
      ArgumentUtility.CheckNotNullAndTypeIsAssignableFrom("storageObjectFactoryType", storageObjectFactoryType, typeof(IStorageObjectFactory));

```

Remotion/Data/DomainObjects/Persistence/Configuration/IStorageSettings.cs
Frazerus: The previous implementation were also properties, meaning only breaking changes is from the old class to the new one.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
MichaelKetting: Yes, the idea is to clean/unify this api. Small stuff that's related to the rewrite should be included to have a clean new API.

Remotion/Data/DomainObjects/Persistence/Configuration/StorageProviderDefinition.cs
MichaelKetting: actually, styleguide says public before private, that goes for state, too. Right now, there's not many cases where this is mixed. we should discuss this.
@a-ctor , @Frazerus

Remotion/Data/DomainObjects.ObjectBinding.UnitTests/BindableDomainObjectMixinTests/GetDisplayNameTest.cs
```diff
@@ -38,8 +39,11 @@ public override void SetUp ()
 
       _bindablePropertyReadAccessStrategyMock = new Mock<IBindablePropertyReadAccessStrategy>(MockBehavior.Strict);
```
Frazerus, line 41: I think we should really think about how this should be handled in the future. I dont think we should fix this completely here, as more changes to IoC just make this problem worse, and those will possibly happen at some point. Thoughts?

MichaelKetting: introduce a fake obect similiar to ISecurityProvider and set it up in the setup fixture? 

and yes, eventually we need a better way to do test-specific IoC overrides.

Evnetually, ClientTransaction.CreateRootTransaction will (should) get an IoC-based version and we then have the option of instantiating our own as needed.

Frazerus: Well, I'll let you have a look at the different changes and we'll talk then :)

MichaelKetting (commented):
partial review

MichaelKetting (approved):

PR #957, opened 2023-11-29
RM-9025 Support ARIA 1.2 for the BocAutoCompleteReferenceValue

a-ctor:
https://re-motion.atlassian.net/browse/RM-9025

Solution implemented with the reference implementation https://raw.githack.com/w3c/aria-practices/aria1.2-combobox-proposal/examples/combobox/combobox-autocomplete-list.html in mind. Tested with JAWS 2021, JAWS 2023, NVDA, and Narrator. See repro in the issue description.

Open points @MichaelKetting:
 - [x] "Update WCAG samples/doc for combobox behavior." What samples/doc?

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
@@ -85,7 +85,6 @@ namespace Remotion.BocAutoCompleteReferenceValue
         max: number;
         isAutoPostBackEnabled: boolean;
         extraParams: Dictionary<unknown>;
-        combobox: HTMLElement;
         selectListID: string;
         informationPopUpID: string;
         dropDownButtonID: string;
@@ -121,8 +120,10 @@ namespace Remotion.BocAutoCompleteReferenceValue
         },
         formatItem: Options["formatItem"];
         formatMatch: Options["formatMatch"];
+        selectListID: Options["selectListID"];
+        informationPopUpID: Options["informationPopUpID"];
+        dropDownButtonID: Options["dropDownButtonID"];
         parse: Options["parse"];
-        combobox: Options["combobox"];
     };
 
     export type PositioningOptions = {
@@ -234,11 +235,6 @@ namespace Remotion.BocAutoCompleteReferenceValue
             serviceMethodSearch: serviceMethodSearch,
             serviceMethodSearchExact: serviceMethodSearchExact,
             data: null,
-            combobox: null,
-            selectListID: null,
-            informationPopUpID: null,
-            // re-motion: clicking this control will display the dropdown list with an assumed input of '' (regardless of textbox value)
-            dropDownButtonID: null
         }, initialOptions as RequiredOptions);
 
         // if highlight is set to false, replace it with a do-nothing function
@@ -1166,19 +1162,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
                 return;
 
             this.element = document.createElement("div");
-            this.element.setAttribute('role', 'listbox');
-            this.element.setAttribute('id', this.options.selectListID);
             this.element.setAttribute('class', this.options.resultsClass);
             this.element.style.position = 'fixed';
             LayoutUtility.Hide(this.element);
 
             this.input.closest('div, td, th, body')!.appendChild(this.element);
 
-            this.options.combobox.setAttribute('aria-owns', this.options.selectListID);
-            const isAria11 = this.options.combobox !== this.input;
-            if (isAria11) {
-                this.options.combobox.setAttribute('aria-controls', this.options.selectListID);
-            }
+            this.input.setAttribute('aria-controls', this.options.selectListID);
 
             const elementStyle = window.getComputedStyle(this.element);
             this.element.dataset['originalMaxHeight'] = "" + parseInt(elementStyle.maxHeight, 10);
@@ -1209,6 +1199,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
             });
 
             this.list = document.createElement("ul");
+            this.list.setAttribute('role', 'listbox');
+            this.list.setAttribute('id', this.options.selectListID);
             innerDiv.appendChild(this.list);
 
             this.list.addEventListener('mouseover', (event) => {
@@ -1480,8 +1472,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
         public hide() {
             if (this.repositionTimer) 
                 clearTimeout(this.repositionTimer);
-            this.options.combobox.setAttribute("aria-expanded", "false");
-            this.input.removeAttribute("aria-activedescendant");
+            this.input.setAttribute("aria-expanded", "false");
+            this.input.setAttribute("aria-activedescendant", "");
             this.element && LayoutUtility.Hide(this.element);
             if (this.listItems) {
                 for (const listItem of this.listItems) {
@@ -1520,7 +1512,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
             this.applyPositionToDropDown();
             
             LayoutUtility.Show(this.element);
-            this.options.combobox.setAttribute("aria-expanded", "true");
+            this.input.setAttribute("aria-expanded", "true");
             
             // re-motion: reposition element 
             if (this.repositionTimer) 
@@ -1617,8 +1609,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
             LayoutUtility.Hide(this.element);
             this.input.closest('div, td, th, body')!.appendChild(this.element);
 
-            if (this.options.combobox.getAttribute('aria-labelledby') !== null) {
-                this.element.setAttribute("aria-labelledby", this.options.combobox.getAttribute('aria-labelledby')!);
+            if (this.input.getAttribute('aria-labelledby') !== null) {
+                this.element.setAttribute("aria-labelledby", this.input.getAttribute('aria-labelledby')!);
             }
 
             const elementStyle = window.getComputedStyle(this.element);
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
@@ -160,11 +160,6 @@ namespace Remotion.ObjectBinding.Web.UI.Controls.BocReferenceValueImplementation
       var script = new StringBuilder(1000);
       script.Append("new BocAutoCompleteReferenceValue(");
       script.AppendFormat("'{0}', ", renderingContext.Control.ClientID);
-      script.AppendFormat(
-          "'#{0} span[{1}={2}]', ",
-          renderingContext.Control.ClientID,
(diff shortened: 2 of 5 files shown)


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -234,11 +235,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
             serviceMethodSearch: serviceMethodSearch,
             serviceMethodSearchExact: serviceMethodSearchExact,
             data: null,
-            combobox: null,
-            selectListID: null,
-            informationPopUpID: null,
             // re-motion: clicking this control will display the dropdown list with an assumed input of '' (regardless of textbox value)
-            dropDownButtonID: null
```
MichaelKetting, line 241: Why did you remove those?
And if removeing it is sensible, why leave the comment?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1480,13 +1472,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
-            this.options.combobox.setAttribute("aria-expanded", "false");
-            this.input.removeAttribute("aria-activedescendant");
+            this.input.setAttribute("aria-expanded", "false");
+            this.input.setAttribute("aria-activedescendant", "");
             this.element && LayoutUtility.Hide(this.element);
             if (this.listItems) {
                 for (const listItem of this.listItems) {
                     listItem.classList.remove(this.CLASSES.ACTIVE);
-                    listItem.setAttribute("aria-selected", "false")
+                    listItem.removeAttribute("aria-selected")
```
MichaelKetting, line 1481: why do you remove it instead of setting it `false`? same at the other places where it gets removed.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1480,13 +1472,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
         public hide() {
             if (this.repositionTimer) 
                 clearTimeout(this.repositionTimer);
-            this.options.combobox.setAttribute("aria-expanded", "false");
-            this.input.removeAttribute("aria-activedescendant");
+            this.input.setAttribute("aria-expanded", "false");
+            this.input.setAttribute("aria-activedescendant", "");
```
MichaelKetting, line 1476: need to check if removing or empty is the "correct" choice here.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1166,19 +1163,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
             this.element = document.createElement("div");
-            this.element.setAttribute('role', 'listbox');
-            this.element.setAttribute('id', this.options.selectListID);
             this.element.setAttribute('class', this.options.resultsClass);
             this.element.style.position = 'fixed';
             LayoutUtility.Hide(this.element);
 
             this.input.closest('div, td, th, body')!.appendChild(this.element);
 
-            this.options.combobox.setAttribute('aria-owns', this.options.selectListID);
```
MichaelKetting, line 1177: not sure if removing the `aria-owns` is correct.

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
```diff
@@ -375,7 +364,17 @@ private TextBox GetTextBox (BocRenderingContext<IBocAutoCompleteReferenceValue>
       textBox.Page = renderingContext.Control.Page!.WrappedInstance;
       textBox.ApplyStyle(renderingContext.Control.CommonStyle);
       renderingContext.Control.TextBoxStyle.ApplyStyle(textBox);
+
+      textBox.Attributes.Add(HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.Combobox);
       textBox.Attributes.Add(HtmlTextWriterAttribute2.AriaAutoComplete, HtmlAriaAutoCompleteAttributeValue.Both);
+      textBox.Attributes.Add(HtmlTextWriterAttribute2.AriaExpanded, HtmlAriaExpandedAttributeValue.False);
+      textBox.Attributes.Add(HtmlTextWriterAttribute2.AriaHasPopup, HtmlAriaHasPopupAttributeValue.True);
```
MichaelKetting, line 371: why did you change this from a listbox to a boolean value?

Remotion/Web/Core/UI/HtmlAriaHasPopupAttributeValue.cs
```diff
@@ -28,5 +28,8 @@ public static class HtmlAriaHasPopupAttributeValue
     public const string Listbox = "listbox";
     public const string Menu = "menu";
     public const string Tree = "tree";
+
+    public const string True = "true";
+    public const string False = "false";
```
MichaelKetting, line 33: I believe those are incorrect.

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
```diff
@@ -375,7 +364,17 @@ private TextBox GetTextBox (BocRenderingContext<IBocAutoCompleteReferenceValue>
...
+      textBox.Attributes.Add(HtmlTextWriterAttribute2.AriaControls, GetDropDownButtonName(renderingContext));
```
MichaelKetting, line 372: this one is new. please explain why the textbox has a controls relationship with the button.

MichaelKetting (changes requested):
I got a couple of questions. some stylisc, some about the aria-1.2 standard.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: Whoops forgot about the comment. I had TypeScript compilation problems because the properties are actually required but were registered as optional here. No idea why this has become a problem just now.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: Same here the reference implementation just clears the value and it is always there to begin with.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: > This is essentially returning to the 1.0 pattern with one small modification -- use aria-controls to specify the input-popup relationship instead of aria-owns.

See https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues. If I understand correctly, `aria-controls` replaces `aria-owns` completely in ARIA 1.2 implemenation

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
a-ctor: That's just a mistake that is fixed by the JavaScript code :p it should reference the result set instead

MichaelKetting: So, I found a mistake in the review that will get fixed :) It was simply compensated for already and now should also get cleaned up. 
(just summarizing to make sure we are talking about the same here)

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
MichaelKetting: yeah, no. I believe there's still the four options available and it's possible `true` was also always an option but "listbox" is more descriptive about the intent. 

> Elements with the role combobox have an implicit [aria-haspopup](https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup) value of listbox. If the combobox popup element has a role other than [listbox](https://www.w3.org/TR/wai-aria-1.2/#listbox), authors MUST specify a value for [aria-haspopup](https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup) that corresponds to the role of its popup.

So, it's actually redundant but specifying it makes things explicit. I like explicit. Also, maybe one day we have a grid in there or a tree. then we need the other options, too :)

Reading further:
> To provide backward compatibility with ARIA 1.0 content, user agents MUST treat an `aria-haspopup` value of `true` as equivalent to a value of `menu`.

We do not want `menu`.

MichaelKetting (commented):
believe all questions have been answered now.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1617,8 +1609,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
             LayoutUtility.Hide(this.element);
             this.input.closest('div, td, th, body')!.appendChild(this.element);
 
-            if (this.options.combobox.getAttribute('aria-labelledby') !== null) {
-                this.element.setAttribute("aria-labelledby", this.options.combobox.getAttribute('aria-labelledby')!);
+            if (this.input.getAttribute('aria-labelledby') !== null) {
+                this.element.setAttribute("aria-labelledby", this.input.getAttribute('aria-labelledby')!);
```
MichaelKetting, line 1613: wait a second, I think this bit of code can simply go away? orginally, it was because I already rendered the aria-labelledby onto the input box and then moved it up to the span. what is this code doing now? please check and either explain or remove because redundant.

MichaelKetting (approved):
I got one final cleanup bit which we should check please. only need to see/discuss if there's a surprise hidden in there.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: This code relates to the info-box and moves the `aria-labelledby` to the opened infobox. I do think that this is still relevant.

MichaelKetting: Then we're done.

PR #980, opened 2023-12-13
feature/RM-8577-Update-log4net-v2.28

KuJo7:
https://re-motion.atlassian.net/browse/RM-8577


KuJo7:
@MichaelKetting pls force merge, because the builds will not run

MichaelKetting (approved):

PR #986, opened 2023-12-18
CSP-support for inline scripts and Javascript event handlers

KUJO7:
https://re-motion.atlassian.net/browse/RM-8550


Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,119 @@
...
+            { "onclick", "click" },
+            { "onchange", "change" },
+            { "onmouseover", "mouseover" },
+            { "onmouseout", "mouseout" },
+            { "onkeyup", "keyup" },
+            { "onkeydown", "keydown" },
+            { "onkeypress", "keypress" }
+        };
+
+    private readonly List<Tuple<string, string>> _registeredEvents = new();
```
MichaelKetting, line 59: we discussed usign a ValueTuple here. please update. while doing so, please introduce names for item1 and item2 as this is supported by c# on ValueTuple.. Makes for better readability.

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,119 @@
...
+
+    public override void AddAttribute (string name, string? value)
+    {
+      if (name.StartsWith("on", StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(value))
+      {
+        if (s_supportedEvents.TryGetValue(name, out var eventType))
+        {
+          if (_registeredEvents.Exists(e => e.Item1 == eventType))
+          {
+            throw new NotSupportedException($"{name} cannot be added more than once.");
```
MichaelKetting, line 102: Message: 
`Event handler '{name}' cannot be registered more than once.`

Values in messages should usually be wrapped in quotes to make them easily distinguishable from the text body.

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,119 @@
...
+      base.RenderBeginTag(tagKey);
+    }
+
+    public override void AddAttribute (string name, string? value)
+    {
+      if (name.StartsWith("on", StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(value))
+      {
+        if (s_supportedEvents.TryGetValue(name, out var eventType))
+        {
+          if (_registeredEvents.Exists(e => e.Item1 == eventType))
```
MichaelKetting, line 100: Hmm... 
```suggestion
          if (_registeredEvents.Exists(eventType.Equals))
```
should work.

if not, please use `e => eventType.Equals(e.Item1)`
The two are semantically identical to the above code, it's just a bit more precise which API we are meant to use. The most compact version also saves two method calls.

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,119 @@
...
+          {
+            throw new NotSupportedException($"{name} cannot be added more than once.");
+          }
+
+          _registeredEvents.Add(new Tuple<string, string>(eventType, value));
```
MichaelKetting, line 105: here we change to ValueTuple and use named items.

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+        else
+        {
+          throw new NotSupportedException($"{name} is not supported.");
```
MichaelKetting, line 109: ArgumentException with message 
`The name of attribute '{name}' indicates a script event but the event type is not supported by `CspEnabledHtmlTextWriter`. 
pass name and value in the arugment exception ctor.

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -23,6 +23,10 @@
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
 using Remotion.Web.Compilation;
+#if !NETFRAMEWORK
```
MichaelKetting, line 26: this class will get rewritten by me before merging. please leave comment open.

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
```diff
@@ -0,0 +1,205 @@
...
+using Moq;
+using NUnit.Framework;
+using Remotion.Web.UnitTests.Core.UI.Controls;
+using Remotion.Web.ContentSecurityPolicy;
+using Remotion.Web.UI;
+
+namespace Remotion.Web.UnitTests.Core.ContentSecurityPolicy
+{
+  [TestFixture]
+  public class CspEnabledHtmlTextWriterTests
```
MichaelKetting, line 29: Test - singluar. Test fixtures are not names as plural.

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
```diff
@@ -0,0 +1,205 @@
...
+      var element2 = _htmlHelper.GetAssertedChildElement(root, "button", 1);
+      _htmlHelper.AssertAttribute(element2, "data-inline-event-target", "eventTargetID2");
+
+      _clientScriptMock.Verify();
+    }
+
+    [Test]
+    public void AddAttribute_WithNotSupportedEventType_ThrowsNotSupportedException ()
+    {
+      Assert.That(
```
MichaelKetting, line 185: note: when changing the exception type, we have helper for assering argument exception messages to make it easy when handling parmaeter names.

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    public void AddAttribute_WithMoreThanOneTag_NonceValueIsRendered ()
+    {
+      _randomNumberGeneratorMock
+          .SetupSequence(m => m.GenerateAlphaNumericNonce())
+          .Returns("eventTargetID1")
+          .Returns("eventTargetID2");
+
+      _clientScriptMock
+          .Setup(m => m.RegisterStartupScriptBlock(_pageMock.Object, typeof(CspEnabledHtmlTextWriter), It.IsAny<string>(), It.IsAny<string>()))
+          .Verifiable();
```
MichaelKetting, line 155: There should be qualified expections set up. Two calls, two eventIDs, two script-strings.

consdering the rest of the setup, we might actually want to move this to line 179 and use only two verify statements with the expected parameters instead of setting it up before the calls. The API returns void and by asserting it at the end / next to the htmlheler-assertAttribute calls, we can better establish the relationship

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    public void AddAttribute_WithMoreThanOneAttribute_NonceValueIsRendered ()
+    {
+      _randomNumberGeneratorMock
+          .Setup(m => m.GenerateAlphaNumericNonce())
+          .Returns("eventTargetID")
+          .Verifiable();
+
+      _clientScriptMock
+          .Setup(m => m.RegisterStartupScriptBlock(_pageMock.Object, typeof(CspEnabledHtmlTextWriter), It.IsAny<string>(), It.IsAny<string>()))
+          .Verifiable();
```
MichaelKetting, line 130: see comment in AddAttribute_WithMoreThanOneTag_NonceValueIsRendered

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
```diff
@@ -0,0 +1,205 @@
...
+      _randomNumberGeneratorMock.Verify(m => m.GenerateAlphaNumericNonce(), Times.Once());
+    }
+
+    [Test]
+    public void AddAttribute_WithMoreThanOneTag_NonceValueIsRendered ()
+    {
+      _randomNumberGeneratorMock
+          .SetupSequence(m => m.GenerateAlphaNumericNonce())
+          .Returns("eventTargetID1")
+          .Returns("eventTargetID2");
```
MichaelKetting, line 151: This is a stub. please rename.

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
```diff
@@ -0,0 +1,205 @@
...
+      _writer.AddAttribute("onchange", "console.info('test2');");
+      _writer.RenderBeginTag(HtmlTextWriterTag.Button);
+      _writer.RenderEndTag();
+
+      var document = _htmlHelper.GetResultDocument();
+      var element = _htmlHelper.GetAssertedChildElement(document, "button", 0);
+      _htmlHelper.AssertAttribute(element, "data-inline-event-target", "eventTargetID");
+
+      _clientScriptMock.Verify();
+      _randomNumberGeneratorMock.Verify(m => m.GenerateAlphaNumericNonce(), Times.Once());
```
MichaelKetting, line 142: drop, not needed.

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTests.cs
MichaelKetting, line 130: There should be two verifies, once for each attribute.

MichaelKetting (changes requested):

Remotion/Web/Core/ContentSecurityPolicy/NonceGenerator.cs
```diff
@@ -0,0 +1,40 @@
...
+#if !NETFRAMEWORK
+using System;
+using System.Security.Cryptography;
+
+namespace Remotion.Web.ContentSecurityPolicy
+{
+  /// <summary>
+  /// Implements <see cref="INonceGenerator"/> to generate alpha numeric nonce values.
+  /// </summary>
+  public class NonceGenerator : INonceGenerator
```
MichaelKetting, line 25: please add a unit test for this. we should excercise the code, make sure it doesn't fail. given how the test works with random data, please suggest a way to test this randomness.

MichaelKetting, line 25: please add a threadsafety xmldoc (see other usages in the code). We want to document this class will be threadsafe for instances and statics.

Remotion/Web/Core/ContentSecurityPolicy/INonceGenerator.cs
```diff
@@ -0,0 +1,29 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+#if !NETFRAMEWORK
+using System;
+
+namespace Remotion.Web.ContentSecurityPolicy
+{
+  /// <summary>
+  /// Represents an API for generating alpha numeric nonce values.
+  /// </summary>
+  public interface INonceGenerator
```
MichaelKetting, line 24: please add a threadsafety xmldoc to the itnerface. see other occurances in the code. we want to document this interface should always be implemented threadsafe.
please add a seealso xmldoc (see other occurances) and reference the implementation. We do this to make it visually easy to find the implementations (at least when they're located in the same assembly)

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+
+      if (_registeredEvents.Count > 0)
+      {
+        var eventTargetID = _nonceGenerator.GenerateAlphaNumericNonce();
+        AddAttribute("data-inline-event-target", eventTargetID);
+
+        foreach (var registeredEvent in _registeredEvents)
+        {
+
+          var script = $"SmartPageContext.Instance.RegisterInlineEvent(\"{eventTargetID}\", \"{registeredEvent.type}\", function (event){{{registeredEvent.value}}});";
```
MichaelKetting, line 76: I always use single quotes for rendered Javascript. that way, I don't need to escape the quotes in the script string.

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      if (_registeredEvents.Count > 0)
+      {
+        var eventTargetID = _nonceGenerator.GenerateAlphaNumericNonce();
+        AddAttribute("data-inline-event-target", eventTargetID);
+
+        foreach (var registeredEvent in _registeredEvents)
+        {
+
```
MichaelKetting, line 75: remove extra blank line

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+          { "onclick", "click" },
+          { "onchange", "change" },
+          { "onmouseover", "mouseover" },
+          { "onmouseout", "mouseout" },
+          { "onkeyup", "keyup" },
+          { "onkeydown", "keydown" },
+          { "onkeypress", "keypress" }
+        };
+
+    private readonly List<(string type, string value)> _registeredEvents = new();
```
MichaelKetting, line 43:
```suggestion
    private readonly List<(string Type, string Value)> _registeredEvents = new();
```

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+            throw new InvalidOperationException($"Event handler '{name}' cannot be registered more than once.");
+          }
+
+          _registeredEvents.Add((eventType, value));
+        }
+        else
+        {
+          throw new ArgumentException(
+              $"The name of attribute '{name}' indicates a script event but the event type is not supported by CspEnabledHtmlTextWriter.",
+              nameof(name));
```
MichaelKetting, line 103:
```suggestion
          throw new ArgumentException(
              $"The name of attribute '{name}' indicates a script event but the event type is not supported.",
              nameof(name));
```

MichaelKetting (changes requested):
Please resolve with fixversion = "6.0.0-alpha.RM-8550" instead of "6.0.0-alpha.4"

added some optimization requests

Remotion/Web/Core/ContentSecurityPolicy/NonceGenerator.cs
KuJo7: I think it should be enough if I write two tests:
1. Returns non null value
2. Returns unique nonce by generating two values

MichaelKetting: Perfect.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -620,8 +620,27 @@ class BocList
...
-    const realTableHead = table.querySelector(':scope > thead')!;
+    const realTableHead = table.querySelector<HTMLElement>(':scope > thead')!;
     const fakeTableHead = realTableHead.cloneNode(true) as HTMLElement;
+
+    // The event handlers on the th elements won't get copied so we need to manually copy them
+    // The order between the two should be identical because we cloned the parent node
+    const realElementsWithEvents = [...realTableHead.querySelectorAll<HTMLAnchorElement>("*[data-inline-event-target]")];
+    const fakeElementsWithEvents = [...fakeTableHead.querySelectorAll<HTMLAnchorElement>("*[data-inline-event-target]")];
+    for (let i = 0; i < realElementsWithEvents.length; i++) {
+      const realElementWithEvents = realElementsWithEvents[i];
```
MichaelKetting:
```suggestion
    for (let i = 0; i < realElementsWithEvents.length; i++)
    {
      const realElementWithEvents = realElementsWithEvents[i];
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -620,8 +620,27 @@ class BocList
...
+      const fakeElementWithEvents = fakeElementsWithEvents[i];
+      for (const key in realElementWithEvents) {
+        if (key.startsWith("on")) {
```
MichaelKetting:
```suggestion
      for (const key in realElementWithEvents)
      {
        if (key.startsWith("on")) {
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -620,8 +620,27 @@ class BocList
...
+          const value = realElementWithEvents[key as keyof HTMLAnchorElement];
+          if (typeof value === "function") {
+            (fakeElementWithEvents as any)[key] = value;
```
MichaelKetting:
```suggestion
          if (typeof value === "function")
          {
            (fakeElementWithEvents as any)[key] = value;
```

Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+
+  private CspDefaultsProvider (
+      bool isCspEnabledDefault,
+      CspHeader cspHeaderDefault,
+      bool isCspReportOnlyEnabledDefault,
+      CspHeader cspReportOnlyHeaderDefault)
+  {
+    ArgumentUtility.CheckNotNull(nameof(cspHeaderDefault), cspHeaderDefault);
+    ArgumentUtility.CheckNotNull(nameof(cspReportOnlyHeaderDefault), cspReportOnlyHeaderDefault);
+
```
MichaelKetting:
```suggestion
```

private code :)

Remotion/Web/Core/ContentSecurityPolicy/ICspDefaultsProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Remotion.Web.ContentSecurityPolicy;
+
+/// <summary>
+/// Provides an interface for retrieving default values for CSP related settings for a <see cref="Page"/>.
+/// </summary>
+/// <threadsafety static="true" instance="true" />
+public interface ICspDefaultsProvider
+{
+  /// <summary>
+  /// Returns default value of <see cref="SmartPage"/>.<see cref="SmartPage.IsCspEnabled"/> for the specified <paramref name="page"/>.
```
MichaelKetting: I'm not sure the CspDefaults-Provider should say "SmartPage". it kind of feels like we're invreting the abstraction here. See comment about parameter type, might be that "ISmartPage" is okay, if we go with the interface. Or "SmartPage" if there is a good reason to not use the interface in the parameter type.

Remotion/Web/Core/ContentSecurityPolicy/ICspDefaultsProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+  /// </summary>
+  bool GetDefaultIsCspEnabledForPage (Page page);
```
MichaelKetting: We should change this to IPage, please. Or ISmartPage if you want to only support SmartPage.

Remotion/Web/Development.WebTesting.IntegrationTests/TreeViewControlObjectTest.cs
```diff
@@ -88,7 +88,7 @@ public void TestNodeIsChecked ()
               .With.Message.EqualTo(
                   AssertionExceptionUtility.CreateExpectationException(
                       Driver,
-                      "The checkbox could not be found: Unable to find xpath: ./tbody/tr/td[a[contains(@onclick, 'TreeView_SelectNode')]]/input[@type='checkbox']").Message));
+                      "The checkbox could not be found: Unable to find xpath: ./tbody/tr/td[a[@data-inline-event-target]]/input[@type='checkbox']").Message));
```
MichaelKetting:
```suggestion
                      "The checkbox could not be found: Unable to find xpath: ./tbody/tr/td[a[contains(@data-event-content-onclick, 'TreeView_SelectNode')]]/input[@type='checkbox']").Message));
```

given we now have the attribute with the actual script body, we should go back to `contains` to preserve maximum compatiblity.

Remotion/Web/Development.WebTesting.WebFormsControlObjects/TreeViewNodeControlObject.cs
```diff
@@ -236,7 +236,7 @@ public UnspecifiedPageObject Click ([CanBeNull] IWebTestActionOptions? actionOpt
     private void ClickNode (IWebTestActionOptions? actionOptions)
     {
       var actualCompletionDetector = MergeWithDefaultActionOptions(Scope, actionOptions);
-      const string nodeClickScopeXpath = "./tbody/tr/td[a[contains(@onclick, 'TreeView_SelectNode')]][last()]/a[last()]";
+      const string nodeClickScopeXpath = "./tbody/tr/td[a[@data-inline-event-target]][last()]/a[last()]";
```
MichaelKetting:
```suggestion
      const string nodeClickScopeXpath = "./tbody/tr/td[a[contains(@data-event-content-onclick, 'TreeView_SelectNode')]][last()]/a[last()]";
```

Remotion/Web/Development.WebTesting.WebFormsControlObjects/TreeViewNodeControlObject.cs
```diff
@@ -249,7 +249,7 @@ private void ClickNode (IWebTestActionOptions? actionOptions)
 
     private ElementScope GetCheckboxScope ()
     {
-      const string xpath = "./tbody/tr/td[a[contains(@onclick, 'TreeView_SelectNode')]]/input[@type='checkbox']";
+      const string xpath = "./tbody/tr/td[a[@data-inline-event-target]]/input[@type='checkbox']";
```
MichaelKetting:
```suggestion
      const string xpath = "./tbody/tr/td[a[contains(@data-event-content-onclick, 'TreeView_SelectNode')]]/input[@type='checkbox']";
```

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -0,0 +1,126 @@
...
+  /// Represents <see cref="HtmlTextWriter"/> which implements Content-Security-Policy
+  /// </summary>
+  public class CspEnabledHtmlTextWriter : HtmlTextWriter
+  {
+    private static readonly IReadOnlyDictionary<string, string> s_supportedEvents =
+        new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+          { "onclick", "onclick" },
+          { "onchange", "onchange" },
+          { "oncontextmenu", "oncontextmenu" },
```
MichaelKetting, line 38:
```suggestion
          // can be removed once WebTreeView learns to use modern html script registration.
          { "oncontextmenu", "oncontextmenu" },
```

Since (or until) we don't add all the events at this time, we should note that we only have the contextmenu because changing remotion right now is too expensive. all the others are there because of WebForms (and then because of re-motion).

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -543,6 +624,40 @@ public bool? EnableSmartFocusing
...
+  /// The value of the header is determined by <see cref="GetCspReportOnlyHeader"/>.
+  /// </summary>
+  [Description("The flag that determines whether a Content-Security-Policy-Report-Only header is set.")]
+  [Category("Behavior")]
+  [DefaultValue(false)]
+  public bool IsCspReportOnlyEnabled
+  {
+    get
+    {
+      _isCspReportOnlyEnabled ??= SafeServiceLocator.Current.GetInstance<ICspDefaultsProvider>().GetDefaultIsCspReportOnlyEnabledForPage(this);
```
MichaelKetting: Please  move SafeServiceLocator.Current.GetInstance<ICspDefaultsProvider>() into a property or resolve it in the ctor into a field. See `MemberCaller`.

Also, please change `SafeServiceLocator.Current` to `ServiceLocator` (here and when getting the NonceProvider)

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -543,6 +624,40 @@ public bool? EnableSmartFocusing
...
+  [DefaultValue(false)]
+  public bool IsCspEnabled
+  {
+    get
+    {
+      _isCspEnabled ??= SafeServiceLocator.Current.GetInstance<ICspDefaultsProvider>().GetDefaultIsCspEnabledForPage(this);
+      return _isCspEnabled.Value;
+    }
+    set { _isCspEnabled = value; }
+  }
```
MichaelKetting: setable properties in webforms: they're special and we cannot asign them while getting. Actually, assigning them durign the get operation is problematic even in non-webforms code :)

the pattern used can be seen with EnableSmartFocusing: we got the settable property EnableSmartFocusing and the protected virtual readonly property IsSmartFocusingEnabled. The same pattern should be used here. (disclaimer: with SmartFocusing, it's also part of an interface, so we used protected instead of public. I think protected is still sensible, though...)

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -543,6 +624,40 @@ public bool? EnableSmartFocusing
...
   }
 
+  /// <summary>
+  /// Gets or sets the flag that determines whether a Content-Security-Policy header is set.
+  /// The value of the header is determined by <see cref="GetCspHeader"/>.
+  /// </summary>
+  [Description("The flag that determines whether a Content-Security-Policy header is set.")]
+  [Category("Behavior")]
+  [DefaultValue(false)]
+  public bool IsCspEnabled
```
MichaelKetting: Please move the new properties after the smart-focusing block. right now it's in the middle of a group of related members.

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -321,6 +378,30 @@ public virtual IEnumerable<string> GetDirtyStates (IReadOnlyCollection<string>?
     return _smartPageInfo.GetDirtyStates(requestedStates);
   }
 
+  /// <summary>
+  /// Returns the Content-Security-Policy header value that should be used if <see cref="IsCspEnabled"/> is set.
+  /// </summary>
+  public virtual CspHeader GetCspHeader ()
```
MichaelKetting: I think we can make GetCspHeader and GetCspReadOnlyHeader protected virtual.

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -240,13 +243,67 @@ public NameValueCollection GetNavigationUrlParameters ()
...
   public SmartPage ()
   {
     _smartPageInfo = new SmartPageInfo(this);
     _validatableControlInitializer = new ValidatableControlInitializer(this);
     _postLoadInvoker = new PostLoadInvoker(this);
     _clientScriptManager = new SmartPageClientScriptManager(base.ClientScript);
+    _nonceGenerator = new NonceGenerator();
+  }
+
+  protected override void OnPreRenderComplete (EventArgs e)
```
MichaelKetting: please move down to after OnInit, possibly all the way down to before ProcessRequest

MichaelKetting (changes requested):
~~Note: checked all but the webtest-testpages.~~

Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+using Remotion.Utilities;
+
+namespace Remotion.Web.ContentSecurityPolicy;
+
+/// <summary>
+/// Implements <see cref="ICspDefaultsProvider"/> to provide a default service with opt-in CSP support with a default header.
+/// Register a custom instance to enable CSP and set a custom CSP header.
+/// </summary>
+/// <threadsafety static="true" instance="true" />
+[ImplementationFor(typeof(ICspDefaultsProvider), Lifetime = LifetimeKind.Singleton)]
```
MichaelKetting, line 15: please add a test fixture `ICspDefaultsProviderTest` for testing the correct IoC configuration.

Remotion/Web/Core/UI/SmartPage.cs
a-ctor: I don't get the problem of settable properties. Could you elaborate on why this is a problem?

Is the `EnableSmartFocusing` just an example or do you want to introduce a virtual get method as well? I would keep just the property and assign the default values in the constructor using the defaults provider and call it a day.

MichaelKetting: @a-ctor  the first reason is that getting the settable property changes the state of the object, which is something we should always avoid because it breaks the idomatic property concept. It's also really impure ;)

The second reason is related to how WebForms Designer works (well...worked). Properties that are persisted in the ASPX get persisted based on the final state they have. This means when the Designer reads the property, it will see a value different from the default and persist it. 

A third reason has to do with consistency. When we start using two wildy separate patterns in the same file, it becomes much more difficult to maintain. 

I believe that the first behavior alone is sufficient to justify the more elaborate style.

Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -240,13 +243,41 @@ public NameValueCollection GetNavigationUrlParameters ()
...
+    if (_cspNonceValue == null)
+     _cspNonceValue = _nonceGenerator.GenerateAlphaNumericNonce();
+
+    return new CspEnabledHtmlTextWriter(
+        this,
+        writer,
+        _nonceGenerator,
+        _cspNonceValue,
+        SafeServiceLocator.Current.GetInstance<IRenderingFeatures>());
   }
```
MichaelKetting, line 281: Please move down to after the OnPreRenderComplete

MichaelKetting (approved):

PR #988, opened 2024-01-04
Feature/rm 7240 support for handling all non success status codes during asynchronous page requests

Frazerus:
https://re-motion.atlassian.net/browse/RM-7240

Not sure what messages I should use. The args.get_error().message just gives a generic message, not sure whether this is due to our setup or typical.


Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -60,6 +60,16 @@ enum SmartPage_DirtyStates
   ClientSide = 'ClientSide'
 }
 
+class SmartPage_ErrorDeclaration {
```
a-ctor, line 63: Please fix all the braces to be on the next line just like in C#. The idea here is that we want to match the C# style to make switching between C# and JS/TS easier. Please check all the places in this file.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,19 +321,42 @@ class SmartPage_Context
...
+    let current = this._errorMessageMap.get(statusCode);
+    if (current == undefined) return;
+
+    current.errorMessage = message;
+  }
+
+  public SmartPage_PageRequestManager_endRequest = (sender: unknown, args: Sys.WebForms.EndRequestEventArgs) =>
   {
-    if (args.get_error() != undefined && (args.get_error() as any).httpStatusCode == '500')
+    if (args.get_error() != undefined)
```
a-ctor, line 333: I know that it was like this before but I would suggest changing that to `!args.get_error()` instead. There are two reasons for this: 1) It also covers other invalid values like `null` and 2) it removes the need to use a comparison operator which might (as is the case here) have type coercion enabled. The other option would be to use `args.get_error() !== undefined` instead.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,19 +321,42 @@ class SmartPage_Context
...
     {
-      const errorMessage = args.get_error().message;
-      args.set_errorHandled(true);
+      let statusCode = (args.get_error() as any).httpStatusCode;
 
-      const errorBody = '<div class="SmartPageErrorBody"><div>' + errorMessage + '</div></div>';
+      let errorDeclaration = this._errorMessageMap.get(statusCode);
 
-      SmartPage_Context.Instance!.ShowMessage("SmartPageServerErrorMessage", errorBody);
+      if (errorDeclaration == undefined || statusCode == 200){
```
a-ctor, line 339: Please always use the non-type coercion variants of `==` and `!=`. These are the triple variants `===` and `!==`. New code should always use the triple variants as there is no implicit type conversion before comparing the values.

a-ctor, line 339: Isn't the second part of the if superfluous since we don't have an entry for any of the 2xx HTTP codes?

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,19 +321,42 @@ class SmartPage_Context
...
+        this.HandleEndRequestError("undefined", args.get_error().name, "An error has occured.", args, true);
+      }
+      else {
+        this.HandleEndRequestError(statusCode, errorDeclaration.shortDescription, errorDeclaration.errorMessage, args, false);
+      }
     }
   }
 
+  private HandleEndRequestError(statusCode: string, errorDescription: string, errorMessage: string, args: Sys.WebForms.EndRequestEventArgs, showBody: boolean): void
```
a-ctor, line 348: I think it would be better to pass in the `SmartPage_ErrorDeclaration` here. This allows us to add new properties without needing to add new parameters to this method. `showBody`, for example, might also be a good property to add to the class instead. That way, we can handle all HTTP status codes the same way and allow overriding the behavior for those values as well.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,19 +321,42 @@ class SmartPage_Context
...
+  {
+    args.set_errorHandled(true);
+
+    let errorBody = '<div class="SmartPageErrorBody"><div>Error ' + statusCode + ', ' + errorDescription + ': ' + errorMessage + '</div>';
+    if (showBody)
+      errorBody += '<iframe sandbox srcdoc="<div>' + args.get_response().get_responseData() + '</div>"></iframe>';
+    errorBody += '</div>';
```
a-ctor, line 355: We need to discuss what parts of the response we consider "dangerous" and should therefore put in the `iframe`. Also spot the XSS problem :D

a-ctor (changes requested):
We need to discuss with Mike if our interpretation of the issue fits reality and what kind of customization is wanted from consumers.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
Frazerus: Comment in ticket hints that there can be a Statuscode 2xx with error, Idk how that would happen either, but well oh well

Remotion/Web/ClientScript/Scripts/SmartPage.ts
Frazerus: I presume putting the response data in the iframe is the XSS problem, but shouldn't that be handled by sandbox? Or do we have a problem by simply calling the 'toString' of the response data? And yea, we do need to talk about whats 'dangerous'

Remotion/Web/ClientScript/Scripts/SmartPage.ts
a-ctor: My point was that since we don't have any 2xx status codes in the `_errorMessageMap` registered. So in this if `errorDeclaration == undefined` will already evaluate to `true` if `statusCode == 200` is `true`, making `statusCode == 200` superfluous.

Something I just saw since I was looking at the line again: In JS all values are either truthy or falsy so null/undefined checks are usually simplified to just `value` instead of  `value !== null` since `null` and `undefined` are both falsy anyway. There are some edge cases for certain falues (see falsy values https://developer.mozilla.org/en-US/docs/Glossary/Falsy), but for objects just `value` is ideal as it checks for `null` and `undefined :)

Remotion/Web/ClientScript/Scripts/SmartPage.ts
a-ctor: What is the potential attach vector? And what input could produce it?

Remotion/Web/Development.WebTesting.TestSite.Shared/UrlMapping.xml
```diff
@@ -43,6 +43,7 @@
   <add id="BrowserSessionTest" type="Remotion.Web.Development.WebTesting.TestSite.Shared::BrowserSessionTestFunction" resource="~/BrowserSessionTest.wxe" />
   <add id="PlatformTest" type="Remotion.Web.Development.WebTesting.TestSite.Shared::PlatformTestFunction" resource="~/PlatformTest.wxe" />
   <add id="CompletionDetectionStrategyTest" type="Remotion.Web.Development.WebTesting.TestSite.Shared::CompletionDetectionStrategyTestFunction" resource="~/CompletionDetectionStrategyTest.wxe" />
+  <add id="SmartPageErrorTest" type="Remotion.Web.Development.WebTesting.TestSite.Shared::SmartPageErrorTestFunction" resource="~/SmartPageErrorTest.wxe" />
```
MichaelKetting, line 46: please add to the Start.aspx index page

Remotion/Web/Development.WebTesting.TestSite.Shared/SmartPageErrorTest.aspx
```diff
@@ -0,0 +1,57 @@
...
+        <Triggers>
+            <asp:AsyncPostBackTrigger ControlID="AsyncPostbackButton" EventName="Click"/>
+        </Triggers>
+        <ContentTemplate>
+            <label for="statusCode">StatusCode:</label>
+            <asp:TextBox runat="server" ID="statusCode"></asp:TextBox>
+            <br/>
+            <asp:Button ID="AsyncPostbackButton" runat="server" OnClick="AsyncOnClick" Text="Trigger Async Postback"></asp:Button>
+            <br/>
+            <asp:Button ID="SyncPostbackButton" OnClick="SyncOnClick" Text="Trigger Normal Postback" runat="server"/>
```
MichaelKetting, line 52: Better naming: 
TriggerAsyncResponseWithPartialRenderingErrorButton
TriggerAsyncResponseWithFullPageRenderingErrorButton

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +263,55 @@ class SmartPage_Context
     }
   };
 
-  public SmartPage_PageRequestManager_endRequest = function (sender: unknown, args: Sys.WebForms.EndRequestEventArgs)
+  public SmartPage_PageRequestManager_endRequest = (sender: unknown, args: Sys.WebForms.EndRequestEventArgs) =>
```
MichaelKetting, line 266: We discussed stripping out the prefix: 
`Sys.WebForms.PageRequestManagerServerErrorException:` before rendering the message. that's still missing.

MichaelKetting, line 266: when you trigger the postback and there's no status code in the text field, an error response is sent but not displayed. that's strange because while it's not intended to send a blank field in the test, the handling on the client should still work. please test and check.

Remotion/Web/Development.WebTesting.IntegrationTests/SmartPageErrorTest.cs
```diff
@@ -0,0 +1,73 @@
...
+      var statusCodeTextBox = home.TextBoxes().GetByID("statusCode");
+      statusCodeTextBox.FillWith("418", new WebTestActionOptions() { CompletionDetectionStrategy = new NullCompletionDetectionStrategy()});
+
+      home.WebButtons().GetByID("AsyncPostbackButton").Click();
+
+      //Call Exists workaround because scope is not updated properly
+      home.Scope.FindCss("div.SmartPageErrorBody > div" + " > span > h1").ExistsWorkaround();
+
+      //Wait for Message header to exist
+      home.Scope.FindCss("div.SmartPageErrorBody > div" + " > span > h1").Exists();
```
MichaelKetting, line 43: please explain

Remotion/Web/Development.WebTesting.IntegrationTests/SmartPageErrorTest.cs
```diff
@@ -0,0 +1,73 @@
...
+
+      var statusCodeTextBox = home.TextBoxes().GetByID("statusCode");
+      statusCodeTextBox.FillWith("418", new WebTestActionOptions() { CompletionDetectionStrategy = new NullCompletionDetectionStrategy()});
+
+      home.WebButtons().GetByID("SyncPostbackButton").Click();
+
+      //Wait for Message header to exist
+      home.Scope.FindCss("div.SmartPageErrorBody > div" + " > span > h1").Exists();
+
+      //Im not sure how to select things when not using the home.Labels() or home.Scopes() methods, any Input? I think theres some way like using CSS selectors, but I couldn't find it
```
MichaelKetting, line 64: pair?

Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionParserTest.cs
```diff
@@ -61,7 +61,10 @@ public void Parse_AsyncPostbackError ()
 
       var result = aspNetRequestErrorDetectionParser.Parse(home.Scope);
 
-      Assert.That(result.HasError, Is.EqualTo(true));
+      var messageField = home.Scope.FindId("SmartPageServerErrorMessage");
+      var parent = messageField.FindTagWithAttribute("div", "class", "SmartPageErrorBody");
+
+      //Assert.That(result.HasError, Is.True);
```
MichaelKetting, line 67: delete?

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +263,55 @@ class SmartPage_Context
...
+    }
+    else
+    {
+      const shortenedErrorMessage = errorMessage.substring(errorMessage.indexOf(':') + 1);
+      errorBody += shortenedErrorMessage;
     }
+
+    if (isSynchronousError)
+    {
+      errorBody += '<iframe sandbox style="width: 100%; height: 80%; border: hidden;" srcdoc="<div>';
```
MichaelKetting, line 306: when there's a synchrous type error, the content is a complete html page. there should be no DIV wrapping the content.

Remotion/Web/Development.WebTesting.TestSite.Shared/SmartPageErrorTestHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Remotion.Web.Development.WebTesting.TestSite.Shared
+{
+  public static class SmartPageErrorTestHelper
+  {
+    public static void Execute (HttpRequest request, HttpResponse response)
+    {
+      if (request.Path.Contains("SmartPageErrorTest.wxe") && request.HttpMethod == "POST" && request.Params["TriggerAsyncResponseWithFullPageRenderingErrorButton"] != null)
+      {
+        response.StatusCode = int.Parse(request.Params["statusCode"]);
```
MichaelKetting, line 26: here, please add a guard int.TryParse so empty string and incorrect data don't fail the request. maybe return a generic 400 result in that case (400 because it's a client error with incorrect data)

Remotion/Web/Development.WebTesting.TestSite.Shared/SmartPageErrorTest.aspx
```diff
@@ -0,0 +1,48 @@
...
+  </head>
+<body>
+<form id="form1" runat="server">
+    <asp:ScriptManager ID="ScriptManager" EnablePartialRendering="true" AsyncPostBackTimeout="3600" runat="server"/>
+    <asp:UpdatePanel ID="UpdatePanel" runat="server">
+        <ContentTemplate>
+            <label for="statusCode">StatusCode:</label>
+            <asp:TextBox runat="server" ID="statusCode"></asp:TextBox>
+            <br/>
+            <asp:Button ID="TriggerAsyncResponseWithPartialRenderingErrorButton" runat="server" OnClick="AsyncOnClick" Text="Trigger Async Postback"></asp:Button>
```
MichaelKetting, line 41:
```suggestion
            <asp:Button ID="TriggerAsyncResponseWithPartialRenderingErrorButton" runat="server" OnClick="AsyncOnClick" Text="Trigger error response with partial rendering result"></asp:Button>
```

Remotion/Web/Development.WebTesting.TestSite.Shared/SmartPageErrorTest.aspx
```diff
@@ -0,0 +1,48 @@
...
+            <br/>
+            <asp:Button ID="TriggerAsyncResponseWithFullPageRenderingErrorButton" OnClick="SyncOnClick" Text="Trigger Normal Postback" runat="server"/>
```
MichaelKetting, line 43:
```suggestion
            <asp:Button ID="TriggerAsyncResponseWithFullPageRenderingErrorButton" OnClick="SyncOnClick" Text="Trigger error response with full page result" runat="server"/>
```

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +258,54 @@ class SmartPage_Context
...
 
-      SmartPage_Context.Instance!.ShowMessage("SmartPageServerErrorMessage", errorBody);
+    //@ts-ignore
+    let serverError: string = Sys.WebForms.Res.PRM_ServerError;
+    let isSynchronousError = errorMessage.includes(serverError.substring(0, serverError.indexOf(':')));
+
+    let errorBody = '<div class="SmartPageErrorBody"><div>';
+
+    const shortenedErrorMessage = errorMessage.substring(errorMessage.indexOf(':') + 1);
+    errorBody += shortenedErrorMessage;
```
MichaelKetting, line 290:
```suggestion
    const errorMessageWithoutExceptionTypePrefix = errorMessage.substring(errorMessage.indexOf(':') + 1);
    errorBody += errorMessageWithoutExceptionTypePrefix;
```

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +258,54 @@ class SmartPage_Context
...
+
+    if (isSynchronousError)
+    {
+      errorBody += ', ' + response.get_statusText() + '<br>';
```
MichaelKetting, line 294:
```suggestion
      errorBody += ', ' + response.get_statusText() + '\n';
```
We don't need a BR because iframe is already a blocklevel element. Adding a newline is still helpful for reading etc.

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +258,54 @@ class SmartPage_Context
...
     }
+
+    if (isSynchronousError)
+    {
```
MichaelKetting, line 298:
```suggestion
```

MichaelKetting (changes requested):
some small stuff

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +258,51 @@ class SmartPage_Context
...
+
+    const errorMessageWithoutExceptionTypePrefix = errorMessage.substring(errorMessage.indexOf(':') + 1);
+    errorBody += errorMessageWithoutExceptionTypePrefix;
 
-      SmartPage_Context.Instance!.ShowMessage("SmartPageServerErrorMessage", errorBody);
+    if (isSynchronousError)
+    {
+      errorBody += ', ' + response.get_statusText() + '\n';
+
+      btoa(response.get_responseData())
```
MichaelKetting, line 296:
```suggestion
```

Remotion/Web/ClientScript/Scripts/SmartPage.ts
```diff
@@ -258,17 +258,51 @@ class SmartPage_Context
...
+    let errorBody = '<div class="SmartPageErrorBody"><div>';
+
+    const errorMessageWithoutExceptionTypePrefix = errorMessage.substring(errorMessage.indexOf(':') + 1);
+    errorBody += errorMessageWithoutExceptionTypePrefix;
 
-      SmartPage_Context.Instance!.ShowMessage("SmartPageServerErrorMessage", errorBody);
+    if (isSynchronousError)
+    {
+      errorBody += ', ' + response.get_statusText() + '\n';
+
```
MichaelKetting, line 295:
```suggestion
      errorBody += ', ' + response.get_statusText();
      errorBody += '\n';
```

MichaelKetting (approved):
JS formatting / cleanup I missed before. don't need to recheck after

a-ctor (approved):

PR #989, opened 2024-01-04
Revert "RM-8773 FillWithAction uses separate calls to SendKeys with G…

Frazerus:
…eckoDriver"

This reverts commit e6622f66a01fc656887cadf1d5fa406510298012.

https://re-motion.atlassian.net/browse/RM-8781


MichaelKetting (approved):

Frazerus:
This has to wait until we update firefox to v114+, then we should be able to simply merge it.

PR #991, opened 2024-01-11
RM-9054 Upgrade firefox to v121

Frazerus:
https://re-motion.atlassian.net/browse/RM-9054


MichaelKetting (approved):
The usaul please run 15 or so browser tests with firefox. now that we have 7 mathcing agents, we can schedule 2-3 buiilds on all matching agents to accomplish this feat.

PR #994, opened 2024-01-16
RM-9054 Upgrade firefox to v121

Frazerus:
https://re-motion.atlassian.net/browse/RM-9054


Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/SavedTestScreenshots/BocListControlObjectTest/Firefox.any.ScreenshotTest_Navigator.png
MichaelKetting, line 1: would like to check what has changed here since it's 12 px.

Remotion/Web/Development.WebTesting.IntegrationTests/MouseTest.cs
```diff
@@ -102,6 +102,8 @@ public void Test_Focus ()
 
       var focusElement = home.Scope.FindId(c_focusDivID);
 
+      Thread.Sleep(250);
```
MichaelKetting, line 105: we need to figure out how to solve this without adding a thread-sleep in the test. normally, this indicates that there is an underlying issue that results in a systematic delay. if that's the case, everyone would be affected.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting.IntegrationTests/MouseTest.cs
MichaelKetting: @a-ctor got ideas how to best analyze/solve this?

a-ctor: We can use `RetryUnilTimeout.Run(...)` instead of waiting. Only downside is that that same assertion error is logged multiple times. So if the test fails the error is printed about 15 times. We can either ignore this or use NUnit's `IsolatedConext` in the delegate to prevent this.

MichaelKetting: @a-ctor the problem is, if the test is flaky, it's reasonably to assume that this will also affect users of the library. we should first find out why the test is flaky and if we can do something in the webtest library to ensure users remain unaffected. Only for cases where our own setup is causing the instability should we use Retry/delay in the test code.

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/SavedTestScreenshots/BocListControlObjectTest/Firefox.any.ScreenshotTest_Navigator.png
Frazerus: The size of the box with the current page number inside has been reduced. This looks reasonable in my eyes. 
image

Frazerus: This seems to be an overall firefox layout change, depending on the size the text box is set to, all values seem to differ between the old and the new version.

MichaelKetting (approved):
Please don't forget to clear-out the ticket's placeholder description (or add a note that the hover has become a tad more flaky)

PR #1002, opened 2024-01-18
RM-6463 Add support for static resource caching

a-ctor:
https://re-motion.atlassian.net/browse/RM-6463


Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+      ArgumentUtility.CheckNotNull(nameof(resourceTypes), resourceTypes);
+
+      _resourcesPhysicalPathLocator = resourcesPhysicalPathLocator;
+      _resourceFileDetailsAppender = resourceFileDetailsAppender;
+      _resourceTypes = resourceTypes;
+
+      _resourceTypeLookup = resourceTypes.Select(e => e.Name).ToHashSet(StringComparer.OrdinalIgnoreCase);
+    }
+
+    public bool HasCacheKey => _cacheKey != null;
```
MichaelKetting, line 141: We already have the DoubleCheckedLockingContainer for this. While I want to eventually phase it out and replace it with lower level APIs *), until such a time, we can use it when having a writable synchronization istuation.

it works the same. you set the implementation delegate in the ctor and when the value is reset to null, it will get reevaluated.

*) most scenarios where we use DCLC are write once or should be refactored to write once and can thus be covered by Lazy<T> these days. The couple of read/write places that remain can then be written with manual lock management.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+        // which could cause a null value return here.
+        cacheKey = _cacheKey;
+        if (cacheKey != null)
+          return cacheKey;
+
+        if (_physicalPath == null)
+        {
+          var physicalPath = _resourcesPhysicalPathLocator.GetResourcesPhysicalPath();
+          physicalPath = Path.GetFullPath(physicalPath);
+          if (!physicalPath.EndsWith("" + Path.DirectorySeparatorChar))
```
MichaelKetting, line 162: why do we need `"" + Path.Dired..` instead of simply testing for `Path.Direc...`

edit:
ah, because of .NET Framework. Okay, please add a conditional compile because we won't find this code again. manually  compare the last char.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+          _physicalPath = physicalPath;
+          _fileSystemWatcher.EnableRaisingEvents = true;
+        }
+
+        var stringBuilder = new StringBuilder();
+        _resourceFileDetailsAppender.AppendResourceFolderDetails(stringBuilder, _physicalPath, _resourceTypes);
+
+        using var sha256 = SHA256.Create();
+        var hashBytes = sha256.ComputeHash(Encoding.UTF8.GetBytes(stringBuilder.ToString()));
+        var hash = string.Concat(hashBytes.Select(e => e.ToString("X2")));
```
MichaelKetting, line 183: for .net, we already got a helper, don't we?

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /// <inheritdoc />
+    public void Dispose ()
+    {
+      _fileSystemWatcher?.Dispose();
+      _fileSystemWatcher = null;
+    }
+
+    private void CheckForPathUpdate (string? path, FileChangeType changeType)
```
MichaelKetting, line 197: Note: I still need to review this method. it looks okay, but I need to understand the semantics :)

MichaelKetting (changes requested):
Got a couple of design-bits

Remotion/Web/Core/Resources/IResourcePathBuilder.cs
```diff
@@ -39,5 +39,7 @@ namespace Remotion.Web.Resources
  public interface IResourcePathBuilder
   {
     string BuildAbsolutePath (Assembly assembly, params string[] assemblyRelativePathParts);
+
+    string BuildCacheableAbsolutePath (Assembly assembly, params string[] assemblyRelativePathParts);
```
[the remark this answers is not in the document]
MichaelKetting: Got more detailed idea:

since we can handle / convert `null` to dot (`"."`) anyway, we can make the `IStaticResourceCacheKeyProvider` part of the ctor signature of the `ResourcePathBuilder` and have just two derived types required for IoC only. One type is the cache-enbled one and accepts the `IStaticResourceCacheKeyProvider`, the other one uses the NullStaticResourceCacheKeyProvider when performing the base ctor call. That way we can register two different `ResourcePathBuilder` for two different `IResourcePathBuilder` interfaces but don't have any implementation differences and thus keep things simple. Should we one day decide to support IoC  with tagged services, this can be further simplified.

Disclaimer: this comment should only simplify the orignal review remarks :)

Remotion/Web/Core/Resources/NullStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+//
+using Remotion.ServiceLocation;
+
+namespace Remotion.Web.Resources
+{
+  /// <summary>
+  /// Represents a nullable <see cref="IStaticResourceCacheKeyProvider"/> according to the "Null Object Pattern".
+  /// </summary>
+  /// <threadsafety static="true" instance="true" />
+  [ImplementationFor(typeof(IStaticResourceCacheKeyProvider), Lifetime = LifetimeKind.Singleton)]
```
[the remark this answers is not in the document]
a-ctor: I think we have every different kind of variation in this statement. Specifying just the singleton lifetime seems pretty common though.

Remotion/Web/Core/Resources/ResourcePathBuilderBase.cs
```diff
@@ -29,6 +29,15 @@ public abstract class ResourcePathBuilderBase : IResourcePathBuilder
   {
     private static readonly ConcurrentDictionary<Assembly, AssemblyName> s_assemblyNameCache = new ConcurrentDictionary<Assembly, AssemblyName>();
 
+    private readonly IStaticResourceCacheKeyProvider _staticResourceCacheKeyProvider;
```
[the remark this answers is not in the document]
a-ctor: How does making the field public improve testability? It is a private field that is only used by base class to pass values to virtual methods.

Remotion/Web/Core/Resources/NullStaticResourceCacheKeyProvider.cs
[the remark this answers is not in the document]
MichaelKetting: Great, thanks!

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+      public void AppendResourceFolderDetails (StringBuilder stringBuilder, string resourceFolder, IReadOnlyCollection<ResourceType> resourceTypes)
+      {
+        // Format of a resource folder is: '~/<assembly name>/<resource type>/...
+        foreach (var assemblyFolder in Directory.EnumerateDirectories(resourceFolder, "*"))
+        {
+          foreach (var resourceType in resourceTypes)
+          {
+            var resourceTypeDirectory = new DirectoryInfo(Path.Combine(assemblyFolder, resourceType.Name));
+            if (!resourceTypeDirectory.Exists)
+              continue;
```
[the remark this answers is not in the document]
a-ctor: Main reason for the explicit loops were a) no extra allocations created by LINQ and b) does not hide the complexity of the operation (3 nested loops). I don't know how much overhead LINQ creates but I feel it is better to not have it in such a case. Readability wise it is fine in my opinion both approaches (imperative/functional) is fine by me.

Remotion/Web/Core/Resources/ResourcePathBuilderBase.cs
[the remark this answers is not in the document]
MichaelKetting: Testing that the correct value has been assigned. In the past, it proved helpful to a) test that the correct assignment is done and b) accessing the injected services sometimes has value beyond unit testing. So, I try to remember to add a public accessor to the class for any serivce that's not for some reason required to be represented as private state.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            foreach (var resourceFile in resourceTypeDirectory.EnumerateFiles("*", SearchOption.AllDirectories))
+            {
+              stringBuilder.Append(resourceFile.FullName);
```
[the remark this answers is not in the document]
a-ctor: Why is this important? I assume it is to get the same cache key on multiple servers? If so, it is very unlikely that we can ensure this is the case anyway as any manual modification will change the last modified date for only one server and not for the rest causing a permanent cache key change...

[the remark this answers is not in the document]
MichaelKetting: usally, the path is the same on the servers. The important aspect is size of the string. It removes extra cpu for this, so it's good for performance.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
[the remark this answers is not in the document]
MichaelKetting: fine. but we will check the patterns in the next devjour fixe

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,251 @@
...
+              stringBuilder.Append(';');
+              stringBuilder.Append(resourceFile.Length);
+              stringBuilder.Append(';');
+              stringBuilder.Append(resourceFile.CreationTimeUtc.ToFileTimeUtc());
+              stringBuilder.Append(';');
+              stringBuilder.Append(resourceFile.LastWriteTimeUtc.ToFileTimeUtc());
```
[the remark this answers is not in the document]
a-ctor: What do you mean with "only a long value."? Both `Ticks` and `.ToFileTimeUtc()` are `long`s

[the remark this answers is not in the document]
MichaelKetting: ah, didn't realize that. okay.

Remotion/Web/Core/Resources/ICacheableResourcePathBuilder.cs
```diff
@@ -0,0 +1,35 @@
...
+  /// <remarks>
+  ///   <para>
+  ///     The default implementation (<see cref="CacheableResourcePathBuilder"/>) builds paths in the following format:
+  ///     &lt;resource root&gt;/[cache_&lt;cache key&gt;/]&lt;Assembly-Name&gt;/part-1/.../part-n.
+  ///   </para><para>
+  ///     The <b>resource root</b> is loaded from the application configuration, <see cref="ResourceRoot"/>, and
+  ///     defaults to <c>/&lt;AppDir&gt;/res</c>, e.g. <c>/WebApplication/res/Remotion.Web/Html/Utilities.js</c>.
+  ///   </para><para>
+  ///     The <b>resource root</b> is mapped to the environment variable <c>REMOTIONRESOURCES</c>,
+  ///     or if the variable does not exist, <c>C:\Remotion.Resources</c>.
```
MichaelKetting, line 31:
```suggestion
```
I think this one's obsolete / no logner existing

Remotion/Development/Web/UnitTesting/Resources/FakeCacheableResourcePathBuilder.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+  /// <summary>
+  /// Fake implementation of the <see cref="ICacheableResourcePathBuilder"/> interface, intended for use in unit testing.
+  /// </summary>
+  public class FakeCacheableResourcePathBuilder : ResourcePathBuilderBase, ICacheableResourcePathBuilder
+  {
+    public FakeCacheableResourcePathBuilder ()
+        : base(new NullStaticResourceCacheKeyProvider())
+    {
+    }
```
MichaelKetting, line 30: without checking the rest of the code: shouldn't we use a FakeStaticResourceCacheKeyProvider here so the FakeCachable is actually having an effect beyond the interface? It might be relevant in certain test scenarios to be able to distinuish the output of the two PathBuilder types?

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,266 @@
...
+      {
+        foreach (var (name, length, creationTime, lastWriteTime) in folderDetails)
+        {
+          // Use only the relative path from the root to calculate the cache key.
+          // This removes a lot of stuff to cache while also allowing multiple servers
+          // to generate the same cache key even if they are store on different paths.
+#if NETFRAMEWORK
+          binaryWriter.Write(name.Substring(_physicalPath.Length));
+#else
+            binaryWriter.Write(name.AsSpan()[_physicalPath.Length..]);
```
MichaelKetting, line 183:
```suggestion
          binaryWriter.Write(name.AsSpan()[_physicalPath.Length..]);
```

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,266 @@
...
+      public string GetMappedResourcesPath ()
+      {
+        var context = _httpContextProvider.GetCurrentHttpContext();
+        return context.Server.MapPath("~/" + _resourceRoot.Value);
+      }
+    }
+
+    public interface IResourceFileDetailsAppender
+    {
+      void AppendResourceFolderDetails (List<(string Name, long Length, long CreationTime, long LastWriteTime)> folderDetails, string resourceFolder, IReadOnlyCollection<ResourceType> resourceTypes);
```
MichaelKetting, line 64: I know I had both this version and the alternative of using a List<string> + List<long> in my suggestions. I realize now that I prefer the Two-List approach because you can pack anything into the cachekey. Does the tuple-version have particular benefits that make it preferable?

Thoughts on the two variants?

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
```diff
@@ -0,0 +1,266 @@
...
+      Deleted,
+      Renamed
+    }
+
+    private readonly IMappedResourcesPathLocator _mappedResourcesPathLocator;
+    private readonly IResourceFileDetailsAppender _resourceFileDetailsAppender;
+    private readonly IReadOnlyList<ResourceType> _resourceTypes;
+
+    private readonly DoubleCheckedLockingContainer<string?> _cacheKey;
+    private readonly HashSet<string> _resourceTypeLookup;
```
MichaelKetting, line 103: Please make this also .net /.netfx contiional with .NET using the IReadOnlySet intead of HashSet. That way, we get threadsafety via compilercheck :)

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
```diff
@@ -0,0 +1,310 @@
...
+
+      using var sha256 = SHA256.Create();
+      var data = sha256.ComputeHash(memoryStream);
+
+#if NETFRAMEWORK
+      return string.Concat(data.Select(e => e.ToString("X2")));
+#else
+      return Convert.ToHexString(data);
+#endif
+    }
```
MichaelKetting, line 308: I'm not sure this is such a great idea / helpful. We just copied the algorithm so any mistakes would get copied too. Would it be sufficeint / better to test the format and that the hash changes when there is a change in one of the affected file properties? 

This would mean we get individual tests, one for each property that affects th cache key.

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
```diff
@@ -0,0 +1,310 @@
...
+
+    private void WaitForCacheKeyChange (TestStaticResourceCacheKeyProvider staticResourceCacheKeyProvider)
+    {
+      for (var i = 0; i < 10; i++)
+      {
+        if (!staticResourceCacheKeyProvider.HasCacheKey)
+          return;
+
+        Thread.Sleep(20);
```
MichaelKetting, line 274: api note: I like to use Timespan.FromMillisconds instead of presuming the reader knows the API. It's not somehting I need to change, it's just somehting for future consideration.

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
```diff
@@ -0,0 +1,310 @@
...
+      folder.CreateDirectory(@"Remotion.Web\Html\subfolder", DateTime.Today, DateTime.Today);
+
+      // Changes to irrelevant folders should be ignored completely
+      folder.CreateDirectory(@"Remotion.Web\UI\subfolder");
+      folder.CreateFile(@"Remotion.Web\UI\subfolder\abc.txt", "testContent", DateTime.Today, DateTime.Today);
+
+      // No better way than to sleep as we expect no notifications to come in and
+      // they can have an arbitrary delay. 100ms should be good enough as tests
+      // with 50ms also worked well.
+      Thread.Sleep(100);
```
MichaelKetting, line 262: could we use our own filesystemwatcher and a monitor.Wait?

MichaelKetting (changes requested):
some questions :)

MichaelKetting:
@a-ctor so I don't forget: please add a task for creating the .NET API helper for the UrlRewrite via Routing  etc.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
a-ctor: This is not an extension point but instead is only here to enable testing. As such, I don't think that using a list here adds any benefit. 

I assume you would use `List<List<object>>` instead? That would require casting all the long objects. If you wanted to hack the hash you could just append something to the name.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
a-ctor: I don't follow what does `IReadOnlySet<T>` add here?

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
MichaelKetting: no, I literally meant to parameters, one List<string> the second one List<long>. To me, it would also simplify testing and it looks weird having a public API like that.

If it's not a good idea, we can keep it as is.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
MichaelKetting: It ensures that the values can only be changed in the ctor, i.e. the type ensures that we know the member is used in a non-modifyable way just by looking at the field declaration. This makes thread safety more easily verifyable.

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
a-ctor: The main idea was to not just have hash values hardcoded in the tests. Instead, one can provide the values that are expected and derive the hash from that. Just checking for changed hashes is, in my opinion, less specific than asserting what kind of values should be used in the hash generation.

The risk that this piece of code has a bug in the hash generation that is not spotted is minimal. It is clear that all values are written to the binary stream so the only error left is that the hash is not calculated for the whole memoryStream. But this case is covered by other test that would break if we supplied a constant hash.

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
a-ctor: image
;)

Remotion/Development/Web/UnitTesting/Resources/FakeCacheableResourcePathBuilder.cs
a-ctor: I looked at the usages again and decided to remove the fake entirely. It was only used because a `ICachedResourcePathBuilder` reference was necessary. I replaced the uses with a stub.

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
a-ctor: Can you elaborate on what you mean by "second list"? I don't know what a second list would add here...

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
MichaelKetting: mhm, okay.
So..compromise: in the test, please add an assert on the expected value (i.e. in this method here) that the Hash string has the expected length and a regex-test for the format. That way, we catch if there's a bug in the test and the hash string ends up being empty or not correctly formatted. 

I know that from reading the code, we can be pretty sure that it's correct. I don't trust mental verification if I can to it in code, though, and I like to follow the rule of not doing the same algorithm in the tests we do in production code. That's how copy&paste errors are introduced. 

Also, test code should be so simple that it's easily verified by reading it, which is why I suggested we drop uneeded details. I can live with being more precise once we add the trip wires. Just for completeness sake on my reasoning: the actual algorithm or order of items added to the hash value isn't important for the unit test. We only want a hashvalue that's collission free. If it's MD5 or SHA256 isn't somehting that's an important detail of the implemetnation since we don't need to do cryptographic safety, only collision-free-ness.

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
MichaelKetting: image

:p

Code should always be readable without IDE hints:
There's too many scenarios where you don't have Rider/Resharper enabled (Teams that don't use those tools, decompiled code, viewed on the website of a source control system or a depedency analyzer. And devs who don't like the inline hits because it messes with alignment and line length experience).

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
MichaelKetting: Ah, thanks. I managed to not understand the comment :o

Remotion/Web/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProvider.cs
a-ctor: Ah okay now I understand your idea. I think the current version, while a bit more verbose, is better as each entry (name, length, creation date, and modified date) is stored side by side which makes the tests more readable. In regards to boxing, we don't do any boxing here as well since we use `ValueTuple<T>` so both solutions are equivalent in that they don't box.

MichaelKetting: Exactly. The boxing was just why it's two lists instead of one list of object. Done.

MichaelKetting (approved):

PR #1004, opened 2024-01-24
RM-9071 Upgrade to C# 12

a-ctor:
https://re-motion.atlassian.net/browse/RM-9071

### DIFF
--- a/Build/ProjectTypes/Analyzer.props
+++ b/Build/ProjectTypes/Analyzer.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
 <!--Langversion 9 is required for compatibility with netstandard2.0 -->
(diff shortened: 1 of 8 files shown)


MichaelKetting (approved):
Should we split the compiler version and the tests?

a-ctor:
> Should we split the compiler version and the tests?

@MichaelKetting I kept them as one commit as they both change the same files. I think its fine as one commit but we can split it if you want.

MichaelKetting:
> > Should we split the compiler version and the tests?
> 
> @MichaelKetting I kept them as one commit as they both change the same files. I think its fine as one commit but we can split it if you want.

H... I see. I still think it's cleaner to have separate commits given the files are feature we add after/separate from the c# upgrade. that way, the commit history tells you "this is for the c# upgrade" and "this is for the new language features". And yes, I'm aware that we could even split it into three parts. I'll leave the aesthetic decision to you.

PR #1005, opened 2024-01-24
RM-9038 Add appsettings.json implementation for WebTestSettings

a-ctor:
https://re-motion.atlassian.net/browse/RM-9038

### DIFF
--- a/Remotion/Web/Development.WebTesting.UnitTests/Configuration/IConfiguration/WebTestSettingsDtoTest.cs
+++ b/Remotion/Web/Development.WebTesting.UnitTests/Configuration/IConfiguration/WebTestSettingsDtoTest.cs
@@ -1,0 +1,253 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+#if !NETFRAMEWORK
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.Json;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+using NUnit.Framework;
+using Remotion.Web.Development.WebTesting.Configuration;
+using Remotion.Web.Development.WebTesting.Configuration.IConfiguration;
+using Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chromium;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration.IConfiguration
+{
+  [TestFixture]
+  public class WebTestSettingsDtoTest
+  {
+    [Test]
+    public void Deserialize_MinimalConfig ()
+    {
+      var webTestSettings = DeserializeFromString(
+          """
+          {
+            "Remotion": {
+              "WebTesting": {
+                "BrowserName": "Edge",
+                "SearchTimeout": "00:00:06",
+                "RetryInterval": "00:00:00.011",
+                "WebApplicationRoot": "http://localhost:60401/",
+                "Hosting": {
+                  "Name": "IisExpress",
+                  "Type": "IisExpressType"
+                },
+                "TestSiteLayout": {
+                  "RootPath": "myRootPath"
+                },
+              }
+            }
+          }
+          """);
+
+      Assert.That(webTestSettings, Is.Not.Null);
+      Assert.That(webTestSettings.BrowserName, Is.EqualTo("Edge"));
+      Assert.That(webTestSettings.SearchTimeout, Is.EqualTo(TimeSpan.FromSeconds(6)));
+      Assert.That(webTestSettings.CommandTimeout, Is.EqualTo(TimeSpan.FromMinutes(1)));
+      Assert.That(webTestSettings.DownloadStartedTimeout, Is.EqualTo(TimeSpan.FromSeconds(10)));
+      Assert.That(webTestSettings.DownloadUpdatedTimeout, Is.EqualTo(TimeSpan.FromSeconds(10)));
+      Assert.That(webTestSettings.VerifyWebApplicationStartedTimeout, Is.EqualTo(TimeSpan.FromMinutes(1)));
+      Assert.That(webTestSettings.RetryInterval, Is.EqualTo(TimeSpan.FromMilliseconds(11)));
+      Assert.That(webTestSettings.AsyncJavaScriptTimeout, Is.EqualTo(TimeSpan.FromSeconds(10)));
+      Assert.That(webTestSettings.Headless, Is.False);
+      Assert.That(webTestSettings.WebApplicationRoot, Is.EqualTo("http://localhost:60401/"));
+      Assert.That(webTestSettings.ScreenshotDirectory, Is.EqualTo(""));
+      Assert.That(webTestSettings.LogsDirectory, Is.EqualTo("."));
+      Assert.That(webTestSettings.CloseBrowserWindowsOnSetUpAndTearDown, Is.False);
+      Assert.That(webTestSettings.RequestErrorDetectionStrategyTypeName, Is.EqualTo("None"));
+
+      var hostingSettings = webTestSettings.Hosting;
+      Assert.That(hostingSettings, Is.Not.Null);
+      Assert.That(hostingSettings.Name, Is.EqualTo("IisExpress"));
+      Assert.That(hostingSettings.Type, Is.EqualTo("IisExpressType"));
+      Assert.That(hostingSettings.Parameters, Is.Empty);
+
+      var testSiteLayout = webTestSettings.TestSiteLayout;
+      Assert.That(testSiteLayout, Is.Not.Null);
+      Assert.That(testSiteLayout.RootPath, Is.EqualTo("myRootPath"));
+      Assert.That(testSiteLayout.Resources, Is.Empty);
+
+      var chromeSettings = webTestSettings.Chrome;
+      Assert.That(chromeSettings, Is.Not.Null);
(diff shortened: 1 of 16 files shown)


Remotion/Web/Development.WebTesting.UnitTests/SetUpFixture.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Remotion.Web.Development.WebTesting.UnitTests
+{
+  [SetUpFixture]
+  public class SetUpFixture
+  {
+    [OneTimeSetUp]
+    public void OneTimeSetUp ()
+    {
+      WebTestSettings.Current = WebTestSettings.CreateAppConfigBasedWebTestSettings();
```
MichaelKetting, line 29: Unit test: wouldn't it be easy/helpful to use json-based here?

Remotion/Web/Development.WebTesting/Configuration/WebTestSettings.cs
```diff
@@ -57,11 +63,35 @@ public static IWebTestSettings Current
       }
     }
```
MichaelKetting, line 64: Something I missed in the previous PR: for cases when setting the current value is a very conscious act, I prefer having a SetCurrent method instead

Remotion/Web/Development.WebTesting/Configuration/WebTestSettings.cs
```diff
@@ -57,11 +63,35 @@ public static IWebTestSettings Current
...
-    private static IWebTestSettings CreateDefaultWebTestSettings ()
+    public static IWebTestSettings CreateAppConfigBasedWebTestSettings ()
     {
       var settings = (WebTestConfigurationSection)ConfigurationManager.GetSection("remotion.webTesting");
       Assertion.IsNotNull(settings, "Configuration section 'remotion.webTesting' missing.");
       return settings;
     }
+
+#if !NETFRAMEWORK
+    public static IWebTestSettings CreateAppSettingsJsonBasedWebTestSettings ()
```
MichaelKetting, line 74: Is this still json-based?

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestChromiumSettingsDto.cs
```diff
@@ -0,0 +1,31 @@
...
+//
+#if !NETFRAMEWORK
+using Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chromium;
+
+namespace Remotion.Web.Development.WebTesting.Configuration.IConfiguration
+{
+  /// <summary>
+  /// DTO object for <see cref="IWebTestChromiumSettings"/>, which is deserialized by the .NET configuration infrastructure.
+  /// </summary>
+  public class WebTestChromiumSettingsDto : IWebTestChromiumSettings
```
MichaelKetting, line 25: With introduction of c#12, we can make it immutable. Is there still a need for the drop suffix?

MichaelKetting (changes requested):
Still need to review the drop. Got some questions / remarks in the meantime.

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestTestSiteLayoutDto.cs
```diff
@@ -0,0 +1,38 @@
...
+  /// <summary>
+  /// DTO object for <see cref="IWebTestTestSiteLayoutSettings"/>, which is deserialized by the .NET configuration infrastructure.
+  /// </summary>
+  public class WebTestTestSiteLayoutDto : IWebTestTestSiteLayoutSettings
+  {
+    /// <inheritdoc />
+    [Required]
+    public string RootPath { get; set; } = "";
+
+    public List<string> Resources { get; set; } = new();
```
MichaelKetting, line 32: can the dto serializer handle ReadOnlyCollection<string>?

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestHostingSettingsDto.cs
```diff
@@ -0,0 +1,53 @@
...
+  {
+    /// <inheritdoc />
+    [Required]
+    public string Name { get; set; } = default!;
+
+    /// <inheritdoc />
+    [Required]
+    public string Type { get; set; } = "";
+
+    public Dictionary<string, string> Parameters { get; set; } = new();
```
MichaelKetting, line 37: Can the json config deserializer handle ReadOnlyDictionary<...>.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/IConfiguration/WebTestSettingsDtoTest.cs
```diff
@@ -0,0 +1,200 @@
...
+using Remotion.Web.Development.WebTesting.Configuration;
+using Remotion.Web.Development.WebTesting.Configuration.IConfiguration;
+using Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chromium;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration.IConfiguration
+{
+  [TestFixture]
+  public class WebTestSettingsDtoTest
+  {
+    private const string c_testConfig = @"{
```
MichaelKetting, line 35: one config is a field, the other is inline. we use both only once. I would prefer a symmetric approach.

MichaelKetting (commented):
Reviewed tests.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/IConfiguration/WebTestSettingsDtoTest.cs
```diff
@@ -0,0 +1,200 @@
...
+  }
+}";
+
+    [Test]
+    public void Deserialize_MinimalConfig ()
+    {
+      var webTestSettings = DeserializeFromString(@"{
+  ""remotion.WebTesting"": {
+    ""BrowserName"": ""Edge"",
+    ""SearchTimeout"": ""00:00:06"",
```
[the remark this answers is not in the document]
a-ctor: Parse error: empty string is not a valid `TimeSpan`

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestHostingSettingsDto.cs
a-ctor: No but we can use `IReadOnlyDictionary<T, K>` instead

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestTestSiteLayoutDto.cs
a-ctor: No but we can use `IReadOnlyList<T>`

Remotion/Web/Development.WebTesting/Configuration/WebTestSettings.cs
a-ctor: What do you mean with "still json-based"? This convenience setup method uses `appsettings.json` as its main config file which is why I chose that name. It does include other configuration sources, like environment variables, but the JSON one is the main source.

Remotion/Web/Development.WebTesting.UnitTests/SetUpFixture.cs
a-ctor: It doesn't really matter here. We already have a `Web.config` that is setup for the tests and is used in unrelated tests. All the new tests do not rely on `WebTestSettings.Current` so a change here would not have an effect on the new tests.

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestHostingSettingsDto.cs
MichaelKetting: then we should do that please.

Remotion/Web/Development.WebTesting/Configuration/IConfiguration/WebTestTestSiteLayoutDto.cs
MichaelKetting: yes please.

Remotion/Web/Development.WebTesting/Configuration/WebTestSettings.cs
MichaelKetting: quick detail question: the code in the method doesn't do anything to actually have a json dependency, does it? the application developer could use any implementation of the config system, even though most use json?

if so, we should not call the public api json-based since that's not the identifying characteristic of the new config api or our implementation in here.

Remotion/Web/Development.WebTesting.UnitTests/SetUpFixture.cs
MichaelKetting: Ah, I see. we use a config file for unit testing some of the features (i.e. WebTestHelper.CreateFromConfiguration). Given we will be deprecating and then removing the xml-based config anyway in the future, wouldn't it be best to switch to either an appsettings based version or create the config in code?

Reasoning: leaving legacy code as the main version we are working with feels like we're pushing cleanup down the road when we could do it right now anyway.

It would be a good idea to one day move access to the static WebTestSettings.Current property into the ctor of WebTestConfigurationFactory so we only have a single access and then we can refector the tests to better isolate this concern.

Remotion/Web/Development.WebTesting/Configuration/WebTestSettings.cs
a-ctor: We might want to talk about the dependencies that we bring in with the addition of adding the hosting infrastructure. It does have a dependency on System.Text.Json for JSON parsing. We provide a default here, if you want to use any kind of other custom JSON parsing, settings, configuration sources, etc. you will have to call this method yourself.

Remotion/Web/Development.WebTesting/Configuration/Legacy/WebTestHostingAdapter.cs
```diff
@@ -42,6 +106,12 @@ public WebTestHostingAdapter (ProviderSettings providerSettings)
     public string Type => _providerSettings.Type;
 
     /// <inheritdoc />
-    public NameValueCollection Parameters => _providerSettings.Parameters;
+    public IReadOnlyDictionary<string, string> Parameters
```
MichaelKetting, line 109: I'll admit, I expected this to end up as a copy operation.

Since the providesettings are private anyway, would a copy+cache of the copied value have been sufficient and resulted in less code? Since we want to kick it anyway and the code looks correct, I don't mind either way.

MichaelKetting (approved):

Remotion/Web/Development.WebTesting/Configuration/Legacy/WebTestHostingAdapter.cs
a-ctor: Copy and cache would be out-of-date when somebody changes the `NameValueCollection`. As that is more likely than replacing the collection in its entirety I wrapped it in as a read-only. Probably this is already overkill but it wasn't that hard to implement so...

PR #1006, opened 2024-01-26
RM-9070 Upgrade chrome to v121

Frazerus:
https://re-motion.atlassian.net/browse/RM-9070


Frazerus:
Chrome changed something with it's dropdown menus during the update.
Our drop down menus are now not shown completely anymore:
image

Before:
DropDownMenuControlObjectTest Chrome Browser ScrollIntoView_WithLongDropDownMenu_ScrollInvisibleItemIntoView

Frazerus:
This might be some margin error instead, but I'm not sure.

MichaelKetting:
> Chrome changed something with it's dropdown menus during the update. Our drop down menus are now not shown completely anymore:

This seems to be related to the avaialble screenheight (see issue in Headless Mode). The DropDownMenu has so many items in the dropdown element that the maximum height is required. This height is calculated from the available screenspace minus a top/bottom margin. So, if the browser after the update has a couple pixel more or less available for the visible html due to window chrome, the popup grows/shrinks in size and then the last / first visible menu item might not have enough space to be fully visible (gets scrolled out of view).

TL;DR: annoying because screenshot needs update but fully expected behavior when the availble viewport size changes.

MichaelKetting (commented):
review was requeted by github, am assuming that's an automation issue?

MichaelKetting:
> Chrome changed something with it's dropdown menus during the update. Our drop down menus are now not shown completely anymore:

Reading the title of the second screenshot, I got some very important details:
The idea was to scroll the menuitem fully into view. apparently, before it was possible, now it isn't. Based on the available screensize, the vertical scrollbar should not scroll fully to the end, instead stop when the red item is fully visible and leaving item99 partially obscured instead. Without checking the test to see if the task in the test is to make the green or the red item fully visible, I can't say if the current behavior is correct or a bug.

If the task is to scroll the green item (item94) into view, it's okay if the red item is partially obscured. 

If the task is to scroll the red item (item91) into view, the test shows a bug with Chrome 121. 

-> please check what the test is supposed to accomplish and then we know if it's okay (screenshot needs patching) or if we found an issue.

MichaelKetting:
@Frazerus in case that the Chrome 121 bug on scrollbars is indeed the culprit we should defer the upgrade v121 (decide separately for Chrome and Edge depending on tests) and stick to v120 for the next Chrome release or two and see how Chromium team decides to handle the bug.
cc @a-ctor

Frazerus:
The first screenshot is done manually, no testing going on there. I think the test is fine, but we cant fully see all the items were were previously supposed to see. 
It's not a scrollbar issue, removing the scrollbar does not change the height.

a-ctor (approved):

PR #1008, opened 2024-01-30
Feature/rm 6099 boc list to initialize the row menus on demand during the event handling phase of the postback

Frazerus:
https://re-motion.atlassian.net/browse/RM-6099


Remotion/ObjectBinding/Web/UI/Controls/BocList.RowMenuSupport.cs
```diff
@@ -80,6 +94,52 @@ private void EnsureRowMenusInitialized ()
       _rowMenus = rowMenus.ToArray();
     }
 
+    private bool TryGetRowMenu (string targetInfo, [NotNullWhen(true)] out string? id)
```
a-ctor, line 97: `TryGetRowMenuId` since we don't actually return a row menu?

Remotion/ObjectBinding/Web/UI/Controls/BocList.RowMenuSupport.cs
```diff
@@ -80,6 +94,52 @@ private void EnsureRowMenusInitialized ()
...
+    }
+
+    private void EnsureRowMenuInitialized (string id)
+    {
+      if (_rowMenusInitialized)
+        return;
+
+      _singleRowMenuInitialized = true;
+
+      if (! AreRowMenusEnabled)
```
a-ctor, line 117:
```suggestion
      if (!AreRowMenusEnabled)
```

Remotion/ObjectBinding/Web/UI/Controls/BocList.RowMenuSupport.cs
```diff
@@ -80,6 +94,52 @@ private void EnsureRowMenusInitialized ()
...
+      var rowMenu = new BocListRowMenu(this, row.ValueRow);
+      rowMenu.ID = GetRowMenuID(row.ValueRow);
+      rowMenu.EventCommandClick += (sender, e) => HandleRowMenuItemClick(sender, e, OnRowMenuItemEventCommandClick);
+      rowMenu.WxeFunctionCommandClick += (sender, e) => HandleRowMenuItemClick(sender, e, OnRowMenuItemWxeFunctionCommandClick);
+
+      _rowMenusPlaceHolder.Controls.Add(rowMenu);
+      WebMenuItem[] menuItems = InitializeRowMenuItems(row.ValueRow.BusinessObject, row.ValueRow.Index);
+      rowMenu.MenuItems.AddRange(menuItems);
+
+      rowMenus.Add(rowMenu);
```
a-ctor, line 138: We are duplicating this part of the code, which makes it easy to forget changes when updating one of the places. I think we should extract this logic into its own method

Remotion/ObjectBinding/Web/UI/Controls/BocList.RowMenuSupport.cs
```diff
@@ -80,6 +94,52 @@ private void EnsureRowMenusInitialized ()
...
+      id = null;
+      return false;
+    }
+
+    private void EnsureRowMenuInitialized (string id)
+    {
+      if (_rowMenusInitialized)
+        return;
+
+      _singleRowMenuInitialized = true;
```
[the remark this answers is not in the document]
Frazerus: Great question: I don't think it matters, as this is merely used for answering things in postbacks. There might be an issue if the same row menu gets initialized twice (same id), but I'm not certain that that would actually problematic. 

We might want to have a list (or dictionary) of created initialized rows instead of this boolean, and check for its emptiness instead.

@MichaelKetting any input?

Remotion/ObjectBinding/Web/UI/Controls/BocList.cs
```diff
@@ -455,7 +455,15 @@ protected override void OnLoad (EventArgs e)
...
+
+        var postBackCollection = Context!.Request.Form;
+        var targetInfo = postBackCollection["__EVENTTARGET"];
+
+        if (targetInfo != null && TryGetRowMenuId(targetInfo, out var row))
+        {
+          EnsureRowMenuInitialized(row);
+        }
+
         EnsureCustomColumnsInitialized(columns);
```
MichaelKetting, line 467: Please check, I think it should be okay to reorder these so the EnsureCustomColumns is before the complicated custom code.

MichaelKetting (changes requested):

Frazerus:
Extra change: line 129 in BocList.RowMenuSupport, we chnaged all other instances of empty array initializes, so why not this too.

MichaelKetting (approved):

a-ctor (approved):

PR #1009, opened 2024-01-31
RM-9072 Fix scrollbar style for Chromium browsers

a-ctor:
https://re-motion.atlassian.net/browse/RM-9072

### DIFF
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
@@ -430,21 +430,25 @@ a.remotion-themed:focus,
   outline: none;
 }
 
-select.remotion-themed,
-textarea.remotion-themed,
-.remotion-themed select,
-.remotion-themed textarea
+/* If either scrollbar-width`/`scrollbar-color` is specified, the `::-webkit-scrollbar` selectors won't work  */
+@supports not selector(::-webkit-scrollbar)
 {
-  scrollbar-width: var(--remotion-themed-scrollbar-style);
-  scrollbar-color: var(--remotion-themed-scrollbar-color-thumb) var(--remotion-themed-background-color-control);
-}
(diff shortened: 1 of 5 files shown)


MichaelKetting (commented):
Note @a-ctor : given the reply by the chromium team, it's possible we will only do this in v5.0, and drop the rounded corner in v6.0. descicison is not final yet.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -514,8 +518,22 @@ img.disabled,
...
-  scrollbar-width: var(--remotion-themed-scrollbar-style);
-  scrollbar-color: var(--remotion-themed-scrollbar-color);
+}
+
+.remotion-themed.scrollable > *
+{
+  scrollbar-color: initial; /* Stop inheriting the scrollbar color to child elements */
+  scrollbar-width: initial; /* Stop inheriting the scrollbar width to child elements */
+}
+
```
MichaelKetting, line 528: Why did you move this one up in the file? it's a deeper nested selector so semantically "after" the others.

MichaelKetting (changes requested):
The changes look good, one detail in comments.

plan for how we proceed:
on develop, we kick webkit-styles for scrollbards instead of doing the changes.

for NovaViso versions of remotion, we do a backport of the changes you did here.

Reasoning:
Chromium team recommends no longer using webkit-styles for this and instead only do with and color while leaving shape to the browser rendering.

This will result in released versions of re-motion retaining the existing behavior (hoping that Chrome won't drop those until the product lifetime for those versions ends). For new release of re-motion, we change the look in chrome. In Edge, user won't notice a difference due to the rounded look of the native scrollbar. 

please update  the ticket and create breaking change note. If sensible, we can do separate tickets for develop and the backports since it's a bugfix in the backports and a task+breakingchange on develop.

PR #1011, opened 2024-01-31
RM-9056 Add label to the list menu region element

a-ctor:
https://re-motion.atlassian.net/browse/RM-9056

### DIFF
--- a/Remotion/Web/UnitTests/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRendererTest.cs
+++ b/Remotion/Web/UnitTests/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRendererTest.cs
@@ -243,6 +243,54 @@ namespace Remotion.Web.UnitTests.Core.UI.Controls.ListMenuImplementation.Renderi
       wrapper.AssertAttributeValueEquals(DiagnosticMetadataAttributes.IsDisabled, (!_control.Object.Enabled).ToString().ToLower());
     }
 
+    [Test]
+    public void Render_WithNoHeading_NoHeaderAndNoLabelledBy ()
+    {
+      SetUpGetPostBackLinkExpectations(true);
+
+      _control.Setup(e => e.Heading).Returns(WebString.Empty);
+      var wrapper = GetAssertedWrapper();
+
+      wrapper.AssertNoAttribute("aria-labelledby");
+      wrapper.AssertChildElementCount(1);
+    }
+
+    [Test]
+    public void Render_WithHeadingButNoHeadingLevel_RendersAndLinksSpan ()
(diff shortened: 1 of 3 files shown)


Remotion/Web/UnitTests/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRendererTest.cs
```diff
@@ -243,6 +243,52 @@ public void RenderDiagnosticMetadataAttributes ()
...
+
+      _control.Setup(e => e.Heading).Returns(WebString.CreateFromText("My heading"));
+      _control.Setup(e => e.HeadingLevel).Returns((HeadingLevel?)null);
+      var wrapper = GetAssertedWrapper();
+
+      wrapper.AssertAttributeValueEquals("aria-labelledby", "MyListMenu_Heading");
+
+      var span = wrapper.GetAssertedChildElement("span", 0);
+      span.AssertAttributeValueEquals("id", "MyListMenu_Heading");
+      span.AssertAttributeValueEquals("class", "screenReaderText");
```
MichaelKetting, line 271: Here got one more bit: when it's a span, please add `hidden` attribute instead of css class `screenreadertext`. The text shouldn't be read just by virtually browsing the screen, it should only show up when reading the region.

Note : I am pretty sure that hidden is correct here. We do that with labels for controls, too, e.g. in bocdatetimevalue. Please verify

MichaelKetting (changes requested):

Remotion/Web/UnitTests/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRendererTest.cs
MichaelKetting: If I read the code correctly, you got the hidden attribute for both span and  heading instead of just the heading. The heading as 'hidden' doesn't have value, please update to only do the span hidden and the heading remains screen reader.

MichaelKetting (changes requested):
See comment on hidden attribute

Remotion/Web/UnitTests/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRendererTest.cs
a-ctor: My bad I misread your comment. So if there is a heading level specified it should be targetable using the arrow keys and be the focus of the region? Because that is the current behavior and it seems kind of strange to target the invisible heading. Looks like this:
image

MichaelKetting (approved):

PR #1013, opened 2024-02-06
Bugfix/rm 9082 wxe resource page steps resolves incorrect type

a-ctor:
https://re-motion.atlassian.net/browse/RM-9082


Remotion/Web/Core/ExecutionEngine/WxeResourcePageStep.cs
```diff
@@ -34,6 +34,9 @@ public class WxeResourcePageStep : WxePageStep
     /// <summary>
     ///   Calls the page using the calling assemby's resource directory.
     /// </summary>
+#if !NETFRAMEWORK
+    [Obsolete("Use a constructor with type/assembly argument instead.", error: true, DiagnosticId = "RMWEB0002")]
```
MichaelKetting, line 38: @a-ctor should we start maintaining a type with the const DiagnosticIDs so we don't acccidentally reuse?

Remotion/Web/Development.Analyzers/RMWEB0002_InvalidWxeResourcePageStepConstructorCodeFix.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Web.Development.Analyzers;
+
+[ExportCodeFixProvider(LanguageNames.CSharp)]
+public class RMWEB0002InvalidWxeResourcePageStepConstructorCodeFix : CodeFixProvider
```
MichaelKetting, line 30: @Frazerus The code fix is meant to take the type, where the being-analyzed ctor of WxeResourcePageStep is used and insert it as the first parameter into this ctor.

Remotion/Web/Development.Analyzers/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Web.Development.Analyzers
+{
+  [DiagnosticAnalyzer(LanguageNames.CSharp)]
+  public class RMWEB0003TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer : DiagnosticAnalyzer
```
MichaelKetting, line 28: @Frazerus 
The analyzer is meant to check that the ctor of WxeResourcePageStep uses the type as ctor parameter where it is used.

MichaelKetting, line 28: @a-ctor  & @Frazerus would it be helpful if our analyzer and code fixer get a summary that decscribes what they will do?

Remotion/Web/Development.Analyzers/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+  {
+    private record SymbolContext (
+        INamedTypeSymbol WxeResourcePageStepSymbol,
+        INamedTypeSymbol SystemTypeSymbol);
+
+    public static readonly DiagnosticDescriptor RMWEB0003DiagnosticDescriptor = new(
+        "RMWEB0003",
+        $"Passing a type from another assembly to the constructor of WxeResourcePageStep is probably a mistake.",
+        "'{0}' is not from the same assembly as the containing type '{1}'. This might be a mistake.",
```
MichaelKetting, line 37: simplify so we don't repeat "mistake".

Remotion/Web/Development.Analyzers.IntegrationTests/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+//
+using System;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Verifier = Remotion.Web.Development.Analyzers.IntegrationTests.CSharpAnalyzerVerifier<Web.Development.Analyzers.RMWEB0003TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer>;
+
+namespace Remotion.Web.Development.Analyzers.IntegrationTests
+{
+  [TestFixture]
+  public class RMWEB0003TypeNotFromSameAssemblyWxeResourcePageStepAnalyzerTests
```
MichaelKetting, line 25: Can we also do a test for the code fix?

Remotion/Web/Development.Analyzers/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Web.Development.Analyzers
+{
+  [DiagnosticAnalyzer(LanguageNames.CSharp)]
+  public class RMWEB0003TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer : DiagnosticAnalyzer
+  {
+    private record SymbolContext (
+        INamedTypeSymbol WxeResourcePageStepSymbol,
+        INamedTypeSymbol SystemTypeSymbol);
```
Frazerus, line 32: I don't see why you would need the SystemTypeSymbol

Remotion/Web/Development.Analyzers/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+            compilationContext.RegisterSyntaxNodeAction(
+                ctx => AnalyzeObjectCreationExpression(ctx, symbolContext),
+                ImmutableArray.Create(SyntaxKind.ObjectCreationExpression, SyntaxKind.ImplicitObjectCreationExpression));
+          });
+    }
+
+    private void AnalyzeObjectCreationExpression (SyntaxNodeAnalysisContext ctx, SymbolContext symbolContext)
+    {
+      var node = (BaseObjectCreationExpressionSyntax)ctx.Node;
+      if (ctx.SemanticModel.GetOperation(node) is not IObjectCreationOperation operation || !SymbolEqualityComparer.Default.Equals(operation.Type, symbolContext.WxeResourcePageStepSymbol))
```
Frazerus, line 71: a bit long, and since we already have so many exit conditions, splitting the or in separate if's probably won't hurt either :)

Frazerus (approved):

Frazerus:
Aren't we missing the new rules in the AnalzyerReleases.Shipped/Unshipped.md?

Remotion/Web/Development.Analyzers.IntegrationTests/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+  {
+    [Test]
+    public async Task TypeFromOtherAssembly_Invalid ()
```
MichaelKetting, line 28: I found two more things: 
1. we need to detect usaes GetType(). If it's a lot of work, we can do it later. Using GetType will likely result in incorrect code once derived types are used. 
2. WxeResourceUserControlStep also needs to be covered. We can do this later if it's too much work.

Remotion/Web/Development.Analyzers.IntegrationTests/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                               new WxeResourcePageStep(typeof(int), "Test.aspx");
+                             }
+                             
+                             public WxeResourcePageStep Step => new(typeof(string), "Test.aspx");
+                           }
+                           """;
+
+      var diagnostic1 = Verifier.Diagnostic()
+          .WithSpan(7, 5, 7, 54)
+          .WithMessage("'int' is not from the same assembly as the containing type 'A'. This might be a mistake.");
```
MichaelKetting, line 46: "Type 'int' is not defined in the same assembly as the containing type 'A'. If this is intentional, you can suppress the diagnostic ."

MichaelKetting (approved):

Remotion/Web/Development.Analyzers.IntegrationTests/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzerTests.cs
a-ctor: 2) done

1) Detection of `.GetType` is a bit more complex that just checking for a `typeof(...)` expression since the `.GetType` is method call and you need to resolve everything to get determine if there is a problem. Right now, we only check for `typeof(xxx)` since that is straightforward, checking for something else gets difficult quickly as you need to check variables involved and on.

How about we have a general warning that we emit if you don't supply a value using a `typeof(xxx)`? This would reduce the complexity and requires consumers to check if they are affect. If they want they can even turn off this warning in the editorconfi

a-ctor:
> Aren't we missing the new rules in the AnalzyerReleases.Shipped/Unshipped.md?

Yes we are missing those except when I add them I get errors. For some reason it does not detect the RMWEB0003 analyzer for the purposes of release tracking. I searched a bit and did not find a good solution. I also did not find a good reason as to why release tracking is necessary so I would keep it like this. In theory, we would not be able to compile without adding entries to the release MD files...

Remotion/Web/Development.Analyzers/WebDiagnosticIds.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.Reflection;
+
+namespace Web;
```
MichaelKetting, line 3: Ah okay, we need to fix the namespace please.

Remotion/Web/Development.Analyzers/RMWEB0002_InvalidWxeResourcePageStepConstructorCodeFix.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Web.Development.Analyzers;
```
MichaelKetting, line 27: Analyzer should start with `Remotion` too.

MichaelKetting (changes requested):
found namespace problem in existing analyzers. we should fix those before we do more analyzers?

Remotion/Web/Development.Analyzers.IntegrationTests/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzerTests.cs
MichaelKetting: Hmm...yeah, and someone who supplies a programmatically resolved type can just use `type.Assembly` instead of `type` to not have to suppress the warning.

a-ctor: Actually, do we really need an analyzer for the type parameter? Initially the reason we wanted it was to prevent copy-paste errors. With the code fix this isn't a problem anymore...

MichaelKetting: In the future, the ctor without the type parameter is going to the way of the Dodo. So new code will need to be written correctly from the start which will still produce copy&paste errors if one's not careful.

Remotion/Web/Development.Analyzers/WebDiagnosticIDs.cs
```diff
@@ -0,0 +1,46 @@
...
+
+  /// <summary>
+  /// Diagnostic ID reported when a type is passed in the constructor of WxeResourcePageStep that does not belong to the calling assembly.
+  /// This is most likely a (copy &amp; paste)-mistake as the passed type usually belongs to the same assembly.
+  /// </summary>
+  /// <example>
+  /// <code>
+  /// new WxeResourcePageStep(typeof(int), "Test.aspx"); // &lt;-- RMWEB0003 emitted
+  /// new WxeResourcePageStep(typeof(string), new WxeVariableReference("test")); // &lt;-- RMWEB0003 emitted
+  /// vs.
```
MichaelKetting, line 40:
```suggestion
  /// // vs.
```

Remotion/Web/Development.Analyzers/RMWEB0003_TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+  [DiagnosticAnalyzer(LanguageNames.CSharp)]
+  public class RMWEB0003TypeNotFromSameAssemblyWxeResourcePageStepAnalyzer : DiagnosticAnalyzer
+  {
+    private record SymbolContext (
+        INamedTypeSymbol WxeResourcePageStepSymbol,
+        INamedTypeSymbol WxeResourceUserPageStepSymbol);
+
+    public static readonly DiagnosticDescriptor RMWEB0003DiagnosticDescriptor = new(
+        WebDiagnosticIDs.RMWEB0003_PossiblyInvalidTypeForWxeResourcePageStepConstructor,
+        $"Passing a type from another assembly to the constructor of WxeResourcePageStep is probably a mistake.",
```
MichaelKetting, line 40:
```suggestion
        $"Passing a type from another assembly to the constructor of WxeResourcePageStep or WxeResourceUserControlStep can indicate a copy&paste mistake.",
```

MichaelKetting (approved):
Some small string stuff. 

And I believe the discussion got all information it needs. If not, discuss tomorrow face2face before merge :)

MichaelKetting:
> > Aren't we missing the new rules in the AnalzyerReleases.Shipped/Unshipped.md?
> 
> Yes we are missing those except when I add them I get errors. For some reason it does not detect the RMWEB0003 analyzer for the purposes of release tracking. I searched a bit and did not find a good solution. I also did not find a good reason as to why release tracking is necessary so I would keep it like this. In theory, we would not be able to compile without adding entries to the release MD files...

@a-ctor  & @Frazerus best guess: because we have a fix but no analyzer for the one with the obsolete-warning? Not sure what this means for future endeavors. Please create a follow-up ticket and we should analyze this and update the guidelines accordingly for future selves.

PR #1018, opened 2024-02-06
RM-9084 Remove the webkit specific CSS scrollbar styling

a-ctor:
https://re-motion.atlassian.net/browse/RM-9084


MichaelKetting (approved):
Please document as breaking change.

PR #1026, opened 2024-02-08
RM-9059 Add content details to exception

KuJo7:
https://re-motion.atlassian.net/browse/RM-9059

### DIFF
--- a/Remotion/Development/UnitTests/Web/UnitTesting/UI/Controls/Rendering/HtmlHelperBaseTest.cs
+++ b/Remotion/Development/UnitTests/Web/UnitTesting/UI/Controls/Rendering/HtmlHelperBaseTest.cs
@@ -15,6 +15,7 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Xml;
 using Moq;
 using NUnit.Framework;
 using Remotion.Development.Web.UnitTesting.UI.Controls.Rendering;
@@ -122,6 +123,19 @@ namespace Remotion.Development.UnitTests.Web.UnitTesting.UI.Controls.Rendering
       Assert.That(document, Is.Not.Null);
       Assert.That(document.DocumentElement, Is.Not.Null);
       Assert.That(document.DocumentElement.Name, Is.EqualTo("TopLevelElement"));
+    }
+
+    [Test]
+    public void Fail_GetResultDocument ()
+    {
+      _htmlHelper.Writer.Write("randomError");
+
+      Assert.That(
+          () => _htmlHelper.GetResultDocument(),
+          Throws.Exception
+              .TypeOf<XmlException>()
+              .With.Message.EqualTo("Data at the root level is invalid. Line 1, position 665.\r\n\r\nContent:\r\n" + c_document + "randomError")
+              .And.With.InnerException.Null);
     }
 
     [Test]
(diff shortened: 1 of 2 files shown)


Remotion/Development/Web/UnitTesting/UI/Controls/Rendering/HtmlHelperBase.cs
```diff
@@ -104,12 +104,20 @@ public string GetDocumentText ()
     /// <returns>The <see cref="XmlDocument"/> built from the document text.</returns>
     public XmlDocument GetResultDocument ()
     {
-      XmlDocument document = new XmlDocument();
-      using (TextReader reader = new StringReader(GetDocumentText().Replace("&nbsp;", "&amp;nbsp;")))
+      var text = GetDocumentText().Replace("&nbsp;", "&amp;nbsp;");
```
MichaelKetting, line 107: Please add a unit test for the new stuff.

MichaelKetting (changes requested):

Remotion/Development/Web/UnitTesting/UI/Controls/Rendering/HtmlHelperBase.cs
```diff
@@ -104,12 +104,20 @@ public string GetDocumentText ()
...
+        XmlDocument document = new XmlDocument();
+        using (TextReader reader = new StringReader(text))
+        {
+          document.Load(reader);
+        }
+        return document;
+      }
+      catch (XmlException ex)
+      {
+        throw new InvalidOperationException(ex.Message + "\nContent: " + text, ex);
```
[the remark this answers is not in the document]
KuJo7: I checked with Lukas and we think the inner exception is not necessary here since it doesn't make any difference in this case.

Remotion/Development/UnitTests/Web/UnitTesting/UI/Controls/Rendering/HtmlHelperBaseTest.cs
```diff
@@ -124,6 +124,16 @@ public void Succeed_GetResultDocument ()
...
     }
 
+    [Test]
+    public void Fail_GetResultDocument ()
+    {
+      _htmlHelper.Writer.Write("randomError");
+
+      Assert.That(
+          () => _htmlHelper.GetResultDocument(),
+          Throws.Exception.With.Message.EqualTo("Data at the root level is invalid. Line 1, position 665.\r\n\r\nContent:\r\n" + c_document + "randomError"));
```
MichaelKetting, line 134:
```suggestion
          Throws.Exception.OfType<XmlException>().With.Message.EqualTo("Data at the root level is invalid. Line 1, position 665.\r\n\r\nContent:\r\n" + c_document + "randomError").And.With.InnerException.IsNull);
```

Note: I think the part with the InnerExcpetion is possible, but I did not acutally try it. If it's not a perfect match, please verify syntax and if it appears to not work at all with the single-line syntax while remaining type safe, we leave it out. The general idea is to document that we intentionally skipped the inner exception

MichaelKetting: I don't need to recheck it but perhaps ask @Frazerus for a once over when you updated to make sure I didn't miss anything?

MichaelKetting (approved):

PR #1027, opened 2024-02-08
bugfix/RM-9089-Missing-Nullability-Annotation-on-WxeFunctionStateManager.GetItem

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-9089

### DIFF
--- a/Remotion/Web/Core/ExecutionEngine/WxeHandler.cs
+++ b/Remotion/Web/Core/ExecutionEngine/WxeHandler.cs
@@ -310,7 +310,7 @@ namespace Remotion.Web.ExecutionEngine
       }
 
       WxeFunctionStateManager functionStateManager = WxeFunctionStateManager.Current;
-      if (functionStateManager.IsExpired(functionToken))
+      if (!functionStateManager.TryGetLiveValue(functionToken, out var functionState))
       {
         if (isPostBackAction)
         {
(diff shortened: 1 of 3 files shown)


Remotion/Web/Core/ExecutionEngine/WxeHandler.cs
```diff
@@ -363,6 +364,22 @@ protected WxeFunctionState CreateNewFunctionState (HttpContext context, Type typ
           throw new WxeException(string.Format("Function missing in WxeFunctionState {0}.", functionState.FunctionToken));
         return functionState;
       }
+
+      static bool TryGetLiveFunctionState (WxeFunctionStateManager functionStateManager, string functionToken, [MaybeNullWhen(false)] out WxeFunctionState functionState)
+      {
+        // When performing WxeFunctionStateManager.CleanupExpired() during HttpApplication.EndRequest, the operation is done after the session's lock was released.
+        // Therefor it is possible that WxeFunctionStateManager.IsExpired(...) returns true for a live WxeFunctionState while WxeFunctionStateManager.GetItem(...) results 
+        // in a null-value. The effects of this can be mitigated by combing the operations into a single try-get operation.
```
a-ctor, line 372: I don't think I understand what this comment is trying to tell me.

> When performing WxeFunctionStateManager.CleanupExpired() during HttpApplication.EndRequest, the operation is done after the session's lock was released.

What is "the operation" in this context? `CleanupExpired` does all of its operations inside the lock. Any item that was expired at the time, has been removed from `_functionStates` and the session.

> Therefor it is possible that WxeFunctionStateManager.IsExpired(...) returns true for a live WxeFunctionState while WxeFunctionStateManager.GetItem(...) results in a null-value. The effects of this can be mitigated by combing the operations into a single try-get operation.

As `IsExpired` and `GetItem` are done with two locks, any combination for `true/false` and `null/not null` is possible here. Because of this, I don't see how the addition of `GetItem` is supposed to solve the problem.

a-ctor (approved):

PR #1028, opened 2024-02-16
RM-9009 Update edge to v119

Frazerus:
Since we can't update to v121 currently
https://re-motion.atlassian.net/browse/RM-9009


Frazerus:
* Boclist MenuBlock: 
image
Dropdown got 2 px wider than before, still looks ok.

Frazerus:
Resolve Border Element B used the same screenshot as chrome, but chrome introduced problems here which are fixed in edge. Therefore split screenshots between chrome and edge.

Frazerus:
WithMarkedBadge:
The sidebar reappeared, we therefore have less space in the browser. I don't think that the sidebar can be easily removed as before, so an extra ticket might be required.

Frazerus:
DropDownMenuTests:
The view got slimmer probably because of the rounded scrollareas needing less space or the sidebar being shown again.
Height changed due to environmental circumstances

PR #1029, opened 2024-02-20
RM-9092 Remove creation date from resource hash calculation

a-ctor:
https://re-motion.atlassian.net/browse/RM-9092


MichaelKetting (commented):
are you going to create a separate ticket for truncating the file path?

MichaelKetting (approved):
are you going to create a separate ticket for truncating the file path?

a-ctor:
@MichaelKetting I discovered that this was already implemented and I just forgot about it. As such, we are done after the creation date is removed.

PR #1031, opened 2024-02-21
Feature/rm 9087 improve safe context analyzers to detect correct usages

Frazerus:
https://re-motion.atlassian.net/browse/RM-9087

### DIFF
--- a/Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
+++ b/Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
@@ -164,6 +164,355 @@ public class A
     }
 
     [Test]
+    public async Task SafeContext_WithTasksParallelAndSafeContextBoundary_NoDiagnostic ()
+    {
+      const string input = @"
+using System;
+using System.Threading.Tasks;
+using System.Linq;
+using Remotion.Context;
+public class A
+{
+  public void Test ()
+  {
+    using var boundary = SafeContext.Parallel.OpenSafeContextBoundary();
+
+    SafeContext.Instance.SetData(""a"", 3);
+
+    Parallel.For(0, 5, i => { i = (int) SafeContext.Instance.GetData(""a"");});
+
+    Parallel.ForEach(Enumerable.Range(5, 10).ToList(), i => Console.Write(i));
+
+    Parallel.ForEachAsync(
+        Enumerable.Range(0, 5),
+        new ParallelOptions(), 
+        (i, token) => ValueTask.CompletedTask
+    );
+
+    Parallel.Invoke(() => Console.Write(""a""));
+  }
+}
+";
+      await Verifier.VerifyAnalyzerAsync(input, true);
+    }
+
+    [Test]
+    public async Task SafeContext_WithTasksParallelAndSafeContextBoundaryFromInstance_NoDiagnostic ()
+    {
+      const string input = @"
+using System;
+using System.Threading.Tasks;
+using System.Linq;
+using Remotion.Context;
+public class A
+{
+  public void Test ()
+  {
+    using var boundary = SafeContext.Instance.OpenSafeContextBoundary();
+
+    SafeContext.Instance.SetData(""a"", 3);
+
+    Parallel.For(0, 5, i => { i = (int) SafeContext.Instance.GetData(""a"");});
+
+    Parallel.ForEach(Enumerable.Range(5, 10).ToList(), i => Console.Write(i));
+
+    Parallel.ForEachAsync(
+        Enumerable.Range(0, 5),
+        new ParallelOptions(), 
+        (i, token) => ValueTask.CompletedTask
+    );
+
+    Parallel.Invoke(() => Console.Write(""a""));
+  }
+}
+";
+      await Verifier.VerifyAnalyzerAsync(input, true);
+    }
+
+    [Test]
+    public async Task SafeContext_WithTasksParallelAndSafeContextBoundaryAfterUsages_Diagnostic ()
+    {
+      const string input = @"
+using System;
+using System.Threading.Tasks;
+using System.Linq;
+using Remotion.Context;
+public class A
+{
+  public void Test ()
+  {
+    SafeContext.Instance.SetData(""a"", 3);
+
+    Parallel.For(0, 5, i => { i = (int) SafeContext.Instance.GetData(""a"");});
+
+    Parallel.ForEach(Enumerable.Range(5, 10).ToList(), i => Console.Write(i));
+
+    Parallel.ForEachAsync(
+        Enumerable.Range(0, 5),
+        new ParallelOptions(), 
+        (i, token) => ValueTask.CompletedTask
+    );
+
+    Parallel.Invoke(() => Console.Write(""a""));
+
+    using var boundary = SafeContext.Parallel.OpenSafeContextBoundary();
+  }
+}
+";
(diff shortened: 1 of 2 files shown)


Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -111,6 +123,89 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+      //Check whether the invocation expression should be covered
+      var invocationSymbolInfo = ctx.SemanticModel.GetSymbolInfo(ctx.Node);
+
+      if (invocationSymbolInfo.Symbol == null)
+        return;
+
+      var invocationOriginalDefinitionSymbol = invocationSymbolInfo.Symbol.OriginalDefinition;
+
+      var match = methodSymbols.FirstOrDefault(m => SymbolEqualityComparer.Default.Equals(m.methodSymbol, invocationOriginalDefinitionSymbol));
+      if (match == default)
```
a-ctor, line 140: I suggest not using `default` for null checks. Yes it evaluates to `null` in this case but that is not clear without knowing the type of the variable.

I treat the `default` vs. `null` decision by considering if I care about the value or not. In this case, a null check, we do want to compare against `null` so using `null` makes the code more explicit. Using `default` indicates that the exact value is not relevant but a value is needed. A good example for this would be the `TryXXX` pattern with its `out` parameter that needs to be assigned even if the operation failed. In such a case, it makes sense to assign `default` as the value is meant to be irrelevant.

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -111,6 +123,89 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+          .FirstAncestorOrSelf<BlockSyntax>(blockSyntax => blockSyntax
+              .ChildNodes()
+              .Where(
+                  child => !checkedChildrenSet.Contains(child))
+              .Select(
+                  child =>
+                  {
+                    checkedChildrenSet.Add(child);
+                    return child;
+                  })
```
a-ctor, line 192: Adding to a `HashSet<T>` returns a bool indicating if the value already existed.
```suggestion
              .Where(child => checkedChildrenSet.Add(child))
```

a-ctor (changes requested):
If you haven't tested the analyzer live please do after addressing the review changes. We should do a smoke test that ensures writing code (as in having incomplete syntax trees) does not break the analyzer. As such, test that writing Parallel statements cause diagnostics and then supress them by writing the using statements.

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -111,6 +123,89 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+              .OfType<LocalDeclarationStatementSyntax>()
+              .Where(statement => statement.UsingKeyword.IsKeyword())
+              .Where(statement => statement.GetLocation().SourceSpan.Start < ctx.Node.GetLocation().SourceSpan.Start)
+              .Select(statement => statement.Declaration.Variables.First().Initializer!.Value)
```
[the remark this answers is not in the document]
Frazerus: oftype ignores them :)

Frazerus:
As for the null / default checks: I thought that first ancestor or self would return the first ancestor if one was found matching the criteria, and if none were found, self would be returned. So for me it wasn't even a null check before 😉

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -14,6 +12,8 @@ namespace Core.Development.Analyzers
   [DiagnosticAnalyzer(LanguageNames.CSharp)]
   public class RMCORE0001_SafeContextAnalyzer : DiagnosticAnalyzer
   {
+    private record SymbolContext (ITypeSymbol? safeContextBoundarySymbol);
```
a-ctor, line 15: Since the property is public please rename to `SafeContextBoundarySymbol` (capital S)

a-ctor (approved):

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -163,6 +163,357 @@ public void Test ()
...
+
+    Parallel.Invoke(() => Console.Write(""a""));
+  }
+}
+";
+      await Verifier.VerifyAnalyzerAsync(input, true);
+    }
+
+    [Test]
+    public async Task SafeContext_WithTasksParallelAndSeveralLocalUsingDeclarations2_Diagnostic ()
```
MichaelKetting, line 375: is the method name incorrect? It should be `NoDiagnostic`?

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -163,6 +163,357 @@ public void Test ()
...
+      var diagnosticInvoke = Verifier.Diagnostic(Analyzer.AlternativeDescriptor)
+          .WithSpan(25, 5, 25, 46)
+          .WithMessage(
+              "'System.Threading.Tasks.Parallel.Invoke' should not be used with SafeContext by itself, use 'Remotion.Context.SafeContext.Parallel.OpenSafeContextBoundary()' to create a safe context boundary.");
+
+      await Verifier.VerifyAnalyzerAsync(input, true, diagnosticFor, diagnosticForEach, diagnosticForEachAsync, diagnosticInvoke);
+    }
+
+    [Test]
+    public async Task SafeContext_WithTasksParallelAndSeveralLocalUsingDeclarations1_Diagnostic ()
```
MichaelKetting, line 340: is the method name incorrect? It should be `NoDiagnostic`?

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -163,6 +163,357 @@ public void Test ()
...
+using System.Threading.Tasks;
+using System.Linq;
+using Remotion.Context;
+public class A
+{
+  public void Test ()
+  {
+    using (SafeContext.Parallel.OpenSafeContextBoundary()) 
+    {
+
```
MichaelKetting, line 423: found an extra blank line ?

Remotion/Core/Development.Analyzers.IntegrationTests/RMCORE0001_SafeContextAnalyzerTests.cs
```diff
@@ -163,6 +163,357 @@ public void Test ()
...
+using System.Threading.Tasks;
+using System.Linq;
+using Remotion.Context;
+public class A
+{
+  public void Test ()
+  {
+    using (SafeContext.Instance.OpenSafeContextBoundary()) 
+    {
+
```
MichaelKetting, line 458: found an extra blank line ?

MichaelKetting (approved):
I think I found some testname-issues and a bit of formatting in the tests.

Did not do a detailed review of the prod-code. trusting @a-ctor here  to better understand what's going on :)

cc @a-ctor

PR #1032, opened 2024-02-21
RM-9074 remove scrollbar on ResolveBorderElementB

Frazerus:
https://re-motion.atlassian.net/browse/RM-9074


Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.Shared.cs
```diff
@@ -329,6 +329,8 @@ public void ResolveBorderElementB ()
           {
             var home = Start();
 
+            if (Helper.MainBrowserSession.Headless)
```
MichaelKetting, line 332: In WebTestHelper.CreateNewBrowserSession we already adjust the size, just to a different default. I am interested in the details why we need a 20% larger size here. 

And could we use WebTestHelper.CreateNewBrowserSession with a "correct" window size passed via the tests. 

Right now, it feels more like a workaround than the offiical solution?

MichaelKetting (commented):
Got an open question on how to best add the special case

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.Shared.cs
Frazerus: our default creates a scollbar here, as the site is too long. We could change the default for all headless screenshot tests as well. I thought since we only needed to change to another value in this test, we might want to keep the default.

Frazerus: We need a 20% larger size because this page does not fit into the standard size. 

The shared screenshot tests in headless should have a different size from the headful tests. We cannot use the window size override in the screenshot tests, as that is already done in the headful tests. 

As for the CreateNewBrowserSession with correct window size: I'm not sure how I would do that without extracting all logic from the IntegrationTest baseclass

Since this only matters for a single test, IMO it makes sense to just add it the test itself.

MichaelKetting: It's using always the same page in this testfixture, why does the size only matter for one test?

Frazerus: Because the site was too small including the test output down below, which added the scrollbar. We do not care about seeing the content of that output, so it can stay hidden.

Frazerus:
See comments for update: https://re-motion.atlassian.net/browse/RM-9074

Remotion/Web/Development.WebTesting.TestSite.Shared/ScreenshotBorderTest.aspx
```diff
@@ -0,0 +1,30 @@
...
+<asp:Content ContentPlaceHolderID="body" runat="server">
+    <ContentTemplate>
+      <%-- General tests --%>
+      <div id="screenshotTarget" style="">
+      </div>
+        
+      <div id="borderElementTop" style="position: absolute; width: 1px; height: 1px; top: 0; left: 50px; background-color: black"></div>
+      <div id="borderElementRight" style="position: absolute; width: 1px; height: 1px; top: 50px; right: 0; background-color: black"></div>
+      <div id="borderElementLeft" style="position: absolute; width: 1px; height: 1px; top: 50px; left: 0; background-color: black"></div>
+      <div id="borderElementBottom" style="position: absolute; width: 1px; height: 1px; bottom: 0; right: 50px; background-color: black"></div>
```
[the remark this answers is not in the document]
Frazerus: Works with absolute, changed to % because it makes more sense

Remotion/Web/Development.WebTesting.IntegrationTests.RequireUI/SavedTestScreenshots/ScreenshotTest/ResolveBorderElementBottom.png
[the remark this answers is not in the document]
Frazerus: I would hope that browsers do not add shading over webpages for the common usecase of just using the browser normally. (chrome already did this with the "Chrome is being automated by ....") I think we should only take care of it when it's needed.

a-ctor (approved):

PR #1035, opened 2024-02-22
RM-9064 Check mousehelper issue with thread sleep

Frazerus:
https://re-motion.atlassian.net/browse/RM-9064

The tests seem to be stable without the Thread.Sleep(...), can't really investigate if the issues doesn't occur. 

We might just want to remove the thread.sleeps for now, until issues occur, then we can properly look into this.

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests.RequireUI/MouseTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests.RequireUI/MouseTest.cs
@@ -62,7 +62,6 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
       var clickElement = home.Scope.FindId(c_clickDivID);
       Helper.BrowserConfiguration.MouseHelper.Hover(clickElement);
       Helper.BrowserConfiguration.MouseHelper.LeftClick();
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests.RequireUI/MouseTest.cs
```diff
@@ -138,7 +130,7 @@ public void Tooltip ()
       // If the test is run behind all the other tests the tooltip will not show.
       // Mouse movement beforehand with a little sleep does the trick
       Helper.BrowserConfiguration.MouseHelper.Hover(tooltipElement);
-      Thread.Sleep(500);
+      //Thread.Sleep(500);
```
a-ctor, line 133: Instead of a comment we should probably remove this line as well.

a-ctor (approved):
Apart from queuing the tests a bunch of times we can't really do anything more. I saw that you queued the builds a few times and they worked so I am guessing it is pretty stable. We can't really do much more than than. If you want we can queue a few more builds over the weekend to be safe but otherwise we'll see what happens in the future.

MichaelKetting (approved):

PR #1039, opened 2024-02-27
RM-9074 remove scrollbar on ResolveBorderElementB

Frazerus:
https://re-motion.atlassian.net/browse/RM-9074

Redo of previous, failed to also run tests on firefox

I hope I'm not going insane here.


MichaelKetting (approved):
Weird observation: I got all files as not-viewed-before. No idea why.

PR #1050, opened 2024-03-12
RM-9122  Refactor PersistenceManager into a service

a-ctor:
https://re-motion.atlassian.net/browse/RM-9122

### DIFF
--- a/Remotion/Data/DomainObjects/Persistence/IPersistenceService.cs
+++ b/Remotion/Data/DomainObjects/Persistence/IPersistenceService.cs
@@ -22,7 +22,8 @@ using Remotion.Data.DomainObjects.Mapping;
 
 namespace Remotion.Data.DomainObjects.Persistence;
 
-public interface IPersistenceManager
+/// <threadsafety static="true" instance="false" />
+public interface IPersistenceService
 {
   ObjectID CreateNewObjectID (StorageProviderManager storageProviderManager, ClassDefinition classDefinition);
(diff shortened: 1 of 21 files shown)


Remotion/Data/DomainObjects/Persistence/IPersistenceService.cs
```diff
@@ -0,0 +1,38 @@
...
+//
+using System;
+using System.Collections.Generic;
+using Remotion.Data.DomainObjects.DataManagement;
+using Remotion.Data.DomainObjects.DataManagement.RelationEndPoints;
+using Remotion.Data.DomainObjects.Mapping;
+
+namespace Remotion.Data.DomainObjects.Persistence;
+
+public interface IPersistenceService
```
MichaelKetting, line 25: We forgot: please add threadsafety xmldoc here and to the class itself.

offcially, we should also have a type summary, buuuuuut....

MichaelKetting (approved):
I noticed we missed the xml doc on threadsafety.

Remotion/Data/DomainObjects/Persistence/IPersistenceService.cs
a-ctor: What exactly should I add? The class does not look to be thread safe so should I add `/// <threadsafety static="false" instance="false" />`?

MichaelKetting (approved):

PR #1051, opened 2024-03-12
RM-8350 Introduce InterfaceDefinition and update ClassDefinition accordingly

a-ctor:
https://re-motion.atlassian.net/browse/RM-8350


MichaelKetting (commented):
partial review, found some parameter namings in the first couple of files. don't know how many more there are that should be optimized to add the relevant semantics

a-ctor:
Stale. We can reopen in the future.

PR #1052, opened 2024-03-12
RM-9137 Space key should not trigger command in ListMenu

Frazerus:
https://re-motion.atlassian.net/browse/RM-9137

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/ListMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/ListMenu.ts
@@ -221,7 +221,6 @@ class ListMenu
         // exit tab strip
         return;
       case 13: //enter
-      case 32: //space
         {
           event.preventDefault();
           event.stopPropagation();


Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -221,7 +221,6 @@ class ListMenu
         // exit tab strip
         return;
       case 13: //enter
```
MichaelKetting, line 223: note, this is a breaking change ticket, too.

MichaelKetting (changes requested):
Right now, it works that when using Tab to focus the menu, SPACE isn't triggering the menu item. However, JAWS still triggers the menu item when using SPACE. This means we can't use it toggle Manual Forms Mode / Focus Mode. We need to figure out if there is something we can do or if this ticket turns out to be a won't fix because it's not helping with JAWS.

Request-Changes is set since this is blocked until we know more.

MichaelKetting:
> Right now, it works that when using Tab to focus the menu, SPACE isn't triggering the menu item. However, JAWS still triggers the menu item when using SPACE. This means we can't use it toggle Manual Forms Mode / Focus Mode. We need to figure out if there is something we can do or if this ticket turns out to be a won't fix because it's not helping with JAWS.
> 
> Request-Changes is set since this is blocked until we know more.

Added extra details in the ticket on the SPACE/ANCHOR behavior.

PR #1056, opened 2024-03-15
Feature/rm 9152 change tree view to improve support for wcag

Frazerus:
https://re-motion.atlassian.net/browse/RM-9152

Problems that are still existent:
* Cannot enter the tree with tab, seems like some tabindex=0 went missing along the way
* Default highlighting doesn't work properly (probably lost focus somewhere)
* We can go through the whole tree while in browse mode using NVDA, whereas in the WCAG example using NVDA, the tree is skipped.

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/TreeView.ts
+++ b/Remotion/Web/ClientScript/Scripts/TreeView.ts
@@ -81,14 +81,14 @@ class WebTreeView
       }
     }
 
-    const focusableTreeNode = treeView.querySelector ('li[tabindex="0"][role=treeitem]');
+    const focusableTreeNode = treeView.querySelector ('a[tabindex="0"][role=treeitem]');
     if (focusableTreeNode === null)
     {
-      const firstSelectedTreeNode = treeView.querySelector ('li[aria-selected=true]');
+      const firstSelectedTreeNode = treeView.querySelector ('a[aria-selected=true]');
       if (firstSelectedTreeNode === null)
       {
         const firstTopLevelTreeNodeList = treeView.querySelector ('ul[role=tree]');
-        const firstTopLevelTreeNode = firstTopLevelTreeNodeList?.querySelector (':scope > li[role=treeitem]');
+        const firstTopLevelTreeNode = firstTopLevelTreeNodeList?.querySelector (':scope > li > span.treeViewNode a[role=treeitem]');
         if (firstTopLevelTreeNode)
         {
           firstTopLevelTreeNode.setAttribute ('tabindex', '0');
@@ -103,14 +103,14 @@ class WebTreeView
 
   private static OnKeyDown (event: KeyboardEvent, treeView: HTMLElement): void
   {
-    const treeNodes = Array.from (treeView.querySelectorAll<HTMLElement> ('li[role=treeitem]'));
+    const treeNodes = Array.from (treeView.querySelectorAll<HTMLElement> ('a[role=treeitem]'));
 
     let activeTreeNodeIndex = -1;
 
     let activeTreeNode = document.activeElement as HTMLElement;
     if (activeTreeNode != null &&
       TypeUtility.IsDefined (activeTreeNode.tagName) &&
-      activeTreeNode.tagName.toUpperCase() === 'LI')
+      activeTreeNode.tagName.toUpperCase() === 'A')
     {
       activeTreeNodeIndex = treeNodes.indexOf (activeTreeNode);
     }
@@ -155,13 +155,14 @@ class WebTreeView
           let hasChildren = activeTreeNode.getAttribute ('aria-expanded') === 'true';
           if (!hasChildren)
           {
-            let parent = activeTreeNode.parentElement;
+            // Selects the <li> that contains this activeTreeNode <a>'s parent <li>
+            let parent = activeTreeNode.closest<HTMLElement>('li')!.parentElement;
             while (parent && !parent.matches ('ul[role=tree]'))
             {
-              if (parent.tagName.toLowerCase() === 'li')
+              if (parent.matches('li'))
               {
-                let newTreeNode = parent;
-                WebTreeView.SetFocus (newTreeNode, activeTreeNode);
+                let newlyFocusedTreeNode = parent.querySelector<HTMLElement>('a[role=treeitem]')!;
+                WebTreeView.SetFocus (newlyFocusedTreeNode, activeTreeNode);
                 break;
               }
 
@@ -187,8 +188,11 @@ class WebTreeView
           let hasChildren = activeTreeNode.getAttribute ('aria-expanded') === 'true';
           if (hasChildren)
           {
-            const newTreeNode = activeTreeNode.querySelector<HTMLElement> (':scope > ul > li')!;
-            WebTreeView.SetFocus (newTreeNode, activeTreeNode);
+            const activeTreeNodeListItem = activeTreeNode.closest<HTMLElement>('li')!;
+            const newTreeNodeListItem = activeTreeNodeListItem.querySelector<HTMLElement> (':scope > ul > li:first-child')!;
+
+            const newNode =  newTreeNodeListItem.querySelector<HTMLElement>('a[role=treeitem]')!;
+            WebTreeView.SetFocus (newNode, activeTreeNode);
           }
           else
           {
@@ -269,8 +273,13 @@ class WebTreeView
 
   private static ToggleExpander (treeNode: HTMLElement): void
   {
-    const expanderSpan = treeNode.querySelector<HTMLElement> ('span')!;
-    const expanderAnchor = expanderSpan.querySelector<HTMLElement> ('a')!;
+    const expanderSpan = treeNode.parentElement!.parentElement!;
+    if (expanderSpan.getAttribute('class') !== 'treeViewNode')
+    {
+      return;
+    }
+
+    const expanderAnchor = expanderSpan.querySelector<HTMLElement> ('a[aria-hidden=true]')!;
     expanderAnchor.click();
   };
 
@@ -278,8 +287,6 @@ class WebTreeView
   {
     treeNode.setAttribute ('tabindex', '0');
     treeNode.focus();
-    let headSpan = treeNode.querySelector<HTMLElement> ('span')!;
-    let headAnchor = headSpan.querySelector<HTMLElement> ('a[id^=Head_]')!;
-    headAnchor.click();
+    treeNode.click();
   };
 }
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
@@ -48,7 +48,7 @@ ul.treeViewRoot ul.treeViewTopLevelNodeChildren
 }
 
 ul.treeViewRoot span.treeViewNodeHeadSelected a:focus,
-ul.treeViewRoot li[role=treeitem]
+ul.treeViewRoot a[role=treeitem]
 {
   outline: none;
 }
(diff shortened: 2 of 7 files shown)


Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -861,13 +870,19 @@ private void RenderNodeChildren (HtmlTextWriter writer, WebTreeNode node, bool i
...
       writer.RenderBeginTag(HtmlTextWriterTag.Ul); // Begin child nodes
 
       RenderNodes(writer, node.Children, false, nodeIDAlgorithm, nestingDepth + 1);
 
       writer.RenderEndTag(); // End child nodes
     }
 
+    private string CreateSubTreeNodeID (string nodeID)
+    {
+      return $"Node_{nodeID}_SubTree";
```
a-ctor, line 883: We shouldn't add anything to the ID in the front. IDs in ASP.NET are hierarchical and container all the IDs of parent elements from top to bottom. e.g. `Body_MyContainer_MyList_MyItem`

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -103,14 +103,15 @@ class WebTreeView
 
   private static OnKeyDown (event: KeyboardEvent, treeView: HTMLElement): void
   {
-    const treeNodes = Array.from (treeView.querySelectorAll<HTMLElement> ('li[role=treeitem]'));
+    const treeNodes = Array.from (treeView.querySelectorAll<HTMLElement> ('a[role=treeitem]'));
+    const parentTreeNodeListItems = Array.from(treeView.querySelectorAll<HTMLElement>('li'));
```
a-ctor, line 107: I don't like that we do two separate queries and assume that their elements relate to each other in the correct order. Since `li` is always a parent of `a[role=treeitem]` we should transform the first list into the second using `.closest("li")`.

Maybe it makes sense to put them into an object to group them so that we don't have to maintain two separate but related arrays. We would need to use the `.findIndex` instead of `indexOf` below instead.

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -187,8 +188,13 @@ class WebTreeView
           let hasChildren = activeTreeNode.getAttribute ('aria-expanded') === 'true';
           if (hasChildren)
           {
-            const newTreeNode = activeTreeNode.querySelector<HTMLElement> (':scope > ul > li')!;
-            WebTreeView.SetFocus (newTreeNode, activeTreeNode);
+            const nodeIndex = treeNodes.indexOf(activeTreeNode);
+            const liNode = parentTreeNodeListItems[nodeIndex];
+            const newListItemChildNode = liNode.querySelector<HTMLElement> (':scope > ul > li')!;
```
a-ctor, line 193: In theory, this CSS selector could return any of the `li` elements if there are many. To make it more apparent that we want the first I would add the `:first-child` selector as well.

a-ctor (commented):
Regarding the functionality it is best to check the behavior with @MichaelKetting to ensure we don't do too many rounds

Remotion/Web/Core/UI/Controls/WebTreeView.cs
Frazerus: There was nothing added to the front in comparison to before, do you mean to say that it should be `Node_SubTree_0000-.....-0000`?

MichaelKetting:
@Frazerus right now, I'm going with nothing-yet-for-me-to-see. Please advise if that's not correct :)
cc @a-ctor

Frazerus:
Should now behave exactly as it did before.

Remotion/Web/ClientScript/Scripts/TreeView.ts
Frazerus: I dont think we need to add them to an object. They are const, so we dont really have to maintain them. It might be overkill having them both to begin with instead of just querying for the required `htmlElement`

Frazerus:
> @Frazerus right now, I'm going with nothing-yet-for-me-to-see. Please advise if that's not correct :) cc @a-ctor

Everything should work now, so you can have a look now :)

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -103,14 +103,15 @@ class WebTreeView
 
   private static OnKeyDown (event: KeyboardEvent, treeView: HTMLElement): void
   {
-    const treeNodes = Array.from (treeView.querySelectorAll<HTMLElement> ('li[role=treeitem]'));
+    const treeNodes = Array.from (treeView.querySelectorAll<HTMLElement> ('a[role=treeitem]'));
+    const parentTreeNodeListItems = Array.from(treeNodes.map<HTMLElement>(a => a.closest('li')!));
```
MichaelKetting, line 107: naming: this sounds like it deals with the parent-TreeNode, not with the containing-LI of this treenode.
-> `treeNodeParentListItems`

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -155,13 +156,14 @@ class WebTreeView
...
             {
-              if (parent.tagName.toLowerCase() === 'li')
+              if (parent.matches('li:has(a[role=treeitem])'))
               {
-                let newTreeNode = parent;
-                WebTreeView.SetFocus (newTreeNode, activeTreeNode);
+                let nodeIndex = parentTreeNodeListItems.indexOf(parent);
+                WebTreeView.SetFocus (treeNodes[nodeIndex], activeTreeNode);
                 break;
               }
```
MichaelKetting, line 169: I'm trying to think through the updated logic and I'm not sure if it couldn't be simplified:
1. take activeTreeNode and set as loop-variable (`parent`)
2. walk up the parent until you hit an `UL` with `role=tree`
3. when you find a `LI`, stop and select the treenode-Anchor inside.

matching `LI` with nested a with a role=treeitem seems overkill.
I also think we can keep the original stucture and simply change the way the `newTreeNode` is picked. Here, if you go with the array or do a fresh selection is simply a performance question and it could very well be that selecting the `A`-child is cheaper than the hashtable lookup. Or at least, won't be felt and we can keep it more straight forward (that's TBD)

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -187,8 +189,13 @@ class WebTreeView
...
           {
-            const newTreeNode = activeTreeNode.querySelector<HTMLElement> (':scope > ul > li')!;
-            WebTreeView.SetFocus (newTreeNode, activeTreeNode);
+            const nodeIndex = treeNodes.indexOf(activeTreeNode);
+            const liNode = parentTreeNodeListItems[nodeIndex];
+            const newListItemChildNode = liNode.querySelector<HTMLElement> (':scope > ul > li:first-child')!;
+
+            const childNodeIndex = parentTreeNodeListItems.indexOf(newListItemChildNode);
+            const newNode = treeNodes[childNodeIndex];
+            WebTreeView.SetFocus (newNode, activeTreeNode);
```
MichaelKetting, line 198: naming: 
nodeIndex -> activeTreeNodeIndex
liNode -> activeTreeNodeListItem
newListItemChildNode -> newTreeNodeListItem

logic: except for the fact that we already computed the relationship between the LI and the Anchor tag, there's no reason for us to not simply expand n the CSS-selector and use it to pick the correct anchor.

Remotion/Web/ClientScript/Scripts/TreeView.ts
MichaelKetting, line 107: I'm not sure we need the LI-mappings this way instead of just selecting them as we need them via parent and child navigation. Right now, it feels overly complex and might not give us a true performance benefit that offsets the complexity.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
MichaelKetting, line 1: The updated markup does not announce the badge-information when navigating the tree. The reason is probably that before, the badge was nested inside the role=treenode, now it is not.

MichaelKetting (changes requested):
The

Remotion/Web/Core/UI/Controls/WebTreeView.cs
MichaelKetting, line 1: The updated markup does not announce the badge-information when navigating the tree. The reason is probably that before, the badge was nested inside the role=treenode, now it is not.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -779,9 +770,14 @@ private void RenderNodes (HtmlTextWriter writer, WebTreeNodeCollection webTreeNo
...
     {
       if (node.IsSelected)
+      {
         writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassNodeHeadSelected);
+      }
       else
+      {
         writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassNodeHead);
+      }
+
```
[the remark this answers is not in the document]
Frazerus: Changed something here earlier, while reverting I it seems I didn't change it back

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -166,17 +166,17 @@ ul.treeViewRoot span.treeViewNode span.treeViewNodeBadge
   align-self: start;
 }
 
-ul.treeViewRoot li:focus > ul > li:first-of-type span.treeViewNode:not(:hover) span.treeViewNodeBadge,
-ul.treeViewRoot li:focus + li span.treeViewNode:not(:hover) span.treeViewNodeBadge
+ul.treeViewRoot li:has(a:focus) > ul > li:first-of-type span.treeViewNode:not(:hover) span.treeViewNodeBadge,
+ul.treeViewRoot li:has(a:focus) + li span.treeViewNode:not(:hover) span.treeViewNodeBadge
```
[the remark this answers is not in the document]
Frazerus: the LI's aren't being focused here, the a's which have the li as a parent element are the ones we select for.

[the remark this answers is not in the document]
Frazerus: This whole thing is also getting updated in a follow up, because the box shadow did not work properly for some time now.

Remotion/Web/ClientScript/Scripts/TreeView.ts
Frazerus: we can't take activeTreeNode and set as a parent, as this is still contained in a different LI which would be matched.

Matching with the nested a + role is probably overkill.

Remotion/Web/ClientScript/Scripts/TreeView.ts
Frazerus: Yeah, I agree.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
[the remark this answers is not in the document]
MichaelKetting: The change-back is missing

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -659,29 +659,20 @@ private void RenderNodes (HtmlTextWriter writer, WebTreeNodeCollection webTreeNo
...
-        if (node.IsSelected)
-        {
-          writer.AddAttribute(HtmlTextWriterAttribute2.AriaSelected, HtmlAriaSelectedAttributeValue.True);
-        }
-        writer.AddAttribute("tabindex", _focusededNode == node ? "0" : "-1");
+
         if (EnableTopLevelGrouping && isTopLevel && !isLastNode && nodes[i].Category != nodes[i + 1].Category)
           writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassNodeCategorySeparator);
 
+        writer.AddAttribute(HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.None);
```
MichaelKetting, line 666: it seems this change/commit should be part of `RM-9152 change tree view to improve support for wcag` because it always had a role. it just got lost inside this feature branch and then re-added as `none` but in a separate comment.

MichaelKetting (changes requested):
syntax tweaks and one bug :)

Remotion/Web/Core/UI/Controls/WebTreeView.cs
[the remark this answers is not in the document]
Frazerus: Just so you know, I AM ashamed when that happens, it's not intentional, I just can't keep my head straight sometimes.

MichaelKetting (approved):

PR #1060, opened 2024-03-19
RM-9136 Use role=tooltip in ListMenu

a-ctor:
https://re-motion.atlassian.net/browse/RM-9136


MichaelKetting (approved):
@a-ctor FYI I replaced anchor-role with button-role in ListMenu. Otherwise, the disabled items do not get announced in the total count.

PR #1061, opened 2024-03-19
RM-5477 Autocomplete scrolling

a-ctor:
https://re-motion.atlassian.net/browse/RM-5477

 - [x] Check if the "loading"/"loaded" announcements are okay and if their behavior is okay this way
 - [x] Add localization for the different announcements
 - [x] Determine how to do pagination in `SecurityManagerAutoCompleteReferenceValueWebService`

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
@@ -90,8 +90,8 @@ namespace Remotion.BocAutoCompleteReferenceValue
         dropDownButtonID: string;
         dataType: string;
         selectFirst(inputValue: string, searchTerm: Nullable<string>): boolean;
-        formatItem(item: Item, index: number, length: number, value: string, term: string): FormatResult;
-        formatMatch(item: Item, index: number, length: number): string | false;
+        formatItem(item: Item, value: string, term: string): FormatResult;
+        formatMatch(item: Item): string | false;
         formatResult?(item: Item): string;
         autoFill: boolean;
         highlight(value: string, term: string): string;
@@ -1072,7 +1072,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
             for (let i = 0, ol = this.options.data.length; i < ol; i++) {
                 let rawValue = this.options.data[i];
 
-                const value = this.options.formatMatch(rawValue, i + 1, this.options.data.length);
+                const value = this.options.formatMatch(rawValue);
                 if (value === false)
                     continue;
 
@@ -1365,7 +1365,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
             for (let i = 0; i < max; i++) {
                 if (!this.data[i])
                     continue;
-                const item = this.options.formatItem(this.data[i].data, i + 1, max, this.data[i].value, this.term);
+                const item = this.options.formatItem(this.data[i].data, this.data[i].value, this.term);
                 const termElement = document.createElement("div");
                 termElement.innerText = this.term;
                 const termAsHtml = termElement.innerHTML;
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
@@ -42,6 +42,13 @@ namespace Remotion.BocAutoCompleteReferenceValue
     {
         Type: 'ValueList';
         Values: Remotion.BocAutoCompleteReferenceValue.Item[];
+        HasMoreSearchResults: boolean;
+    }
+
+    export interface BocAutoCompleteReferenceValueCacheableSearchResult
+    {
+        cacheRow: CacheRow;
+        hasMoreSearchResults: boolean;
     }
 
     export type Item = {
@@ -97,7 +104,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
         highlight(value: string, term: string): string;
         scroll: boolean;
         repositionInterval: number;
-        parse(data: BocAutoCompleteReferenceValueSearchResult): Remotion.BocAutoCompleteReferenceValue.CacheRow;
+        parse(data: BocAutoCompleteReferenceValueSearchResult): BocAutoCompleteReferenceValueCacheableSearchResult;
         handleRequestError(err: Sys.Net.WebServiceError): void;
         clearRequestError(err?: Sys.Net.WebServiceError): void;
         serviceUrl: string;
@@ -795,7 +802,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
                 if (dropDownTriggered)
                     searchString = options.searchStringValidationParams.getDropDownSearchString(searchString);
 
-                const successHandler = function (q: string, data: Nullable<CacheRow>) {
+                const successHandler = function (q: string, data: BocAutoCompleteReferenceValueCacheableSearchResult) {
                     stopLoading();
                     receiveData (q, data);
                     if (!select.visible() && options.noDataFoundMessage) {
@@ -811,7 +818,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
                     closeDropDownListAndSetValue(state.previousValue, false);
                 };
 
-                requestData(searchString, successHandler, failureHandler);
+                requestData(searchString, 0, successHandler, failureHandler);
             } else {
                 stopLoading();
                 select.hide();
@@ -886,12 +893,12 @@ namespace Remotion.BocAutoCompleteReferenceValue
             state.lastKeyPressValue = null;
         };
 
-        function receiveData(q: string, data: Nullable<CacheRow>): void {
+        function receiveData(q: string, data: BocAutoCompleteReferenceValueCacheableSearchResult): void {
             informationPopUp.hide();
             if (data && hasFocus) {
                 select.display(data, q);
-                if (data.length) {
-                    autoFill(q, data[0].result);
+                if (data.cacheRow.length) {
+                    autoFill(q, data.cacheRow[0].result);
                     select.show();
                     if (options.selectFirst (input.value, q)) {
                         select.selectItem (0);
@@ -904,7 +911,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
             }
         };
 
-        function requestData(term: string, success: (term: string, result: Nullable<CacheRow>) => void, failure: (term: string) => void) {
+        function requestData(term: string, offset: number, success: (term: string, result: BocAutoCompleteReferenceValueCacheableSearchResult) => void, failure: (term: string) => void) {
             // re-motion: cancel an already running request
             abortRequest();
 
@@ -917,10 +924,10 @@ namespace Remotion.BocAutoCompleteReferenceValue
 
             options.clearRequestError();
 
-            const data = cache.load(term);
-            // recieve the cached data
-            if (data && data.length) {
-                success(term, data);
+            const cachedSearchResult = cache.load(term, offset);
+            // receive the cached data
+            if (cachedSearchResult && cachedSearchResult.cacheRow.length) {
+                success(term, cachedSearchResult);
 
                 // re-motion: if a webservice url and a method name have been supplied, try loading the data now
             } else if ((typeof options.serviceUrl == "string") && (options.serviceUrl.length > 0)
@@ -933,6 +940,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
                 //           under "JSON, objects, and strings: oh my!" for details.
                 const params: Dictionary<unknown> = {
                     searchString: term,
+                    completionSetOffset: offset,
                     completionSetCount: options.max
                 };
                 for (const propertyName in options.extraParams)
@@ -942,7 +950,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
                                             function(result: BocAutoCompleteReferenceValueSearchResult) {
                                                 executingRequest = null;
                                                 const parsed = options.parse(result);
-                                                cache.add(term, parsed);
+                                                cache.add(term, offset, parsed);
                                                 success(term, parsed);
                                             } as any,
                                             function(err: Sys.Net.WebServiceError) {
@@ -987,9 +995,9 @@ namespace Remotion.BocAutoCompleteReferenceValue
                                                         executingRequest = null;
                                                         let parsed: Nullable<CacheRowEntry> = null;
                                                         if (result != null) {
-                                                            const resultArray: BocAutoCompleteReferenceValueSearchResultWithValueList = { Type: "ValueList", Values: [result] };
+                                                            const resultArray: BocAutoCompleteReferenceValueSearchResultWithValueList = { Type: "ValueList", Values: [result], HasMoreSearchResults: false };
                                                             const parsedArray = options.parse(resultArray);
-                                                            parsed = parsedArray[0];
+                                                            parsed = parsedArray.cacheRow![0];
                                                         }
                                                         success(term, parsed);
                                                     } as any,
@@ -1037,25 +1045,37 @@ namespace Remotion.BocAutoCompleteReferenceValue
 
     export class Cache {
         private readonly options: Options;
-        private readonly data = new Map<string, CacheRow>();
+        private readonly data = new Map<string, BocAutoCompleteReferenceValueCacheableSearchResult>();
 
         constructor(options: Options) {
             this.options = options;
         }
 
-        public add(q: string, value: CacheRow): void {
+        public add(q: string, offset: number, value: BocAutoCompleteReferenceValueCacheableSearchResult): void {
             if (this.data.size > this.options.cacheLength) {
                 this.flush();
             }
 
-            this.data.set(q, value);
+            if (offset === 0) {
+                this.data.set(q, value);
+            } else {
+                const existingData = this.data.get(q);
+                if (existingData && existingData.cacheRow.length === offset) {
+                    existingData.cacheRow.push(...value.cacheRow);
+                    existingData.hasMoreSearchResults = value.hasMoreSearchResults;
+                }
+                // ignore offset data that does not append to an existing cache line
+            }
         }
 
         public flush(): void {
             this.data.clear();
         }
 
-        public load(q: string): Nullable<CacheRow> {
+        public load(q: string, offset: number): Nullable<BocAutoCompleteReferenceValueCacheableSearchResult> {
+            if (offset !== 0)
+                return null;
+
             return this.data.get(q) || null;
         }
     }
@@ -1372,9 +1392,9 @@ namespace Remotion.BocAutoCompleteReferenceValue
             return -1;
         }
 
-        public display(d: CacheRow, q: string): void {
+        public display(d: BocAutoCompleteReferenceValueCacheableSearchResult, q: string): void {
             this.init();
-            this.data = d;
+            this.data = d.cacheRow;
             this.term = q;
             this.fillList();
         }
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
@@ -103,7 +103,6 @@ namespace Remotion.BocAutoCompleteReferenceValue
         serviceUrl: string;
         serviceMethodSearch: string;
         serviceMethodSearchExact: string;
-        data: Nullable<Item[]>;
     };
 
     // This type contains the properties that are required to be passed into the .autocomplete() function
@@ -234,7 +233,6 @@ namespace Remotion.BocAutoCompleteReferenceValue
             serviceUrl: serviceUrl,
             serviceMethodSearch: serviceMethodSearch,
             serviceMethodSearchExact: serviceMethodSearchExact,
-            data: null,
         }, initialOptions as RequiredOptions);
 
         // if highlight is set to false, replace it with a do-nothing function
@@ -595,10 +593,7 @@ namespace Remotion.BocAutoCompleteReferenceValue
             },
             setOptions: function(newOptions: Options): AutoCompletePublicApi {
                 Object.assign(options, newOptions);
-                // if we've updated the data, repopulate
-                if ("data" in newOptions)
-                    cache.populate();
-                
+
                 return this;
             },
             showInformationPopUp: function (message: string): AutoCompletePublicApi {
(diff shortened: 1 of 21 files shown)


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
MichaelKetting, line 1: Announcements:
When searching for "A" in Edge/JAWS, the loading-message is announced before the loaded-message on the first go-around. I can reproduce this after reloading the page. We should make the loaded-announcement impolite.

Note on repro: It works "correctly" when I use longer lists, i.e. with "s" the first call already returns 21 items instead of 11. But for "a" or "n", I only get 11 items on the first call and then the expansion of the result set flips the message order. 

If you can figure out the underlying reason, we can also fix that problem instead of the politeness.

MichaelKetting, line 1: Announcements:
Is there an easy way to re-announce the lst item in the list after we are done loading and the element is still the active one. That way, we would get the new count immediately. Might be that it doesn't work well or is way too complicated. So, just an idea. Might also be a future improvement if doable but extra work.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocAutoCompleteReferenceValue.UI.css
```diff
@@ -98,6 +98,14 @@
   background-repeat: no-repeat;
 }
 
+.ac_placeholder_loading > div:first-child
+{
+  background-image: url('../Image/sprite.svg#AutoCompleteReferenceValueRequestInProgressIndicator');
```
MichaelKetting, line 103: do you think we should de-duplciate the url into a css variable local to the commen ancestor of ac_loading and ac_placeholder_loading?

Remotion/ObjectBinding/Web/Services/BocAutoCompleteReferenceValueSearchResultWithValueList.cs
MichaelKetting, line 1: one day, this could be a `record` instead of a `class` :)

Remotion/ObjectBinding/Web/Services/IBocAutoCompleteReferenceValueWebService.cs
```diff
@@ -29,13 +29,16 @@ public interface IBocAutoCompleteReferenceValueWebService : IBocReferenceValueBa
     /// Retrieves a list of objects based on the <paramref name="searchString"/> and the search context (i.e. the <paramref name="businessObjectClass"/> etc).
     /// </summary>
     /// <param name="searchString">The <see cref="string"/> all returned values must match.</param>
+    /// <param name="completionSetOffset">
+    ///   The number of items to skip before returning results or <see langword="null" /> if no offset should be used and results should be returned from the beginning.
+    /// </param>
```
MichaelKetting, line 34:
```suggestion
    /// <param name="completionSetOffset">
    ///   The number of items to skip before returning results or zero if no offset should be used and results should be returned from the beginning.
    /// </param>
```
it's a non-nullable int.

SecurityManager/Clients.Web/UI/SecurityManagerAutoCompleteReferenceValueWebService.asmx.cs
```diff
@@ -94,6 +95,7 @@ public static void BindServiceToControl (BocAutoCompleteReferenceValue control)
 
       var businessObjectClassWithIdentity = GetBusinessObjectClassWithIdentity(businessObjectClass);
       var referenceProperty = GetReferenceProperty(businessObjectProperty, businessObjectClassWithIdentity);
+      // TODO how to add pagination here correctly
```
MichaelKetting, line 98: our SecurityManager Service is going to ignore this feature, at least for now. 

please make a follow-up ticket to properly integrate with Linq-skip/take clauses.

SecurityManager/Clients.Web/UI/SecurityManagerAutoCompleteReferenceValueWebService.asmx.cs
```diff
@@ -103,7 +105,8 @@ public static void BindServiceToControl (BocAutoCompleteReferenceValue control)
         var referencingObject = businessObjectClassWithIdentity.GetObject(businessObject);
         var result = referenceProperty.SearchAvailableObjects(referencingObject, securityManagerSearchArguments);
         var resultArray = result.Cast<IBusinessObjectWithIdentity>().Select(o => new BusinessObjectWithIdentityProxy(o)).ToArray();
-        return BocAutoCompleteReferenceValueSearchResult.CreateForValueList(resultArray);
+        var hasMoreSearchResults = completionSetCount.HasValue && resultArray.Length >= completionSetCount.Value;
```
MichaelKetting, line 108: for now, it's false. should be "fixed" in the follow up ticket.

MichaelKetting (commented):
parital review, still need to check scripts and webtests (and test pages)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -783,6 +788,20 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
 
+        function requestMoreData(term: string, offset: number): Promise<BocAutoCompleteReferenceValueCacheableSearchResult> {
+            return new Promise((resolve, reject) => {
+                const successHandler = (term: string, result: BocAutoCompleteReferenceValueCacheableSearchResult) => {
+                    resolve(result);
+                };
+                const failureHandler = (error: string) => {
+                    select.hide();
+                    reject();
+                };
```
MichaelKetting, line 799: Is this triggering the error popup? (this is what would be the expected behavior as a user)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1174,6 +1152,16 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
 
+            this.politeAnnouncementElement = document.createElement("ul");
+            this.politeAnnouncementElement.ariaLive = "polite";
+            this.politeAnnouncementElement.classList.add("screenReaderText");
+            this.element.appendChild(this.politeAnnouncementElement);
+
+            this.assertiveAnnouncementElement = document.createElement("ul");
+            this.assertiveAnnouncementElement.ariaLive = "assertive";
+            this.assertiveAnnouncementElement.classList.add("screenReaderText");
+            this.element.appendChild(this.assertiveAnnouncementElement);
```
MichaelKetting, line 1163: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22

I'm wondering if instead of an UL, we shouldn't use a simple div with aria-role=status and update the text content of this DIV. As I understand it, for what we need, a UL isn't the best fit.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1356,41 +1376,50 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
+
+            if (hasMoreData) {
+                const placeholder = document.createElement("li");
+                placeholder.role = "option";
+                placeholder.ariaSelected = "false";
+                placeholder.id = `${this.options.selectListID}_placeholder`;
+
+                placeholder.dataset["isAnnotation"] = "true";
+                placeholder.addEventListener("click", () => this.loadMoreDataIfAvailable());
+                placeholder.innerHTML = "<div></div><div><span>...</span></div>";
```
MichaelKetting, line 1409: This might bite us when we support more complicated layouts in the drop down.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/SearchServiceResultItem.cs
```diff
@@ -57,5 +57,42 @@ public string DisplayName
...
+
+    public static bool operator == (SearchServiceResultItem? left, SearchServiceResultItem? right)
+    {
+      return Equals(left, right);
+    }
+
+    public static bool operator != (SearchServiceResultItem? left, SearchServiceResultItem? right)
+    {
+      return !Equals(left, right);
+    }
```
MichaelKetting, line 96: statics are before isntance members.

MichaelKetting, line 96: I understand that you built this so you can do unit tests for the entire object. in webtests, I don't think anyone's building this kind of complex items. they would just test based on simpler subsets.

MichaelKetting (changes requested):
okay, checked the production code and got some remarks. 

test code is next.

Marked as request-changes without knowing if theres' actual changes needed or just queistons.

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocAutoCompleteReferenceValueControlObjectTest.cs
```diff
@@ -494,34 +495,21 @@ public void TestGetSearchServiceResults ()
 
       var bocAutoComplete = home.AutoCompletes().GetByLocalID("PartnerField_Normal");
 
-      var searchResults = bocAutoComplete.GetSearchServiceResults("D", 1);
+      var searchResults = bocAutoComplete.GetSearchServiceResults("D", 0, 1);
```
MichaelKetting, line 498: breakign change ticket for this, please.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
MichaelKetting, line 1: Announcement: add Resource-Entry for the loading and loaded strings.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
MichaelKetting: Yes, it is.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -162,17 +167,17 @@ class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes
         {
           const row = document.createElement('li');
 
+          const imgContainer = document.createElement('div');
+          row.append(imgContainer);
```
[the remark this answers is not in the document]
a-ctor: In this case, we create new elements that are not attached to the DOM yet (and never will because we only use it to generate HTML), so there are no updates to be made.

Even if they were attached to the DOM, shouldn't happen during JavaScript execution as the page is not rendered during execution. Thats why an infinite loop will freeze the page.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: I haven't tested `role=status` but I assume the reading behavior should mimic an attribute with the `aria-live="..."` set. The problem I faced was that text changes were not read correctly if the text changed to quickly. The list is a workaround to ensure that the message is announce again when a new message is added (even if that message is the same as the previous message).

a-ctor: Regarding best fit: Since you can't browse this element it doesn't really matter what we pick here. `role=status` is probably best if it is a visual element that you can browse so that you are informed that this element might change in the future. As we just want the message and you can't browse to the element anyways, I think it does not matter what element we use for the announcement.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/SearchServiceResultItem.cs
a-ctor: Even operators? I don't think they are a good fit above as they can't be called by name and always have operations related to the instance data, which fits better down below with the instance methods.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
a-ctor: The `loaded` message is already set to `assertive` that is why `loaded` can be read before `loading`. I tried to guard against this by removing the `loading` element when `loaded` is read, but that seems to not work and `loading` is still queued even if its element is removed.

It seems like there is no good way to mix and match `polite` and `assertive` to get a consistent result. I would suggest trying to go back to a single element that is read, choosing either `polite` or `assertive` and using it for both `loading` and `loaded` messages. That way, new announcements might override the old ones and only `loaded` is read. That would also solve the ordering problem.

a-ctor: As far as I know there is only a way to start reading the whole element (so the full list with `aria-atomic`) but there is not option to read just one element. You have to write your own message and put it in the `aria-live` region.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
MichaelKetting: Note for finishing the PR: we should use a DIV/SPAN instead of a UL, it just doesn't seem right. 
Note for finishing the PR: we already have a working example in `InformationPopUp` class that uses `aria-atomic` and `role=alert` and replaces the contents. adjust to this pattern to better symmetry of used patterns.

MichaelKetting: Note on best fit: alert vs status since it's not an error, status might really be correct, regardless of it being visible to seeing users.

Remotion/ObjectBinding/Web.Development.WebTesting/ControlObjects/SearchServiceResultItem.cs
MichaelKetting: See ObjectID: yes.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
MichaelKetting: Suggestion matches with the other comment. should do that.

MichaelKetting: We switch to aria-atomic and replace contents. See other comment.

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/IBocAutoCompleteReferenceValue.cs
```diff
@@ -36,6 +36,8 @@ public interface IBocAutoCompleteReferenceValue : IBocReferenceValueBase
     string? ValidSearchStringRegex { get; }
     string? ValidSearchStringForDropDownRegex { get; }
     string? SearchStringForDropDownDoesNotMatchRegexMessage { get; }
+    string? LoadingMoreMatchesMessage { get; }
+    string? LoadedMoreMatchesMessage { get; }
```
MichaelKetting, line 40: revert please. we don't need the option to configure the messages. the renderer loads them forn the resources, that's sufficient.
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocAutoCompleteReferenceValue.cs
```diff
@@ -632,6 +634,30 @@ public int SelectionUpdateDelay
...
+    /// </summary>
+    [Category("AutoComplete")]
+    [DefaultValue("")]
+    [Description("The message announced to screen reader users when new matches were loaded from the server and the result list has been expanded.")]
+    public string? LoadedMoreMatchesMessage
+    {
+      get { return _loadedMoreMatches; }
+      set { _loadedMoreMatches = StringUtility.EmptyToNull(value); }
+    }
+
```
MichaelKetting, line 660:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocAutoCompleteReferenceValue.cs
```diff
@@ -119,6 +119,8 @@ public enum ResourceIdentifier
     private string? _validSearchStringRegex;
     private string? _validSearchStringForDropDownRegex;
     private string? _searchStringForDropDownDoesNotMatchRegexMessage;
+    private string? _loadingMoreMatches;
+    private string? _loadedMoreMatches;
```
MichaelKetting, line 123:
```suggestion
```

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
```diff
@@ -270,6 +285,12 @@ private string GetResourcesAsJson (BocAutoCompleteReferenceValueRenderingContext
       jsonBuilder.Append(", ");
       jsonBuilder.Append("NoDataFoundMessage : ");
       AppendStringValueOrNullToScript(jsonBuilder, resourceManager.GetText(ResourceIdentifier.NoDataFoundMessage));
+      jsonBuilder.Append(", ");
+      jsonBuilder.Append("LoadingMoreMatchesMessage : ");
+      AppendStringValueOrNullToScript(jsonBuilder, loadingMoreMatchesMessage);
+      jsonBuilder.Append(", ");
+      jsonBuilder.Append("LoadedMoreMatchesMessage : ");
+      AppendStringValueOrNullToScript(jsonBuilder, loadedMoreMatchesMessage);
```
MichaelKetting, line 293:
```suggestion
      AppendStringValueOrNullToScript(jsonBuilder, resourceManager.GetText(ResourceIdentifier.LoadedMoreMatchesMessage));
```

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
```diff
@@ -270,6 +285,12 @@ private string GetResourcesAsJson (BocAutoCompleteReferenceValueRenderingContext
       jsonBuilder.Append(", ");
       jsonBuilder.Append("NoDataFoundMessage : ");
       AppendStringValueOrNullToScript(jsonBuilder, resourceManager.GetText(ResourceIdentifier.NoDataFoundMessage));
+      jsonBuilder.Append(", ");
+      jsonBuilder.Append("LoadingMoreMatchesMessage : ");
+      AppendStringValueOrNullToScript(jsonBuilder, loadingMoreMatchesMessage);
```
MichaelKetting, line 290:
```suggestion
      AppendStringValueOrNullToScript(jsonBuilder, resourceManager.GetText(ResourceIdentifier.LoadingMoreMatchesMessage));
```

Remotion/ObjectBinding/Web/UI/Controls/BocReferenceValueImplementation/Rendering/BocAutoCompleteReferenceValueRenderer.cs
```diff
@@ -261,6 +265,17 @@ private string GetResourcesAsJson (BocAutoCompleteReferenceValueRenderingContext
...
+          StringUtility.EmptyToNull(renderingContext.Control.LoadingMoreMatchesMessage)
+          ?? resourceManager
+              .GetText(ResourceIdentifier.LoadingMoreMatchesMessage)
+              .ToString(WebStringEncoding.HtmlWithTransformedLineBreaks);
+      var loadedMoreMatchesMessage =
+          StringUtility.EmptyToNull(renderingContext.Control.LoadedMoreMatchesMessage)
+          ?? resourceManager
+              .GetText(ResourceIdentifier.LoadedMoreMatchesMessage)
+              .ToString(WebStringEncoding.HtmlWithTransformedLineBreaks);
+
```
MichaelKetting, line 278:
```suggestion
```

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite.Shared/Controls/BocAutoCompleteReferenceValueWebService.asmx.cs
```diff
@@ -77,6 +78,17 @@ public IconProxy GetIcon (string businessObjectClass, string businessObject, str
       if (searchString == "throw")
         throw new InvalidOperationException("I'm always going to throw an exception if you search for 'throw'!");
 
+      if (searchString == "testlist")
+      {
+        var count = completionSetCount ?? 10;
+        var resultItems = Enumerable.Range(0, 20)
+            .Skip(completionSetOffset)
+            .Take(count)
+            .Select(e => new BusinessObjectWithIdentityProxy { DisplayName = $"Person {e}", UniqueIdentifier = e.ToString() })
```
MichaelKetting, line 87: Recommendation: BocAutoComplete works best when the items match the search string. So, prefix with "testlist "?

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.UI.ts
```diff
@@ -1588,6 +1653,33 @@ namespace Remotion.BocAutoCompleteReferenceValue
...
+        private clearAnnouncements() {
+            this.announcementElement.innerText = "";
+        }
+
+        private clearQueuedLoadingAnnouncement() {
+            if (this.queuedLoadingAnnouncementTimeout !== null) {
+                clearTimeout(this.queuedLoadingAnnouncementTimeout);
+                this.queuedLoadingAnnouncementTimeout = null;
+            }
+        }
```
MichaelKetting, line 1682: I would like to do a quick rundown of this. announcementelement changes form UL to simple text?
Loading-message delay?

MichaelKetting (commented):
Looks good, got some revert-requests. want to do a quick check together on some details tomorrow.

MichaelKetting (changes requested):
Looks good, got some revert-requests. want to do a quick check together on some details tomorrow.

MichaelKetting (approved):

PR #1062, opened 2024-03-20
RM-9114   Implement Log4NetLoggerProvider based on Log4NetLogManager

KuJo7:
https://re-motion.atlassian.net/browse/RM-9114

### DIFF
--- a/Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
+++ b/Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
@@ -1,0 +1,41 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using log4net.Core;
+using Microsoft.Extensions.Logging;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLogger
+    : Microsoft.Extensions.Logging.ILogger
+{
+  public void Log<TState> (Microsoft.Extensions.Logging.LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+  {
+    throw new NotImplementedException();
+  }
+
+  public bool IsEnabled (Microsoft.Extensions.Logging.LogLevel logLevel)
+  {
+    throw new NotImplementedException();
+  }
+
+  public IDisposable? BeginScope<TState> (TState state)
+      where TState : notnull
+  {
+    throw new NotImplementedException();
+  }
+}
--- a/Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
+++ b/Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
@@ -1,0 +1,42 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Reflection;
+using log4net.Appender;
+using log4net.Config;
+using log4net.Core;
+using log4net.Layout;
+using log4net.Repository.Hierarchy;
+using Microsoft.Extensions.Logging;
+using Remotion.Reflection;
+using Remotion.Utilities;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLoggerProvider : ILoggerProvider
+{
+  public Microsoft.Extensions.Logging.ILogger CreateLogger (string categoryName)
+  {
+    ArgumentUtility.CheckNotNull("categoryName", categoryName);
+
+    return new Log4NetLogger();
+  }
+
+  public void Dispose ()
+  {
+  }
+}
--- a/Remotion/Core/Core/Core.Core.csproj
+++ b/Remotion/Core/Core/Core.Core.csproj
@@ -20,6 +20,7 @@
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Remove="JetBrains.Annotations" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
     <PackageReference Include="Remotion.ThirdParty.Dependency.Jetbrains.Annotations" Version="2022.1.0" />
     <PackageReference Include="Remotion.ThirdPartyDependency.CommonServiceLocator" Version="2.0.6" />
   </ItemGroup>
--- a/Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
+++ b/Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
@@ -1,0 +1,37 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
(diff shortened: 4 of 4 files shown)


Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,60 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using log4net.Core;
+using Microsoft.Extensions.Logging;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLogger
+    : Microsoft.Extensions.Logging.ILogger
```
Frazerus, line 24: We should note that this lengthy thing can be cleaned up once log4net is kicked out. Please note in the right jira ticket :)

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+    XmlConfigurator.Configure();
+  }
+
+  public void InitializeConsole ()
+  {
+    var appender = CreateConsoleAppender();
+    BasicConfigurator.Configure(appender);
+  }
+
+  public void InitializeConsole (Microsoft.Extensions.Logging.LogLevel defaultThreshold, params LogThreshold[] logThresholds)
```
Frazerus, line 43: This might actually not be required anymore. @MichaelKetting? 
Only DependDB usecases are re-motion itself.

If useless now, we can remove the convert method in `Log4NetLogger` in this task.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+        throw new ArgumentException(message, "logThresholds");
+      }
+
+      log4netLogger.Level = Log4NetLog.Convert(logThreshold.Threshold);
+    }
+  }
+
+  public void Dispose ()
+  {
+    throw new NotImplementedException();
```
Frazerus, line 94: We do need to implement this ;)

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
Frazerus: This should happen to all useages where we need to more closely specify which one to use.

Frazerus (changes requested):

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
MichaelKetting: All three Initialize-Methods:

originally they where instance to allow interface implementations. This is now an obsolete concept. If an application developer wants to use log4net, they can use the underlying APIs so we can drop all the Initialize-methods on this type, leaving only the interface-declared methods as public api.

@Frazerus 
Please remind me, did I already mention that we also want to keep a copy of the original APIs in Remotion.Obsolete to look them up via DependDB? it mihgt be in one of the later tasks.

Frazerus: You did not mention that before concerning the initialize things. For (almost) everything else we should have the APIs still in the extension methods to provide backwards compatibility.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  public Microsoft.Extensions.Logging.ILogger CreateLogger (string categoryName)
```
Frazerus, line 97: We are missing tests here :)

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,60 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using log4net.Core;
+using Microsoft.Extensions.Logging;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLogger
```
Frazerus, line 23: This class does not require unittests yet, as this is just a stub implementation for now.

Frazerus (changes requested):
Short notes because I got feedback from Mike :)

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
MichaelKetting: @Frazerus this sounds about right. The reasoning is:
1. only stuff without a log4net dependency can get moved into an extension method in Remotion.Extensions.
2. those methods could indeed get moved to an extension method in Remtoion.Logging.Log4Net, class `Log4NetLoggerProviderExtensions` and marked as obsolete. I went with the default of deleting based on recommendation plus my own thought on "there's not good place to put them" in combination with almost no usage at all. If you feel it's better to move them as obsolete to extension method, I'm okay with that, too.

Frazerus (approved):

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+    : Microsoft.Extensions.Logging.ILogger
+{
+  public void Log<TState> (Microsoft.Extensions.Logging.LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+  {
+    throw new NotImplementedException();
```
MichaelKetting, line 28: we documented in the jira ticket "stays empty as needed" -> please replace the NotImplementedException with a NOP. In the unit tests, for now we test as `Throws.Nothing`.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public bool IsEnabled (Microsoft.Extensions.Logging.LogLevel logLevel)
```
MichaelKetting, line 31: returns `false`, unit tests assert this. later, will be implemented fully functional.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+  {
+    throw new NotImplementedException();
+  }
+
+  public IDisposable? BeginScope<TState> (TState state)
```
MichaelKetting, line 36: returns null, unit test tests null and since we don't have a Scope for log4net, it will remain that way.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using log4net.Core;
+using Microsoft.Extensions.Logging;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLogger
```
MichaelKetting, line 23: xmldoc is either now or in next task.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+using log4net.Core;
+using log4net.Layout;
+using log4net.Repository.Hierarchy;
+using Microsoft.Extensions.Logging;
+using Remotion.Reflection;
+using Remotion.Utilities;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLoggerProvider : ILoggerProvider
```
MichaelKetting, line 30: xmldoc please.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+using System;
+using System.Reflection;
+using log4net.Appender;
+using log4net.Config;
+using log4net.Core;
+using log4net.Layout;
+using log4net.Repository.Hierarchy;
+using Microsoft.Extensions.Logging;
+using Remotion.Reflection;
+using Remotion.Utilities;
```
MichaelKetting, line 26: It seems like there should be a namespace cleanup?

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using log4net.Core;
```
MichaelKetting, line 18: namespace cleanup?

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLogger
+    : Microsoft.Extensions.Logging.ILogger
+{
+  public void Log<TState> (Microsoft.Extensions.Logging.LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
```
MichaelKetting, line 26: is there benefit to not using Microosoft.Extensions.Logging as an included namespace?

MichaelKetting (changes requested):

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
MichaelKetting: Ah, I see. Would be cleaner to do this one here already or stick to the using version mainly to keep expections. Can live with it.

KuJo7:
@MichaelKetting I noted all the points you mentioned and we can do the cleanup in the second task, since some points are already done

MichaelKetting (approved):

PR #1063, opened 2024-03-21
Feature/rm 9131 introduce public query id and query metadata to query factory create linq query

Frazerus:
https://re-motion.atlassian.net/browse/RM-9131

The implementations of the different CreateLinqQuery<T> cannot use optional parameters because of expression tree limitations. (CS0854). I'm not sure what the best way to create those methods is, but this is one of them which the benefits backwards compatability and allows the user to only write the id and metadata when they care about it. 

We could also have overloads for only ID and only Metadata, but if another param is ever needed, it's going out of hand quickly.

We also need to check whether the default value for the id should be a predetermined string or null until it is required.

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
@@ -15,6 +15,7 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Moq;
 using NUnit.Framework;
@@ -113,14 +114,36 @@ namespace Remotion.Data.DomainObjects.UnitTests.Queries
       var id = "id";
       var statement = "stmt";
       var parameterCollection = new QueryParameterCollection();
+      var metaData = new Dictionary<string, object>
+                     {
+                       { "dummyKey", "dummyValue" }
+                     };
 
-      IQuery query = QueryFactory.CreateScalarQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection);
+      IQuery query = QueryFactory.CreateScalarQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, metaData);
       Assert.That(query.CollectionType, Is.Null);
       Assert.That(query.ID, Is.EqualTo(id));
       Assert.That(query.Parameters, Is.SameAs(parameterCollection));
       Assert.That(query.QueryType, Is.EqualTo(QueryType.ScalarReadOnly));
       Assert.That(query.Statement, Is.EqualTo(statement));
       Assert.That(query.StorageProviderDefinition, Is.SameAs(TestDomainStorageProviderDefinition));
+      Assert.That(query.Metadata, Is.SameAs(metaData));
+    }
+
+    [Test]
+    public void CreateScalarQuery_WithNullMetadata_EmptyMetadataCollection ()
+    {
+      var id = "id";
+      var statement = "stmt";
+      var parameterCollection = new QueryParameterCollection();
+
+      IQuery query = QueryFactory.CreateScalarQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, null);
+      Assert.That(query.CollectionType, Is.Null);
+      Assert.That(query.ID, Is.EqualTo(id));
+      Assert.That(query.Parameters, Is.SameAs(parameterCollection));
+      Assert.That(query.QueryType, Is.EqualTo(QueryType.ScalarReadOnly));
+      Assert.That(query.Statement, Is.EqualTo(statement));
+      Assert.That(query.StorageProviderDefinition, Is.SameAs(TestDomainStorageProviderDefinition));
+      Assert.That(query.Metadata, Is.Empty);
     }
 
     [Test]
@@ -130,14 +153,37 @@ namespace Remotion.Data.DomainObjects.UnitTests.Queries
       var statement = "stmt";
       var parameterCollection = new QueryParameterCollection();
       var collectionType = typeof(OrderCollection);
+      var metaData = new Dictionary<string, object>
+                     {
+                       { "dummyKey", "dummyValue" }
+                     };
 
-      IQuery query = QueryFactory.CreateCollectionQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, collectionType);
+      IQuery query = QueryFactory.CreateCollectionQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, collectionType, metaData);
       Assert.That(query.ID, Is.EqualTo(id));
       Assert.That(query.CollectionType, Is.SameAs(collectionType));
       Assert.That(query.Parameters, Is.SameAs(parameterCollection));
       Assert.That(query.QueryType, Is.EqualTo(QueryType.CollectionReadOnly));
       Assert.That(query.Statement, Is.EqualTo(statement));
       Assert.That(query.StorageProviderDefinition, Is.SameAs(TestDomainStorageProviderDefinition));
+      Assert.That(query.Metadata, Is.SameAs(metaData));
+    }
+
+    [Test]
+    public void CreateCollectionQuery_WithNullMetadata_EmptyMetadataCollection ()
+    {
+      var id = "id";
+      var statement = "stmt";
+      var parameterCollection = new QueryParameterCollection();
+      var collectionType = typeof(OrderCollection);
+
+      IQuery query = QueryFactory.CreateCollectionQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, collectionType, null);
+      Assert.That(query.ID, Is.EqualTo(id));
+      Assert.That(query.CollectionType, Is.SameAs(collectionType));
+      Assert.That(query.Parameters, Is.SameAs(parameterCollection));
+      Assert.That(query.QueryType, Is.EqualTo(QueryType.CollectionReadOnly));
+      Assert.That(query.Statement, Is.EqualTo(statement));
+      Assert.That(query.StorageProviderDefinition, Is.SameAs(TestDomainStorageProviderDefinition));
+      Assert.That(query.Metadata, Is.Empty);
     }
 
     [Test]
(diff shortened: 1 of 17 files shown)


Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -86,8 +91,8 @@ public class QueryDefinition
   ///   <paramref name="statement"/> is an empty string.
   /// </exception>
   /// <exception cref="System.ArgumentOutOfRangeException"><paramref name="queryType"/> is not a valid enum value.</exception>
-  public QueryDefinition (string queryID, StorageProviderDefinition storageProviderDefinition, string statement, QueryType queryType)
-    : this(queryID, storageProviderDefinition, statement, queryType, null)
+  public QueryDefinition (string queryID, StorageProviderDefinition storageProviderDefinition, string statement, QueryType queryType, IReadOnlyDictionary<string, object>? metaData = null)
+    : this(queryID, storageProviderDefinition, statement, queryType, null, metaData)
```
martin-herwich, line 95: I think we should merge the two constructors, because `collectionType` is also effectively an optional parameter and should be implemented as such. (The current implementation is probably older than the "optional parameter" syntactic feature.)
It just seems awkward to mix two styles of optional parameters.

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -269,6 +286,7 @@ protected virtual void GetObjectData (SerializationInfo info, StreamingContext c
       info.AddValue("Statement", _statement);
       info.AddValue("QueryType", _queryType);
       info.AddValue("CollectionType", _collectionType);
+      info.AddValue("Metadata", _metadata);
```
martin-herwich, line 289: Since we can have any type of `object` in our metadata, what happens if one of our values is not serializable?

Remotion/Data/DomainObjects/Queries/QueryFactory.cs
```diff
@@ -63,10 +66,40 @@ public static class QueryFactory
...
+    /// customer (which includes an implicit join between <c>Order</c> and <c>Customer</c> objects).
+    /// <code>
+    /// var query =
+    ///     from o in QueryFactory.CreateLinqQuery&lt;Order&gt; ()
+    ///     where o.OrderNumber &lt;= 4
+    ///     orderby o.Customer.Name
+    ///     select o;
+    /// var result = query.ToArray();
+    /// </code>
+    /// </example>
```
martin-herwich, line 92: The code example does not fit this override, because it uses the override without parameters.

Remotion/Data/DomainObjects/Queries/QueryFactory.cs
```diff
@@ -201,16 +235,17 @@ public static IQuery CreateQueryFromConfiguration (string id, QueryParameterColl
     /// <param name="storageProviderDefinition">The <see cref="StorageProviderDefinition"/> used to execute the query.</param>
     /// <param name="statement">The scalar query statement.</param>
     /// <param name="queryParameterCollection">The parameter collection to be used for the query.</param>
+    /// <param name="metaData">The metadata dictionary supplied to the query definition.</param>
```
martin-herwich, line 238: Since we are going to create the `QueryDefintion`, this should be: "The metadata dictionary that will be supplied to the `QueryDefintion`".

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -76,6 +78,9 @@ public class QueryDefinition
   /// <param name="queryType">
   /// One of the <see cref="QueryType"/> enumeration constants.
   /// </param>
+  /// <param name="metaData">
+  /// Adds metadata to the query.
```
martin-herwich, line 82: What is the intended use for the metadata? This comment should explain to the reader why he might want to use the parameter.

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -52,8 +53,9 @@ public class QueryDefinition
   private readonly QueryType _queryType;
   private readonly Type? _collectionType;
   private readonly StorageProviderDefinition _storageProviderDefinition;
+  private IReadOnlyDictionary<string,object> _metadata;
 
-  // Note: _ispartOfQueryConfiguration is used only during the deserialization process. 
+  // Note: _ispartOfQueryConfiguration is used only during the deserialization process.
```
martin-herwich, line 58: We should capitalize the P in `_isPartOfQueryConfiguration`

martin-herwich (changes requested):

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
Frazerus: I have no clue, when using metadata correctly, it should not have any non serializable things inside (mostly strings or rather enums to cover the metadata aspect.)
But other than that, I'm not sure how this should behave.
@MichaelKetting?

MichaelKetting: @Frazerus & @martin-herwich : if  a value is not serializable, the serialization will fail. For enums and strings as values, that's a non-issue. If someone decides to put complex datastructures into the metadata, all bets are off.

For the backport, the projects don't use serialization. For re-motion v7, we will be dropping serialization anyway. For our own default value (an empty readonly dictionary, we should make sure the type is serializable. 

Perhaps it would be a good idea/helpful to note in the xmldoc for the parameter that the metadata dictionary and it's values must be serializable if the project uses serialization.

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -109,6 +80,10 @@ public QueryDefinition (string queryID, StorageProviderDefinition storageProvide
   /// <param name="collectionType">If <paramref name="queryType"/> specifies a collection to be returned, <paramref name="collectionType"/> specifies the type of the collection.
   /// If <paramref name="queryType"/> is <see langword="null"/>, <see cref="DomainObjectCollection"/> is used.
   /// </param>
+  /// <param name="metaData">
+  /// Adds metadata to a query which should be used to communicate additional information for the query, like whether it should connect to the read/write or readonly instance of
+  /// the db.
```
martin-herwich, line 85: This comment suggests that there are two instances of the database (db), but it is only the connection that is designated as read-only, the database is always the same. Therefore I suggest the following:
"... like whether the query should be executed using a read-only connection."

Frazerus: Huh, seems like I misunderstood how this is supposed to be used. Thanks

martin-herwich (approved):

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Queries/LinqProviderComponentFactory.cs
```diff
@@ -65,14 +66,14 @@ public virtual IQueryParser CreateQueryParser ()
...
-    public virtual IQueryExecutor CreateQueryExecutor (StorageProviderDefinition providerDefinition)
+    public virtual IQueryExecutor CreateQueryExecutor (StorageProviderDefinition providerDefinition, string id, IReadOnlyDictionary<string, object> metadata)
     {
       var queryGenerator = providerDefinition.Factory.CreateDomainObjectQueryGenerator(
           providerDefinition,
           _methodCallTransformerProvider,
           _resultOperatorHandlerRegistry,
           MappingConfiguration.Current);
-      return new DomainObjectQueryExecutor(providerDefinition, queryGenerator);
+      return new DomainObjectQueryExecutor(providerDefinition, queryGenerator, id, metadata);
```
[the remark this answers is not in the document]
Frazerus: I feel like that was a reference which I didnt get :(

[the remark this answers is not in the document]
MichaelKetting: Asterix & Brits :)

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -228,6 +209,14 @@ public QueryType QueryType
     get { return _collectionType; }
   }
 
+  /// <summary>
+  /// Used to e.g. provide diagnostic information or query hints.
```
MichaelKetting, line 213:
```suggestion
  /// Gets the provided diagnostic information or query hints for this <b>QueryDefinition</b>.
```

Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
```diff
@@ -113,47 +114,114 @@ public void CreateScalarQuery ()
...
 
-      IQuery query = QueryFactory.CreateCollectionQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, collectionType);
+      IQuery query = QueryFactory.CreateCollectionQuery(id, TestDomainStorageProviderDefinition, statement, parameterCollection, collectionType, null);
       Assert.That(query.ID, Is.EqualTo(id));
       Assert.That(query.CollectionType, Is.SameAs(collectionType));
       Assert.That(query.Parameters, Is.SameAs(parameterCollection));
       Assert.That(query.QueryType, Is.EqualTo(QueryType.CollectionReadOnly));
       Assert.That(query.Statement, Is.EqualTo(statement));
       Assert.That(query.StorageProviderDefinition, Is.SameAs(TestDomainStorageProviderDefinition));
+      Assert.That(query.Metadata.Count, Is.Zero);
```
MichaelKetting, line 186:
```suggestion
      Assert.That(query.Metadata, Is.Empty);
```

MichaelKetting (approved):
some small tweaks. review done.

SecurityManager/Core/Domain/AccessControl/AccessEvaluation/SecurityContextRevisionBasedCacheBase.cs
```diff
@@ -84,7 +84,8 @@ private IExecutableQuery<IEnumerable<T>> CreateExecutableQuery<T> (string key, F
             "<dynamic query>",
             queryExecutor.StorageProviderDefinition,
             queryModel,
-            Enumerable.Empty<FetchQueryModelBuilder>());
+            Enumerable.Empty<FetchQueryModelBuilder>(),
+            queryExecutor.Metadata);
```
MichaelKetting, line 88: In https://re-motion.atlassian.net/browse/RM-9164 you mentioned every place the metadata is exposed. Please update to mention where the metadata is part of the public api only, not when it is used internally.

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -44,6 +45,8 @@ public class QueryDefinition
   private static readonly NonPersistentProviderDefinition s_dummyStorageProviderDefinition =
       new NonPersistentProviderDefinition("NonSerializedStorageProviderDefinition", new NonPersistentStorageObjectFactory());
 
+  private static readonly IReadOnlyDictionary<string, object> s_emptyMetadata = new Dictionary<string, object>();
```
MichaelKetting, line 48: Here we still need to use a ReadOnly instance.

MichaelKetting (changes requested):

MichaelKetting (approved):

PR #1067, opened 2024-03-26
RM-9140 Add delay to remove flakiness of ResolveBorderElement tests

Frazerus:
https://re-motion.atlassian.net/browse/RM-9140


Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.Shared.cs
```diff
@@ -313,6 +313,11 @@ public void ResolveBorderElement (string testSide, string testName)
             if (Helper.BrowserConfiguration.IsChromium())
               Thread.Sleep(100);
 
+            // Due to race conditions, we need to wait for the Browser to fully finish drawing it's contents before we take the screenshot.
```
MichaelKetting, line 316: Looking at the line 313, I think I know why there's no (or fewer) issues in Chrome :D

For the temp workaround, the Assembly-Name-Test is sufficient. Testing for a Has-UI flag would be nicer but likely overkill in the workaround.

MichaelKetting (approved):

PR #1069, opened 2024-03-27
RM-9115   Implement Log4NetLogger based on Microsoft.Extensions.Logging.Abstractions

KuJo7:
https://re-motion.atlassian.net/browse/RM-9115


Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using Remotion.Utilities;
+using MicrosoftLogging = Microsoft.Extensions.Logging;
```
Frazerus, line 19: We can have a normal using statement here.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLoggerProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLoggerProvider : MicrosoftLogging.ILoggerProvider
```
Frazerus, line 23: We should probably add some xml docs here to give information about 
* that the class implements the `Microsoft.Extensions.Logging.ILoggerProvider` interface
* that this class exists because we wanted an implementation for the interface with log4net

Remotion/Core/Core/Core.Core.csproj
```diff
@@ -20,6 +20,7 @@
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Remove="JetBrains.Annotations" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
```
Frazerus, line 23: Since we are only adding things to Core.Extensions, this should probably be added there, not in `Core.Core` for now.

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.Extensions.Logging;
+using NUnit.Framework;
+using Remotion.Logging.Log4Net;
+
+namespace Remotion.Extensions.UnitTests.Logging.Log4NetTests;
+
+public class Log4NetLoggerProviderTest
+{
+  [Test]
+  public void CreateLogger_WithValidCategoryName_ReturnsLoggerInstance ()
```
Frazerus, line 27: We should figure out what the category name is used for and using that knowledge verify that it is used correctly here.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+using System;
+using System.Globalization;
+using log4net.Core;
+using log4net.Util;
+using MicrosoftLogging = Microsoft.Extensions.Logging;
+using RemotionLogging = Remotion.Logging;
+
+namespace Remotion.Logging.Log4Net;
+
+public class Log4NetLogger : MicrosoftLogging.ILogger
```
Frazerus, line 26: We still need to implement the methods on `ILog` in the extension methods for this class.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+
+  public IDisposable? BeginScope<TState> (TState state)
+      where TState : notnull
+  {
+    return null;
+  }
+
+  private LoggingEvent CreateLoggingEvent (Level level, int? eventID, object? message, Exception? exceptionObject)
+  {
+    var loggingEvent = new LoggingEvent(typeof(RemotionLogging.Log4NetLog), null, _logger.Name, level, message, exceptionObject);
```
Frazerus, line 93: We don't want to use `Log4NetLog` here, as it would create a dependency on something that we want to remove.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+      case MicrosoftLogging.LogLevel.Critical:
+        return Level.Critical;
+      default:
+        throw new ArgumentException(string.Format("LogLevel does not support value {0}.", logLevel), "logLevel");
+    }
+  }
+
+  public static Level Convert (RemotionLogging.LogLevel logLevel)
+  {
+    switch (logLevel)
```
Frazerus, line 56: We can use switch expressions here and in the other convert method, making it more cocise and IMO more readable: 
```
return logLevel switch 
{
 Loglevel.Debug => Level.Debug, 
.... 
}
```

Frazerus: We should also support the rest of the `Microsoft.Extensions.LogLevels`, there should be appropriate Levels in log4net

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+    return null;
+  }
+
+  private LoggingEvent CreateLoggingEvent (Level level, int? eventID, object? message, Exception? exceptionObject)
+  {
+    var loggingEvent = new LoggingEvent(typeof(RemotionLogging.Log4NetLog), null, _logger.Name, level, message, exceptionObject);
+
+    if (eventID.HasValue)
+    {
+      if (eventID < 0 || eventID > 0xFFFF)
```
Frazerus, line 97: I'm not sure this makes check makes sense in our infrastructure. 

Please find out why the event id can only be between those numbers and check how the microsoft API behaves.

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerTest.cs
```diff
@@ -0,0 +1,154 @@
...
+  private TextWriter _originalConsoleOut;
+  private StringWriter _fakeConsoleOut;
+
+  [SetUp]
+  public void SetUp ()
+  {
+    _originalConsoleOut = Console.Out;
+    _fakeConsoleOut = new StringWriter();
+    Console.SetOut(_fakeConsoleOut);
+
```
Frazerus, line 44:
```suggestion

```

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerTest.cs
```diff
@@ -0,0 +1,154 @@
...
+  [TearDown]
+  public void TearDown ()
+  {
+    log4net.LogManager.ResetConfiguration();
+    Console.SetOut(_originalConsoleOut);
+  }
+
+  [Test]
+  public void Convert_WithValidLogLevel_ReturnsConvertedLogLevel ()
```
Frazerus, line 55: I think this could also be done using test cases, which would probably lead to a more readable test.

We should also split the act and assert parts of this test.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+      case RemotionLogging.LogLevel.Error:
+        return Level.Error;
+      case RemotionLogging.LogLevel.Fatal:
+        return Level.Fatal;
+      default:
+        throw new ArgumentException(string.Format("LogLevel does not support value {0}.", logLevel), "logLevel");
+    }
+  }
+
+  public void Log<TState> (MicrosoftLogging.LogLevel logLevel, MicrosoftLogging.EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
```
Frazerus, line 73: We should add a test where we properly use the formatter.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+    if (_logger.IsEnabledFor(level))
+      _logger.Log(CreateLoggingEvent(level, eventId.Id, formatter(state, exception), exception));
+  }
+
+  public bool IsEnabled (MicrosoftLogging.LogLevel logLevel)
+  {
+    return _logger.IsEnabledFor(Convert(logLevel));
+  }
+
+  public IDisposable? BeginScope<TState> (TState state)
```
Frazerus, line 85: We are missing a test for this, unless we don't want one for this? I'm not sure in this case @MichaelKetting?

Frazerus (changes requested):

Frazerus:
I didn't notice that some of the files were already from the previous PR and that this extends the 9114 PR.

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
KuJo7: it's just the category name for messages produced by the logger, so I think its fine

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
KuJo7: As I checked, the eventID is a 32 bit int, but its value must be between 0 and 65535
So, I guess the checks we have, should make sense

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
MichaelKetting: I believe you're talking about the logger's name here
https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger-1?view=dotnet-plat-ext-8.0


This maps 1:1 to the log4net logger name passed during construction as name or category. it is essential for configuring/filtering logging events.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
KuJo7: I fixed this in the already existing test

Remotion/Core/Extensions/Logging/LogExtensions.cs
```diff
@@ -53,6 +55,22 @@ public static void LogFormat (this ILog log, LogLevel logLevel, Enum messageEnum
...
+    /// <param name="exceptionObject"></param>
+    /// <param name="args"></param>
+    public static void LogFormat (this ILogger logger, MicrosoftLogging.LogLevel logLevel, Enum messageEnum, Exception? exceptionObject, params object[] args)
+    {
+      ArgumentUtility.CheckNotNull("logger", logger);
+
+      logger.Log(logLevel,s_globalizationService.Value.GetEnumerationValueDisplayName(messageEnum), exceptionObject, args);
+    }
+
+
```
Frazerus, line 73:
```suggestion

```

Remotion/Core/Extensions/Logging/LogExtensions.cs
```diff
@@ -53,6 +55,22 @@ public static void LogFormat (this ILog log, LogLevel logLevel, Enum messageEnum
           args);
     }
 
+    /// <summary>
+    /// Log a message and event id derived from the <paramref name="messageEnum"/> with the specified <paramref name="logLevel"/>.
+    /// </summary>
+    /// <param name="logger"></param>
```
Frazerus, line 61: Please add info or remove param docs

Frazerus (approved):

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
```diff
@@ -31,7 +30,6 @@ public void CreateLogger_WithValidCategoryName_ReturnsLoggerInstance ()
 
     var logger = loggerProvider.CreateLogger(categoryName);
 
-    Assert.That(logger, Is.Not.Null);
     Assert.That(logger, Is.InstanceOf<Log4NetLogger>());
   }
```
MichaelKetting, line 34: assertion for the log4net logger's name matching categoryName

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -15,27 +15,118 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 //
 using System;
+using System.Globalization;
 using log4net.Core;
-using Microsoft.Extensions.Logging;
+using log4net.Util;
+using MicrosoftLogging = Microsoft.Extensions.Logging;
+using RemotionLogging = Remotion.Logging;
```
MichaelKetting, line 22: I thought we would do type alising, not namespace alisaing 
MicrosoftLogger
RemotionLogger
MicorosftLogLevel
RemotionLogLevel
etc

Remotion/Core/Core/Core.Core.csproj
```diff
@@ -20,7 +20,6 @@
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Remove="JetBrains.Annotations" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
```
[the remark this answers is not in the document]
Frazerus: We do need it, but later, here we only create the log4net things in core.extensions.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -15,27 +15,118 @@
...
+    else if (eventID > 0xFFFF)
+      safeEventID  = 0xFFFF;
+    else
+      safeEventID = eventID;
+
+    Level level = Level.Error;
+    if (_logger.IsEnabledFor(level))
+    {
+      if (_logger.IsEnabledFor(level))
+      {
```
MichaelKetting, line 118: This is a copy of the original code but since we are loosing the trackable history, we should remove the redundancy to avoid questions about this in the future.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -15,27 +15,118 @@
...
+            "eventID", string.Format("An event id of value {0} is not supported. Valid event ids must be within a range of 0 and 65535.", eventID));
+      }
+
+      loggingEvent.Properties["EventID"] = eventID;
+    }
+
+    return loggingEvent;
+  }
+
+  private void LogLoggingError (int eventID, Exception? exceptionObject, object? message)
```
MichaelKetting, line 104: the message in the new infrastructure is of type `string` (not nullable) not `object` (nullable). Please update. After this change, we can drop the `ToString` further down.

MichaelKetting (changes requested):
@Frazerus I have some questions on the test design which we should discuss before @KuJo7 checks them out.

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
```diff
@@ -15,27 +15,118 @@
...
+/// </summary>
+public class Log4NetLogger : MicrosoftLogging.ILogger
 {
-  public void Log<TState> (Microsoft.Extensions.Logging.LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+  private readonly log4net.Core.ILogger _logger;
+
+  public Log4NetLogger (string categoryName)
+  {
+    _logger = log4net.LogManager.GetLogger(categoryName).Logger;
+  }
```
MichaelKetting, line 36: I missed this in the first review (sorry): we need to pass the log4net logger in the ctor, not resolve it manually inside the ctor. this would create a dependency we do not want.

Remotion/Core/Extensions.UnitTests/Logging/Log4NetTests/Log4NetLoggerProviderTest.cs
MichaelKetting: It's okay to export the log4net Logger on the public API.

Naming: Logger.

Checkout Log4NetLogManagerTest.GetLogger_WithNameAsString_ReturnsSameLoggerTwice on how we solved this in the original code base

Remotion/Core/Extensions/Logging/Log4Net/Log4NetLogger.cs
Frazerus: We already found this locally, would've been fixed anyways :)

PR #1072, opened 2024-03-29
RM-9144 Possible Race-Condition in Smartpage PageToken field

Frazerus:
https://re-motion.atlassian.net/browse/RM-9144

### DIFF
--- a/Remotion/Web/Core/UI/SmartPageImplementation/SmartPageInfo.cs
+++ b/Remotion/Web/Core/UI/SmartPageImplementation/SmartPageInfo.cs
@@ -295,7 +295,7 @@ namespace Remotion.Web.UI.SmartPageImplementation
 
     private void PreRenderSmartPage ()
     {
-      var smartPageTokenValue = (long)(DateTime.UtcNow - s_pageTokenStartDate).TotalMilliseconds;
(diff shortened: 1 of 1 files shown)


Remotion/Web/Core/UI/SmartPageImplementation/SmartPageInfo.cs
```diff
@@ -295,7 +295,7 @@ private void Page_PreRenderComplete (object? sender, EventArgs eventArgs)
 
     private void PreRenderSmartPage ()
     {
-      var smartPageTokenValue = (long)(DateTime.UtcNow - s_pageTokenStartDate).TotalMilliseconds;
+      var smartPageTokenValue = (DateTime.UtcNow - s_pageTokenStartDate).Ticks;
```
MichaelKetting, line 298: before I approve, please check if increasing the token-length has any adverse effects on the browser state management. We are using the token as part of a key for the clientside storage API and a longer key will increase memory consumption. we need to make sure that the increase is negligible. Please check the avaialble discussio in the original tickets on this and do a quick performance test.

MichaelKetting (commented):
there's an open question before we can approve.

MichaelKetting (approved):

PR #1075, opened 2024-04-02
RM-9157 Badge of treeNode correctly shows dropshadow

Frazerus:
https://re-motion.atlassian.net/browse/RM-9157

### DIFF
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
@@ -174,8 +174,8 @@ ul.treeViewRoot span.treeViewNode span.treeViewNodeBadge
   align-self: start;
 }
 
-ul.treeViewRoot li:has(a:focus) > ul > li:first-of-type span.treeViewNode:not(:hover) span.treeViewNodeBadge,
-ul.treeViewRoot li:has(a:focus) + li span.treeViewNode:not(:hover) span.treeViewNodeBadge
+ul.treeViewRoot li:has(> span.treeViewNode span[class^="treeViewNodeHead"] a[role=treeitem]:focus) > ul > li:first-of-type > span.treeViewNode:not(:hover) span.treeViewNodeBadge,
+ul.treeViewRoot li:has(> span.treeViewNode span[class^="treeViewNodeHead"] a[role=treeitem]:focus) + li:first-of-type > span.treeViewNode:not(:hover) span.treeViewNodeBadge
 {
   /* Simulate continued box-shadow from previous element due to the badge-element hiding the shadow-effect. */
   --rounding-value: 0.5px;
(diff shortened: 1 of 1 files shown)


Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
```diff
@@ -174,8 +174,8 @@ ul.treeViewRoot span.treeViewNode span.treeViewNodeBadge
   align-self: start;
 }
 
-ul.treeViewRoot li:has(a:focus) > ul > li:first-of-type span.treeViewNode:not(:hover) span.treeViewNodeBadge,
-ul.treeViewRoot li:has(a:focus) + li span.treeViewNode:not(:hover) span.treeViewNodeBadge
+ul.treeViewRoot li:has(> span.treeViewNode span[class^="treeViewNodeHead"] a[role=treeitem]:focus) > ul > li:first-of-type > span.treeViewNode:not(:hover) span.treeViewNodeBadge,
+ul.treeViewRoot li:has(> span.treeViewNode span[class^="treeViewNodeHead"] a[role=treeitem]:focus) + li:first-of-type > span.treeViewNode:not(:hover) span.treeViewNodeBadge
```
MichaelKetting, line 178: since we don't do substring matching, I would prefer to not open pandora's box right now. can we do either two selectors or an OR?

MichaelKetting, line 178: I think the match for role=treeitem is overqualiciation.

MichaelKetting (approved):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/TreeView.css
Frazerus: I think we should just remove the `span[class^='treeVieNodeHead"]`, it's not really required as we only care about the first `span `being a direct child of `li`

Frazerus: I think we should keep this, with the above changes I requested, we still need it, and it's probably cleaner this way

MichaelKetting: oh? I thought it was also thhat the anchor should be the head-anchor not a child-item-anchor?

MichaelKetting: hmm..yes, there's no short way to do this.

I looked into the perf-stats and now I'm worried about using `aria-selected` instead of treenodeheadselected css class. Maybe a ticket for this to be checked later?
image

Frazerus: it can't be a child item anchor, the anchor with role=treeitem is unique inside the treeViewNode, so it should be fine.

MichaelKetting: Ah, I checked but did not see :D

Then we can simplify to `> span.treeviewNode a:focus` ?

Frazerus: If we think that it's impossible that the <a> expander has focus, then yes, otherwise we have the same problem again of not being specific enough

PR #1080, opened 2024-04-03
Feature/rm 9175 add decorator to i read only storage provider instances that checks the query type

Frazerus:
https://re-motion.atlassian.net/browse/RM-9175

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
@@ -1,0 +1,128 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.DataManagement;
+using Remotion.Data.DomainObjects.Persistence;
+using Remotion.Data.DomainObjects.Queries;
+using Remotion.Data.DomainObjects.Queries.Configuration;
+using Remotion.Development.NUnit.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence
+{
+  public class ReadOnlyStorageProviderDecoratorTest
+  {
+    private static QueryType[] s_readOnlyTestCases = { QueryType.CollectionReadOnly, QueryType.CustomReadOnly, QueryType.ScalarReadOnly };
+    private static QueryType[] s_readWriteTestCases = { QueryType.CollectionReadWrite, QueryType.CustomReadWrite, QueryType.ScalarReadWrite };
+
+    [Test]
+    [TestCaseSource(nameof(s_readOnlyTestCases))]
+    public void ExecuteCollectionQuery_WithReadOnlyQueryType_CallsInnerStorageProvider (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderMock = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderMock.Object);
+      var expectedResult = new DataContainer[1];
+      innerReadOnlyStorageProviderMock.Setup(_ => _.ExecuteCollectionQuery(queryStub.Object)).Returns(expectedResult);
+
+      var result = readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object);
+
+      Assert.That(result, Is.SameAs(expectedResult));
+    }
+
+    [Test]
+    [TestCaseSource(nameof(s_readOnlyTestCases))]
+    public void ExecuteCustomQuery_WithReadOnlyQueryType_CallsInnerStorageProvider (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+      var expectedResult = new IQueryResultRow[1];
+      innerReadOnlyStorageProviderStub.Setup(_ => _.ExecuteCustomQuery(queryStub.Object)).Returns(expectedResult);
+
+      var result = readOnlyStorageProviderDecorator.ExecuteCustomQuery(queryStub.Object);
+
+      Assert.That(result, Is.SameAs(expectedResult));
+    }
+
+    [Test]
+    [TestCaseSource(nameof(s_readOnlyTestCases))]
+    public void ExecuteScalarQuery_WithReadOnlyQueryType_CallsInnerStorageProvider (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+      innerReadOnlyStorageProviderStub.Setup(_ => _.ExecuteScalarQuery(queryStub.Object)).Returns("Successful!");
+
+      var result = readOnlyStorageProviderDecorator.ExecuteScalarQuery(queryStub.Object);
+
+      Assert.That(result, Is.EqualTo("Successful!"));
+    }
+
+    [Test]
+    [TestCaseSource(nameof(s_readWriteTestCases))]
+    public void ExecuteCollectionQuery_WithReadWriteQueryType_ThrowsArgumentException (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+
+      Assert.That(
+          () => readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object),
+          Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+              $"Query '{queryStub.Object.ID}' has query type '{queryStub.Object.QueryType}' which cannot be used in a read-only query scenario.",
+              "query"));
+    }
+
+    [Test]
+    [TestCaseSource(nameof(s_readWriteTestCases))]
+    public void ExecuteScalarQuery_WithReadWriteQueryType_ThrowsArgumentException (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+
+      Assert.That(
+          () => readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object),
+          Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+              $"Query '{queryStub.Object.ID}' has query type '{queryStub.Object.QueryType}' which cannot be used in a read-only query scenario.",
+              "query"));
+    }
+
+    [Test]
+    [TestCaseSource(nameof(s_readWriteTestCases))]
+    public void ExecuteCustomQuery_WithReadWriteQueryType_ThrowsArgumentException (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+
+      Assert.That(
+          () => readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object),
+          Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+              $"Query '{queryStub.Object.ID}' has query type '{queryStub.Object.QueryType}' which cannot be used in a read-only query scenario.",
+              "query"));
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


Remotion/Data/DomainObjects/Persistence/ReadOnlyStorageProviderDecorator.cs
```diff
@@ -0,0 +1,118 @@
...
+
+      return InnerStorageProvider.ExecuteCollectionQuery(query);
+    }
+
+    public IEnumerable<IQueryResultRow> ExecuteCustomQuery (IQuery query)
+    {
+      ArgumentUtility.CheckNotNull("query", query);
+
+      if (query.QueryType != QueryType.CustomReadOnly)
+        throw new ArgumentException($"Query '{query}' with query type '{query.QueryType}' cannot be used with an IReadOnlyStorageProvider.ExecuteCustomQuery.", nameof(query));
```
martin-herwich, line 88: The decorator should only throw when the QueryType is read-write; the decorator is not responsible for detecting the wrong result type - that is done by the decorated StorageProvider.
Use a Check-method to throw when the QueryType is read-write, and use that in all three Execute methods.
Update the unit tests accordingly.

SecurityManager/Core.UnitTests/Persistence/SecurityManagerSqlStorageObjectFactoryTest.cs
```diff
@@ -80,22 +80,24 @@ public void CreateReadOnlyStorageProvider ()
...
     {
+      //I have no clue how to fix this test. Any input?
       using (
           MixinConfiguration.BuildFromActive().ForClass(typeof(RdbmsProvider)).Clear().AddMixins(typeof(SecurityManagerRdbmsProviderTestMixin)).
               EnterScope())
       {
         var result = _securityManagerSqlStorageObjectFactory.CreateReadOnlyStorageProvider(_rdbmsProviderDefinition, _persistenceExtensionStub.Object);
 
-        Assert.That(Mixin.Get<SecurityManagerRdbmsProviderTestMixin>(result), Is.Not.Null);
+        Assert.That(Mixin.Get<RdbmsProvider>(result), Is.Not.Null);
```
martin-herwich, line 100: The `result` is the `ReadOnlyStorageProviderDecorator`; you need the decorated `SecurityManagerRdbmsProvider` for the call to `Mixin.Get<SecurityManagerRdbmsProviderTestMixin>()`.
This is because `RdbmsProvider` - and therefore its derived class `SecurityManagerRdbmsProvider` - is the mixin target, which is decorated within `SqlStorageObjectFactory.GetReadOnlyStorageProvider()`.

martin-herwich (changes requested):

SecurityManager/Core.UnitTests/Persistence/SecurityManagerSqlStorageObjectFactoryTest.cs
Frazerus: For the updated: I'm not sure this is how we actually want the API to look, but I'm probably missing something.

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
martin-herwich, line 1: WithCorrectType/WithInvalidQueryType: "Correct" and "Invalid" are imprecise labels for our test cases. It should be "ReadOnly" and "ReadWrite". Also, consistently use either "QueryType" or "Type" only.

martin-herwich (approved):
Please rename the tests, the rest is fine.

SecurityManager/Core.UnitTests/Persistence/SecurityManagerSqlStorageObjectFactoryTest.cs
martin-herwich: That is fine. We don't usually want to get the `Mixin` objects, we only do that here to test that the instance is indeed mixed, i.e. created using `ObjectFactory` (instead of `new`). Since the mixed instance is now decorated, the test must cast to the decorator to get to the mixed object.
For developers using the `IReadOnlyStorageProvider`, these details are hidden behind the interface, and everything works as intended.

Frazerus: Ok, thanks for the explanation! :)

SecurityManager/Core.UnitTests/Persistence/SecurityManagerSqlStorageObjectFactoryTest.cs
```diff
@@ -95,7 +96,9 @@ public void CreateReadOnlyStorageProviderWithMixin ()
       {
         var result = _securityManagerSqlStorageObjectFactory.CreateReadOnlyStorageProvider(_rdbmsProviderDefinition, _persistenceExtensionStub.Object);
 
-        Assert.That(Mixin.Get<SecurityManagerRdbmsProviderTestMixin>(result), Is.Not.Null);
+        var innerReadOnlyStorageProvider = result.As<ReadOnlyStorageProviderDecorator>().InnerStorageProvider;
```
MichaelKetting, line 99:
```suggestion
        Assert.That(result, Is.TypeOf(typeof(ReadOnlyStorageProviderDecorator)));
        var innerReadOnlyStorageProvider = result.As<ReadOnlyStorageProviderDecorator>().InnerStorageProvider;
```

Remotion/Data/DomainObjects/Persistence/ReadOnlyStorageProviderDecorator.cs
```diff
@@ -0,0 +1,125 @@
...
+  public class ReadOnlyStorageProviderDecorator : IReadOnlyStorageProvider
+  {
+    public ReadOnlyStorageProviderDecorator (IReadOnlyStorageProvider innerStorageProvider)
+    {
+      ArgumentUtility.CheckNotNull("innerStorageProvider", innerStorageProvider);
+
+      InnerStorageProvider = innerStorageProvider;
+    }
+
+    public IReadOnlyStorageProvider InnerStorageProvider { get; }
```
MichaelKetting, line 41: state, please move to before ctor.

Remotion/Data/DomainObjects/Persistence/ReadOnlyStorageProviderDecorator.cs
```diff
@@ -0,0 +1,125 @@
...
+      InnerStorageProvider.Rollback();
+    }
+
+    private void CheckQueryType (IQuery query)
+    {
+      _ = query.QueryType switch
+      {
+        QueryType.CollectionReadWrite or QueryType.ScalarReadWrite or QueryType.CustomReadWrite => throw new ArgumentException(
+            $"Read-write query '{query}' cannot be used with IReadOnlyStorageProvider.", nameof(query)),
+        _ => true
```
MichaelKetting, line 121: I think this would read better as an IF statement. The switch-syntax without actually using the result values makes it very difficult to understand what the actual intent is without already knowing it.

We should consider giving more context with the ArgumentException:
"Query '{query.ID}' has query type '{query.Type}' which cannot be used in a read-only query scenario."

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderManagerTest.cs
```diff
@@ -55,8 +55,10 @@ public void GetMandatory_WithProviderID ()
 #pragma warning restore CS0618 // Type or member is obsolete
 
       Assert.That(provider, Is.Not.Null);
-      Assert.That(provider.GetType(), Is.EqualTo(typeof(RdbmsProvider)));
-      Assert.That(provider.As<RdbmsProvider>().StorageProviderDefinition.Name, Is.EqualTo(c_testDomainProviderID));
+      Assert.That(provider.GetType(), Is.EqualTo(typeof(ReadOnlyStorageProviderDecorator)));
```
MichaelKetting, line 58: I did not notice this before, but we should use 
Assert.That(provider, Is.InstanceOf<ReadOnlyStorageProviderDecorator>());
here and in line 70

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
```diff
@@ -0,0 +1,131 @@
...
+          () => readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object),
+          Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+              $"Read-write query '{queryStub.Object}' cannot be used with IReadOnlyStorageProvider.",
+              "query"));
+    }
+
+    [Test]
+    [TestCase(QueryType.CollectionReadWrite)]
+    [TestCase(QueryType.CustomReadWrite)]
+    [TestCase(QueryType.ScalarReadWrite)]
```
MichaelKetting, line 116: you could use a TestCaseSource, supplying an array defined once in a static field. That way, no need to repeat the list of unsupported (or supported) QueryType values.

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
```diff
@@ -0,0 +1,131 @@
...
+    public void ExecuteScalarQuery_WithReadOnlyQueryType_CallsInnerStorageProvider (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+
+      Assert.That(() => readOnlyStorageProviderDecorator.ExecuteScalarQuery(queryStub.Object), Throws.Nothing);
+
+      innerReadOnlyStorageProviderStub.Verify(_ => _.ExecuteScalarQuery(queryStub.Object), Times.Once);
```
MichaelKetting, line 74: even though Moq supports this, calling Verify on a test double that's named 'stub' is semantically incorrect and will lead to false assumptions about test correctness in the future. 

In general, I recommend supplying a fake return value and asserting that for the returnvalue instead of asserting the actual call whenever this is possible. Another benefit of this is that right now, the test actually does not assert that the correct value is returned. You could have the ExecuteSqualarQuery method on the decorator return NULL and it would not fail the test.

Remidner: affects all three tests, please.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
Frazerus: I would keep the Verify setup for the collection tests because the setup with datacontainers and object ids is too convoluted for this otherwise very simple test.

SecurityManager/Core.UnitTests/Persistence/SecurityManagerSqlStorageObjectFactoryTest.cs
```diff
@@ -80,10 +80,11 @@ public void CreateReadOnlyStorageProvider ()
     {
       var result = _securityManagerSqlStorageObjectFactory.CreateReadOnlyStorageProvider(_rdbmsProviderDefinition, _persistenceExtensionStub.Object);
 
-      Assert.That(result, Is.TypeOf(typeof(SecurityManagerRdbmsProvider)));
-      Assert.That(result.As<SecurityManagerRdbmsProvider>().PersistenceExtension, Is.SameAs(_persistenceExtensionStub.Object));
-      Assert.That(result.As<SecurityManagerRdbmsProvider>().StorageProviderDefinition, Is.SameAs(_rdbmsProviderDefinition));
-      Assert.That(result.As<SecurityManagerRdbmsProvider>().ConnectionString, Is.EqualTo(_rdbmsProviderDefinition.ReadOnlyConnectionString));
+      Assert.That(result, Is.TypeOf(typeof(ReadOnlyStorageProviderDecorator)));
```
MichaelKetting, line 83:
```suggestion
      Assert.That(result, Is.InstanceOf<ReadOnlyStorageProviderDecorator>());
```
I prefer the generic syntax, shorter :)
as for TypeOf vs InstanceOf: one tests exact match, the other allows derived types. Occasionally, it's relevant, other times it's a question of which is used more often.

Remotion/Data/DomainObjects/Persistence/ReadOnlyStorageProviderDecorator.cs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    public void Rollback ()
+    {
+      InnerStorageProvider.Rollback();
+    }
+
+    private void CheckQueryType (IQuery query)
+    {
+      if (query.QueryType is QueryType.CollectionReadWrite or QueryType.CustomReadWrite or QueryType.ScalarReadWrite)
```
MichaelKetting, line 117: I forgot while writing the initial review comment:

Do you think it has added value to check for allowed instead of forbidden QueryType values?

With allowed values, we make sure to list only those we actually support.

With denied values, if there's a new value introduced that's also incorrect, the argument check would miss it.

I'm not sure if it's worth it though at this point, but in my opionion it's the safer and thus cleaner approach. If you think the argument is compelling enough to actually update the code, then please do. 

@martin-herwich  might also have an opinion on it.

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
```diff
@@ -0,0 +1,126 @@
...
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var resultRowStub = new Mock<IQueryResultRow>();
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+      innerReadOnlyStorageProviderStub.Setup(_ => _.ExecuteCustomQuery(queryStub.Object)).Returns(new[] { resultRowStub.Object });
+
+      var result = readOnlyStorageProviderDecorator.ExecuteCustomQuery(queryStub.Object);
+
+      Assert.That(result.Single(), Is.EqualTo(resultRowStub.Object));
```
MichaelKetting, line 60:
```suggestion
      Assert.That(result, Is.EqualTo(new[] { resultRowStub.Object }));
```
I prefer this syntax to the `Single` statement because it gives better diagnostics in there's zero or more than one item in the result.

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
```diff
@@ -0,0 +1,126 @@
...
+    [Test]
+    [TestCaseSource(nameof(s_readOnlyTestCases))]
+    public void ExecuteCustomQuery_WithReadOnlyQueryType_CallsInnerStorageProvider (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var resultRowStub = new Mock<IQueryResultRow>();
+      var innerReadOnlyStorageProviderStub = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderStub.Object);
+      innerReadOnlyStorageProviderStub.Setup(_ => _.ExecuteCustomQuery(queryStub.Object)).Returns(new[] { resultRowStub.Object });
```
MichaelKetting, line 56: Tipp: if you set this up as reutrns an array of IQueryResultRow, you can simply test for `SameAs` and the don't need to fiddle with the actual result content or fake objects.

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
```diff
@@ -0,0 +1,126 @@
...
+    public void ExecuteCollectionQuery_WithReadOnlyQueryType_CallsInnerStorageProvider (QueryType queryType)
+    {
+      var queryStub = new Mock<IQuery>();
+      queryStub.Setup(_ => _.QueryType).Returns(queryType);
+      var innerReadOnlyStorageProviderMock = new Mock<IReadOnlyStorageProvider>();
+      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderMock.Object);
+
+      Assert.That(() => readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object), Throws.Nothing);
+
+      innerReadOnlyStorageProviderMock.Verify(_ => _.ExecuteCollectionQuery(queryStub.Object), Times.Once);
```
MichaelKetting, line 44:
```suggestion
      var readOnlyStorageProviderDecorator = new ReadOnlyStorageProviderDecorator(innerReadOnlyStorageProviderMock.Object);
      var expectedResult = new DataContainer[1];
      innerReadOnlyStorageProviderStub.Setup(_ => _.ExecuteCollectionQuery(queryStub.Object)).Returns(expectedResult);
      Assert.That(readOnlyStorageProviderDecorator.ExecuteCollectionQuery(queryStub.Object), Is.SameAs(expectedResult));
```

Hope I didn't mess up the syntax.

MichaelKetting (approved):
Small tweaks plus a suggestion.

Remotion/Data/DomainObjects.UnitTests/Persistence/ReadOnlyStorageProviderDecoratorTest.cs
Frazerus: Ahh that makes sense. Thanks!

PR #1112, opened 2024-04-19
Feature/rm 9179 refactor query parameter handling

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9179

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactoryTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactoryTest.cs
@@ -1,0 +1,145 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Persistence;
+using Remotion.Data.DomainObjects.Persistence.Configuration;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.Queries;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Data.DomainObjects.UnitTests.Persistence.Configuration;
+using Remotion.Development.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.Parameters;
+
+[TestFixture]
+public class ObjectIDDataParameterDefinitionFactoryTest : StandardMappingTest
+{
+  private Mock<IDataParameterDefinitionFactory> _nextDataParameterDefinitionFactoryMock;
+
+  [SetUp]
+  public override void SetUp ()
+  {
+    base.SetUp();
+
+    _nextDataParameterDefinitionFactoryMock = new Mock<IDataParameterDefinitionFactory>();
+  }
+
+  [Test]
+  public void Initialize ()
+  {
+    var storageProviderDefinition = new TestableStorageProviderDefinition("name", Mock.Of<IStorageObjectFactory>());
+    var storageTypeInformationProvider = Mock.Of<IStorageTypeInformationProvider>();
+    var storageSettings = Mock.Of<IStorageSettings>();
+
+    var result = new ObjectIDDataParameterDefinitionFactory(storageProviderDefinition, storageTypeInformationProvider, storageSettings, _nextDataParameterDefinitionFactoryMock.Object);
+
+    Assert.That(result.StorageProviderDefinition, Is.SameAs(storageProviderDefinition));
+    Assert.That(result.StorageTypeInformationProvider, Is.SameAs(storageTypeInformationProvider));
+    Assert.That(result.StorageSettings, Is.SameAs(storageSettings));
+  }
+
+  [Test]
+  public void CreateDataParameterDefinition_WithObjectIDInSameProviderDefinition ()
+  {
+    var storageProviderDefinition = DomainObjectIDs.Customer1.StorageProviderDefinition;
+    var otherObjectID = DomainObjectIDs.Customer2;
+
+    var storageTypeInformationStub = new Mock<IStorageTypeInformation>();
+
+    var storageTypeInformationProviderStub = new Mock<IStorageTypeInformationProvider>();
+    storageTypeInformationProviderStub
+        .Setup(_ => _.GetStorageTypeForID(true))
+        .Returns(storageTypeInformationStub.Object);
+
+    var factory = new ObjectIDDataParameterDefinitionFactory(
+        storageProviderDefinition,
+        storageTypeInformationProviderStub.Object,
+        StorageSettings,
+        _nextDataParameterDefinitionFactoryMock.Object);
+
+    _nextDataParameterDefinitionFactoryMock
+        .Setup(_ => _.CreateDataParameterDefinition(It.IsAny<QueryParameter>()))
+        .Throws(new AssertionException("Should not be called."));
+
+    var result = factory.CreateDataParameterDefinition(new QueryParameter("other", otherObjectID));
+
+    Assert.That(result, Is.InstanceOf<ObjectIDDataParameterDefinition>());
+    Assert.That(result.As<ObjectIDDataParameterDefinition>().ValueStorageTypeInformation, Is.SameAs(storageTypeInformationStub.Object));
+  }
+
+  [Test]
+  public void CreateDataParameterDefinition_WithObjectIDInOtherProviderDefinition ()
+  {
+    var providerDefinition = new TestableStorageProviderDefinition("bla", Mock.Of<IStorageObjectFactory>());
+    var otherObjectID = DomainObjectIDs.Order5;
+
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation(dotNetType: typeof(string));
+
+    var storageTypeInformationProviderStub = new Mock<IStorageTypeInformationProvider>();
+    storageTypeInformationProviderStub
+        .Setup(_ => _.GetStorageTypeForSerializedObjectID(true))
+        .Returns(storageTypeInformation);
+
+    var factory = new ObjectIDDataParameterDefinitionFactory(
+        providerDefinition,
+        storageTypeInformationProviderStub.Object,
+        StorageSettings,
+        _nextDataParameterDefinitionFactoryMock.Object);
+
+    _nextDataParameterDefinitionFactoryMock
+        .Setup(_ => _.CreateDataParameterDefinition(It.IsAny<QueryParameter>()))
+        .Throws(new AssertionException("Should not be called."));
+
+    var result = factory.CreateDataParameterDefinition(new QueryParameter("other", otherObjectID));
+
+    Assert.That(result, Is.InstanceOf<SerializedObjectIDDataParameterDefinition>());
+    Assert.That(result.As<SerializedObjectIDDataParameterDefinition>().StorageTypeInformation, Is.SameAs(storageTypeInformation));
+  }
+
+  [Test]
+  public void CreateDataParameterDefinition_WithSimpleValue_HandsQueryParameterToNextFactory ()
+  {
+    var providerDefinition = new TestableStorageProviderDefinition("bla", Mock.Of<IStorageObjectFactory>());
+    var dummy = "Dummy";
+    var queryParameter = new QueryParameter("dummy", dummy);
+
+    var storageTypeInformation = Mock.Of<IStorageTypeInformation>();
+    var storageTypeInformationProviderStub = new Mock<IStorageTypeInformationProvider>();
+    storageTypeInformationProviderStub
+        .Setup(_ => _.GetStorageType(dummy))
+        .Returns(storageTypeInformation);
+
+    var expectedDataParameterDefinition = Mock.Of<IDataParameterDefinition>();
+    _nextDataParameterDefinitionFactoryMock
+        .Setup(_ => _.CreateDataParameterDefinition(queryParameter))
+        .Returns(expectedDataParameterDefinition);
+
+    var dataParameterDefinitionFactory = new ObjectIDDataParameterDefinitionFactory(
+        providerDefinition,
+        storageTypeInformationProviderStub.Object,
+        StorageSettings,
+        _nextDataParameterDefinitionFactoryMock.Object);
+
+    var result = dataParameterDefinitionFactory.CreateDataParameterDefinition(queryParameter);
+
+    Assert.That(result, Is.SameAs(expectedDataParameterDefinition));
+  }
+}
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
@@ -1,0 +1,163 @@
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.ComponentModel;
+using System.Data;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Data.DomainObjects.UnitTests.TestDomain;
+using Remotion.Development.NUnit.UnitTesting;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.Parameters;
+
+[TestFixture]
+public class ObjectIDDataParameterDefinitionTest : StandardMappingTest
+{
+  [Test]
+  public void GetParameterValue_WithObjectID_ReturnsObjectIDValue ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateUniqueIdentifierStorageTypeInformation();
+    var dummyValue = DomainObjectIDs.Order1;
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformation);
+
+    var result = objectIDDataParameterDefinition.GetParameterValue(dummyValue);
+
+    Assert.That(result, Is.EqualTo(dummyValue.Value));
+  }
+
+  [Test]
+  public void GetParameterValue_WithNull_ReturnsDBNull ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateUniqueIdentifierStorageTypeInformation(true);
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformation);
+
+    Assert.That(objectIDDataParameterDefinition.GetParameterValue(null), Is.EqualTo(DBNull.Value));
+  }
+
+  [Test]
+  public void GetParameterValue_WithTypeMismatchValue_ThrowsArgumentException ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateIntStorageTypeInformation();
+    var dummyValue = new ObjectID(typeof(Customer), Guid.NewGuid());
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformation);
+
+    Assert.That(
+        () => objectIDDataParameterDefinition.GetParameterValue(dummyValue),
+        Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+            $"Parameter 'parameterObject.Value' has type '{typeof(Guid)}' when type '{typeof(int)}' was expected.",
+            "parameterObject.Value"));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithIntValue_SetsNameValueType ()
+  {
+    var parameterValue = 42;
+    var commandStub = new Mock<IDbCommand>();
+    var dataParameterStub = new Mock<IDbDataParameter>();
+
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateIntStorageTypeInformation();
+
+    commandStub
+        .Setup(_ => _.CreateParameter())
+        .Returns(dataParameterStub.Object);
+
+    dataParameterStub.SetupProperty(_ => _.ParameterName);
+    dataParameterStub.SetupProperty(_ => _.DbType);
+    dataParameterStub.SetupProperty(_ => _.Value);
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformation);
+
+    var result = objectIDDataParameterDefinition.CreateDataParameter(commandStub.Object, "dummy", parameterValue);
+
+    Assert.That(result, Is.SameAs(dataParameterStub.Object));
+    Assert.That(result.ParameterName, Is.EqualTo("dummy"));
+    Assert.That(result.Value, Is.EqualTo(42));
+    Assert.That(result.DbType, Is.EqualTo(DbType.Int32));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithGuidValue_SetsNameValueType ()
+  {
+    var parameterValue = Guid.NewGuid();
+    var commandStub = new Mock<IDbCommand>();
+    var dataParameterStub = new Mock<IDbDataParameter>();
+
+    var storageTypeInformation = new StorageTypeInformation(
+        typeof(Guid),
+        "test",
+        DbType.Guid,
+        false,
+        null,
+        typeof(Guid),
+        new DefaultConverter(typeof(Guid)));
+
+    commandStub
+        .Setup(_ => _.CreateParameter())
+        .Returns(dataParameterStub.Object);
+
+    dataParameterStub.SetupProperty(_ => _.ParameterName);
+    dataParameterStub.SetupProperty(_ => _.DbType);
+    dataParameterStub.SetupProperty(_ => _.Value);
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformation);
+
+    var result = objectIDDataParameterDefinition.CreateDataParameter(commandStub.Object, "dummy", parameterValue);
+
+    Assert.That(result, Is.SameAs(dataParameterStub.Object));
+    Assert.That(result.ParameterName, Is.EqualTo("dummy"));
+    Assert.That(result.Value, Is.EqualTo(parameterValue));
+    Assert.That(result.DbType, Is.EqualTo(DbType.Guid));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithDBNullValue_SetsNameValueType ()
+  {
+    var commandStub = new Mock<IDbCommand>();
+    var dataParameterStub = new Mock<IDbDataParameter>();
+
+    commandStub
+        .Setup(_ => _.CreateParameter())
+        .Returns(dataParameterStub.Object);
+
+    var storageTypeInformation = new StorageTypeInformation(
+        typeof(Guid),
+        "test",
+        DbType.Guid,
+        true,
+        null,
+        typeof(Guid),
+        Mock.Of<TypeConverter>()
+    );
+
+    dataParameterStub.SetupProperty(_ => _.ParameterName);
+    dataParameterStub.SetupProperty(_ => _.DbType);
+    dataParameterStub.SetupProperty(_ => _.Value);
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformation);
+
+    var result = objectIDDataParameterDefinition.CreateDataParameter(commandStub.Object, "dummy", DBNull.Value);
+
+    Assert.That(result, Is.SameAs(dataParameterStub.Object));
+    Assert.That(result.ParameterName, Is.EqualTo("dummy"));
+    Assert.That(result.Value, Is.EqualTo(DBNull.Value));
+    Assert.That(result.DbType, Is.EqualTo(storageTypeInformation.StorageDbType));
+  }
+}
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
@@ -1,0 +1,147 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Data;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Infrastructure.ObjectIDStringSerialization;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Development.NUnit.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.Parameters;
+
+[TestFixture]
+public class SerializedObjectIDDataParameterDefinitionTest : StandardMappingTest
+{
+  [Test]
+  public void Initialize ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation();
+
+    var parameterDefinition = new SerializedObjectIDDataParameterDefinition(storageTypeInformation);
+
+    Assert.That(parameterDefinition.StorageTypeInformation, Is.SameAs(storageTypeInformation));
+  }
+
+  [Test]
+  public void Initialize_WithIncorrectStorageTypeInformation ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateIntStorageTypeInformation();
+
+    Assert.That(
+        () => new SerializedObjectIDDataParameterDefinition(storageTypeInformation),
+        Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+            $"Parameter 'storageTypeInformation.DotNetType' is a '{typeof(int)}', which cannot be assigned to type '{typeof(string)}'.",
+            "storageTypeInformation.DotNetType"));
+  }
+
+  [Test]
+  public void GetParameterValue_WithObjectID_ReturnsSerializedObjectID ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation(storageDbType: DbType.AnsiString, storageTypeLength: 255);
+
+    var objectID = DomainObjectIDs.Order1;
+    var expectedValue = ObjectIDStringSerializer.Instance.Serialize(objectID);
+
+    var serializedObjectIDDataParameterDefinition = new SerializedObjectIDDataParameterDefinition(storageTypeInformation);
+
+    var result = serializedObjectIDDataParameterDefinition.GetParameterValue(objectID);
+
+    Assert.That(result, Is.EqualTo(expectedValue));
+  }
+
+  [Test]
+  public void GetParameterValue_WithNull_ReturnsDBNull ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation();
+
+    var serializedObjectIDDataParameterDefinition = new SerializedObjectIDDataParameterDefinition(storageTypeInformation);
+
+    Assert.That(serializedObjectIDDataParameterDefinition.GetParameterValue(null), Is.EqualTo(DBNull.Value));
+  }
+
+  [Test]
+  public void GetParameterValue_WithTypeMismatchedValue_ThrowsArgumentException ()
+  {
+    var dummyValue = "Not an ObjectID";
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation();
+
+    var serializedObjectIDDataParameterDefinition = new SerializedObjectIDDataParameterDefinition(storageTypeInformation);
+
+    Assert.That(
+        () => serializedObjectIDDataParameterDefinition.GetParameterValue(dummyValue),
+        Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+            $"Parameter 'value' has type '{typeof(string)}' when type '{typeof(ObjectID)}' was expected.",
+            "value"));
+  }
+
+  [Test]
+  public void CreateDataParameter_SetsNameValueTypeSize_AnsiString255 ()
+  {
+    var commandStub = new Mock<IDbCommand>();
+    var dataParameterStub = new Mock<IDbDataParameter>();
+
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation(storageDbType: DbType.AnsiString, storageTypeLength: 255);
+
+    var objectID = DomainObjectIDs.Order1;
+    var testValue = ObjectIDStringSerializer.Instance.Serialize(objectID);
+
+    commandStub.Setup(_ => _.CreateParameter()).Returns(dataParameterStub.Object);
+
+    dataParameterStub.SetupProperty(_ => _.ParameterName);
+    dataParameterStub.SetupProperty(_ => _.DbType);
+    dataParameterStub.SetupProperty(_ => _.Value);
+    dataParameterStub.SetupProperty(_ => _.Size);
+
+    var serializedObjectIDDataParameterDefinition = new SerializedObjectIDDataParameterDefinition(storageTypeInformation);
+
+    var result = serializedObjectIDDataParameterDefinition.CreateDataParameter(commandStub.Object, "dummy", testValue);
+
+    Assert.That(result, Is.SameAs(dataParameterStub.Object));
+    Assert.That(result.ParameterName, Is.EqualTo("dummy"));
+    Assert.That(result.Value, Is.EqualTo(testValue));
+    Assert.That(result.DbType, Is.EqualTo(DbType.AnsiString));
+    Assert.That(result.Size, Is.EqualTo(255));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithDBNullValue_SetsNameValueTypeSize_AnsiString255 ()
+  {
+    var commandStub = new Mock<IDbCommand>();
+    var dataParameterStub = new Mock<IDbDataParameter>();
+
+    commandStub.Setup(_ => _.CreateParameter()).Returns(dataParameterStub.Object);
+
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation(storageDbType: DbType.AnsiString, storageTypeLength: 255);
+
+    dataParameterStub.SetupProperty(_ => _.ParameterName);
+    dataParameterStub.SetupProperty(_ => _.DbType);
+    dataParameterStub.SetupProperty(_ => _.Value);
+    dataParameterStub.SetupProperty(_ => _.Size);
+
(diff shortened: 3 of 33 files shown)


Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    var objectIDDataParameterDefinition = new ObjectIDDataParameterDefinition(storageTypeInformationStub.Object);
+
+    var result = objectIDDataParameterDefinition.GetParameterValue(dummyValue);
+
+    Assert.That(result, Is.EqualTo(DBNull.Value));
+  }
+
+  [Test]
+  public void GetParameterValue_WithInvalidValue_ThrowsArgumentException ()
```
MichaelKetting, line 70: `WithTypeMismatch` isntead of `Invalid`?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,178 @@
...
+
+
+  [Test]
+  public void CreateDataParameter_WithIntValue ()
+  {
+    var parameterValue = 42;
+    var commandMock = new Mock<IDbCommand>(MockBehavior.Strict);
+    var dataParameterMock = new Mock<IDbDataParameter>(MockBehavior.Strict);
+
+    var storageTypeInformation = new StorageTypeInformation(
```
MichaelKetting, line 94: could use objectmother if helpful

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,178 @@
...
+        DbType.Int32,
+        false,
+        null,
+        typeof(int),
+        new DefaultConverter(typeof(int)));
+
+    commandMock.Setup(mock => mock.CreateParameter()).Returns(dataParameterMock.Object).Verifiable();
+
+    dataParameterMock.SetupSet(mock => mock.DbType = storageTypeInformation.StorageDbType).Verifiable();
+    dataParameterMock.SetupSet(mock => mock.Value = parameterValue).Verifiable();
```
MichaelKetting, line 106: I like to use linebreaks for more complicated (or longer) mock setups. Inparticular also verifiables to make it easier discoverable.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,178 @@
...
+    var storageTypeInformation = new StorageTypeInformation(
+        typeof(int),
+        "test",
+        DbType.Int32,
+        false,
+        null,
+        typeof(int),
+        new DefaultConverter(typeof(int)));
+
+    commandMock.Setup(mock => mock.CreateParameter()).Returns(dataParameterMock.Object).Verifiable();
```
MichaelKetting, line 103: this one we don't need to verify. we test the result.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionTest.cs
MichaelKetting, line 106: for these, we actually want to test if the correct properties where set. I would suggest using verify calls in the assert block.

I do like uto set up my verify when doing method calls or mixed stuff. when having an object that simply gets proeprties assigned, the assert block feels more natural.

this will affect the other tests too, here and in SimpleDataParamDefTest.

MichaelKetting: Update: I have seen you're using strict mock for the parameter. okay, then you need to use this syntax. not 100% sure if it's the best way to test this / overkill. might very well be, though.

We also should consider that we will eventually change the parameter to a SqlDataParameter (at least for the SimpleDataParamDefTest) whch will change the way we have to do the assertions.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,170 @@
...
+    var storageTypeInformation = new StorageTypeInformation(
+        typeof(string),
+        "test",
+        DbType.AnsiString,
+        false,
+        -1,
+        typeof(string),
+        new DefaultConverter(typeof(string)));
+
+    var testValue = $"Order|{Guid.NewGuid()}|System.Guid";
```
MichaelKetting, line 129: should we use `ObjectIDStringSerializer.Instance.Serialize` instead?
(or `ObectID.ToString()`)
Having the serialization format hardcoded here feels like duplication

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/IDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+  /// Defines the API for creating <see cref="IDataParameterDefinition"/> instances.
+  /// </summary>
+  public interface IDataParameterDefinitionFactory
+  {
+    /// <summary>
+    /// Creates a <see cref="IDataParameterDefinition"/> for the given <paramref name="queryParameter"/>.
+    /// </summary>
+    IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter);
+  }
+
```
MichaelKetting, line 28:
```suggestion
```

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SimpleDataParameterDefinition.cs
```diff
@@ -0,0 +1,59 @@
...
+  public SimpleDataParameterDefinition (IStorageTypeInformation storageTypeInformation)
+  {
+    ArgumentUtility.CheckNotNull(nameof(storageTypeInformation), storageTypeInformation);
+
+    StorageTypeInformation = storageTypeInformation;
+  }
+
+  public object GetParameterValue (object? parameterObject)
+  {
+    ArgumentUtility.CheckType(nameof(parameterObject), parameterObject, StorageTypeInformation.DotNetType);
```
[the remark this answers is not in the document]
MichaelKetting: Note: Since some implementations use a fixed type here, it might very well be that ArgEx is correct due to symmetry.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinition.cs
```diff
@@ -0,0 +1,62 @@
...
+
+  public IDbDataParameter CreateDataParameter (IDbCommand command, object parameterValue)
+  {
+    ArgumentUtility.CheckNotNull(nameof(command), command);
+    ArgumentUtility.CheckNotNull(nameof(parameterValue), parameterValue);
+
+    var parameter = command.CreateParameter();
+    parameter.Value = parameterValue;
+    parameter.DbType = DbType.AnsiString;
+    parameter.Size = SqlStorageTypeInformationProvider.StorageTypeLengthRepresentingMax;
```
MichaelKetting, line 59: In StorageTypeInfo, we are using varchar(255) for serialized objectid

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlFulltextDataParameterDefinitionDecorator.cs
```diff
@@ -0,0 +1,63 @@
...
+  public object GetParameterValue (object? parameterObject)
+  {
+    return InnerDataParameterDefinition.GetParameterValue(parameterObject);
+  }
+
+  public IDbDataParameter CreateDataParameter (IDbCommand command, object parameterValue)
+  {
+    var parameter = InnerDataParameterDefinition.CreateDataParameter(command, parameterValue);
+
+    if (parameterValue is string stringValue)
```
MichaelKetting, line 48: in the original codebase, we also do this for char[]

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlFulltextDataParameterDefinitionDecoratorTest.cs
```diff
@@ -0,0 +1,100 @@
...
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Parameters;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.Parameters;
+
+[TestFixture]
+public class SqlFulltextDataParameterDefinitionDecoratorTest
+{
+  [Test]
+  public void MaxSize_AnsiString_SizeSetTo8000 ()
```
MichaelKetting, line 29: we should also have a test that tests larger than 8k and 4k length.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlDataParameterDefinitionFactoryDecoratorTests.cs
```diff
@@ -0,0 +1,44 @@
...
+  {
+    var queryParameter = new QueryParameter("dummy", "Dummy");
+    var parameterDefinitionStub = new Mock<IDataParameterDefinition>();
+
+    var factoryMock = new Mock<IDataParameterDefinitionFactory>(MockBehavior.Strict);
+    factoryMock.Setup(_ => _.CreateDataParameterDefinition(queryParameter)).Returns(parameterDefinitionStub.Object);
+
+    var factoryDecorator = new SqlDataParameterDefinitionFactoryDecorator(factoryMock.Object);
+
+    var result = factoryDecorator.CreateDataParameterDefinition(queryParameter);
```
MichaelKetting, line 40: shouldn't we only add the sqldecorator for type string and char[]?

Remotion/Data/DomainObjects/Queries/QueryParameterWithDataDefinition.cs
```diff
@@ -15,36 +15,27 @@
...
 using Remotion.Utilities;
 
 namespace Remotion.Data.DomainObjects.Queries
 {
   /// <summary>
-  /// <see cref="QueryParameterWithType"/> associated a <see cref="QueryParameter"/> with a <see cref="StorageTypeInformation"/>.
+  /// <see cref="QueryParameterWithDataDefinition"/> associated a <see cref="QueryParameter"/> with a <see cref="DataParameterDefinition"/>.
   /// </summary>
-  public struct QueryParameterWithType
+  public struct QueryParameterWithDataDefinition
```
MichaelKetting, line 26: not sure about the name. It feels clunky and not inutive.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/DataParameterDefinitionFactoryTest.cs
```diff
@@ -0,0 +1,140 @@
...
+      var storageTypeInformationProviderStub = new Mock<IStorageTypeInformationProvider>();
+      storageTypeInformationProviderStub.Setup(_ => _.IsSupported(null)).Returns(false);
+
+      var providerDefinition = new TestableStorageProviderDefinition("bla", Mock.Of<IStorageObjectFactory>());
+
+      var queryParameterDefinitionFactory = new DataParameterDefinitionFactory(providerDefinition, storageTypeInformationProviderStub.Object, StorageSettings);
+
+      Assert.That(
+          () => queryParameterDefinitionFactory.CreateDataParameterDefinition(new QueryParameter("dummy", null)),
+          Throws.InstanceOf<NotSupportedException>().With.Message.EqualTo($"The type '(unknown)' of the given value '(null)' cannot be used as a query parameter."));
```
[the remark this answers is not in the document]
martin-herwich: It's 166 characters. My margin is set to 180, everything is within that limit.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/DataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+      _simpleDataParameterDefinitionFactory = new SimpleDataParameterDefinitionFactory(StorageTypeInformationProvider);
+      _objectIDDataParameterDefinitionFactory = new ObjectIDDataParameterDefinitionFactory(StorageProviderDefinition, StorageTypeInformationProvider, StorageSettings);
+    }
+
+    public IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter)
+    {
+      ArgumentUtility.CheckNotNull(nameof(queryParameter), queryParameter);
+
+      var value = queryParameter.Value;
+      var objectID = value as ObjectID;
```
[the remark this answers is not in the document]
martin-herwich: While `SimpleDataParameterDefinitionFactory` does support null values, that is an implementation detail of that type. If we check for `null` early in order to return ` _simpleDataParameterDefinitionFactory.CreateDataParameterDefinition(null)`, we assume that this works.
If we don't want to assume this, then we are back to checking `_simpleDataParameterDefinitionFactory.IsSupported(null)` and throwing a `NotSupportedException`. This also assumes that `SimpleDataParameterDefinitionFactory` is the only implementation that can handle `null` values.
I don't like it.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
```diff
@@ -38,6 +38,14 @@ public SqlStorageTypeInformationProvider ()
     {
     }
 
+    public bool IsSupported (Type? type)
```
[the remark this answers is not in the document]
martin-herwich: The problem with `typeof(object)` is that it should not be supported, whereas `null` should be.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SimpleDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+
+/// <summary>
+/// Can create <see cref="IDataParameterDefinition"/> instances that can handle simple (atomic) parameter values.
+/// </summary>
+public class SimpleDataParameterDefinitionFactory
```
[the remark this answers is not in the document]
martin-herwich: I mirrored the design of `DataStoragePropertyDefinitionFactory` (minus the interfaces for the partial factory types and the `ObjectIDWithoutClassID` implementation).
Idea: make `SimpleDataParameterDefinitionFactory` and `ObjectIDDataParameterDefinitionFactory` implement a new interface `ISingleValueDataParameterDefinitionFactory`, inject a collection of `ISingleValueDataParameterDefinitionFactory` into `DataStoragePropertyDefinitionFactory` and use them in a chain of responsibility in `CreateStoragePropertyDefinition`.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/DataParameterDefinitionFactory.cs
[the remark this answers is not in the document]
MichaelKetting: ah, yeah, I kind of spaced and only focused on the elvis operator usages. forget it. first we need to figure out IsSupported in the other comment.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SimpleDataParameterDefinitionFactory.cs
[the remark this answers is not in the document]
MichaelKetting: Ah, yes. Hmm... further consideration: make just IDataStoragePropertyDefinitionFactory types and the chain of responsibility as decorators. so, each type will either handle the QueryParameter or pass it on. 

Having two interfaces, once for orchestration of the chain and one as the chain items feels overkill for the design? Only advantage is that the chain would be implemented as an iteration instead of a pseduo-recursive call. 

I like that we could simply use a decorator chain here. Unless it's ugly? thoughts...?

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/DataParameterDefinitionFactory.cs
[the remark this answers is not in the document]
martin-herwich: As we talked about, this is going to be redesigned completely.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SimpleDataParameterDefinitionFactory.cs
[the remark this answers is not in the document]
martin-herwich: As we talked about, this is going to be redesigned into a decorator-style chain of responsibility.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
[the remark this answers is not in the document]
martin-herwich: As we talked about, this is going to be redesigned, removing `IsSupported` completely.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinition.cs
```diff
@@ -0,0 +1,66 @@
...
+    ValueStorageTypeInformation = storageTypeInformation;
+  }
+
+  /// <inheritdoc />
+  public object GetParameterValue (object? parameterObject)
+  {
+    var objectID = ArgumentUtility.CheckType<ObjectID>(nameof(parameterObject), parameterObject);
+    var value = objectID?.Value;
+    if (value != null && !ValueStorageTypeInformation.DotNetType.IsInstanceOfType(value))
+      throw ArgumentUtility.CreateArgumentTypeException("parameterObject.Value", value.GetType(), ValueStorageTypeInformation.DotNetType);
```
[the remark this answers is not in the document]
martin-herwich: As we discussed, this should remain an `ArgumentException`.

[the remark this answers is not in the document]
Frazerus: Checktype does not support nullability in the way we need here, so we need to check for that extra.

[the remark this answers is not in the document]
MichaelKetting: Am not sure if there's not an expectation mismatch. We need a check that ignores the test of value is null and otherwise, tests if if Value is assignable to the expected type.

This should be covered by 
public static object? CheckType ([InvokerParameterName] string argumentName, [NoEnumeration] object? actualValue, Type expectedType)

I see that the question is whether we want to handle StorageTypeInformation.DotNetType being an Int32, but the value being null and if this should actually be a DotNetType Nullable<Int32> instead.  From reading the code (mistake is possible) I think that DotNetType would be Nullable<int32> in such cases. Or it would be a correct argumentexception otherwise.

Please check/verify.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlDataParameterDefinitionFactoryDecoratorTests.cs
martin-herwich: No, because we need the factory-decorator later. However, in the `SqlDataParameterDefinitionFactoryDecorator`, we can only add the Fulltext-decorator for `string or char[]`

MichaelKetting: Oh, yeah, my bad. That's what I meant to write :)

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SimpleDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+
+  public SimpleDataParameterDefinitionFactory (IStorageTypeInformationProvider storageTypeInformationProvider)
+  {
+    ArgumentUtility.CheckNotNull(nameof(storageTypeInformationProvider), storageTypeInformationProvider);
+
+    StorageTypeInformationProvider = storageTypeInformationProvider;
+  }
+
+  public IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter)
+  {
```
MichaelKetting, line 39: argumentcheck

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinition.cs
```diff
@@ -0,0 +1,66 @@
...
+public class SerializedObjectIDDataParameterDefinition : IDataParameterDefinition
+{
+  public IStorageTypeInformation StorageTypeInformation { get; }
+
+  public SerializedObjectIDDataParameterDefinition (IStorageTypeInformation storageTypeInformation)
+  {
+    ArgumentUtility.CheckNotNull(nameof(storageTypeInformation), storageTypeInformation);
+
+    if (storageTypeInformation.DotNetType != typeof(string))
+      throw ArgumentUtility.CreateArgumentTypeException("storageTypeInformation.DotNetType", storageTypeInformation.DotNetType, typeof(string));
```
MichaelKetting, line 40:
```suggestion
    ArgumentUtility.CheckTypeIsAssignableFrom("storageTypeInformation.DotNetType", storageTypeInformation.DotNetType, typeof(string));
```

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+    return new SerializedObjectIDDataParameterDefinition(storageTypeInfo);
+  }
+
+  public IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter)
+  {
+    var objectID = queryParameter.Value as ObjectID;
+    if (objectID != null)
+    {
+      return CreateDataParameterDefinition(objectID);
+    }
```
MichaelKetting, line 79:
```suggestion
    if (queryParameter.Value is ObjectID objectID)
      return CreateDataParameterDefinition(objectID);
```

In case you want to use the new, more compact syntax.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+  public IDataParameterDefinition CreateDataParameterDefinition (ObjectID objectID)
+  {
+    ArgumentUtility.CheckNotNull(nameof(objectID), objectID);
+
+    var relatedStorageProviderDefinition = StorageSettings.GetStorageProviderDefinition(objectID.ClassDefinition);
+
+    if (StorageProviderDefinition != relatedStorageProviderDefinition)
+      return CreateCrossProviderRelationStoragePropertyDefinition();
+    else
+     return CreateSameProviderRelationStoragePropertyDefinition();
```
MichaelKetting, line 58:
```suggestion
    if (StorageProviderDefinition == relatedStorageProviderDefinition)
      return CreateSameProviderRelationStoragePropertyDefinition();
    else
      return CreateCrossProviderRelationStoragePropertyDefinition();
```
I prefer to use the positive condition unless the negative one is a guard condition or there is a realy awkward imbalance. "equals" is easier to reason over than "not equals" because you need to put to details on the mental stack :)

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/IStorageTypeInformationProvider.cs
```diff
@@ -26,6 +26,7 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building
   /// </summary>
   public interface IStorageTypeInformationProvider
   {
+
```
MichaelKetting, line 29:
```suggestion
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,145 @@
...
+  [Test]
+  public void Initialize_WithIncorrectStorageTypeInformation ()
+  {
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateIntStorageTypeInformation();
+
+    Assert.That(
+        () => new SerializedObjectIDDataParameterDefinition(storageTypeInformation),
+        Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+            string.Format("Parameter '{0}' has type '{2}' when type '{1}' was expected.", "storageTypeInformation.DotNetType", typeof(string), typeof(int)),
+            "storageTypeInformation.DotNetType"));
```
MichaelKetting, line 50: Please inline the values as much as possible. I like to not use string-format in unit test assertions since they hide errors, e.g. if you flip parameter order. If sensible, using interpolated strings is now an option, e.g. when the type name is to complicated. Here, with the text form being `System.String` and `System.Int32`, I'm not sure even that's warrented.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,145 @@
...
+  {
+    var commandMock = new Mock<IDbCommand>(MockBehavior.Strict);
+    var dataParameterMock = new Mock<IDbDataParameter>(MockBehavior.Strict);
+
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation(storageDbType: DbType.AnsiString, storageTypeLength: 255);
+
+    var objectID = DomainObjectIDs.Order1;
+    var testValue = ObjectIDStringSerializer.Instance.Serialize(objectID);
+
+    commandMock.Setup(mock => mock.CreateParameter()).Returns(dataParameterMock.Object).Verifiable();
```
MichaelKetting, line 104:
```suggestion
    commandMock.Setup(mock => mock.CreateParameter()).Returns(dataParameterMock.Object);
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,145 @@
...
+        () => serializedObjectIDDataParameterDefinition.GetParameterValue(dummyValue),
+        Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+            $"Parameter 'value' has type '{typeof(string)}' when type '{typeof(ObjectID)}' was expected.",
+            "value"));
+  }
+
+  [Test]
+  public void CreateDataParameter ()
+  {
+    var commandMock = new Mock<IDbCommand>(MockBehavior.Strict);
```
MichaelKetting, line 96:
```suggestion
    var commandMock = new Mock<IDbCommand>();
```
Do we need this as a strict mock or could it simply be a stub?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,145 @@
...
+    Assert.That(result, Is.SameAs(dataParameterMock.Object));
+    Assert.That(result.DbType, Is.EqualTo(DbType.AnsiString));
+    Assert.That(result.Value, Is.EqualTo(testValue));
+    Assert.That(result.Size, Is.EqualTo(255));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithNullValue ()
+  {
+    var commandMock = new Mock<IDbCommand>(MockBehavior.Strict);
```
MichaelKetting, line 123: same as above, not sure if 'strict' is necessary for the commandMock or if we can simply make it a stub (rename!)

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,145 @@
...
+    var dataParameterMock = new Mock<IDbDataParameter>(MockBehavior.Strict);
+
+    commandMock
+        .Setup(mock => mock.CreateParameter())
+        .Returns(dataParameterMock.Object);
```
MichaelKetting, line 128: If we a) drop the `Verifiable` on the previous test case and b) thus keep the statement single-line, we can to that here, too. simple setup and simple return -> no fuss. 

Noticed / mentioned because the file is new and the style differs be two tests.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactoryTest.cs
```diff
@@ -0,0 +1,141 @@
...
+        storageTypeInformationProviderStub.Object,
+        StorageSettings,
+        _nextDataParameterDefinitionFactoryMock.Object);
+
+    var result = factory.CreateDataParameterDefinition(otherObjectID);
+
+    Assert.That(result, Is.InstanceOf<ObjectIDDataParameterDefinition>());
+    Assert.That(result.As<ObjectIDDataParameterDefinition>().ValueStorageTypeInformation, Is.SameAs(storageTypeInformationStub.Object));
+
+    _nextDataParameterDefinitionFactoryMock.Verify(_ => _.CreateDataParameterDefinition(It.IsAny<QueryParameter>()), Times.Never());
```
MichaelKetting, line 83: just a note / stylistic stuff: I like to use a setup with a `Throws` of an AssertionException instead of verify with times-never. 

Reasoning: it fails immediately with a nice stacktrace and it's easier to discover when reading the setup. (it's not an afterthought).

`.Throws(new AssertionException("Should not be called."))`

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlDataParameterDefinitionFactoryDecoratorTests.cs
```diff
@@ -0,0 +1,44 @@
...
+public class SqlDataParameterDefinitionFactoryDecoratorTests
+{
+  [Test]
+  public void DecoratesWith_SqlFulltextDataParameterDefinitionDecorator ()
+  {
+    var queryParameter = new QueryParameter("dummy", "Dummy");
+    var parameterDefinitionStub = new Mock<IDataParameterDefinition>();
+
+    var factoryMock = new Mock<IDataParameterDefinitionFactory>(MockBehavior.Strict);
+    factoryMock.Setup(_ => _.CreateDataParameterDefinition(queryParameter)).Returns(parameterDefinitionStub.Object);
```
MichaelKetting, line 36: strict-mock vs stub (see note on other tests)

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlFulltextDataParameterDefinitionDecoratorTest.cs
```diff
@@ -0,0 +1,172 @@
...
+  {
+    yield return new string('x', 8001);
+    yield return new string('x', 8001).ToCharArray();
+  }
+
+  private static IEnumerable<object> GetValuesLength4001 ()
+  {
+    yield return new string('x', 4001);
+    yield return new string('x', 4001).ToCharArray();
+  }
```
MichaelKetting, line 71: please move to top or bottom (or leave right in front of the individual tests. right now it's randomly in the middle.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlDataParameterDefinitionFactoryDecoratorTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using Remotion.Data.DomainObjects.Queries;
+using Remotion.Development.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.Parameters;
+
+[TestFixture]
+public class SqlDataParameterDefinitionFactoryDecoratorTests
+{
+  [Test]
+  public void DecoratesWith_SqlFulltextDataParameterDefinitionDecorator ()
```
MichaelKetting, line 30: please add second test for `char[]` and one for other types.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlDataParameterDefinitionFactoryDecorator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+  public SqlDataParameterDefinitionFactoryDecorator (IDataParameterDefinitionFactory innerFactory)
+  {
+    ArgumentUtility.CheckNotNull(nameof(innerFactory), innerFactory);
+
+    _innerFactory = innerFactory;
+  }
+
+  public IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter)
+  {
```
MichaelKetting, line 38: argnull-check

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+    ArgumentUtility.CheckNotNull(nameof(storageTypeInformationProvider), storageTypeInformationProvider);
+    ArgumentUtility.CheckNotNull(nameof(storageSettings), storageSettings);
+    NextDataParameterDefinitionFactory = nextDataParameterDefinitionFactory;
+
+    StorageSettings = storageSettings;
+    StorageProviderDefinition = storageProviderDefinition;
+    StorageTypeInformationProvider = storageTypeInformationProvider;
+  }
+
+  public IDataParameterDefinition CreateDataParameterDefinition (ObjectID objectID)
```
MichaelKetting, line 49: I'm confused about this method. it's not part of the interface, so how/where is it used? If it's just for testing, I think the public api is simple enough to allow testing without much overhead.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  private IDataParameterDefinition CreateCrossProviderRelationStoragePropertyDefinition ()
+  {
+    var storageTypeInfo = StorageTypeInformationProvider.GetStorageTypeForSerializedObjectID(true);
+    return new SerializedObjectIDDataParameterDefinition(storageTypeInfo);
+  }
+
+  public IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter)
+  {
```
MichaelKetting, line 74: argcheck please

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+    return new ObjectIDDataParameterDefinition(storageTypeInformation);
+  }
+
+  private IDataParameterDefinition CreateCrossProviderRelationStoragePropertyDefinition ()
+  {
+    var storageTypeInfo = StorageTypeInformationProvider.GetStorageTypeForSerializedObjectID(true);
+    return new SerializedObjectIDDataParameterDefinition(storageTypeInfo);
+  }
+
+  public IDataParameterDefinition CreateDataParameterDefinition (QueryParameter queryParameter)
```
MichaelKetting, line 73: please move upwards in the file

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -564,6 +572,16 @@ protected virtual IDataContainerValidator CreateDataContainerValidator (RdbmsPro
       return new DataStoragePropertyDefinitionFactory(valueStoragePropertyDefinitionFactory, relationStoragePropertyDefinitionFactory);
     }
 
+    protected virtual IDataParameterDefinitionFactory CreateDataParameterDefinitionFactory (
+        StorageProviderDefinition storageProviderDefinition,
+        IStorageTypeInformationProvider storageTypeInformationProvider)
+    {
```
MichaelKetting, line 578: argchecks please

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/QueryCommandFactory.cs
```diff
@@ -92,39 +97,18 @@ public IDataStoragePropertyDefinitionFactory DataStoragePropertyDefinitionFactor
...
-            "The query parameter '{0}' is mapped to {1} database-level values. Only values that map to a single database-level value can be used "
-            + "as query parameters.",
-            parameter.Name,
-            columnValues.Length);
-        throw new InvalidOperationException(message);
-      }
-
-      var adaptedParameter = new QueryParameter(parameter.Name, columnValues[0].Value, parameter.ParameterType);
-      return new QueryParameterWithType(adaptedParameter, columnValues[0].Column.StorageTypeInfo);
+      var dataParameterDefinition = _dataParameterDefinitionFactory.CreateDataParameterDefinition(parameter);
```
MichaelKetting, line 102: argcheck please (was missing before)

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -48,23 +48,24 @@ public override IDbCommand Create (IRdbmsProviderCommandExecutionContext command
...
-          statement = statement.Replace(parameterWithType.QueryParameter.Name, queryParameterValueAsString);
+          statement = statement.Replace(parameterWithDefinition.QueryParameter.Name, queryParameterValueAsString);
         }
         else
         {
-          var parameter = parameterWithType.StorageTypeInformation.CreateDataParameter(command, queryParameterValue);
-          parameter.ParameterName = parameterWithType.QueryParameter.Name;
+          var dataParameterValue = parameterWithDefinition.DataParameterDefinition.GetParameterValue(queryParameterValue);
+          var parameter = parameterWithDefinition.DataParameterDefinition.CreateDataParameter(command, dataParameterValue);
+          parameter.ParameterName = parameterWithDefinition.QueryParameter.Name;
```
MichaelKetting, line 68: it feels weird that CreateDataParameter isn't setting the name since it already has all the relevant information. please check if this should be changed. One reason that might argue against it would be that we have to duplicate the setting of the name-property into all implementations. since we do need to set value etc, too, might be with it. 

Another reason against it might be that it clashes when we do structured data or collection parameters.

If we don't do it internally, I wonder if this might warrent an inline comment to explain why we don't do it since it does feel like it should be handled inside CreateDataParameter.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/QueryCommandFactoryTest.cs
```diff
@@ -194,75 +157,57 @@ public void CreateForCustomQuery ()
...
+              new[]
+              {
+                      new QueryParameterWithDataParameterDefinition(
+                              new QueryParameter(_queryParameter1.Name, DomainObjectIDs.Order1, _queryParameter1.ParameterType),
+                              _dataParameterDefinition1),
+                      new QueryParameterWithDataParameterDefinition(_queryParameter2, _dataParameterDefinition2),
+                      new QueryParameterWithDataParameterDefinition(
+                              new QueryParameter(_queryParameter3.Name, DomainObjectIDs.Official1, _queryParameter3.ParameterType),
+                              _dataParameterDefinition3)
+              };
```
MichaelKetting, line 184: I think the indent is broken here. should only be two indents / four whitespaces , it got reformatted to double.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
martin-herwich: It makes sense to do it within `IDataParameterDefinition.CreateDataParameter()`, so I implemented it in all 3 basic implementations.
It will not clash with structured data types or collections, because it is still a single parameter with one name; we just have to implement setting the name a fourth time.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/QueryCommandFactoryTest.cs
martin-herwich: ReSharper auto-detects 4 spaces of indentation instead of 2 - no idea why. After disabling auto-detection, reformatting did the indentation correctly.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
MichaelKetting:
```
var dataParameterValue = parameterWithDefinition.DataParameterDefinition.GetParameterValue(queryParameterValue);
var parameter = parameterWithDefinition.DataParameterDefinition.CreateDataParameter(command, dataParameterValue);
parameter.ParameterName = parameterWithDefinition.QueryParameter.Name;
```
I realized a misunderstanding on my part: I read the reviewed code as the DataParameterDefinition already owning the QueryParameter object and thus the name. Which isn't the case. no we have a new method argument and the duplication. I'm okay with that if you say it's cleaner / more helpful than what we had before. Am just mentioning it because it's a review-mistake.

Am commenting/unresolving this while checking the rest. Assuming there won't be any other details I'll approve it as is and leave the final decision on the dataParameterName up to you.

Remotion/Data/DomainObjects/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,78 @@
...
+        : NextDataParameterDefinitionFactory.CreateDataParameterDefinition(queryParameter);
+  }
+
+  private IDataParameterDefinition CreateDataParameterDefinition (ObjectID objectID)
+  {
+    var relatedStorageProviderDefinition = StorageSettings.GetStorageProviderDefinition(objectID.ClassDefinition);
+    return StorageProviderDefinition == relatedStorageProviderDefinition
+        ? CreateSameProviderRelationStoragePropertyDefinition()
+        : CreateCrossProviderRelationStoragePropertyDefinition();
+  }
```
MichaelKetting, line 65: Stylistic bit:

I prefer using IF/ELSE instead of ternary operator for regular decisions on code. the main advantage is that most code coverage tools only work on a statement level so we can't check if both branches are covered. 

The other aspect is that ternary is meant for low-key decisions / simple values, not cases where control flow should be a prominent part of the system. I'm aawre that pattern matching, switch expressions and Linq dilute this concept further and further.

So, bascially: if you can put it in a single line and it's just a reult value decision (0 or 1 for instance), ternerary is okay. Using it for control flow, I try to stick to IF/ELSE because it adds some weight and space to the whole thing. 

But yeah, it's possible that missing code coverage is the main reason I don't like to use ternary for testable controlflow.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    var objectID = DomainObjectIDs.Order1;
+    var testValue = ObjectIDStringSerializer.Instance.Serialize(objectID);
+
+    commandStub.Setup(mock => mock.CreateParameter()).Returns(dataParameterMock.Object);
+
+    dataParameterMock.SetupProperty(mock => mock.ParameterName);
+    dataParameterMock.SetupProperty(mock => mock.DbType);
+    dataParameterMock.SetupProperty(mock => mock.Value);
+    dataParameterMock.SetupProperty(mock => mock.Size);
```
MichaelKetting, line 109: stylistic input: for new code, we use `_` instead of `mock` or `stub` in expressions to make the varable less prominent.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/SerializedObjectIDDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,147 @@
...
+        Throws.InstanceOf<ArgumentException>().With.ArgumentExceptionMessageEqualTo(
+            $"Parameter 'value' has type '{typeof(string)}' when type '{typeof(ObjectID)}' was expected.",
+            "value"));
+  }
+
+  [Test]
+  public void CreateDataParameter ()
+  {
+    var commandStub = new Mock<IDbCommand>();
+    var dataParameterMock = new Mock<IDbDataParameter>();
```
MichaelKetting, line 97: since we don't do verify / strict anymore, it should be dataParameterStub instead of dataParameterMock.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Parameters/ObjectIDDataParameterDefinitionFactoryTest.cs
```diff
@@ -0,0 +1,145 @@
...
+        .Returns(storageTypeInformationStub.Object);
+
+    var factory = new ObjectIDDataParameterDefinitionFactory(
+        storageProviderDefinition,
+        storageTypeInformationProviderStub.Object,
+        StorageSettings,
+        _nextDataParameterDefinitionFactoryMock.Object);
+
+    _nextDataParameterDefinitionFactoryMock
+        .Setup(mock => mock.CreateDataParameterDefinition(It.IsAny<QueryParameter>()))
```
MichaelKetting, line 79: stylistic mix and mingle of `mock` and `_` naming is something I try to avoid at least within the same method.

MichaelKetting (approved):
Some stylistic input. am done reviewing :)

PR #1113, opened 2024-04-25
Feature/rm 9187 add support for collection parameters to query command factory

Frazerus:
https://re-motion.atlassian.net/browse/RM-9187

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
@@ -20,6 +20,7 @@ using System.Collections;
 using System.Collections.Generic;
 using Moq;
 using NUnit.Framework;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Parameters;
@@ -32,40 +33,41 @@ namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.Para
 [TestFixture]
 public class SqlTableValuedDataParameterDefinitionFactoryTests
 {
-  [Test]
-  public void CreateParameterDefinition_WithCollectionOfInt_ReturnsDefinitionForInt ()
+  private static IEnumerable<(object, Type, bool, IStorageTypeInformation)> GetTestCasesForCreateParameterDefinition ()
   {
-    var nextFactoryStub = new Mock<IDataParameterDefinitionFactory>();
-
-    var storageTypeInformation = StorageTypeInformationObjectMother.CreateIntStorageTypeInformation();
-    var storageTypeInformationProviderStub = new Mock<IStorageTypeInformationProvider>();
-    storageTypeInformationProviderStub.Setup(_ => _.GetStorageType(typeof(int))).Returns(storageTypeInformation);
-
-    var factory = new SqlTableValuedDataParameterDefinitionFactory(storageTypeInformationProviderStub.Object, nextFactoryStub.Object);
-
-    var queryParameter = new QueryParameter("IntArray", new[] { 1, 2, 3, 4 });
-    var result = factory.CreateDataParameterDefinition(queryParameter);
-
-    Assert.That(result, Is.InstanceOf<SqlTableValuedDataParameterDefinition>());
-    Assert.That(result.As<SqlTableValuedDataParameterDefinition>().StorageTypeInformation, Is.SameAs(storageTypeInformation));
+    yield return (Array.Empty<int>(), typeof(int), false, StorageTypeInformationObjectMother.CreateIntStorageTypeInformation());
+    yield return (Array.AsReadOnly(Array.Empty<int>()), typeof(int), false, StorageTypeInformationObjectMother.CreateIntStorageTypeInformation());
+    yield return (new ArrayList(), typeof(object), false, StorageTypeInformationObjectMother.CreateIntStorageTypeInformation());
+    yield return (new List<int>(), typeof(int), false, StorageTypeInformationObjectMother.CreateIntStorageTypeInformation());
+    yield return (Mock.Of<ICollection>(), typeof(object), false, StorageTypeInformationObjectMother.CreateStorageTypeInformation());
+    yield return (Mock.Of<ICollection<DateTime>>(), typeof(DateTime), false, StorageTypeInformationObjectMother.CreateDateTimeStorageTypeInformation());
+    yield return (Mock.Of<IReadOnlyCollection<Guid>>(), typeof(Guid), false, StorageTypeInformationObjectMother.CreateUniqueIdentifierStorageTypeInformation());
+    yield return (Mock.Of<ISet<string>>(), typeof(string), true, StorageTypeInformationObjectMother.CreateVarchar100StorageTypeInformation());
+#if NET5_0_OR_GREATER
+    yield return (Mock.Of<IReadOnlySet<int>>(), typeof(int), true, StorageTypeInformationObjectMother.CreateIntStorageTypeInformation());
+#endif
   }
 
   [Test]
-  public void CreateParameterDefinition_WithCollection_ReturnsDefinitionForObject ()
+  [TestCaseSource(nameof(GetTestCasesForCreateParameterDefinition))]
+  public void CreateParameterDefinition ((object, Type, bool, IStorageTypeInformation) testCase)
   {
-    var nextFactoryStub = new Mock<IDataParameterDefinitionFactory>();
+    var value = testCase.Item1;
+    var dotNetType = testCase.Item2;
+    var isDistinct = testCase.Item3;
+    var storageTypeInformation = testCase.Item4;
 
-    var storageTypeInformation = StorageTypeInformationObjectMother.CreateStorageTypeInformation();
     var storageTypeInformationProviderStub = new Mock<IStorageTypeInformationProvider>();
-    storageTypeInformationProviderStub.Setup(_ => _.GetStorageType(typeof(object))).Returns(storageTypeInformation);
+    storageTypeInformationProviderStub.Setup(_ => _.GetStorageType(dotNetType)).Returns(storageTypeInformation);
 
+    var nextFactoryStub = new Mock<IDataParameterDefinitionFactory>();
     var factory = new SqlTableValuedDataParameterDefinitionFactory(storageTypeInformationProviderStub.Object, nextFactoryStub.Object);
 
-    var collection = new Mock<ICollection>();
-    var queryParameter = new QueryParameter("Collection", collection.Object);
+    var queryParameter = new QueryParameter("Dummy", value);
     var result = factory.CreateDataParameterDefinition(queryParameter);
 
     Assert.That(result, Is.InstanceOf<SqlTableValuedDataParameterDefinition>());
+    Assert.That(result.As<SqlTableValuedDataParameterDefinition>().IsDistinct, Is.EqualTo(isDistinct));
     Assert.That(result.As<SqlTableValuedDataParameterDefinition>().StorageTypeInformation, Is.SameAs(storageTypeInformation));
   }
 
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
@@ -1,0 +1,98 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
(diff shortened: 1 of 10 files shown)


Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Development.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.Parameters;
+
+[TestFixture]
+public class SqlTableValuedDataParameterDefinitionFactoryTests
+{
+  [Test]
+  public void CreateParameterDefinition_WithCollectionOfInt_ReturnsDefinitionForInt ()
```
MichaelKetting, line 36: Naming
```suggestion
  public void CreateParameterDefinition_WithArray_ReturnsDefinitionForItemType ()
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Parameters;
+using Remotion.Data.DomainObjects.Queries;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Development.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.Parameters;
+
+[TestFixture]
+public class SqlTableValuedDataParameterDefinitionFactoryTests
```
MichaelKetting, line 33: I'm missing the tests for ISet<T> and IReadOnlySet<T> and IReadOnlyCollection<T> and ICollection<T>.

The existing test for Array should remain.

I wonder if we should have integration tests for some of the derived collection types.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinition.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    var convertedValues = new ArrayList();
+    foreach (var item in enumerable)
+    {
+      if (item != null && !StorageTypeInformation.DotNetType.IsInstanceOfType(item))
+        throw ArgumentUtility.CreateArgumentItemTypeException(nameof(value), convertedValues.Count, StorageTypeInformation.DotNetType, item.GetType());
+
+      var convertedValue = StorageTypeInformation.ConvertToStorageType(item);
+      convertedValues.Add(convertedValue);
+    }
```
MichaelKetting, line 64: I don't think we should do a copy here. we can simply iterate and perform `tvpValue.AddRecord` instead.

We have use cases with thousands of rows in the parameter and doing two copies is more expensive than doing one copy.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    var sqlParameter = dataParameter.As<SqlParameter>();
+    Assert.That(sqlParameter.DbType, Is.EqualTo(DbType.Object));
+    Assert.That(sqlParameter.SqlDbType, Is.EqualTo(SqlDbType.Structured));
+    Assert.That(sqlParameter.TypeName, Is.EqualTo(tvpValue.TableTypeName));
+    Assert.That(sqlParameter.Value, Is.SameAs(tvpValue));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithEmptyTableValuedParameter_SetsParameterValueToDbNull ()
```
MichaelKetting, line 120:
```suggestion
  public void CreateDataParameter_WithEmptyTableValuedParameter_SetsParameterValueToDBNull ()
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinition.cs
```diff
@@ -0,0 +1,107 @@
...
+  /// <see cref="SqlTableValuedParameterValue"/>.
+  /// </summary>
+  public IDbDataParameter CreateDataParameter (IDbCommand command, string parameterName, object parameterValue)
+  {
+    ArgumentUtility.CheckNotNull(nameof(command), command);
+    ArgumentUtility.CheckNotNullOrEmpty(nameof(parameterName), parameterName);
+    var tvpValue = ArgumentUtility.CheckNotNullAndType<SqlTableValuedParameterValue>(nameof(parameterValue), parameterValue);
+
+    var sqlParameter = (SqlParameter)command.CreateParameter();
+    sqlParameter.Value = tvpValue.Any() ? tvpValue : DBNull.Value;
```
MichaelKetting, line 94: I'm not certain that null is correct for an empty TVP. 
We should also test for Count/Length instead of empty. depending on Any-implementation it would isntantiate an iterator isntead of first testing for IReadOnlyCollection (if there's any version that does that).

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,136 @@
...
+    Assert.That(
+        () => parameterDefinition.GetParameterValue(items),
+        Throws.InstanceOf<ArgumentException>().With
+            .ArgumentExceptionMessageEqualTo($"Item 0 of parameter 'value' has the type '{typeof(string)}' instead of '{typeof(int)}'.", "value"));
+  }
+
+  [Test]
+  public void CreateDataParameter_WithTableValuedParameter_SetsSqlParameterProperties ()
+  {
+    var items = new[] { 1, 2, 3, 5, 8, 13, 21 }.ToList();
```
MichaelKetting, line 101:
```suggestion
    List<int> items = [1, 2, 3, 5, 8, 13, 21];
```

MichaelKetting: Or, actually, just drop the ToList() since the items aren't used beyond the foreach.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,136 @@
...
+    var items = new[] { "Value1", "Value2" };
+    var storageTypeInformation = StorageTypeInformationObjectMother.CreateIntStorageTypeInformation();
+
+    var parameterDefinition = new SqlTableValuedDataParameterDefinition(storageTypeInformation, true);
+
+    Assert.That(
+        () => parameterDefinition.GetParameterValue(items),
+        Throws.InstanceOf<ArgumentException>().With
+            .ArgumentExceptionMessageEqualTo($"Item 0 of parameter 'value' has the type '{typeof(string)}' instead of '{typeof(int)}'.", "value"));
+  }
```
MichaelKetting, line 96: wouldn't this test be more explicit if we did a non-generic collection with items `{1, "stringValue", 2}`?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,136 @@
...
+      Assert.That(tvpValue.TableTypeName, Is.EqualTo($"TVP_{storageTypeInformation.StorageDbType}"));
+
+    Assert.That(tvpValue.ColumnMetaData.Count, Is.EqualTo(1));
+
+    var sqlMetaData = tvpValue.ColumnMetaData.Single();
+    Assert.That(sqlMetaData.Name, Is.EqualTo("Value"));
+
+    var conversionHelper = new SqlParameter();
+    conversionHelper.DbType = storageTypeInformation.StorageDbType;
+    var sqlDbType = conversionHelper.SqlDbType;
```
MichaelKetting, line 78: I would prefer if we had this as part of the testset instead of calculated. the fewer computations a test has to do, the less likely a bug is introduced. also, it documents what is expected. With the current test, if there's a bug with using SqlParamerter (e.g. it always does DbType.Object or something) we might not find it if the prod-code uses the same apis.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+using Remotion.Reflection;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Parameters;
+
+/// <summary>
+/// Can create <see cref="IDataParameterDefinition"/> instances that can handle <see cref="ICollection"/>, <see cref="ICollection{T}"/>, and <see cref="IReadOnlyCollection{T}"/>
+/// parameter values.
+/// </summary>
+public class SqlTableValuedDataParameterDefinitionFactory : IDataParameterDefinitionFactory
```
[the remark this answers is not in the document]
martin-herwich: The other factories are also just factories, we decided to see this as a chain of responsibility rather than decoration.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    if (IsCollection(queryParameter.Value, out var itemType, out var isDistinct))
+    {
+      var storageTypeInformation = StorageTypeInformationProvider.GetStorageType(itemType);
+      return new SqlTableValuedDataParameterDefinition(storageTypeInformation, isDistinct);
+    }
+    return NextDataParameterDefinitionFactory.CreateDataParameterDefinition(queryParameter);
+  }
+
+  private bool IsCollection (object? value, [MaybeNullWhen(false)] out Type itemType, out bool isDistinct)
```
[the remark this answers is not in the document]
martin-herwich: `TryIsCollection` seems weird to me, I suggest `TryGetCollectionInfo`.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinitionFactoryTests.cs
martin-herwich: I have refactored the tests into a single method with `[TestCaseSource]`, and added cases for `ReadOnlyCollection<int>`, `ArrayList`, and `List<int>`, as well as Mocks for the collection and set interfaces.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinition.cs
martin-herwich: Do you mean we should only check the items for their type here, and then re-iterate the `enumerable` after we created the `SqlTableValuedParameterValue`? Or do we iterate only after creating the `SqlTableValuedParameterValue`, but still throw an `ArgumentItemTypeException` at the end of the method?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedDataParameterDefinition.cs
martin-herwich: I looked it up in the existing Acta Nova implementation - the correct value is `null`, not `DBNull.Value`, so I will change that accordingly.

MichaelKetting: Partially. ICollection<T> and ICollection are covered in .NET 8. IReadOnlyCollection is not covered.

Frazerus: > I looked it up in the existing Acta Nova implementation - the correct value is `null`, not `DBNull.Value`, so I will change that accordingly. I remember that an empty collection results in an error.
> 
> With regards to `.Any()`, I would question the value of this micro-optimization vs. the value of the statement's readability. Are you sure that preventing an iterator instantiation is worth having the code look like this: `sqlParameter.Value = tvpValue.Count == 0 ? null : tvpValue;` or `sqlParameter.Value = tvpValue.Count != 0 ? tvpValue : null;` instead of: `sqlParameter.Value = tvpValue.Any() ? tvpValue : null;`

@MichaelKetting?

MichaelKetting: Discussed with @martin-herwich we use .NET NULL, not DBNULL for EMPTY and will have a separate task to do integrtion tests for these cases.

MichaelKetting (approved):

PR #1114, opened 2024-04-25
Feature/rm 9191 move create db parameter from i storage type information to i sql dialect

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9191


Remotion/Data/DomainObjects/Persistence/Rdbms/ISqlDialect.cs
```diff
@@ -26,9 +28,30 @@ public interface ISqlDialect
...
+    /// <param name="parameterName">The dialect-conforming name for the parameter; use <see cref="GetParameterName"/> to obtain a valid name.</param>
+    /// <param name="value">The value to create the parameter for.</param>
+    /// <returns>A parameter holding the given <paramref name="value"/> (possibly converted), for the given <paramref name="command"/>.</returns>
+    /// <remarks>
+    /// <para>
+    /// The returned parameter's <see cref="IDataParameter.Value"/> is set to (a possibly converted version of) <paramref name="value"/>.
+    /// </para>
+    /// <para>
+    /// The parameter's <see cref="IDataParameter.ParameterName"/> is not set, and the parameter is not added to the command's 
+    /// <see cref="IDbCommand.Parameters"/> collection.
```
a-ctor, line 51: The XML doc about the parameter name not being set is outdated

Remotion/Data/DomainObjects/Persistence/Rdbms/ISqlDialect.cs
```diff
@@ -26,9 +28,30 @@ public interface ISqlDialect
...
     string StatementDelimiter { get; }
 
+    /// <summary>Gets the dialect-conforming variant of the given parameter <paramref name="name"/>.</summary>
     string GetParameterName (string name);
 
     /// <summary> Surrounds an identifier with delimiters according to the database's syntax. </summary>
     string DelimitIdentifier (string identifier);
+
+    /// <summary>
+    /// Creates an <see cref="IDbDataParameter"/> for the given <paramref name="value"/> and <see cref="IDbCommand"/>.
```
a-ctor, line 38: We might want to list the other newly added parameters here as well.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/Specifications/UpdatedColumnsSpecification.cs
```diff
@@ -56,8 +56,9 @@ public void AppendColumnValueAssignments (StringBuilder statement, IDbCommand db
       var columnsWithParameters = _columnValues.Select(
           cv =>
           {
-            var parameter = cv.Column.StorageTypeInfo.CreateDataParameter(dbCommand, cv.Value);
-            parameter.ParameterName = sqlDialect.GetParameterName(cv.Column.Name);
+            var storageTypeInfo = cv.Column.StorageTypeInfo;
+            var parameterName = sqlDialect.GetParameterName(cv.Column.Name);
+            var parameter = sqlDialect.CreateDataParameter(dbCommand, storageTypeInfo, parameterName, cv.Value);
```
a-ctor, line 61: The usage pattern of `CreateDataParameter` is always the same. We start with a column, get the `storageTypeInfo` and `parameterName` and pass it to `CreateDataParameter`. We might want to consider if it makes more sense to pass the column into `CreateDataParameter` instead. What do you think?

a-ctor (approved):

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/Specifications/UpdatedColumnsSpecification.cs
Frazerus: I get what you mean, but I honestly can't see which is cleaner. The current pattern is more testable though. @martin-herwich?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
```diff
@@ -50,5 +53,61 @@ public virtual string DelimitIdentifier (string identifier)
...
+          if (fulltextCompatibleMaxLength.HasValue)
+          {
+            var isNullValue = parameter.Value == null || parameter.Value == DBNull.Value;
+            var isStringAndValueDoesNotExceedCompatibleMaxLength =
+                parameter.Value is string stringValue && stringValue.Length <= fulltextCompatibleMaxLength;
+            var isCharArrayAndValueDoesNotExceedCompatibleMaxLength =
+                parameter.Value is char[] charsValue && charsValue.Length <= fulltextCompatibleMaxLength;
+
+            if (isNullValue || isStringAndValueDoesNotExceedCompatibleMaxLength || isCharArrayAndValueDoesNotExceedCompatibleMaxLength)
+              parameter.Size = fulltextCompatibleMaxLength.Value;
```
MichaelKetting, line 92: We could simplify to the following.
```suggestion
            var isStringAndValueDoesNotExceedCompatibleMaxLength =
            bool isTextAndValueDoesNotExceedCompatibleMaxLength = dbDataParameter.Value switch
            {
                string stringValue => stringValue.Length <= _fulltextCompatibleMaxLength,
                char[] charsValue => charsValue.Length <= _fulltextCompatibleMaxLength,
                _ => false
            };
            if (isNullValue || isTextAndValueDoesNotExceedCompatibleMaxLength)
              parameter.Size = fulltextCompatibleMaxLength.Value;
```

I also considered using this just to get "Length or null" from the DataParameter.Value, but it felt easier to read this way. the boolean test in the IF is clearer that way / naming the variables would be awkward. I do accept good version that do not have the comparison twice, though.

Actually, we can integrate the null test into the switch-pattern too and make it one happy boolean value which would look like this:
```
      bool hasFulltextCompatibleMaxLength = dbDataParameter.Value switch
      {
          null => true,
          DBNull => true,
          string stringValue => stringValue.Length <= _fulltextCompatibleMaxLength,
          char[] charsValue => charsValue.Length <= _fulltextCompatibleMaxLength,
          _ => false
      };
      if (hasFulltextCompatibleMaxLength)
        parameter.Size = fulltextCompatibleMaxLength.Value;
```
Which one is best really depends on whether you wanto be explicit about the null-behavior in the if condition.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
```diff
@@ -50,5 +53,61 @@ public virtual string DelimitIdentifier (string identifier)
...
+          }
+        }
+        else
+        {
+          var isStringAndValueExceedsParameterSize = convertedValue is string stringValue && stringValue.Length > parameterSize;
+          var isCharArrayAndValueExceedsParameterSize = convertedValue is char[] charsValue && charsValue.Length > parameterSize;
+          var isByteArrayAndValueExceedsParameterSize = convertedValue is byte[] bytesValue && bytesValue.Length > parameterSize;
+
+          if (!(isStringAndValueExceedsParameterSize || isCharArrayAndValueExceedsParameterSize || isByteArrayAndValueExceedsParameterSize))
+            parameter.Size = parameterSize;
```
MichaelKetting, line 106: The same simplification as above could be applied here, too.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -49,5 +57,219 @@ public void GetParameterName ()
...
+        dataParameterMock.SetupSet(_ => _.Size = storageTypeSize.Value).Verifiable();
+
+      var result = _dialect.CreateDataParameter(commandMock.Object, storageTypeInformation, "@dummy", "value");
+
+      commandMock.Verify();
+      dataParameterMock.Verify();
+
+      Assert.That(result, Is.SameAs(dataParameterMock.Object));
+      if (!storageTypeSize.HasValue)
+        dataParameterMock.VerifySet(_ => _.Size = It.IsAny<int>(), Times.Never());
```
MichaelKetting, line 98: We got a strict mock above. we don't need to verify-never and do a strict mock. We can have either a stub and just verify the set at the end (or make it property behavior and just assert the properties at the end of the test) or do a strict mock and not set up the stuff we want to fail.

Right now, this last line will never happen because eitehr it's not expected to happen and then it won't get asserted or it would fail when called due to the strict-behavior.

MichaelKetting: Reading more into the strict mock: since we can't actually test the correct value of Size when it's not been set (the default being 0), we do would be doing a VerifySet at the end. We can still do regular Assert for the other properties if we want to. Either way, no need to be strict and Verify. One or the other suffices.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -49,5 +57,219 @@ public void GetParameterName ()
...
+      dataParameterMock.SetupSet(_ => _.DbType = storageTypeInformation.StorageDbType).Verifiable();
+      dataParameterMock.SetupSet(_ => _.Value = "converted value").Verifiable();
+
+      var result = _dialect.CreateDataParameter(commandMock.Object, storageTypeInformation, "@dummy", "value");
+
+      commandMock.Verify();
+      dataParameterMock.Verify();
+
+      Assert.That(result, Is.SameAs(dataParameterMock.Object));
+      dataParameterMock.VerifySet(_ => _.Size = It.IsAny<int>(), Times.Never());
```
MichaelKetting, line 130: strict mock, this line will never fail so using a strict-mock isn't helpful.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -49,5 +57,219 @@ public void GetParameterName ()
...
+      commandMock.Verify();
+      dataParameterMock.Verify();
+
+      Assert.That(result, Is.SameAs(dataParameterMock.Object));
+      if (!storageTypeSize.HasValue)
+        dataParameterMock.VerifySet(_ => _.Size = It.IsAny<int>(), Times.Never());
+    }
+
+    [Test]
+    public void CreateDataParameter_WithStringValueExceedingFixedSize_DoesNotInitializeSize ()
```
MichaelKetting, line 102: is there a reason why you do this via three tests instead of just using a TestCase or TestCaseSource? Since you're using testcase for the previous test it feels weird to not do it here, too for string/char[]/byte[]

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -49,5 +57,219 @@ public void GetParameterName ()
...
+          "test",
+          DbType.Boolean,
+          false,
+          storageTypeSize,
+          typeof(int),
+          _typeConverterStub.Object);
+
+      _typeConverterStub.Setup(_ => _.ConvertTo(null, null, "value", storageTypeInformation.StorageType)).Returns("");
+
+      commandMock.Setup(_ => _.CreateParameter()).Returns(dataParameterMock.Object).Verifiable();
```
MichaelKetting, line 83: the usual question applies: isn't mock overkill for command / CreateParameter?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -24,11 +28,15 @@ namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer
...
     private SqlDialect _dialect;
+    private Mock<TypeConverter> _typeConverterStub;
+    private StorageTypeInformation _storageTypeInformation;
 
     [SetUp]
     public void SetUp ()
     {
       _dialect = new SqlDialect();
+      _typeConverterStub = new Mock<TypeConverter>();
+      _storageTypeInformation = new StorageTypeInformation(typeof(bool), "test", DbType.Boolean, false, null, typeof(int), _typeConverterStub.Object);
```
MichaelKetting, line 39: please create the StorageTypeInfo in each test. we already create a custom instance most of the time and it leads to possible copy&paste bugs / harder readability when mixing fields and variables for the same concept in the test fixture. If you need the same isnance more than once in the test fixture but a special one in other cases, use a factory method for it.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -49,5 +57,219 @@ public void GetParameterName ()
...
+      dataParameterMock.VerifySet(_ => _.Size = It.IsAny<int>(), Times.Never());
+    }
+
+    [Test]
+    public void CreateDataParameter_WithNullResult ()
+    {
+      var commandMock = new Mock<IDbCommand>(MockBehavior.Strict);
+      var dataParameterMock = new Mock<IDbDataParameter>(MockBehavior.Strict);
+
+      _typeConverterStub.Setup(_ => _.ConvertTo(null, null, "value", _storageTypeInformation.StorageType)).Returns((object)null);
```
MichaelKetting, line 205: here, for isntance, it's actually difficult to see that the parameter has no size. you need to scroll up to the setup to understand this.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
```diff
@@ -49,5 +57,219 @@ public void GetParameterName ()
...
+      Assert.That(result, Is.SameAs(dataParameterMock.Object));
+    }
+
+    [Test]
+    public void CreateDataParameter_WithNullInput ()
+    {
+      var commandMock = new Mock<IDbCommand>(MockBehavior.Strict);
+      var dataParameterMock = new Mock<IDbDataParameter>(MockBehavior.Strict);
+
+      _typeConverterStub.Setup(_ => _.ConvertTo(null, null, null, _storageTypeInformation.StorageType)).Returns("converted value");
```
MichaelKetting, line 259: shouldn't this be DbNullValue? I believe right now, we're missing the DBNullValue test in this fixture.
We might also be missing DBNullValue with parameter size?

MichaelKetting (changes requested):

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/Specifications/ComparedColumnsSpecificationTest.cs
```diff
@@ -37,10 +37,11 @@ public class ComparedColumnsSpecificationTest
     private ColumnDefinition _column2;
     private object _value2;
     private Mock<IStorageTypeInformation> _storageTypeInformationMock2;
+
```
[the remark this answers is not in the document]
martin-herwich: The line break is meaningful here.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/SqlDialectTest.cs
martin-herwich: The reason is that we are dealing with different types here, which affects the original value, the converted value, and the `StorageTypeInformation`. It is more readable as three separate test cases.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
```diff
@@ -50,5 +53,65 @@ public virtual string DelimitIdentifier (string identifier)
...
+              DbType.String => 4000,
+              // AnsiStringFixedLength and StringFixedLength represent char(...) and nchar(...), which do not support "max" as size
+              _ => default(int?)
+          };
+
+          if (fulltextCompatibleMaxLength.HasValue)
+          {
+            var hasFullTextCompatibleMaxLength = convertedValue switch
+            {
+                DBNull => true,
```
MichaelKetting, line 85: I checked the original code and we had
`var isNullValue = parameter.Value == null || parameter.Value == DBNull.Value;`
with the switch, we need to also have `null -> true` for semantic equality. It may be that this case never happens but we should handle it as it was before to avoid sematnic changes.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
```diff
@@ -50,5 +53,65 @@ public virtual string DelimitIdentifier (string identifier)
...
+      if (storageTypeInformation.StorageTypeLength.HasValue)
+      {
+        var parameterSize = storageTypeInformation.StorageTypeLength.Value;
+        if (parameterSize == SqlStorageTypeInformationProvider.StorageTypeLengthRepresentingMax)
+        {
+          var fulltextCompatibleMaxLength = storageTypeInformation.StorageDbType switch
+          {
+              DbType.AnsiString => 8000,
+              DbType.String => 4000,
+              // AnsiStringFixedLength and StringFixedLength represent char(...) and nchar(...), which do not support "max" as size
```
MichaelKetting, line 77: instead of a comment, we could implement those two now and assign null, with a comment that they don't support max. might make it easier to understand and isn't that much longer. thoughts?
```suggestion
              DbType.AnsiString => 8000,
              DbType.String => 4000,
              DbType.AnsiStringFixedLength  => default(int?), // represents char(...), which do not support "max" 
              DbType.StringFixedLength => default(int?), // represents nchar(...), which do not support "max" 
```

MichaelKetting, line 77: We also had explciit tests for the FixedLength types to better document the intended behavior. I believe those are gone now?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
```diff
@@ -50,5 +53,65 @@ public virtual string DelimitIdentifier (string identifier)
...
+          };
+
+          if (fulltextCompatibleMaxLength.HasValue)
+          {
+            var hasFullTextCompatibleMaxLength = convertedValue switch
+            {
+                DBNull => true,
+                string stringValue => stringValue.Length <= fulltextCompatibleMaxLength.Value,
+                char[] charsValue => charsValue.Length <= fulltextCompatibleMaxLength.Value,
+                _ => false
```
MichaelKetting, line 88: I switched this to true, no test was red

MichaelKetting (changes requested):
Sorry, some small stuff

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
Frazerus: Those did not exist

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
Frazerus: Changed to exception since this case is theoretically impossible as long as the StorageTypeInfo is correct

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
```diff
@@ -50,5 +53,67 @@ public virtual string DelimitIdentifier (string identifier)
...
+
+          if (fulltextCompatibleMaxLength.HasValue)
+          {
+            var hasFullTextCompatibleMaxLength = convertedValue switch
+            {
+                DBNull => true,
+                null => true,
+                string stringValue => stringValue.Length <= fulltextCompatibleMaxLength.Value,
+                char[] charsValue => charsValue.Length <= fulltextCompatibleMaxLength.Value,
+                _ => throw new NotSupportedException($"The type '{convertedValue.GetType()}' is not fulltext compatible.")
```
MichaelKetting, line 90:
```suggestion
                _ => false
```
We should keep `false` instead of exception because not all string-cases would be checked for string-ness and it is weird to throw if the value type is e.g. int and the DbType is AnsyString but not if the value type is int and the DbType is AnsiStringFixedLength.

MichaelKetting (approved):
I noticed the last-minute excpetion change in SqlDialect. Please revert to boolean result (see comment for explanation).

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SqlDialect.cs
martin-herwich: We only enter this code block if there is a `fulltextCompatibleMaxLength`, which is only the case for `DbTypes` `String` and `AnsiString`. If the converted value is neither `string` nor `char[]` nor `null`/`DbNull`, the error is that the (.NET) storage type is not compatible with the `DbType`, which is only relevant here because we worry about fulltext search.
If there is a crazy `IStorageTypeInformation` that claims its `StorageDbType` is `DbType.String`, but it converts values to `byte[]`, this will trigger the exception, regardless of the actual length of the `byte[]`.

MichaelKetting: I understand that it's likely a bug if this happens. My intention here is that the place guarding against this isn't well defined here and we already have a simple and well-defined behavior: if it's a string and the storagetype length is below the max-size, use the storage type length and otherwise leave the original default. If we add a guard here, some error scenarios would fail here and others, which don't trigger max length but have incompatible types, would fail elsewhere. that would be inconsistent and make it more difficult to identy the root cause of the issue.

The second part was the message itself which implies that the value must be fulltext compatible while it's only about the parameter length. 

So, we the general idea being that we want to only inject special handling for some very specific usecases also means we shouldn't touch other usecases but rather pass them through unchanged. 

I can't say how much of this is due to the fact that that was the original behavior and how much is a general practice. It's very well possible that a competely greenfield design might result in different concepts, but those would probably work with an entirely differnt code structure, making the datatype / length / fulltext-option very specific. Since we don't do this here, it's best to see the maxlength for the fulltext compatible parameter as an add-on for text and nothing else.

Some more background information on the orginal requirements can be found here https://re-motion.atlassian.net/issues/RM-3211

martin-herwich: I reverted to `false`, and also renamed the variable to `replaceMaxSizeWithFulltextCompatibleSize`, so that false is the semantically correct result for the condition in question.

PR #1125, opened 2024-05-14
RM-9206 Upgrade Microsoft.CodeAnalysis.CSharp.Workspaces to 4.9.2

KuJo7:
https://re-motion.atlassian.net/browse/RM-9206

### DIFF
--- a/Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
+++ b/Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
@@ -7,6 +7,7 @@
     <AssemblyName>Remotion.Development.Analyzers</AssemblyName>
     <AssemblyTitle>Remotion Analyzers</AssemblyTitle>
     <Description>Analyzer ensuring best practices when using the Remotion assemblies.</Description>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Remove="$(MsBuildThisFileFullPath)"/>
--- a/Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
+++ b/Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
@@ -17,7 +17,7 @@ namespace Core.Development.Analyzers
     public static readonly DiagnosticDescriptor AlternativeDescriptor = new DiagnosticDescriptor(
         "RMCORE0001",
         "Use SafeContext instead of typical API",
-        "'{0}' should not be used with SafeContext, use 'Remotion.Context.SafeContext.{1}' instead.",
+        "'{0}' should not be used with SafeContext, use 'Remotion.Context.SafeContext.{1}' instead",
         "Usage",
         DiagnosticSeverity.Warning,
         true);
@@ -25,7 +25,7 @@ namespace Core.Development.Analyzers
     public static readonly DiagnosticDescriptor WarningDescriptor = new DiagnosticDescriptor(
         "RMCORE0001",
         "Use typical API with SafeContext",
-        "'{0}' should not be used with SafeContext by itself, use '{1}' to create a safe context boundary.",
+        "'{0}' should not be used with SafeContext by itself, use '{1}' to create a safe context boundary",
         "Usage",
         DiagnosticSeverity.Warning,
         true);
--- a/Remotion/Web/Development.Analyzers/AnalyzerReleases.Unshipped.md
+++ b/Remotion/Web/Development.Analyzers/AnalyzerReleases.Unshipped.md
@@ -1,0 +1,6 @@
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+RMWEB0003 | Usage | Error | WxeResourcePageStepAnalyzer
+RMWEB0004 | Usage | Error | WxeResourcePageStepAnalyzer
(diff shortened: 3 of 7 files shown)


Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
```diff
@@ -17,15 +17,15 @@ private record SymbolContext (ITypeSymbol? SafeContextBoundarySymbol);
     public static readonly DiagnosticDescriptor AlternativeDescriptor = new DiagnosticDescriptor(
         "RMCORE0001",
         "Use SafeContext instead of typical API",
-        "'{0}' should not be used with SafeContext, use 'Remotion.Context.SafeContext.{1}' instead.",
+        "'{0}' should not be used with SafeContext, use 'Remotion.Context.SafeContext.{1}' instead",
```
MichaelKetting, line 20: Micorsoft now uses a form without a trailing period in the analyzers? it would be best to have this as a separate commit with an explainatory commit message.

Context: I remember a discussion (or change) about this from 12-24 months ago.  I would like to understand when and how this was changed (or if we always did it "wrong").
cc @a-ctor & @Frazerus

Remotion/Web/Development.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,6 @@
+### New Rules
```
MichaelKetting, line 1: @a-ctor  & @Frazerus There was a discussion about not needed this if I remember correctly. Apparently, this semantic has now changed. Thoughts?

Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
```diff
@@ -7,6 +7,7 @@
     <AssemblyName>Remotion.Development.Analyzers</AssemblyName>
     <AssemblyTitle>Remotion Analyzers</AssemblyTitle>
     <Description>Analyzer ensuring best practices when using the Remotion assemblies.</Description>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
```
MichaelKetting, line 10: What does the new property do? what happens by setting it true (or false)?

Remotion/Web/Development.Analyzers/WxeResourcePageStepAnalyzer.cs
```diff
@@ -93,7 +93,7 @@ private void AnalyzeObjectCreationExpression (SyntaxNodeAnalysisContext ctx, Sym
 
       if (operation.Arguments[0].Value is not ITypeOfOperation typeOfOperation)
       {
-        if (SymbolEqualityComparer.Default.Equals(operation.Arguments[0].Parameter.Type, symbolContext.SystemTypeSymbol))
+        if (SymbolEqualityComparer.Default.Equals(operation.Arguments[0].Parameter!.Type, symbolContext.SystemTypeSymbol))
```
MichaelKetting, line 96: I didn't look into this beyond the code I'm seeing. Typically, overriding nullability is a bad thing. It may be correct here. Would like a second opinion @a-ctor

MichaelKetting (changes requested):
Requesting changes, at least for the commit history, possibly code, too.

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
Frazerus: This has always been with `.` at the end here, I do not remember a discussion about this however.

MichaelKetting: I just found a current example and yeah, Microsoft doesn't place a `.` at the end. It probably dates back to Marcel. 
Question for @a-ctor  and @Frazerus change this now, later or not at all?
Question for @KuJo7 : is there a ready-made assertion in the Microsoft-Library which forced this change or did you simply notice it and adapt it to be consistent with Microsoft?

a-ctor: With a newer version of roslyn a check for `<EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>` was introduced and this is what causes all those changes now as it has an error severity. I would do the changes now as they don't seem to be too much work.

Remotion/Web/Development.Analyzers/AnalyzerReleases.Unshipped.md
a-ctor: We did not need it because compile did not complain back then. Now it does so we should add them as is normal with all analyzer projects. To this day, I don't know if they have any technical relevance or are just here for documentation purposes...

Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
a-ctor: As the name suggest it enforces analyzer best practices that are not enforced before. With the update the analyzer class had an error that complained that this property should be set. If we don't want to do all the work now we might be able to set this to `false`, which might disable the initial error as well.

Remotion/Web/Development.Analyzers/WxeResourcePageStepAnalyzer.cs
a-ctor: We should probably use `?.` instead as that won't fail and the equality still works in our favor.

Remotion/Core/Development.Analyzers/Core.Development.Analyzers.csproj
MichaelKetting: Good. 
@KuJo7 We should do a separate commit that fixes the trailing periods and introduces the build property and probably the unshipped.md file and document in the comment message that the change was required due to stronger enforcement of rules. Otherwise, we don't know why it was done when we look at the code history again. We accept that the build wouldn't work for this first revision without the changes since it's only in the feature branch. in theory, we could do the fix-commit first and then the upgrade-commit, but I don't think it's worth it.

Remotion/Core/Development.Analyzers/RMCORE0001_SafeContextAnalyzer.cs
MichaelKetting: See note in Core.Development.Analyzers.csproj

MichaelKetting (approved):

PR #1133, opened 2024-05-14
Upgrade MSBuild.Extension.Pack to 1.9.0

KuJo7:
https://re-motion.atlassian.net/browse/RM-9212


MichaelKetting (changes requested):
In the RMTEAM-task, it's stated that the buildscript dependencies should not be touched. 

As far as I remember, the MSBuildExtension.Pack is a dependency introduced by the build script. Please verify and then wont-fix.

PR #1140, opened 2024-05-28
RM-9214 Fix ErrorMessage to be an optional parameter

KuJo7:
https://re-motion.atlassian.net/browse/RM-9214


martin-herwich:
The task was not ready to be implemented, because customizing the technical error message is not useful without also being able to customize the localized error message.

PR #1146, opened 2024-06-05
RM-9155 Screenreader informs user on click when no children were foun…

Frazerus:
…d in treeView

https://re-motion.atlassian.net/browse/RM-9155

Only contains the additions needed for the second option, as the first one might be overkill.

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/TreeView.ts
+++ b/Remotion/Web/ClientScript/Scripts/TreeView.ts
@@ -16,11 +16,29 @@
 // 
 class WebTreeView
 {
-  public static Initialize (treeViewOrSelector: CssSelectorOrElement<HTMLElement>): void
+  public static Initialize (treeViewOrSelector: CssSelectorOrElement<HTMLElement>, updateMessageText: Nullable<string>): void
   {
     ArgumentUtility.CheckNotNull ('treeViewOrSelector', treeViewOrSelector);
 
-    const treeView = ElementResolverUtility.ResolveSingle (treeViewOrSelector);
+    const treeView = ElementResolverUtility.ResolveSingle(treeViewOrSelector);
+
+    if (updateMessageText)
+    {
+      const treeAlertBox = treeView.querySelector<HTMLElement>('span[aria-live=assertive] span');
+      if (treeAlertBox)
+      {
+        treeAlertBox.setAttribute("aria-hidden", 'false');
+
+        setTimeout(() => treeAlertBox.innerHTML = updateMessageText, 0);
+
+        // we need a timeout to reset aria-hidden, as the screenreader needs time to read the value of the alert,
+        // but it should not be able to find the alert text afterward.
+        setTimeout(
+            () => treeAlertBox.setAttribute("aria-hidden", 'true'),
+            7_000
+        );
+      }
+    }
 
     treeView.addEventListener ('keydown', function (event)
     {
--- a/Remotion/Web/Core/UI/Controls/WebTreeView.cs
+++ b/Remotion/Web/Core/UI/Controls/WebTreeView.cs
@@ -65,6 +65,8 @@ namespace Remotion.Web.UI.Controls
     private const string c_nodeIconTPlus = "sprite.svg#TreeViewTPlus";
     private const string c_nodeIconWhite = "sprite.svg#TreeViewWhite";
 
+    private const string c_possibleChildrenDescriptorIDPostfix = "PossibleChildrenAriaDescriptor";
+
     #endregion
 
     /// <summary> The separator used for the node path. </summary>
@@ -89,7 +91,9 @@ namespace Remotion.Web.UI.Controls
     public enum ResourceIdentifier
     {
       /// <summary> Additional text for improved screen reader support in Internet Explorer.</summary>
-      ScreenReaderNodeSelectedLabelText
+      ScreenReaderNodeSelectedLabelText,
+      ScreenReaderNoItemFoundText,
+      ScreenReaderPossiblyNoChildrenText
     }
 
     // statics
@@ -143,6 +147,7 @@ namespace Remotion.Web.UI.Controls
     private bool _hasTreeNodeMenusCreated;
     private int _menuCounter;
     private string? _assignedLabelID;
+    private WebString? _clickedNodeWithoutChildrenText;
 
     private readonly IRenderingFeatures _renderingFeatures;
 
@@ -222,27 +227,30 @@ namespace Remotion.Web.UI.Controls
     /// <param name="eventArgument"> The path to the clicked tree node. </param>
     private void HandleExpansionCommandEvent (string eventArgument)
     {
-      string[] pathSegments;
-      WebTreeNode? clickedNode = ParseNodePath(eventArgument, out pathSegments);
-      if (clickedNode != null)
-      {
-        _focusededNode = clickedNode;
-        if (clickedNode.IsEvaluated)
-        {
-          clickedNode.IsExpanded = !clickedNode.IsExpanded;
-          if (clickedNode.IsExpanded && EnableLookAheadEvaluation)
-            clickedNode.EvaluateChildren();
-        }
-        else
-        {
-          clickedNode.EvaluateExpand();
-          if (EnableLookAheadEvaluation)
-            clickedNode.EvaluateChildren();
-        }
+      WebTreeNode? clickedNode = ParseNodePath(eventArgument, out _);
+      if (clickedNode == null)
+        return;
 
-        if (clickedNode.IsExpanded)
-          InitializeTreeNodeMenus(clickedNode.Children);
+      _focusededNode = clickedNode;
+      if (clickedNode.IsEvaluated)
+      {
+        clickedNode.IsExpanded = !clickedNode.IsExpanded;
+        if (clickedNode.IsExpanded && EnableLookAheadEvaluation)
+          clickedNode.EvaluateChildren();
       }
+      else
+      {
+        clickedNode.EvaluateExpand();
+        var hasChildren = clickedNode.Children.Count > 0;
+        if (!hasChildren)
+          _clickedNodeWithoutChildrenText = clickedNode.Text;
+
+        if (EnableLookAheadEvaluation)
+          clickedNode.EvaluateChildren();
+      }
+
+      if (clickedNode.IsExpanded)
+        InitializeTreeNodeMenus(clickedNode.Children);
     }
 
     /// <summary> Handles the click command. </summary>
@@ -585,11 +593,18 @@ namespace Remotion.Web.UI.Controls
     /// <summary> Overrides the parent control's <c>RenderContents</c> method. </summary>
     protected override void RenderContents (HtmlTextWriter writer)
     {
+      var escapedUpdateMessageTextOrNull = _clickedNodeWithoutChildrenText switch
+      {
+          null => "null",
+          _ => "'" + ScriptUtility.EscapeClientScript(
+              _clickedNodeWithoutChildrenText.GetValueOrDefault().GetValue() + " " + GetResourceManager().GetString(ResourceIdentifier.ScreenReaderNoItemFoundText)) + "'"
+      };
+
       ((IControl)this).Page!.ClientScript.RegisterStartupScriptBlock(
           this,
           typeof(WebTreeView),
           Guid.NewGuid().ToString(),
-          string.Format("WebTreeView.Initialize ('#{0}');", ClientID));
+          $"WebTreeView.Initialize ('#{ClientID}', {escapedUpdateMessageTextOrNull});");
 
       ResolveNodeIcons();
 
@@ -611,8 +626,38 @@ namespace Remotion.Web.UI.Controls
         RenderNodes(writer, _nodes, true, nodeIDAlgorithm, 0);
       }
       writer.RenderEndTag();
+
+      RenderAriaTreeAlertBox(writer);
+      RenderAriaPossibleChildrenDescription(writer);
+
       foreach (DropDownMenu menu in _menuPlaceHolder.Controls)
         menu.RenderAsContextMenu(writer);
+    }
+
+    private void RenderAriaTreeAlertBox (HtmlTextWriter writer)
+    {
+      writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassScreenReaderText);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaLive, HtmlAriaLiveAttributeValue.Assertive);
+      writer.RenderBeginTag(HtmlTextWriterTag.Span);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
+      writer.RenderBeginTag(HtmlTextWriterTag.Span);
+      writer.RenderEndTag();
+      writer.RenderEndTag();
+    }
+
+    private void RenderAriaPossibleChildrenDescription (HtmlTextWriter writer)
+    {
+      writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassScreenReaderText);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
+      writer.AddAttribute(HtmlTextWriterAttribute.Id, CreatePossibleChildrenDescriptorID());
+      writer.RenderBeginTag(HtmlTextWriterTag.Span);
+      GetResourceManager().GetText(ResourceIdentifier.ScreenReaderPossiblyNoChildrenText).WriteTo(writer);
+      writer.RenderEndTag();
+    }
+
+    private string CreatePossibleChildrenDescriptorID ()
+    {
+      return $"{ID}_{c_possibleChildrenDescriptorIDPostfix}";
     }
 
     /// <summary> Renders the <paremref name="nodes"/> onto the <paremref name="writer"/>. </summary>
@@ -802,6 +847,9 @@ namespace Remotion.Web.UI.Controls
       }
       if (node.IsExpanded)
         writer.AddAttribute(HtmlTextWriterAttribute2.AriaOwns, CreateNodeChildrenID(nodeID));
+
+      if (!isEvaluated)
+        writer.AddAttribute(HtmlTextWriterAttribute2.AriaDescribedBy, CreatePossibleChildrenDescriptorID());
 
       writer.RenderBeginTag(HtmlTextWriterTag.A);
       if (_treeNodeRenderMethod == null)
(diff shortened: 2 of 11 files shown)


Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -615,6 +627,28 @@ protected override void RenderContents (HtmlTextWriter writer)
...
+      writer.Write("&nbsp;");
+      writer.RenderEndTag();
+      writer.RenderEndTag();
+    }
+
+    private void RenderAriaPossibleChildrenDescription (HtmlTextWriter writer)
+    {
+      writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassScreenReaderText);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
+      writer.AddAttribute(HtmlTextWriterAttribute.Id, $"{c_possibleChildrenDescriptorIDPrefix}_{ID}");
```
MichaelKetting, line 646: ID-format: `ClientID_prefix` (please fix naming.
please extract ID into separate method since we will be using it twice in the code (once when rendering and once when referencing)

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -615,6 +627,28 @@ protected override void RenderContents (HtmlTextWriter writer)
         menu.RenderAsContextMenu(writer);
     }
 
+    private void RenderAriaTreeAlertBox (HtmlTextWriter writer)
+    {
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaLive, HtmlAriaLiveAttributeValue.Assertive);
+      writer.RenderBeginTag(HtmlTextWriterTag.Span);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
+      writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassScreenReaderText + " AriaTreeAlerts");
```
MichaelKetting, line 635: so we need the second css class beside screenreader text?

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -615,6 +627,28 @@ protected override void RenderContents (HtmlTextWriter writer)
...
+      writer.RenderBeginTag(HtmlTextWriterTag.Span);
+      writer.Write("&nbsp;");
```
MichaelKetting, line 637: why the nbsp?

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -615,6 +627,28 @@ protected override void RenderContents (HtmlTextWriter writer)
...
+      writer.RenderEndTag();
+    }
+
+    private void RenderAriaPossibleChildrenDescription (HtmlTextWriter writer)
+    {
+      writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassScreenReaderText);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaHidden, HtmlAriaHiddenAttributeValue.True);
+      writer.AddAttribute(HtmlTextWriterAttribute.Id, $"{c_possibleChildrenDescriptorIDPrefix}_{ID}");
+      writer.RenderBeginTag(HtmlTextWriterTag.Span);
+      writer.Write(GetResourceManager().GetString(ResourceIdentifier.ScreenReaderPossiblyNoChildrenText));
```
MichaelKetting, line 648: we got better alternatives:
`GetResourceManager.GetText(...).WriteTo(writer)`
This ensures html encoding is done properly.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
MichaelKetting: if it's for selecting the element, please use an ID.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -174,6 +179,7 @@ public WebTreeView (IControl? ownerControl)
       _renderingFeatures = SafeServiceLocator.Current.GetInstance<IRenderingFeatures>();
       _fallbackNavigationUrlProvider = SafeServiceLocator.Current.GetInstance<IFallbackNavigationUrlProvider>();
       _labelReferenceRenderer = SafeServiceLocator.Current.GetInstance<ILabelReferenceRenderer>();
+      _updateMessageText = string.Empty;
```
MichaelKetting, line 182: should be type `WebString?`

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -222,27 +228,30 @@ void IPostBackEventHandler.RaisePostBackEvent (string eventArgument)
...
+        clickedNode.IsExpanded = !clickedNode.IsExpanded;
+        if (clickedNode.IsExpanded && EnableLookAheadEvaluation)
+          clickedNode.EvaluateChildren();
+      }
+      else
+      {
+        clickedNode.EvaluateExpand();
+        var hasChildren = clickedNode.Children.Count > 0;
+        if (!hasChildren)
+          _updateMessageText = clickedNode.Text.ToString() + " " + GetResourceManager().GetString(ResourceIdentifier.ScreenReaderNoItemFoundText);
```
MichaelKetting, line 247: clickedNode.Text + GetResourceManageer.GetText(...)

`WebString` combining: we may have examples.
@a-ctor

MichaelKetting (changes requested):
partial review. Script and Tests are todo.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
Frazerus: Thought I needed it, but it seems I didn't even know what it did, just thought that because we had it in some other areas, it would make sense here.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
Frazerus: Not sure, it is only used in Typescript, would that even make sense? We need to convert it to a string anyways?

Remotion/Web/Core/UI/Controls/WebTreeView.cs
Frazerus: Talked with @a-ctor, we do not have examples. 

IMO: 
Adding a combiner for `WebString`s would be overkill, and we need the end output of `_updateMessageText` to be a string anyways, unless there's some way to put webstring directly into TS

Remotion/Web/Core/UI/Controls/WebTreeView.cs
MichaelKetting: oh, yes, you are correct. I missed this detail in the review. Alas, we now have to make sure the Javascript parameter value is escaped in RenderContents.

MichaelKetting: `ScriptUtiity.EscapeClientScript`

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -16,11 +16,28 @@
...
+      if (treeAlertBox)
+      {
+        treeAlertBox.setAttribute("aria-hidden", 'false');
+        // alert box needs to be unset as we use additive aria-live, meaning only added changes are used by the screen reader
+        treeAlertBox.innerHTML = '';
+        treeAlertBox.innerHTML = updateMessageText;
+
+        setTimeout(
+            () => treeAlertBox.setAttribute("aria-hidden", 'true'),
+            7_000
```
MichaelKetting, line 37: we should add a comment why we wait 7 seconds.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
MichaelKetting, line 247: WebString.ToString() isn't correct here as it would perform exscaping. we want to use  GetValue and then do the scriptencoding.

MichaelKetting (changes requested):

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -16,11 +16,28 @@
...
 {
-  public static Initialize (treeViewOrSelector: CssSelectorOrElement<HTMLElement>): void
+  public static Initialize (treeViewOrSelector: CssSelectorOrElement<HTMLElement>, updateMessageText: string): void
   {
     ArgumentUtility.CheckNotNull ('treeViewOrSelector', treeViewOrSelector);
 
-    const treeView = ElementResolverUtility.ResolveSingle (treeViewOrSelector);
+    const treeView = ElementResolverUtility.ResolveSingle(treeViewOrSelector);
+
+    if (updateMessageText.length > 0)
```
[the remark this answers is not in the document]
Frazerus: using null for the updateMessageText doesn't work: The update message text is empty if it's not set as we set it like `$"WebTreeView.Initialize ('#{ClientID}', '{escapedUpdateMessageText}');");`

Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocTreeViewControlObjectTest.cs
```diff
@@ -107,6 +109,29 @@ public void WebTreeViewScreenshot_WithMarkedBadge ()
...
+      Assert.That(ariaTreeAlertSpan.InnerHTML, Is.EqualTo("&nbsp;"));
+
+      notExpandableNode.Expand();
+      ariaTreeAlertSpan = webTreeView.Scope.FindAllCss(".AriaTreeAlerts").Single();
+      Assert.That(ariaTreeAlertSpan.InnerHTML, Is.EqualTo("Does not contain any items."));
+
+      Thread.Sleep(10_000);
+
+      ariaTreeAlertSpan = webTreeView.Scope.FindAllCss(".AriaTreeAlerts").Single();
+      Assert.That(ariaTreeAlertSpan.InnerHTML, Is.EqualTo("&nbsp;"));
```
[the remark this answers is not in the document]
Frazerus: Ah nevermind, this is after the reset

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -16,11 +16,28 @@
...
+    {
+      const treeAlertBox = treeView.querySelector<HTMLElement>('#AriaTreeAlerts');
```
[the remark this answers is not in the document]
MichaelKetting: I'm still confused because in the code I've seen, this was a css class, not an id.

Remotion/Web/ClientScript/Scripts/TreeView.ts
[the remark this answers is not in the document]
MichaelKetting: yeah, I usually do a null-test when building the string and we use null not empty to communicate missing values.

Okay, the code is only RenderBase
```
 protected void AppendStringValueOrNullToScript (StringBuilder scriptBuilder, string? stringValue)
    {
      if (string.IsNullOrEmpty(stringValue))
        scriptBuilder.Append("null");
      else
        scriptBuilder.Append("'").Append(ScriptUtility.EscapeClientScript(stringValue)).Append("'");
    }
```

Frazerus:
Aria live seems to work more consistently than role=alert, so I would stick with this, as it essentially does the same thing (we dont care about aria-atomic) https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -585,11 +593,18 @@ protected override void Render (HtmlTextWriter writer)
     /// <summary> Overrides the parent control's <c>RenderContents</c> method. </summary>
     protected override void RenderContents (HtmlTextWriter writer)
     {
+      var escapedUpdateMessageTextOrNull = _clickedNodeWithoutChildrenText switch
+      {
+          null => "null",
+          _ => "'" + ScriptUtility.EscapeClientScript(_clickedNodeWithoutChildrenText + " " + GetResourceManager().GetString(ResourceIdentifier.ScreenReaderNoItemFoundText)) + "'"
+      };
+
+
```
MichaelKetting, line 602:
```suggestion
```

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -585,11 +593,18 @@ protected override void Render (HtmlTextWriter writer)
     /// <summary> Overrides the parent control's <c>RenderContents</c> method. </summary>
     protected override void RenderContents (HtmlTextWriter writer)
     {
+      var escapedUpdateMessageTextOrNull = _clickedNodeWithoutChildrenText switch
+      {
+          null => "null",
+          _ => "'" + ScriptUtility.EscapeClientScript(_clickedNodeWithoutChildrenText + " " + GetResourceManager().GetString(ResourceIdentifier.ScreenReaderNoItemFoundText)) + "'"
+      };
```
MichaelKetting, line 600: code design : in null-or-else situations, my go to would be ?: instead of a switch with pattern matching. 

It's okay like this, it just feels like overkill because it has five lines instead of three (or one line if no linewrap is needed)

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -143,6 +147,7 @@ public enum ResourceIdentifier
     private bool _hasTreeNodeMenusCreated;
     private int _menuCounter;
     private string? _assignedLabelID;
+    private string? _clickedNodeWithoutChildrenText;
```
MichaelKetting, line 150:
```suggestion
    private WebString? _clickedNodeWithoutChildrenText;
```

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -222,27 +227,30 @@ void IPostBackEventHandler.RaisePostBackEvent (string eventArgument)
...
+        clickedNode.IsExpanded = !clickedNode.IsExpanded;
+        if (clickedNode.IsExpanded && EnableLookAheadEvaluation)
+          clickedNode.EvaluateChildren();
+      }
+      else
+      {
+        clickedNode.EvaluateExpand();
+        var hasChildren = clickedNode.Children.Count > 0;
+        if (!hasChildren)
+          _clickedNodeWithoutChildrenText = clickedNode.Text.GetValue();
```
MichaelKetting, line 246:
```suggestion
          _clickedNodeWithoutChildrenText = clickedNode.Text;
```

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -585,11 +593,18 @@ protected override void Render (HtmlTextWriter writer)
     /// <summary> Overrides the parent control's <c>RenderContents</c> method. </summary>
     protected override void RenderContents (HtmlTextWriter writer)
     {
+      var escapedUpdateMessageTextOrNull = _clickedNodeWithoutChildrenText switch
+      {
+          null => "null",
+          _ => "'" + ScriptUtility.EscapeClientScript(_clickedNodeWithoutChildrenText + " " + GetResourceManager().GetString(ResourceIdentifier.ScreenReaderNoItemFoundText)) + "'"
```
MichaelKetting, line 599:
```suggestion
          _ => "'" + ScriptUtility.EscapeClientScript(_clickedNodeWithoutChildrenText.GetValue() + " " + GetResourceManager().GetString(ResourceIdentifier.ScreenReaderNoItemFoundText)) + "'"
```

MichaelKetting: This way, the information about escaping isn't distributed in the code.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -611,10 +626,40 @@ protected override void RenderContents (HtmlTextWriter writer)
...
+      RenderAriaPossibleChildrenDescription(writer);
+
       foreach (DropDownMenu menu in _menuPlaceHolder.Controls)
         menu.RenderAsContextMenu(writer);
     }
 
+    private void RenderAriaTreeAlertBox (HtmlTextWriter writer)
+    {
+      writer.AddAttribute(HtmlTextWriterAttribute.Class, CssClassScreenReaderText);
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaLive, HtmlAriaLiveAttributeValue.Assertive);
```
MichaelKetting, line 640: we discussed whether role=alert is better than aria-live. Did it not work or was there another reason that spoke against it?

Remotion/Web/ClientScript/Scripts/TreeView.ts
```diff
@@ -16,11 +16,29 @@
...
+    {
+      const treeAlertBox = treeView.querySelector<HTMLElement>('span[aria-live=assertive] span');
+      if (treeAlertBox)
+      {
+        treeAlertBox.setAttribute("aria-hidden", 'false');
+
+        setTimeout(() => treeAlertBox.innerHTML = updateMessageText, 0);
+
+        // we need a timeout to reset aria-hidden, as the screenreader needs time to read the value of the alert but it should
+        // not be able to find the alert text afterward.
```
MichaelKetting, line 35:
```suggestion
        // we need a timeout to reset aria-hidden, as the screenreader needs time to read the value of the alert
        // but it should not be able to find the alert text afterward.
```
better readability of the comment

MichaelKetting (approved):
done after comments are resolved.
~~role=alert vs aria-live is just a discussion note I would like to clear up so we have the solution we feel is best and understand why.~~

Remotion/Web/Core/UI/Controls/WebTreeView.cs
MichaelKetting: Just seen the comment earlier on this. 

> Aria live seems to work more consistently than role=alert, so I would stick with this, as it essentially does the same thing (we dont care about aria-atomic) https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role
@a-ctor  do you have any additional thoughts on this?

Otherwise, done.

Remotion/Web/Core/UI/Controls/WebTreeView.cs
Frazerus: I think ?: is horrible readability wise in cases where we have one option as long as the second one, splitting the ? : always feels off IMO

Remotion/Web/Core/UI/Controls/WebTreeView.cs
Frazerus: Since role=alert basically just sets aria-atomic and aria-live, aria-atomic might even induce the problem the screenreader seems to have

PR #1147, opened 2024-06-07
Feature/rm 9229 update license for security manager from agpl to lgpl

Frazerus:
https://re-motion.atlassian.net/browse/RM-9229

Second commit are only changers to license headers in src files.


SecurityManager/Clients.Web/SecurityManager.Clients.Web.csproj
```diff
@@ -11,9 +11,6 @@
     <AssemblyName>Remotion.SecurityManager.Clients.Web</AssemblyName>
     <RootNamespace>$(AssemblyName)</RootNamespace>
   </PropertyGroup>
-  <ItemGroup>
-    <None Update="..\SecurityManager AGPL.licenseheader" Pack="true" PackagePath="res/" Visible="false" />
-  </ItemGroup>
```
MichaelKetting, line 16: still need this, just switch to remoton lgpl file.

MichaelKetting: It delivers the lhm file to the res folder on the consumer side, so they, too have the correct header when our res files are linked in their solution.

Build/Customizations/Properties.props
```diff
@@ -2,15 +2,12 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <!-- Special settings for re-mix: -->
-    <AssemblyInfoFileForMixins>$(SolutionDirectory)Remotion\Mixins\AssemblyInfoShared.cs</AssemblyInfoFileForMixins>
```
MichaelKetting, line 5: we still have the remotion AssemblyInfoShared file on the disk we should remember to delete it some time. ticket please :)

MichaelKetting (approved):
one fix, one ticket, otherwise, done.

PR #1149, opened 2024-06-11
RM-9052 Autocomplete OnChange checks for same invalid value

Frazerus:
https://re-motion.atlassian.net/browse/RM-9052


Frazerus:
@a-ctor I talked with Mike, if you think he doesn't need to see this change, we can merge it after your review.

a-ctor (approved):
Looks good to me. I think we can merge this fix directly as it is very simple and specific.

PR #1150, opened 2024-06-11
Feature/rm 9188 replace xml with tvp when generating where clause

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9188


Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
```diff
@@ -82,5 +82,84 @@ IEnumerator<SqlDataRecord> IEnumerable<SqlDataRecord>.GetEnumerator ()
...
+      var ours = _columnMetaData[i];
+      var theirs = other._columnMetaData[i];
+      if (!theirs.Name.Equals(ours.Name))
+        return false;
+
+      if (!theirs.DbType.Equals(ours.DbType))
+        return false;
+
+      if (!theirs.MaxLength.Equals(ours.MaxLength))
+        return false;
```
MichaelKetting, line 111: please use `==` instead of `.Equals`
reasons: 
earlier, you already use `!=`
DbType.Equals uses the Object overload and thus boxing.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
```diff
@@ -82,5 +82,84 @@ IEnumerator<SqlDataRecord> IEnumerable<SqlDataRecord>.GetEnumerator ()
...
+      var ours = _records[rec];
+      var theirs = other._records[rec];
+      if (ours.FieldCount != theirs.FieldCount)
+        return false;
+
+      for (int col = 0; col < ours.FieldCount; col++)
+      {
+        var ourValue = ours[col];
+        var theirValue = theirs[col];
+        if (!ourValue.Equals(theirValue))
```
MichaelKetting, line 125:
```suggestion
        if (!ourValue.Equals(theirValue))
```
```suggestion
        if (!ourValue == theirValue)
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
```diff
@@ -82,5 +82,84 @@ IEnumerator<SqlDataRecord> IEnumerable<SqlDataRecord>.GetEnumerator ()
   }
 
   public int Count => _records.Count;
+
+  public bool Equals (SqlTableValuedParameterValue? other)
```
MichaelKetting, line 86: Is there benefit to having this on the production code? if you need to compare in the tests, just add a helper there.

On production, we are makign a promise that it's a well implemented and helpful api. if there is no use case for comparing the Values, we should not do it, given the amount of prod-code this generates.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
```diff
@@ -0,0 +1,111 @@
...
+    var storageTypeInfo = ColumnDefinition.StorageTypeInfo;
+
+    var parameterName = GetParameterName(sqlDialect);
+    var parameter = (SqlParameter)sqlDialect.CreateDataParameter(command, storageTypeInfo, parameterName, null);
+
+    // use the parameter we need anyway to find out the SqlDbType for the column
+    parameter.DbType = storageTypeInfo.StorageDbType;
+    var sqlMetaData = new SqlMetaData("Value", parameter.SqlDbType);
+
+    var typeName = $"TVP_{storageTypeInfo.StorageDbType}";
```
MichaelKetting, line 70: Shouldn't we get the typename from a lookup/resolver?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
```diff
@@ -0,0 +1,111 @@
...
+        var convertedValue = storageTypeInfo.ConvertToStorageType(value);
+        tvpValue.AddRecord(convertedValue);
+      }
+
+      parameter.Value = tvpValue;
+    }
+    else
+    {
+      parameter.Value = DBNull.Value;
+    }
```
MichaelKetting, line 89: didn't I see this logic already somewhere? in particular, empty-tvp is dbnull

MichaelKetting (changes requested):
got some questions which will possibly result in changes.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
martin-herwich: These are `object`-typed variables, so I think using `.Equals()` is safer at this point.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
martin-herwich: The test infrastructure checks whether the parameter value equals the expected parameter value - see `SqlProviderGeneratedSqlTestHelper.CheckCommand()`.
If we want to get rid of the `Equals()` implementation in the production code, we would have to inject an IEqualityComparer<SqlTableValuedParameterValue> with each table-valued parameter, and use that for equality assertions instead of `Is.EqualTo()`.
Is that what you had in mind?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
martin-herwich: Eventually, yes. But that has been built in the context of RM-9216, which has not been merged yet - see https://github.com/re-motion/Framework/pull/1152
We will have to inject an `ISingleScalarStructuredTypeDefinitionProvider` into the `SqlDbCommandBuilderFactory` so that it can be injected into the `SqlTableValuedParameterColumnSpecification` objects. Then we can use `TableTypeDefinition.CreateTableValuedParameterValue()` here.
Perhaps we should do RM-9216 first, and make adjustments to this PR based on the new design?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
martin-herwich: Yes, and it's wrong here - the value should ne `null`, not `DBNull.Value` - we can delete the `else`-branch.
The other place is in `SqlTableValuedDataParameterDefinition`.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
MichaelKetting: I was thinking, we could just add a check-method in the test fixture that compares the expected and actual ones and gives Assert-errors inline. Similiar to CheckClassDefinition etc.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
MichaelKetting: ah, okay. so, right now the code is locked on one specific pattern/usecase and the next PR fixes this. 

I can live with both orderings. officially cleaner would be to have the infrastructure first and then use the infrastructure here. but if thats' either more than 5min of work, it's probably not worth it? Was not sure if that was a bug or by design. perhaps add a comment for now if we keep the original order of PRs?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
MichaelKetting: and the rest of the copy/convert logic? is this also implemented in SqlTable..ParamDef? What would be required to unify the logic, assuming that it is not just accidentally the same logic but the same business case.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
MichaelKetting: ah, I understand. this is needed because we check if the parameter value matches the mock expectations.

I would suggest, to make the test easier/diagnostics better, that we match any tvp parameter, back up the value in the callback and assert it at the end of the test. that way, we get good diangostics if there's an error in the parameter's contents and not simply a failed mock.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecificationTest.cs
```diff
@@ -0,0 +1,116 @@
...
+    Assert.That(_sqlCommand.Parameters.Count, Is.EqualTo(1));
+    Assert.That(_sqlCommand.Parameters[0], Is.SameAs(_sqlParameter));
+
+    Assert.That(_sqlParameter.SqlDbType, Is.EqualTo(SqlDbType.Structured));
+    Assert.That(_sqlParameter.TypeName, Is.EqualTo("TVP_Guid"));
+    Assert.That(_sqlParameter.Value, Is.EqualTo(DBNull.Value));
+  }
+
+  [Test]
+  public void AppendComparisons ()
```
MichaelKetting, line 104: i think we need a second test for emptycollection

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/IntegrationTests/SqlProviderGeneratedSqlTest.cs
```diff
@@ -81,13 +83,17 @@ public void LoadDataContainers_SingleID ()
     [Test]
```
MichaelKetting, line 83: please add test with empty collection.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Parameters/SqlTableValuedParameterValue.cs
martin-herwich: I still don't understand. `SqlProviderGeneratedSqlTestHelper.CheckCommand()` is used by 19 tests, only 4 of which use a TVP - the others also need to check for equality of the parameter value. If we want to continue using a common helper method, we have to make an equality comparison for parameter values of type `SqlTableValuedParameterValue`.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
martin-herwich: We could try to put the conversion into `SqlTableValuedParameterValue`, but that would require injecting the conversion delegates for each column along with the `SqlMetaData`, using either a wrapper type or `ValueTuple` as item type for the `IReadOnlyCollection<T>`.
This would mean eliminating the conversion from `RecordDefinition.GetColumnValues()`.
I can do that in a separate commit, so you can see what it would look like.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecificationTest.cs
martin-herwich: I don't think so, because at this point, the collection of values is hidden behind a parameter, and the SQL command text is the same for empty and non-empty collections.
Empty collections must be handled differently when creating the parameter, because ADO.NET cannot deal with an empty `SqlDataRecord[]`, but that limitation only affects the `SqlParameter.Value`.

MichaelKetting: mhm... i'm not sure where it would fit, but some kind of itnegration test (or larger-scope unit test) would definitely be warrented to show everything works well together. It felt like here would be a good spot for thsi and the test would be cheap. you have a better / more meaningful place? 

And yes, there would be some redundancy, but the empty/null case is so special that it feels like it should be covered explicitly.

oh, and another reason why "here" would be good: we already test empty with the parameters, so testing with comparision, too, would be symetrical. I'm speaking from the experience of having looked for some documentation of expected behavior and tests are extremly helpful on this. So, cheap redundancy gets trumped by "acts as good docs".

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/IntegrationTests/SqlProviderGeneratedSqlTest.cs
martin-herwich: Unfortunately, this does not execute any SQL command, because in `LookupCommandFactory.CreateForSortedMultiIDLookup()`, the `objectIDList` is empty, resulting ina `MultiObjectLoadCommand<DataContainer?>` with an empty collection of `dbCommandBuildersAndReaders`, wherefore nothing is done in `MultiObjectLoadCommand<T>.Execute()`.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
```diff
@@ -72,12 +78,20 @@ public SqlDbCommandBuilderFactory (ISqlDialect sqlDialect)
       ArgumentUtility.CheckNotNull("selectedColumns", selectedColumns);
       ArgumentUtility.CheckNotNull("orderedColumns", orderedColumns);
 
-      var comparedValues = GetValuesForSingleColumnDefinition(comparedColumnValueTable);
+      var (columnDefinition, values) = GetValuesForSingleColumnDefinition(comparedColumnValueTable);
```
MichaelKetting, line 81:
```suggestion
      var (columnDefinition, comparedValues) = GetValuesForSingleColumnDefinition(comparedColumnValueTable);
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
```diff
@@ -72,12 +78,20 @@ public SqlDbCommandBuilderFactory (ISqlDialect sqlDialect)
...
+          tableType,
+          [RecordPropertyDefinition.ScalarAsValue(tableType.Properties.Single())]);
+
+      var dataParameterDefinition = new SqlTableValuedDataParameterDefinition(recordDefinition);
 
       return new SelectDbCommandBuilder(
           table,
           new SelectedColumnsSpecification(selectedColumns),
-          new SqlXmlSetComparedColumnSpecification(comparedValues.Item1, comparedValues.Item2),
+          new SqlTableValuedParameterComparedColumnSpecification(columnDefinition, values, dataParameterDefinition),
```
MichaelKetting, line 94:
```suggestion
          new SqlTableValuedParameterComparedColumnSpecification(columnDefinition, comparedValues, dataParameterDefinition),
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
```diff
@@ -72,12 +78,20 @@ public SqlDbCommandBuilderFactory (ISqlDialect sqlDialect)
...
       ArgumentUtility.CheckNotNull("orderedColumns", orderedColumns);
 
-      var comparedValues = GetValuesForSingleColumnDefinition(comparedColumnValueTable);
+      var (columnDefinition, values) = GetValuesForSingleColumnDefinition(comparedColumnValueTable);
+      var tableType = (TableTypeDefinition)_tableTypeDefinitionProvider.GetStructuredTypeDefinition(columnDefinition.StorageTypeInfo.DotNetType, false);
+
+      var recordDefinition = new RecordDefinition(
+          columnDefinition.Name,
+          tableType,
+          [RecordPropertyDefinition.ScalarAsValue(tableType.Properties.Single())]);
```
MichaelKetting, line 87: The fact that there can only be a single property in tableType.Properties is ensured by how we call GetStructuredTypeDefinition, correct?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
```diff
@@ -0,0 +1,88 @@
...
+      ColumnDefinition comparedColumn,
+      IEnumerable<object?> objectValues,
+      SqlTableValuedDataParameterDefinition dataParameterDefinition)
+  {
+    ArgumentUtility.CheckNotNull(nameof(comparedColumn), comparedColumn);
+    ArgumentUtility.CheckNotNull(nameof(objectValues), objectValues);
+    ArgumentUtility.CheckNotNull(nameof(dataParameterDefinition), dataParameterDefinition);
+
+    ComparedColumnDefinition = comparedColumn;
+    ObjectValues = objectValues.ToArray();
```
MichaelKetting, line 56:
```suggestion
    ObjectValues = objectValues;
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/Specifications/SqlTableValuedParameterComparedColumnSpecification.cs
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /// <inheritdoc />
+  public void AddParameters (IDbCommand command, ISqlDialect sqlDialect)
+  {
+    ArgumentUtility.CheckNotNull(nameof(command), command);
+    ArgumentUtility.CheckNotNull(nameof(sqlDialect), sqlDialect);
+
+    var parameterValue = DataParameterDefinition.GetParameterValue(ObjectValues);
+    var parameter = DataParameterDefinition.CreateDataParameter(command, GetParameterName(sqlDialect), parameterValue);
```
MichaelKetting, line 67:
```suggestion
    var parameterName = GetParameterName(sqlDialect);
    var parameterValue = DataParameterDefinition.GetParameterValue(ObjectValues);
    var parameter = DataParameterDefinition.CreateDataParameter(command, parameterName, parameterValue);
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/IntegrationTests/SqlTableValuedParameterValueChecker.cs
```diff
@@ -0,0 +1,67 @@
+using System;
```
MichaelKetting, line 1: lic header

MichaelKetting (approved):
naming and an unecessary enumatble-copy

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
martin-herwich: Correct. The `columnDefinition.StorageTypeInfo.DotNetType` must be a scalar type, so the returned `TableTypeDefinition` cannot be anything else than a table type with a single column for that type.
However, a twisted mind could hypothetically replace the `ISingleScalarStructuredTypeDefinitionProvider` with an implementation that returns something else. If there are more than one `IRdbmsStoragePropertyDefinition`, `.Single()` in line 87 would throw; if it is not a `SimpleStoragePropertyDefinition`, `SqlTableValuedParameterValue.AddRecord()` will throw an `ArgumentException`.

MichaelKetting: ah, I see the tableProvider is interface ISingle... 

Do you want to add `Assertion.IsTrue(tableType.Properties.Count == 1, "Table type returned by ISingleScalarStructuredTypeDefinitionProvider should only contain a single property.")` to make it more explicit to better document our contract?

MichaelKetting (approved):

PR #1164, opened 2024-06-20
Feature/rm 9237 add mixin crossreferencer to re motion

Frazerus:
https://re-motion.atlassian.net/browse/RM-9237


Remotion/Mixins/XRef/xml_utilities/saxon/IKVM.OpenJDK.Charsets.dll
MichaelKetting, line 1: note: saxon und stylessheets und images shouldn't be needed since we don't generate the html, only serialize the xml.

Remotion/Mixins/XRef/Mixins.XRef.Reflectors.csproj
MichaelKetting, line 1: folder name, projectfile name

MichaelKetting:
> Forgot to fix the API, will follow up

waiting for update :)

MichaelKetting (approved):

PR #1165, opened 2024-06-20
Feature/rm 9222 use table valued parameters in queries with constant collection expression

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9222

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/ObjectIDIntegrationTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/ObjectIDIntegrationTest.cs
@@ -208,6 +208,25 @@ namespace Remotion.Data.DomainObjects.UnitTests.Linq.IntegrationTests
           where possibleItems.Contains(o.ID.Value)
           select o;
 
+      Assert.That(
+          () => CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3),
+          Throws.InstanceOf<NotSupportedException>().With.Message.EqualTo(
+              $"The parameter value's type 'System.Object[]' cannot be mapped by "
+              + "Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters.SimpleTypeQueryParameterRecordDefinitionFinder. "
+              + "Either use a collection of scalar values, or use an implementation of "
+              + "Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters.IQueryParameterRecordDefinitionFinder "
+              + "that supports 'System.Object[]'."));
+    }
+
+    [Test]
+    public void QueryWithContainsInWhere_OnCollection_WithObjectIDValuesAsGuid ()
+    {
+      var possibleItems = new[] { (Guid)DomainObjectIDs.Order1.Value, (Guid)DomainObjectIDs.Order3.Value };
+      var orders =
+          from o in QueryFactory.CreateLinqQuery<Order>()
+          where possibleItems.Contains((Guid)o.ID.Value)
+          select o;
+
       CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3);
     }
 
@@ -255,8 +274,12 @@ namespace Remotion.Data.DomainObjects.UnitTests.Linq.IntegrationTests
           select o;
       Assert.That(
           () => CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3),
-          Throws.InstanceOf<NotSupportedException>()
-              .With.Message.EqualTo("Objects of type 'Remotion.Data.DomainObjects.UnitTests.TestDomain.Order_AssembledTypeProxy_1' cannot be used as data parameter value."));
+          Throws.InstanceOf<NotSupportedException>().With.Message.EqualTo(
+              "The parameter value's type 'Remotion.Data.DomainObjects.DomainObject[]' cannot be mapped by "
+              + "Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters.SimpleTypeQueryParameterRecordDefinitionFinder. "
+              + "Either use a collection of scalar values, or use an implementation of "
+              + "Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters.IQueryParameterRecordDefinitionFinder "
+              + "that supports 'Remotion.Data.DomainObjects.DomainObject[]'."));
     }
 
     [Test]
--- a/Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
@@ -1,0 +1,318 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Persistence.Rdbms;
+using Remotion.Data.DomainObjects.Queries;
+using Remotion.Data.DomainObjects.UnitTests.TestDomain;
+using Remotion.Development.NUnit.UnitTesting;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Linq.IntegrationTests;
+
+[TestFixture]
+public class TableValuedParameterIntegrationTest : IntegrationTestBase
+{
+  [Test]
+  public void Contains_WithEmptyCollection ()
+  {
+    var array = Array.Empty<Guid>();
+    var collectionStub = new Mock<ICollection<Guid>>();
+    collectionStub.As<IEnumerable>().Setup(_ => _.GetEnumerator()).Returns(() => array.GetEnumerator());
+    collectionStub.Setup(_ => _.GetEnumerator()).Returns(() => ((IEnumerable<Guid>)array).GetEnumerator());
+    var possibleItems = collectionStub.Object;
+
+    var orders =
+        from o in QueryFactory.CreateLinqQuery<Order>()
+        where possibleItems.Contains((Guid)o.ID.Value)
+        select o;
+
+    CheckQueryResult(orders, []);
+  }
+
+  [Test]
+  public void Contains_WithCollectionOfT ()
+  {
+    var possibleItems = SetupCollection<ICollection<Guid>>([ (Guid)DomainObjectIDs.Order1.Value, (Guid)DomainObjectIDs.Order3.Value, (Guid)DomainObjectIDs.Order3.Value ]);
+
+    var orders =
+        from o in QueryFactory.CreateLinqQuery<Order>()
+        where possibleItems.Contains((Guid)o.ID.Value)
+        select o;
+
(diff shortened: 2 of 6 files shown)


Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/ObjectIDIntegrationTest.cs
```diff
@@ -202,10 +202,10 @@ public void Query_UsesClassID_OnReferencedValue ()
     [Test]
     public void QueryWithContainsInWhere_OnCollection_WithObjectIDValues ()
     {
-      var possibleItems = new[] { DomainObjectIDs.Order1.Value, DomainObjectIDs.Order3.Value };
+      var possibleItems = new[] { (Guid)DomainObjectIDs.Order1.Value, (Guid)DomainObjectIDs.Order3.Value };
       var orders =
           from o in QueryFactory.CreateLinqQuery<Order>()
-          where possibleItems.Contains(o.ID.Value)
+          where possibleItems.Contains((Guid)o.ID.Value)
           select o;
```
MichaelKetting, line 209: Since you have changed the test spec in the integration test, I think we instead need to copy the test and do both versions, once with object and once with Guid

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/ObjectIDIntegrationTest.cs
```diff
@@ -255,8 +255,9 @@ where possibleItems.Contains(o)
...
       Assert.That(
           () => CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3),
-          Throws.InstanceOf<NotSupportedException>()
-              .With.Message.EqualTo("Objects of type 'Remotion.Data.DomainObjects.UnitTests.TestDomain.Order_AssembledTypeProxy_1' cannot be used as data parameter value."));
+          Throws.InstanceOf<InvalidOperationException>().With.Message.EqualTo(
+              "The parameter value's type 'Remotion.Data.DomainObjects.DomainObject[]' cannot be mapped by "
+              + "Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters.SimpleTypeQueryParameterRecordDefinitionFinder."));
```
MichaelKetting, line 260: getting an invalidopex instead of a notsupportedex here in the itnegration test feels like a regression. there is also a not very helpful bit of information with an internal type. I would prefer the original message or something close to it, e.g. `A collection of objects of type '<type>' cannot be used as a data parameter value. Only collections of simple values are supported.` (not sure whether to use use 'simple values' or 'scalar values' in the phrasing.

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,239 @@
...
+using Remotion.Data.DomainObjects.Queries;
+using Remotion.Data.DomainObjects.UnitTests.TestDomain;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Linq.IntegrationTests;
+
+[TestFixture]
+public class TableValuedParameterIntegrationTest : IntegrationTestBase
+{
+  [Test]
+  public void Contains_WithEmptyCollection_UsesTableValuedParameter ()
```
MichaelKetting, line 32: since this is already the itnegration test named "TVP..", we can drop the suffix "UsesTVP"

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,239 @@
...
+    collectionStub.As<IEnumerable>().Setup(_ => _.GetEnumerator()).Returns(() => array.GetEnumerator());
+    collectionStub.Setup(_ => _.GetEnumerator()).Returns(() => ((IEnumerable<Guid>)array).GetEnumerator());
+    var possibleItems = collectionStub.Object;
+
+    var orders =
+        from o in QueryFactory.CreateLinqQuery<Order>()
+        where possibleItems.Contains((Guid)o.ID.Value)
+        select o;
+
+    CheckQueryResult(orders);
```
MichaelKetting, line 45:
```suggestion
    CheckQueryResult(orders, Array.Empty<ObjectID>);
```

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,239 @@
...
+[TestFixture]
+public class TableValuedParameterIntegrationTest : IntegrationTestBase
+{
+  [Test]
+  public void Contains_WithEmptyCollection_UsesTableValuedParameter ()
+  {
+    var array = Array.Empty<Guid>();
+    var collectionStub = new Mock<ICollection<Guid>>();
+    collectionStub.As<IEnumerable>().Setup(_ => _.GetEnumerator()).Returns(() => array.GetEnumerator());
+    collectionStub.Setup(_ => _.GetEnumerator()).Returns(() => ((IEnumerable<Guid>)array).GetEnumerator());
```
MichaelKetting, line 37: Would it have been helpful to extract this into a generic helper method for this test fixture and reuse?

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,239 @@
...
+    var objects =
+        from o in QueryFactory.CreateLinqQuery<ClassWithAllDataTypes>()
+        where values.Contains(o.ExtensibleEnumProperty)
+        select o;
+
+    CheckQueryResult(
+        objects,
+        new ObjectID(typeof(ClassWithAllDataTypes), new Guid("583EC716-8443-4B55-92BF-09F7C8768529")));
+  }
+
```
MichaelKetting, line 207: I think we should have a string example as well. and possilby one with binary, if there is a good (=short) value available in the test data..

MichaelKetting, line 207: I think we should also have an example with a nullable value using NULL.

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,239 @@
...
+    var possibleItems = (IEnumerable<Guid>)collectionStub.Object;
+
+    var orders =
+        from o in QueryFactory.CreateLinqQuery<Order>()
+        where possibleItems.Contains((Guid)o.ID.Value)
+        select o;
+
+    CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3);
+  }
+
```
MichaelKetting, line 98: I'm missing a  test with ISet / IReadOnlySet.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/StructuredTypeDefinitionFinderForCollectionOfScalars.cs
```diff
@@ -73,21 +73,29 @@ public StructuredTypeDefinitionFinderForCollectionOfScalars (ISingleScalarStruct
       if (value is not IEnumerable)
         return result;
```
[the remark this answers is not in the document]
martin-herwich: When you say "move down", do you mean create an "amend" commit for RM-9216?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
```diff
@@ -278,16 +277,16 @@ private TypeConverter GetTypeConverter (Type nullableValueType, Type underlyingT
           new AdvancedEnumConverter(enumType));
     }
 
-    private string GetStorageTypeStringForVarType (string varType, int? maxLength)
+    private string GetStorageTypeStringForVarType (string varType, int maxLength)
```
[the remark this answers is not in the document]
martin-herwich: Should the bugfix commit be in its own pull request? If so, target branch "develop" or our "quest" branch RM-7194?

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/StructuredTypeDefinitionFinderForCollectionOfScalars.cs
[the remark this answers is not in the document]
MichaelKetting: yes, amend earlier PR (if that's sensible/correct)

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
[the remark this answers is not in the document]
MichaelKetting: yes, dedicated PR. it can go into develop or be part of the PRs n the quest branch. either way, we wil integrate those anyway as individual merges on develop so it's only a question of when it will show up in the next build. I'm okay either way.

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/StructuredTypeDefinitionFinderForCollectionOfScalars.cs
```diff
@@ -73,21 +73,29 @@ public StructuredTypeDefinitionFinderForCollectionOfScalars (ISingleScalarStruct
...
+
+      if (value is ICollection)
+      {
+        if(type.CanAscribeTo(typeof(IEnumerable<>)))
+          result.itemType = type.GetAscribedGenericArguments(typeof(IEnumerable<>)).Single();
+        else
+          result.itemType = typeof(object);
       }
 
       return result;
```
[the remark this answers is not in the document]
martin-herwich: I think that the potential performance benefit it definitely outweighed by the reduced readability. This code is only run when there is a database query to begin with, so we are already micro-optimizing. Plus, we have early returns for all scalar values, so this is only for a query parameter with a multi-item value.

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/ObjectIDIntegrationTest.cs
martin-herwich: I am happy to switch back to NotSupportedException, but if I recall correctly, we talked about this.
As for the internal class: this is an extension point; the exception is only thrown when the developer tries to use a parameter value that is a collection of objects whose type has no corresponding `TableTypeDefinition`; the exception is supposed to help the developer realize that he needs to decorate `SimpleTypeQueryParameterRecordDefinitionFinder` with a class that maps the item type to a `TableTypeDefinition`.

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,318 @@
...
+    var orders =
+        from o in QueryFactory.CreateLinqQuery<Order>()
+        where possibleItems.Contains((Guid)o.ID.Value)
+        select o;
+
+    Assert.That(() => CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3),
+        Throws.InstanceOf<RdbmsProviderException>()
+            .With.Message.Contains("Error while executing SQL command: Violation of UNIQUE KEY constraint")
+            .With.Message.Contains("Cannot insert duplicate key in object 'dbo.@1'. The duplicate key value is (83445473-844a-4d3f-a8c3-c27f8d98e8ba)."));
```
MichaelKetting, line 116: just a note: this might one day break due to changes in sql server messages. if it does, we might want to look into checking less details in the message.

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/TableValuedParameterIntegrationTest.cs
```diff
@@ -0,0 +1,318 @@
...
+    var orders =
+        from o in QueryFactory.CreateLinqQuery<Order>()
+        where possibleItems.Contains((Guid)o.ID.Value)
+        select o;
+
+    CheckQueryResult(orders, DomainObjectIDs.Order1, DomainObjectIDs.Order3);
+  }
+
+  [Test]
+  public void Contains_WithSetOfT_UsesTableTypeWithUniqueConstraint ()
```
MichaelKetting, line 103: liking the added tests

MichaelKetting (approved):

PR #1168, opened 2024-06-24
RM-9242 Integration Tests for SQL queries with table-valued parameters

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9242

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/IntegrationTests/TableValuedParametersIntegrationTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/IntegrationTests/TableValuedParametersIntegrationTest.cs
@@ -1,0 +1,335 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.DomainImplementation;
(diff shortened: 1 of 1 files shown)


Remotion/Data/DomainObjects.UnitTests/IntegrationTests/TableValuedParametersIntegrationTest.cs
```diff
@@ -0,0 +1,318 @@
+using System;
```
MichaelKetting, line 1: lic-header please

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/TableValuedParametersIntegrationTest.cs
```diff
@@ -0,0 +1,318 @@
...
+        Throws.InstanceOf<RdbmsProviderException>()
+            .With.Message.Contains("Error while executing SQL command: Violation of UNIQUE KEY constraint")
+            .With.Message.Contains("Cannot insert duplicate key in object 'dbo.@1'. The duplicate key value is (83445473-844a-4d3f-a8c3-c27f8d98e8ba)."));
+  }
+#endif
+
+  [Test]
+  public void Contains_UsingTableValuedParameter_NullValue ()
+  {
+    var orderNumbers = new int?[] { 2, null, 99 };
```
MichaelKetting, line 134: Reminder needed: we discussed the option of having null values and updated the type in the actual tvp type in the database. Why did we decide against supporting it as a value in memory?

MichaelKetting (changes requested):
got a question about nulls

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/TableValuedParametersIntegrationTest.cs
martin-herwich: What would the null value be used for? Keep in mind that this is about all columns in the row being `NULL`, because the collection item being `null` is what causes the exception.

At the moment, we only support TVP for `WHERE IN` clauses, where `NULL` just does not work:
```
declare @tbl table ([Value] int NULL)
insert into @tbl values (NULL)
select * from [TableWithAllDataTypes] where NaInt32WithNullValue IN (SELECT [Value] FROM @tbl)
```
does not return any row, even though there are 2 rows where `NaInt32WithNullValue` is `NULL`

For `JOIN` clauses, `NULL`-rows hardly make sense because there is nothing to join on; even though we can write a join condition that joins on `NULL`, this seems like a purely hypothetical scenario.

For `FROM` clauses, `NULL`-rows may make some sense in some fringe scenarios, but any scenario where we select from a table-valued parameter without joining (and thus filtering out the `NULL`-rows, see above) seems completely redundant.

martin-herwich: As discussed, we do not want to support `null` items, but we should throw a `NotSupportedException` that explains to the developer why `null` items are not supported.

MichaelKetting (approved):

PR #1170, opened 2024-06-27
Feature/rm 9247 add editorconfig license information instead of licenseheader files

Frazerus:
https://re-motion.atlassian.net/browse/RMTEAM-877


martin-herwich (approved):

SharedSource/Core/SharedSource.Core.csproj
MichaelKetting, line 1: You dropped the ReSharper licenseheader file in share source but not the Apache2.0 licenseheader for the other shared source files / did not add an editor.cofnig for them (Core/Development/UnitTests). It looks like this is an oversight?

MichaelKetting (changes requested):
I found a left-over shared-source import in 
Remotion\ObjectBinding\Web.Development.WebTesting.IntegrationTests\App_Packages
after reviwing the list of licenseheader-files.

It's something we can simply switch to using a locally linked usage like it is done with SharedSource/NUnit in the same project. Since we're touching the licenseheader files, it would be great if you could create a ticket and clean this up in the current sprint. Should only take a couple of minutes :)

SharedSource/Core/SharedSource.Core.csproj
Frazerus: Yea, I did not find that license file when looking through with solutionwide search. Weird. Fixed now

MichaelKetting (approved):

PR #1182, opened 2024-07-24
feature/RM-9263-Upgrade-System.Text.Json-v6.0

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-9263

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


Directory.Packages.props
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
```
a-ctor, line 5: Do you want to enable that proactively? Just so you know it probably does not do anything right now as it only works for `<PackageReference>`. So right now we don't actually need this change

a-ctor (approved):

Directory.Packages.props
MichaelKetting: @a-ctor am confused: the problem exists in Remotion.Web.Development.WebTesting via transitive dependency and this project is using SDK Style project format with PackageReference.

PR #1185, opened 2024-08-02
Feature/rm 9277 add support for asp net core hosting

Frazerus:


Build/Build.cs
```diff
@@ -263,10 +263,10 @@ public override void ConfigureTestMatrix (TestMatricesBuilder builder)
...
         new TestDimension[,] // todo docker images need to be wired to the config file
         {
-            { Chrome, NET48, Debug, x86, NoDB, EnforcedLocalMachine(Docker_Win_NET48) },
-            { Chrome, NET481, Release, x86, NoDB, EnforcedLocalMachine(Docker_Win_NET481) },
-            { Firefox, NET48, Release, x64, NoDB, EnforcedLocalMachine(Docker_Win_NET481) },
-            { Edge, NET48, Release, x64, NoDB, EnforcedLocalMachine(Docker_Win_NET48) },
+            { Chrome, NET8_0, Debug, x86, NoDB, EnforcedLocalMachine(Docker_Win_NET8_0) },
+            { Chrome, NET8_0, Release, x86, NoDB, EnforcedLocalMachine(Docker_Win_NET8_0) },
+            { Firefox, NET8_0, Release, x64, NoDB, EnforcedLocalMachine(Docker_Win_NET8_0) },
+            { Edge, NET8_0, Release, x64, NoDB, EnforcedLocalMachine(Docker_Win_NET8_0) },
```
a-ctor, line 269: Using .NET here as target might not work if the test projects target `net8.0-windows` instead. Until we do the .NET support we should be able to keep using `NET48`/`NET481` instead of `NET8_0`

a-ctor (changes requested):

Remotion/Web/Development.WebTesting.UnitTests/App.config
```diff
@@ -13,7 +13,6 @@
       headless="true"
       webApplicationRoot="http://localhost:60401/"
       screenshotDirectory=".\WebTestingOutput">
-    <hosting name="IisExpress" type="IisExpress" port="60401" />
     <testSiteLayout rootPath=".\Some\Path">
```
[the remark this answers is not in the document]
Frazerus: Just a missclick it seems

Build/Build.cs
Frazerus: Sure. Changed it because we have a test which checks for the same platform for server and instance, it seems the server is always run in 64bit. Do we still need to configure something here? Why do we even support 32 bit anymore?

Frazerus: Changing to NET8_0 fixed this

a-ctor (approved):

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/AspNetDockerContainerConfigurationParameters.cs
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Collections.Generic;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.HostingStrategies.DockerHosting;
+
+public class AspNetDockerContainerConfigurationParameters : DockerContainerConfigurationParameters
```
MichaelKetting, line 9: xmldoc please

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/AspNetDockerContainerConfigurationParameters.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+  public string PathToExecutable { get; set; }
```
MichaelKetting, line 11: ExecutablePath

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/AspNetDockerContainerConfigurationParameters.cs
```diff
@@ -0,0 +1,28 @@
...
+      int webApplicationPort,
+      string dockerImageName,
+      string? dockerIsolationMode,
+      string? hostname,
+      bool is32BitProcess,
+      IReadOnlyCollection<string> mounts,
+      string pathToExecutable)
+      : base(absoluteWebApplicationPath, webApplicationPort, dockerImageName, dockerIsolationMode, hostname, is32BitProcess, mounts)
+  {
+    ArgumentUtility.CheckNotNull("pathToExecutable", pathToExecutable);
```
MichaelKetting, line 24:
```suggestion
    ArgumentUtility.CheckNotNullOrEmpty("pathToExecutable", pathToExecutable);
```

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerCommandLineClient.cs
```diff
@@ -86,20 +89,29 @@ public void Pull (string imageName)
 
       if (mounts.Any())
       {
-        var mountFlags = string.Join(" ", mounts.Select(kvp => $@"-v ""{kvp.Key}"":""{kvp.Value}"""));
+        var mountFlags = string.Join(" ", mounts.Select(kvp => $@"-v ""{kvp.Key.Trim('\\')}"":""{kvp.Value.Trim('\\')}"""));
```
MichaelKetting, line 92:
```suggestion
        var mountFlags = string.Join(" ", mounts.Select(kvp => $@"-v ""{kvp.Key.Trim('\\')}"":""{kvp.Value.Trim('\\')}"""));
```

The key should only be single-word, no backslashes, correct? It's true that this is defensive coding. The question is if we are trying to make the code robust against formatting issues (e.g. trailing backslash in a path) or protect it against injection of exploits. I would suggest only protecting against formatting, thus the key would be taken at face value, no quotes, not handling of special characters.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerCommandLineClient.cs
```diff
@@ -54,19 +54,22 @@ public void Pull (string imageName)
...
         string? args)
     {
       ArgumentUtility.CheckNotNull("ports", ports);
       ArgumentUtility.CheckNotNull("mounts", mounts);
+      ArgumentUtility.CheckNotNull("environmentVariables", environmentVariables);
       ArgumentUtility.CheckNotNullOrEmpty("imageName", imageName);
       ArgumentUtility.CheckNotEmpty("hostname", hostname);
       ArgumentUtility.CheckNotEmpty("entryPoint", entryPoint);
-      ArgumentUtility.CheckNotEmpty("args", args);
+      ArgumentUtility.CheckNotEmpty("workingDirectory", workingDirectory);
```
MichaelKetting, line 72:
```suggestion
      ArgumentUtility.CheckNotEmpty("workingDirectory", workingDirectory);
      ArgumentUtility.CheckNotEmpty("args", args);
```

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerCommandLineClient.cs
```diff
@@ -54,19 +54,22 @@ public void Pull (string imageName)
...
         bool remove,
         string? entryPoint,
+        string? workingDirectory,
         string? args)
     {
       ArgumentUtility.CheckNotNull("ports", ports);
       ArgumentUtility.CheckNotNull("mounts", mounts);
+      ArgumentUtility.CheckNotNull("environmentVariables", environmentVariables);
       ArgumentUtility.CheckNotNullOrEmpty("imageName", imageName);
       ArgumentUtility.CheckNotEmpty("hostname", hostname);
```
MichaelKetting, line 70: please add a not-empty guard for isolationMode

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerCommandLineClient.cs
```diff
@@ -75,8 +78,8 @@ public void Pull (string imageName)
       if (remove)
         commandBuilder.Append("--rm").Append(' ');
 
-      if (isolationMode != null)
-        commandBuilder.Append($@"--isolation=""{isolationMode}""").Append(' ');
+      if (!string.IsNullOrEmpty(isolationMode))
```
MichaelKetting, line 81: revert please

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerCommandLineClient.cs
```diff
@@ -86,20 +89,29 @@ public void Pull (string imageName)
...
+        commandBuilder.Append($@"--workdir ""{workingDirectory.Trim('\\')}""").Append(' ');
+
       if (hostname != null)
         commandBuilder.Append($@"--hostname ""{hostname}""").Append(' ');
 
       commandBuilder.Append(imageName).Append(' ');
 
-      if (args != null)
-        commandBuilder.Append(args);
+      if (!string.IsNullOrEmpty(args))
```
MichaelKetting, line 113: revert please.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerCommandLineClient.cs
MichaelKetting: The idea was that a string typically is null to indicate not-there and empty is forbidden unless it's a valid value.

Down in the methid, the checks for not-null vs string.isnullorempty - that's a complete discussion and there's no hard and fast rule. probably, most of the time you just check for string.isnullorempty because it frees up the code to never run into a bug there. since the code already uses the not-null check instead, swithching this now seems odd.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/IDockerClient.cs
```diff
@@ -53,11 +55,13 @@ public interface IDockerClient
...
         [NotNull] IDictionary<int, int> ports,
         [NotNull] IDictionary<string, string> mounts,
+        IDictionary<string, string> environmentVariables,
         [NotNull] string imageName,
         [CanBeNull] string? isolationMode,
         [CanBeNull] string? hostname,
         bool remove,
         [CanBeNull] string? entryPoint,
+        string? workingDirectory,
         [CanBeNull] string? args);
```
MichaelKetting, line 65: please add the jetbrains annotations for consistency.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHostingStrategy.cs
```diff
@@ -57,49 +83,57 @@ public class DockerHostingStrategy : IHostingStrategy
...
-          absoluteWebApplicationPath,
-          port,
-          dockerImageName,
-          dockerIsolationMode,
-          hostname,
-          is32BitProcess,
-          mounts);
+      return innerType switch
+      {
+        null => CreateIisDockerContainer(),
```
MichaelKetting, line 88: do we need/want a default here? In particular cosnidering we might be going to switch to aspnetcore anyway in the future. Since the type is new, it's gonna be a breaking change and a required parameter, removing any ambiguity or questions.

In practice, the change wouild be harmless anyway since there's no usage yet.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHostingStrategy.cs
```diff
@@ -32,19 +32,45 @@ namespace Remotion.Web.Development.WebTesting.HostingStrategies
...
+      var dockerImageName = properties["dockerImageName"];
+      var dockerIsolationMode = properties["dockerIsolationMode"];
+      var dockerPullTimeout = TimeSpan.Parse(properties["dockerPullTimeout"]);
+      var hostname = properties["hostname"];
+#if NET6_0_OR_GREATER
+      var innerType = properties.GetValueOrDefault("innerType");
+#else
+      string? innerType;
+      properties.TryGetValue("innerType", out innerType);
+#endif
```
MichaelKetting, line 73: is this conditional helpful since it's just syntax?

MichaelKetting (changes requested):
I hope all the questions and suggestions make sense and are consistent. There's a couple of things which didn't get a deep discussion up front but I'm hoping the suggestions will simplify the concepts for our code and the user code.

Remotion/Web/Development.WebTesting.UnitTests/HostingStrategies/Configuration/TestSiteLayoutConfigurationTest.cs
```diff
@@ -57,13 +61,16 @@ public void CreateFromWebTestConfigurationSection_WithAbsolutePaths ()
       webTestSettingsStub
           .Setup(m => m.TestSiteLayout.Resources)
           .Returns(new List<string>() { @"C:\Some\Resource", @"Some\Other\Resource" });
-
+      webTestSettingsStub
+          .Setup(m => m.TestSiteLayout.PathFromRootToExecutable)
+          .Returns(@".\Some\Executable.exe");
```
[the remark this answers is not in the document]
Frazerus: Fixed by adding same behaviour as with resources.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingStrategy.cs
```diff
@@ -0,0 +1,81 @@
...
+    if (_testSiteLayoutConfiguration.ExecutablePath == _testSiteLayoutConfiguration.RootPath)
+    {
+      throw new ArgumentException(
+          "Executable path was not set while trying to host with ASP.NET core. Please add 'pathFromRootToExecutable' property to the testSiteLayout configuration parameter.");
+    }
+
+    var processWrapper = new AspNetCoreHostingProcessWrapper(
+        Path.Combine(_testSiteLayoutConfiguration.ExecutablePath),
+        _testSiteLayoutConfiguration.RootPath,
+        $"http://localhost:{_port}");
```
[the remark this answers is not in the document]
Frazerus: Https is used regardless, I think this should overall just be a different task.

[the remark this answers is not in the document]
Frazerus: The hostname settings is seemingly only used in docker hosting.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingStrategy.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    _webServerTask = processWrapper.RunWebServerAsync(_cancellationTokenSource.Token);
+  }
+
+  /// <inheritdoc/>
+  public void StopAndUndeployWebApplication ()
+  {
+    _cancellationTokenSource.Cancel();
+    if (_webServerTask != null && !_webServerTask.Wait(1000))
```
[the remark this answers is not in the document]
Frazerus: I honestly think that 1000 is easier to read. Everyone knows that these timings are always in milliseconds, and when you start reading something with a possible waiting time, you already expect there to be 1000 by default anyways.

Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/TestSiteLayoutConfiguration.cs
```diff
@@ -41,6 +43,8 @@ public TestSiteLayoutConfiguration ([NotNull] IWebTestSettings webTestSettings)
       Resources = webTestSettings.TestSiteLayout.Resources
           .Select(resourceElement => EnsureRootedPath(RootPath, resourceElement))
           .Select(rootedPath => new TestSiteResource(rootedPath)).ToArray();
+      var pathToExecutable = Path.Combine(RootPath, webTestSettings.TestSiteLayout.PathFromRootToExecutable);
```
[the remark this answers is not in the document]
Frazerus: It is already called Executable Path afterwards, and this is the `TestSiteLayoutConfiguration`, and this is the only usage where we read the config for any of these infos. I don't see the issue here.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingStrategy.cs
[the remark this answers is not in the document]
MichaelKetting: Compromise: `TimeSpan.FromMilliSeconds(1000)` ?

Reasoning: I did a lot of work with Microsoft stuff and there's a complete disregard for consistent units in the configurations. You get mixture of seconds and minutes or bytes and kilobytes, weird defaults ,etc. This is why we have always tried to remember to add the unit to the name of the property in the configration so you see it without thinking about it. 

It's true that the normal expectation is things are specified in milliseconds, but I have learned that it's easier to be explict all the time rather than having to remember if this is an oversight or a special case or actually a correct setting. The longer read doesn't hurt since it's not complicated and you immediately know what's going on.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/IDockerClient.cs
Frazerus: Why do we even use these? The types already give all the information anyways

Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/TestSiteLayoutConfiguration.cs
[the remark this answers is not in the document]
MichaelKetting: It's about expectations / easy of reading: "PathFromRootToExecutable" is a mouthful and you need to mentally parse the semantic of the config parameter's name and then use this information to build the correct value. This means everyone who sees this for the first time has to think about it. 

On the other hand, if we name it `ExecutablePath` here, too and require that the value is provided in its complete form, you see all the information in one place, no extra mental stack space required. 

I understand that the value is shorter with the current solution, but I'm not convinved that it's easier to maintain. 

Microsoft decided to do `processPath` in it's web.config
https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/web-config?view=aspnetcore-8.0

> Path to the executable that launches a process listening for HTTP requests. Relative paths are supported. If the path begins with `.`, the path is considered to be relative to the site root.

They do support both absolute and relative paths, so we could argue this should be an option. And since I only just stumbled (re-found) this setting: considering there's precedent, should we use `processPath` instead of ExecutablePath and mimick the semantics? (Note: the `.` semantic implies `.\bin\foo.exe` which is equivalent with `bin\foo.exe`, they just specified explicitly it for unknown purproses.)

And please remember: right now it is already possible to overrule the relative semantics by supplying an absolute path, so the naming is misleading anyway.

Remotion/Web/Development.WebTesting/HostingStrategies/IisDockerContainerWrapper.cs
```diff
@@ -28,109 +23,27 @@ namespace Remotion.Web.Development.WebTesting.HostingStrategies
...
-C:\Windows\System32\inetsrv\appcmd.exe set apppool /apppool.name:DefaultAppPool /enable32bitapponwin64:{_configurationParameters.Is32BitProcess};
-C:\Windows\System32\inetsrv\appcmd.exe add site /name:WebTestingSite /physicalPath:""{_configurationParameters.AbsoluteWebApplicationPath}"" /bindings:""http://*:{_configurationParameters.WebApplicationPort}"";
-C:\ServiceMonitor.exe w3svc;
-""");
-    }
-
-    /// <inheritdoc />
-    public void Dispose ()
+        IDockerClient docker,
+        DockerContainerConfigurationParameters configurationParameters)
```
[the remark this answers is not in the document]
Frazerus: I think that is useless code duplication, we only have a single usage in the docker hosting strategy, the main class where all the docker logic happens. It should be obvious enough.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/IDockerClient.cs
MichaelKetting: This goes back to when ReSharper wasn't considering the nullable reference type attributes in .net code. We can eventually remove them, but I am lobbying for keeping things consistent. starting to mix in the same signature isn't helpful so it's easier to apply them consistently when extending existing code and doing a general clean-up later. We also need to then verify that nullability works correctly for ReSharper even when the feature is disabled in the c# compiler. And yes, projects keep NRT disabled due to exessive nullability inconsistencies while the ReSharper attributes are still helpful. And since the web test library is intended to help test automation engineers, we try to be as graceful as possible.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHostingStrategy.cs
Frazerus: Since Netframework is dropping directly after this, I think it makes sense to go for the new syntax already.

Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/TestSiteLayoutConfiguration.cs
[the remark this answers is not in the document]
Frazerus: I can work with processPath

Remotion/Web/Development.WebTesting/HostingStrategies/IisDockerContainerWrapper.cs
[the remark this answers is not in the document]
MichaelKetting: okay, I can live with it. The code is isolated and likely never used outside of our infrastructure. 

Question though: would you make the same choice if the type didn't have that many ctor parameters and it would thus be less code to create a matching inheritance hierarchy?

Rule of thumb: it typically makes code relations easier to understand and follow if there's symetric inheritance hierarchies. You see from the naming which types work together.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHostingStrategy.cs
MichaelKetting: okay

Remotion/Web/Development.WebTesting/HostingStrategies/IisDockerContainerWrapper.cs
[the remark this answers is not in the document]
Frazerus: I think what matters more here is that this probably will never be used outside of our infrastructure, and would just mean another file to maintain.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingProcessWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Remotion.Utilities;
+using Remotion.Web.Development.WebTesting.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.HostingStrategies;
+
+/// <summary>
+/// Wraps an Asp Net Core instance hosting the web application under test.
```
MichaelKetting, line 13:
```suggestion
/// Wraps an ASP.NET Core instance hosting the web application under test.
```

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingProcessWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+  private readonly string _workingDirectory;
+  private readonly string _url;
+  private const int c_exitTimeoutInMilliseconds = 10_000;
+
+  /// <summary>
+  /// Initializes the wrapper, does not yet run the Asp.Net Core process.
+  /// </summary>
+  /// <param name="exePath">Absolute file path to the executable which runs the web server.</param>
+  /// <param name="workingDirectory">Working directory for the Asp.Net core process.</param>
+  /// <param name="url">The URL under which the test sites will be accessible from.</param>
```
MichaelKetting, line 27:
```suggestion
  /// <summary>
  /// Initializes the wrapper, does not yet run the ASP.NET Core process.
  /// </summary>
  /// <param name="exePath">Absolute file path to the executable which runs the web server.</param>
  /// <param name="workingDirectory">Working directory for the ASP.NET core process.</param>
  /// <param name="url">The URL under which the test sites will be accessible from.</param>
```

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingStrategy.cs
```diff
@@ -0,0 +1,84 @@
...
+/// </summary>
+public class AspNetCoreHostingStrategy : IHostingStrategy
+{
+  private readonly ITestSiteLayoutConfiguration _testSiteLayoutConfiguration;
+  private readonly int _port;
+  private readonly CancellationTokenSource _cancellationTokenSource;
+  private Task? _webServerTask;
+
+  /// <summary>
+  /// Initialize a new instance of the type.
```
MichaelKetting, line 27:
```suggestion
  /// Initializes a new instance of the type.
```

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingStrategy.cs
```diff
@@ -0,0 +1,84 @@
...
+    ArgumentUtility.CheckNotNull(nameof(testSiteLayoutConfiguration), testSiteLayoutConfiguration);
+    ArgumentUtility.CheckNotNull(nameof(port), port);
+
+    _testSiteLayoutConfiguration = testSiteLayoutConfiguration;
+    _port = port;
+    _cancellationTokenSource = new CancellationTokenSource();
+  }
+
+  /// <summary>
+  /// Initialize a new instance of the type. Used by the when the AspNetCore hosting type is selected.
```
MichaelKetting, line 42:
```suggestion
  /// Initializes a new instance of the type. Used by the when the <c>AspNetCore<c> hosting type is selected.
```

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingStrategy.cs
```diff
@@ -0,0 +1,84 @@
...
+  /// <inheritdoc/>
+  public void DeployAndStartWebApplication ()
+  {
+    if (_webServerTask != null)
+      return;
+
+    if (_testSiteLayoutConfiguration.ProcessPath == null)
+    {
+      throw new ArgumentException(
+          "Executable path was not set while trying to host with ASP.NET core. Please add 'processPath' property to the testSiteLayout configuration parameter.");
```
MichaelKetting, line 63:
```suggestion
          "The executable's path was not set while trying to host with ASP.NET Core. Please add the 'processPath' attribute to the 'testSiteLayout' configuration.");
```

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetDockerContainerWrapper.cs
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Remotion.Web.Development.WebTesting.HostingStrategies.DockerHosting;
+
+namespace Remotion.Web.Development.WebTesting.HostingStrategies;
+
+/// <summary>
+/// Represents an Asp.Net Core Docker container and manages its lifecycle.
```
MichaelKetting, line 11:
```suggestion
/// Represents an ASP.NET Core Docker container and manages its lifecycle.
```

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetDockerContainerWrapper.cs
```diff
@@ -0,0 +1,43 @@
...
+/// </summary>
+public class AspNetDockerContainerWrapper : DockerContainerWrapperBase
+{
+  private readonly string _processPath;
+
+  public AspNetDockerContainerWrapper (IDockerClient docker, AspNetDockerContainerConfigurationParameters configurationParameters)
+      : base(docker, configurationParameters)
+  {
+    _processPath = configurationParameters.ProcessPath ?? throw new ArgumentException(
+        "Executable path was not set while trying to host with ASP.NET core. Please add 'processPath' property to the testSiteLayout configuration parameter.");
```
MichaelKetting, line 21:
```suggestion
        "The executable's path was not set while trying to host with ASP.NET Core. Please add the 'processPath' attribute to the 'testSiteLayout' configuration.");
```

MichaelKetting, line 21: Given the length of the line, I would prefer to split this into an argument check and an assignment. Am not sure that combining these actually makes the code easier to read.

Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/ITestSiteLayoutConfiguration.cs
```diff
@@ -36,5 +36,10 @@ public interface ITestSiteLayoutConfiguration
     /// </summary>
     [NotNull]
     IReadOnlyList<ITestSiteResource> Resources { get; }
+
+    /// <summary>
+    /// Gets the path from the <see cref="RootPath"/> to the test site executable. Used with Kestrel-based hosting.
```
MichaelKetting, line 41:
```suggestion
    /// Gets the path to the test site's executable. The path may be relative to the <see cref="RootPath"/>. Used with Kestrel-based hosting.
```

Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/TestSiteLayoutConfiguration.cs
```diff
@@ -41,6 +43,18 @@ public TestSiteLayoutConfiguration ([NotNull] IWebTestSettings webTestSettings)
...
+
+      ProcessPath = GetRootedProcessPathOrNull(RootPath, webTestSettings.TestSiteLayout.ProcessPath);
+    }
+
+    private string? GetRootedProcessPathOrNull (string rootPath, string? processPath)
+    {
+      if (processPath == null)
+        return null;
+      if (!processPath.EndsWith(".exe"))
+        return null;
```
MichaelKetting, line 55: I would suggest this is an exception, otherwise you'd get difficult to trace errors.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerContainerWrapperBase.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.IO;
+using System.Threading;
+using log4net;
+using Remotion.Utilities;
+using Remotion.Web.Development.WebTesting.HostingStrategies.DockerHosting;
+
+namespace Remotion.Web.Development.WebTesting.HostingStrategies
+{
+  /// <summary>
+  /// Represents a docker container base which manages the lifecycle of a docker container.
```
MichaelKetting, line 14:
```suggestion
  /// Represents a Docker container base which manages the lifecycle of a Docker container.
```

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHostingStrategy.cs
```diff
@@ -57,49 +75,62 @@ public class DockerHostingStrategy : IHostingStrategy
...
+            is32BitProcess,
+            mounts,
+            testSiteLayoutConfiguration.ProcessPath);
+
+        return new AspNetDockerContainerWrapper(docker, configurationParameters);
+      }
+
+      DockerContainerWrapperBase ThrowNoInnerTypeException ()
+      {
+        throw new InvalidOperationException($"Could not resolve inner Type '{innerType}'. (Possible values: 'iisExpress' or 'aspnetcore')");
```
MichaelKetting, line 117: please check. I think this should be a ConfigurationException, but only if there's a) precendent here and b) ConfigurationException is actually available in .NET 8 without the extension packs.

MichaelKetting (approved):
Some string fix requests mainly. don't need to see it again.

PR #1186, opened 2024-08-05
RM-9281 Update CoreForms web dependency

a-ctor:
https://re-motion.atlassian.net/browse/RM-9281


MichaelKetting (approved):
The upgrade to v8.2.x on develop is sensible anyway , regardless of how we deal with the framework-dependent security fix.

PR #1187, opened 2024-08-07
RM-9238 Not selected tree nodes are now set as not selected

Frazerus:
https://re-motion.atlassian.net/browse/RM-9238

### DIFF
--- a/Remotion/Web/Core/UI/Controls/WebTreeView.cs
+++ b/Remotion/Web/Core/UI/Controls/WebTreeView.cs
@@ -834,8 +834,10 @@ namespace Remotion.Web.UI.Controls
       writer.AddAttribute(HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.TreeItem);
       writer.AddAttribute("tabindex", _focusededNode == node ? "0" : "-1");
 
-      if(node.IsSelected)
-        writer.AddAttribute(HtmlTextWriterAttribute2.AriaSelected, HtmlAriaSelectedAttributeValue.True);
+      var ariaSelectedAttributeValue = node.IsSelected
+        ? HtmlAriaSelectedAttributeValue.True
+        : HtmlAriaSelectedAttributeValue.False;
+      writer.AddAttribute(HtmlTextWriterAttribute2.AriaSelected, ariaSelectedAttributeValue);
 
       bool hasChildren = node.Children.Count > 0;
       bool isEvaluated = node.IsEvaluated;


Remotion/Web/Core/UI/Controls/WebTreeView.cs
```diff
@@ -834,8 +834,8 @@ private void RenderNodeLabel (HtmlTextWriter writer, WebTreeNode node, string no
       writer.AddAttribute(HtmlTextWriterAttribute2.Role, HtmlRoleAttributeValue.TreeItem);
       writer.AddAttribute("tabindex", _focusededNode == node ? "0" : "-1");
 
-      if(node.IsSelected)
-        writer.AddAttribute(HtmlTextWriterAttribute2.AriaSelected, HtmlAriaSelectedAttributeValue.True);
+      var ariaSelectedAttributeValue = node.IsSelected ? HtmlAriaSelectedAttributeValue.True : HtmlAriaSelectedAttributeValue.False;
+        writer.AddAttribute(HtmlTextWriterAttribute2.AriaSelected, ariaSelectedAttributeValue);
```
a-ctor, line 838:
```suggestion
      var ariaSelectedAttributeValue = node.IsSelected
        ? HtmlAriaSelectedAttributeValue.True
        : HtmlAriaSelectedAttributeValue.False;
      writer.AddAttribute(HtmlTextWriterAttribute2.AriaSelected, ariaSelectedAttributeValue);
```

a-ctor (approved):

Remotion/ObjectBinding/Web.IntegrationTests/BocTreeView/BocTreeViewTest.cs
```diff
@@ -48,6 +48,25 @@ public void WebTreeView_ExpansionOfEmptyEmitsNoItemFoundText ()
     Assert.That(() => ariaTreeAlertSpan.GetAttribute("aria-hidden"), Is.EqualTo("true").After(7).Seconds.PollEvery(200).MilliSeconds);
   }
 
+  [Test]
+  public void WebTreeView_Selected_HasAriaSelectedAttributeTrue ()
```
MichaelKetting, line 52: Since the change is static code rendered by the server, I would suggest not using a webtest to verify this. I know we don't test the rendering for TreeView any other way, but I would still dropp the test due to the runtime effort involved and setting a precent (i.e. you suddenly find webtests for stuff that should be tested as a unit test and are unsure when to use what). In theory, we could start building unit tests for WebTreeView rendering but the usual disclaimer applies: is this change big enough to warrent the expense?

MichaelKetting: If you want to keep the test, I would recommend adding a comment that this is only since we don't have a unit test for it.

MichaelKetting (approved):

PR #1189, opened 2024-08-08
RM-9257 Nova viso has remotion-scrollable-theme to use scrollbar them…

Frazerus:
[…e without remotion themed
](https://re-motion.atlassian.net/browse/RM-9257)


Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -471,13 +471,15 @@ img.disabled,
   filter: grayscale(100%);
 }
 
+.remotion-scrollable-theme,
```
a-ctor, line 474: Not sure about the same I prefer `remotion-scrollable` as you explicitly don't want any other theming so the `-theme` suffix implies something else.

a-ctor (approved):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
MichaelKetting, line 1: 1. In CssClassDefinitions.Scrollable, please update to the new css class.

2. In WebTreeViewRender, please replace
 `CssClass += " " + CssClassDefinition.Themed + " " + CssClassDefinition.Scrollable;` 
with
 `CssClass += " " + CssClassDefinition.Scrollable;`

On SingleTestTreeView.aspx, the expanded tree would get scrollbars. after the change, this should still work :)

MichaelKetting (approved):
please remember to document the css class in the task.
Found relevant changes in C# code, please update. Don't need to see again.

PR #1190, opened 2024-08-13
RM-7956 Switch Data.DomainObjects.Web.Test to NET 8.0

Frazerus:
https://re-motion.atlassian.net/browse/RM-7956


Remotion/Data/DomainObjects.Web.Test/App.config
```diff
@@ -6,13 +6,22 @@
...
+        <add assembly="System.Drawing.Primitives"/>
+      </assemblies>
+    </compilation>
+    <httpRuntime requestValidationMode="2.0"/>
     <authentication mode="Windows"/>
+    <httpModules>
+      <add name="Session" type="System.Web.SessionState.SessionStateModule"/>
+    </httpModules>
   </system.web>
   <system.webServer>
```
a-ctor, line 26: `<system.WebServer>` should be obsolete. But we need to move the handlers over just like in the other PRs.

Remotion/Data/DomainObjects.Web.Test/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+        virtualPath = "/";
+
+      LegacyAspNetInitialization.License = "I hereby confirm that I use CoreForms only for trial purposes and have read and accept the CoreForms Trial License.";
+
+      var legacyAspNetInitializationOptions = new LegacyAspNetInitializationOptions(virtualPath: virtualPath, physicalPath: Environment.CurrentDirectory);
+      LegacyAspNetInitialization.Initialize(legacyAspNetInitializationOptions);
+
+      var webControls = (Hashtable)System.Web.Configuration.WebConfigurationManager.GetSection("system.web/webControls");
+      webControls["clientScriptsLocation"] = virtualPath.TrimEnd('/') + webControls["clientScriptsLocation"];
+
```
a-ctor, line 29: Why do we need this magic? It feels like we need to do something here because that looks a bit hacky 😅

a-ctor (changes requested):

Remotion/Data/DomainObjects.Web.Test/Program.cs
MichaelKetting: That's me and I and to make it so I can specify the virtual directory of the application from outside . That way, I can host below root as well as in root. I was hoping to get it completed before now, alas...

This change will go into all the applications. If you want you can strip it out for now and keep it in a separate PR that does hosting cleanup for all applications.

Frazerus: Removed it for now, added a new issue

a-ctor (approved):

PR #1192, opened 2024-08-13
RM-7958 Switch SecurityManager.Clients.Web.Test to NET 8.0

Frazerus:
https://re-motion.atlassian.net/browse/RM-7958

### DIFF
--- a/SecurityManager/Clients.Web.Test/App.config
+++ b/SecurityManager/Clients.Web.Test/App.config
@@ -18,11 +18,17 @@
       <controls>
         <add tagPrefix="remotion" namespace="Remotion.ObjectBinding.Web.UI.Controls" assembly="Remotion.ObjectBinding.Web" />
         <add tagPrefix="remotion" namespace="Remotion.Web.UI.Controls" assembly="Remotion.Web" />
+        <add tagPrefix="asp" namespace="System.Web.UI.WebControls" assembly="CoreForms.Web" />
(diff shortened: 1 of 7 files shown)


SecurityManager/Clients.Web.Test/App.config
```diff
@@ -54,8 +60,11 @@
...
       -->
     <customErrors mode="Off" />
-    <httpRuntime requestValidationMode="2.0" targetFramework="4.8" />
+    <httpRuntime requestValidationMode="2.0"/>
     <globalization requestEncoding="utf-8" responseEncoding="utf-8" culture="de-AT" uiCulture="de-AT" />
+    <httpHandlers>
+      <add verb="*" path="*.wxe" type="Remotion.Web.ExecutionEngine.WxeHandler, Remotion.Web" />
+    </httpHandlers>
   </system.web>
   <system.webServer>
```
a-ctor, line 69: I don't think we need the `system.webServer` tag in the `App.config`. This was read by IIS in the `Web.config` and used for running in integrated mode but should no longer be necessary because we don't have the same IIS integration. This is also the reason why you had to move the `handlers` section up. 

We also might need the other registered handler for something.

a-ctor (changes requested):
SVG files won't load correctly because they are served with an incorrect MIME type. Search for `Application_PostRequestHandlerExecute` on how to fix this issue.

a-ctor (approved):

PR #1197, opened 2024-08-21
Feature/rm 9288 add type converter for object

Frazerus:
https://re-motion.atlassian.net/browse/RM-9288

### DIFF
--- a/Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
+++ b/Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.ComponentModel;
+using System.Globalization;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Infrastructure;
+
+/// <summary> Specialization of <see cref="TypeConverter"/> for conversions from and to <see cref="ObjectID"/>. </summary>
+public class ObjectIDConverter : TypeConverter
+{
+  public override bool CanConvertFrom (ITypeDescriptorContext? context, Type sourceType)
+  {
+    ArgumentUtility.CheckNotNull(nameof(sourceType), sourceType);
+
+    return sourceType == typeof(string) || base.CanConvertFrom(context, sourceType);
+  }
+
+  public override bool CanConvertTo (ITypeDescriptorContext? context, Type? destinationType)
+  {
+    if (destinationType == null)
+      return false;
+
+    return destinationType == typeof(string) || base.CanConvertTo(context, destinationType);
+  }
+
+  public override object? ConvertFrom (ITypeDescriptorContext? context, CultureInfo? culture, object? value)
+  {
+    if (value == null)
+      return null;
+
+    if (value is string valueString)
+    {
+      return ObjectID.Parse(valueString);
+    }
+
+    return base.ConvertFrom(context, culture, value);
+  }
+
+  public override object? ConvertTo (ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+  {
+    ArgumentUtility.CheckNotNull(nameof(destinationType), destinationType);
+
+    if (value == null)
+      return null;
+
+    var objectID = (ObjectID)value;
+
+    if (destinationType == typeof(string))
+      return objectID.ToString();
+
+    return base.ConvertTo(context, culture, value, destinationType);
+  }
+}
--- a/Remotion/Data/DomainObjects.UnitTests/Infrastructure/ObjectIDConverterTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Infrastructure/ObjectIDConverterTest.cs
@@ -1,0 +1,180 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.ComponentModel.Design.Serialization;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Infrastructure;
+using Remotion.Data.DomainObjects.UnitTests.TestDomain;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Infrastructure;
+
+[TestFixture]
+public class ObjectIDConverterTest : StandardMappingTest
+{
+  [Test]
+  public void CanConvertFrom_String ()
+  {
+    var converter = new ObjectIDConverter();
+
+    var result = converter.CanConvertFrom(null, typeof(string));
+
+    Assert.That(result, Is.True);
+  }
+
+  [Test]
+  public void CanConvertFrom_InstanceDescriptor ()
+  {
+    var converter = new ObjectIDConverter();
+
+    var result = converter.CanConvertFrom(null, typeof(InstanceDescriptor));
+
+    Assert.That(result, Is.True);
+  }
+
+  [Test]
+  [TestCase(typeof(int))]
+  [TestCase(typeof(object))]
+  [TestCase(typeof(Enum))]
+  [TestCase(typeof(DateTime))]
+  public void CanConvertFrom_OtherTypes (Type type)
+  {
+    var converter = new ObjectIDConverter();
+
+    var result = converter.CanConvertFrom(null, type);
+
+    Assert.That(result, Is.False);
+  }
+
+  [Test]
+  public void CanConvertTo_String ()
+  {
+    var converter = new ObjectIDConverter();
+
+    var result = converter.CanConvertTo(null, typeof(string));
+
+    Assert.That(result, Is.True);
+  }
+
+  [Test]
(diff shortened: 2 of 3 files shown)


Remotion/Data/DomainObjects/Utilities/ObjectIDConverter.cs
```diff
@@ -0,0 +1,56 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.ComponentModel;
+using System.Globalization;
+using Remotion.Utilities;
+
+namespace Remotion.Data.DomainObjects.Utilities;
```
a-ctor, line 8: Not sure if we need a new namespace for just this file. With `WebStringConverter` we just placed it by `WebString`. As such, I would suggest moving it to the root too. ping @MichaelKetting

Remotion/Data/DomainObjects/Utilities/ObjectIDConverter.cs
```diff
@@ -0,0 +1,56 @@
...
+
+public class ObjectIDConverter : TypeConverter
```
a-ctor, line 10: Missing XMl doc

Remotion/Data/DomainObjects/Utilities/ObjectIDConverter.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+  private readonly Type _stringType = typeof(string);
```
a-ctor, line 12: Not sure why `typeof(string)` is in a field. I don't think it adds any benefit and can be inlined instead. If you really want to use a field to remove the duplication, it should be `static` instead. But `typeof(...)` is a special case where the compiler can generate efficient code anyways so a field is not needed anyway. For example:

```
public Type M() => typeof(string);
```

compiles down to a constant move:

```
C.M()
    L0000: mov eax, 0xa000014
    L0005: ret
```

Remotion/Data/DomainObjects/Utilities/ObjectIDConverter.cs
```diff
@@ -0,0 +1,56 @@
...
+
+  public override bool CanConvertFrom (ITypeDescriptorContext? context, Type sourceType)
+  {
+    ArgumentUtility.CheckNotNull(nameof(sourceType), sourceType);
+
+    return sourceType == _stringType;
```
a-ctor, line 18: I don't know if we need it but in `WebStringConverter` we also did base calls as `TypeConverter` has special handling for `InstanceDescriptor`. It's probably not necessary but I would add the base calls for consistency. Also affects other methods as well.

a-ctor (changes requested):

Remotion/Data/DomainObjects/Utilities/ObjectIDConverter.cs
MichaelKetting: Namespace...

I don't think it has to be in the root namespace. 


I suggest we move the converter to the infrastructure namespace.

Remotion/Data/DomainObjects/Utilities/ObjectIDConverter.cs
MichaelKetting: I concur with consistency.

MichaelKetting: Hmm... for practical purposes, we don't need to worry about `InstanceDescriptor`. It's used with the WebForms designer etc. 

If you want to handle it, you can do it identically to PlainTextStringConverter implementation, the same semantics apply for both cases (a string representation gets convertered to an object).

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Remotion.Data.DomainObjects.Infrastructure;
+
+/// <summary> Specialization of <see cref="TypeConverter"/> for conversions from and to <see cref="ObjectID"/>. </summary>
+public class ObjectIDConverter : TypeConverter
+{
+  /// <summary> Test: Is <paramref name="sourceType"/> a <see cref="String"/>? Only <see cref="String"/>s can be converted to <see cref="ObjectID"/>s.</summary>
+  /// <param name="context"> An <see cref="ITypeDescriptorContext"/> that provides a format context. </param>
+  /// <param name="sourceType"> The <see cref="Type"/> of the value to be converted into an <see cref="ObjectID"/> type. </param>
+  /// <returns> <see langword="true"/> if the conversion is supported. </returns>
```
a-ctor, line 16: As these are implementations we don't need to reintroduce the XML docs. We can add `/// <inheritdoc />` if you want but I think that is also the default so...

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+  /// <param name="value"> The source value. </param>
+  /// <returns> An <see cref="ObjectID"/> value.  </returns>
+  /// <exception cref="NotSupportedException"> The conversion could not be performed. </exception>
+  public override object? ConvertFrom (ITypeDescriptorContext? context, CultureInfo? culture, object? value)
+  {
+    if (value == null)
+      return null;
+
+    if (!CanConvertFrom(context, value.GetType()))
+      throw new NotSupportedException($"Value of type '{value.GetType()}' cannot be converted to type 'ObjectID'.");
```
a-ctor, line 48: I don't think we need this anymore? As we now do the base call, it should handle the case where we have an incorrect type.

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    if (value is string valueString)
+    {
+      if (ObjectID.TryParse(valueString, out var objectID))
+        return objectID;
+      else
+        throw new NotSupportedException($"String '{valueString}' cannot be converted to type 'ObjectID'.");
```
a-ctor, line 55: I would just use `ObjectID.Parse` here, I don't think the specific exception adds much here.

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+  public override object? ConvertTo (ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+  {
+    ArgumentUtility.CheckNotNull(nameof(destinationType), destinationType);
+
+    if (value == null)
+      return null;
+
+    if (!CanConvertTo(destinationType))
+      throw new NotSupportedException($"This TypeConverter cannot convert ObjectIDs to type '{destinationType}'.");
+
```
a-ctor, line 77: Same as above, not needed as we have the base call.

a-ctor (changes requested):

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
MichaelKetting: The official FormatException actually contains more helpful details, so it's actually the better choice, not just the less wordy implementation.

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
MichaelKetting: I'm not sure that accurate. Looking at 
```
public virtual object ConvertTo(
      ITypeDescriptorContext context,
      CultureInfo culture,
      object value,
      Type destinationType)
    {
      if (destinationType == (Type) null)
        throw new ArgumentNullException(nameof (destinationType));
      if (!(destinationType == typeof (string)))
        throw this.GetConvertToException(value, destinationType);
      if (value == null)
        return (object) string.Empty;
      return culture != null && culture != CultureInfo.CurrentCulture && value is IFormattable formattable ? (object) formattable.ToString((string) null, (IFormatProvider) culture) : (object) value.ToString();
    }
```
And there's an exception for anything but string.

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public override object? ConvertTo (ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType)
+  {
+    ArgumentUtility.CheckNotNull(nameof(destinationType), destinationType);
+
+    if (value == null)
+      return null;
+
+    if (!CanConvertTo(destinationType))
```
MichaelKetting, line 48: since we're using explicit checks most of the time, plesae update to test for string explicity. We can also do a hardcast for value (see WebStringConverter)

Remotion/Data/DomainObjects/Infrastructure/ObjectIDConverter.cs
Frazerus: Thats weird then, I tried it and got no exception in the test, but maybe I messed up the test setup somehow.

Frazerus: Yea, worked now, I probably messed up the test somehow

a-ctor (approved):

PR #1199, opened 2024-08-23
RM-9295 Fix bug in remotion static file helper method

a-ctor:
https://re-motion.atlassian.net/browse/RM-9295


MichaelKetting (approved):
The code looks okay but I freely admit that I'm not deep enough into ASP.NET Core to understand the implications. This would make a great topic for a Dev Jour fixe or similiar. Please put it on the agenda or schedule a separate know how meeting for it.

PR #1200, opened 2024-08-23
Bugfix/rm 8853 boc boolean value boc enum value and boc list use incorrect elements as target for the label for attribute

Frazerus:
https://re-motion.atlassian.net/browse/RM-8853

Implementation for the non hover issues

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocBooleanValue.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocBooleanValue.ts
@@ -106,13 +106,10 @@ class BocBooleanValue_Resource
             description = trueDescription;
     } // RM-7676: Handle invalid check box states in BocBooleanValue_Resource.SelectNextCheckboxValue
 
-
     checkboxSpan.setAttribute('aria-checked', checkedState!);
 
-    const isHover = icon.src == icon.dataset['srcHover'];
-    icon.dataset["src"] = iconSrc!;
-    icon.dataset['srcHover'] = iconHoverSrc!;
-    icon.src = isHover ? icon.dataset["src"] : icon.dataset['srcHover'];
+    icon.style.setProperty("--standard-background-image", "url('" + iconSrc! + "')");
+    icon.style.setProperty("--hover-background-image", "url('" + iconHoverSrc! + "')");
 
     if (label == null)
     {
@@ -201,17 +198,5 @@ class BocBooleanValue
       event.cancelBubble = true;
       event.returnValue = false;
     }
-  }
-
-  public static OnMouseOver (context: HTMLElement): void
-  {
-    const img = context.querySelector('img')!;
-    img.src = img.dataset['srcHover']!;
-  }
-
-  public static OnMouseOut (context: HTMLElement): void
-  {
-    const img = context.querySelector('img')!;
-    img.src = img.dataset["src"]!;
   }
 }
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocBooleanValue.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocBooleanValue.css
@@ -6,6 +6,7 @@
   vertical-align: middle;
   display: inline-block;
   white-space: nowrap;
+  --standard-background-image: initial;
 }
 
 .bocBooleanValue.readOnly
@@ -50,6 +51,11 @@
 .bocBooleanValue:not(.readOnly) span[role=checkbox]
 {
   cursor: default;
+}
+
+.bocBooleanValue input + span[role=checkbox] img
+{
+  background-image: var(--standard-background-image);
 }
 
 .bocBooleanValue img
--- a/Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
+++ b/Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
@@ -445,7 +445,17 @@ namespace Remotion.ObjectBinding.Web.UI.Controls
     [Browsable(false)]
     public string? FocusID
     {
-      get { return IsReadOnly ? null : GetValueName(); }
+      get
+      {
+        if (IsReadOnly)
+          return null;
+
+        // Note: The ID is calculated by System.Web.UI.ListBox
+        if (ListControlStyle.ControlType == ListControlType.RadioButtonList)
+          return GetValueName() + "_0";
+
+        return GetValueName();
+      }
     }
 
     /// <summary>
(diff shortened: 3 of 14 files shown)


Remotion/ObjectBinding/Web/UI/Controls/BocEnumValue.cs
```diff
@@ -445,7 +445,15 @@ public override Control TargetControl
     [Browsable(false)]
     public string? FocusID
     {
-      get { return IsReadOnly ? null : GetValueName(); }
+      get
+      {
+        if (IsReadOnly)
+          return null;
+        if (ListControlStyle.ControlType == ListControlType.RadioButtonList)
+          return Controls.Count > 0 ? GetValueName() + "_0" : null;
```
MichaelKetting, line 453:
```suggestion
        if (IsReadOnly)
          return null;
        if (ListControlStyle.ControlType == ListControlType.RadioButtonList)
          return Controls.Count > 0 ? GetValueName() + "_0" : null;
```
```suggestion
        if (IsReadOnly)
          return null;

        if (ListControlStyle.ControlType == ListControlType.RadioButtonList)
          return Controls.Count > 0 ? GetValueName() + "_0" : null;
```

MichaelKetting, line 453: Please add a comment that the ID is calculated by System.Web.U.ListBox. Some day I want to rewrite the rendering without relying on ASP.NET WebForms, then we could conceivably remove this duplication.

Remotion/Web/Core/UI/Controls/SmartLabel.cs
```diff
@@ -184,12 +184,15 @@ protected virtual void LoadResources (IResourceManager resourceManager)
...
-      if (((ISmartControl)target).UseLabel)
-        return ((IFocusableControl)target).FocusID;
+      if (!smartTarget.UseLabel)
+        return null;
+
+      if (target is IFocusableControl focusableControl)
+        return focusableControl.FocusID;
 
-      return null;
+      return target.ClientID;
```
MichaelKetting, line 195: I updated the docs in https://re-motion.atlassian.net/browse/RM-8853 for the "best" implementation. I think it already covers most of the notes but I added a case for IFocusableControl without ISmartControl. Please verify the logic against the updated spec and check if you want to extract this into a separate bug. If you see something strange, please question :)

MichaelKetting (commented):
Looks good so far. Please double-check against amended spec in https://re-motion.atlassian.net/browse/RM-8853

Frazerus:
### No hover:
image

### Old hover effect:
Hover over label:
image
Hover over specific item:
image

### Trial hover effect on bocEnumValue
Hover over label:
image
Hover over specific item:
image
Hover over selected item:
image

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -276,6 +284,8 @@ input[type="password"].remotion-themed,
   background-image: url('../Image/sprite.svg#CheckBoxFalse');
 }
 
+.remotion-themed input[type=checkbox]:not(:disabled):hover + label::before,
+.remotion-themed input[type=checkbox]:not(:disabled):hover + span::before,
```
MichaelKetting, line 288: A few lines up we have 
```
.remotion-themed input[type=checkbox] + label:not(:hover)::before,
.remotion-themed input[type=checkbox] + span:not(:hover)::before,
```

I think are should hadd the not-hover rule too.
```
.remotion-themed input[type=checkbox]:not(:hover) + label::before,
.remotion-themed input[type=checkbox]:not(:hover) + span::before,
```
Please check this for the other added rules, too. 

The not-hover rules are added to make it explicit that it's not-hover / the default scenario, so while it's possible that there is no actual effect, having the rules symetrical makes the reasoning easier to follow. OR there is actually a selector-priority-based reason why we need them.

MichaelKetting (commented):
partial review. some fixed to come :)

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
MichaelKetting: Note: already pushed as a new commit to add missing stuff.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -19,6 +19,7 @@
   --color-primary-highlight: #218FC9;
   --color-primary-text: #FFFFFF;
   --color-highlight: #DEF3FE;
+  --color-sibling-hover: #f3faff;
```
MichaelKetting, line 22: Considering the set of colors was created based on a color palette and design, I prefer not to add new colors, in particular colors used only once. Am pushing a version where the entire background is highlighted when hovering the label and only the respective OPTION is highlighted when hovering the ELEMENT element.

MichaelKetting (changes requested):
I tweaked the CSS. please counter-review.
Some of the review-comments are already completed and only left open for reviewing.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
MichaelKetting: Comment completed. Please review.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
MichaelKetting: Comment completed. please review.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
Frazerus: Looks good

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocBooleanValue.ts
```diff
@@ -106,13 +106,10 @@ class BocBooleanValue_Resource
...
 
-
     checkboxSpan.setAttribute('aria-checked', checkedState!);
 
-    const isHover = icon.src == icon.dataset['srcHover'];
-    icon.dataset["src"] = iconSrc!;
-    icon.dataset['srcHover'] = iconHoverSrc!;
-    icon.src = isHover ? icon.dataset["src"] : icon.dataset['srcHover'];
+    icon.style.setProperty("--bocbooleanvalue-standard-background", "url('" + iconSrc! + "')");
+    icon.style.setProperty("--bocbooleanvalue-hover-background", "url('" + iconHoverSrc! + "')");
```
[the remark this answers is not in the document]
Frazerus: In that case, we would be unable to use the resouce manager thingy to change the urls, we need to set css urls programmatically.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocBooleanValue.css
MichaelKetting, line 1: We don't need hover in NovaGray :)

MichaelKetting (commented):
There's 19 or so red tests left so I'll wait with the approval until they're done.

MichaelKetting (approved):

PR #1214, opened 2024-08-28
Feature/rm 9268 upgrade nunit to v4.2

Frazerus:
https://re-motion.atlassian.net/browse/RM-9268

### DIFF
--- a/.nuke/parameters.ci.json
+++ b/.nuke/parameters.ci.json
@@ -8,5 +8,5 @@
   "TestTargetRuntimes": [ "NET8_0_WINDOWS" ],
   "TestBrowsers": [ "NoBrowser", "Chrome" ],
   "TestSqlServers": [ "NoDB", "SqlServer2016" ],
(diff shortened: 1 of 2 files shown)


a-ctor (approved):

Frazerus:
Already tested with multiple test categories ignored on TC.

Frazerus:
Well, I added the test category to be ignored for a TC run, checked that there were 1400 less tests and then removed the change, as we do not want to just ignore these tests I assume.

MichaelKetting:
> Well, I added the test category to be ignored for a TC run, checked that there were 1400 less tests and then removed the change, as we do not want to just ignore these tests I assume.

Resolved misunderstanding: we do want this change back in the CI runs because the change was removed with the update in the build infrastructure and should be re-integrated now. The browser-testbuilds run entire testsuite anyway.

MichaelKetting (approved):

PR #1216, opened 2024-09-04
RM-9312 Fix argument order in HtmlHelper asserts

a-ctor:
https://re-motion.atlassian.net/browse/RM-9312

### DIFF
--- a/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/HtmlHelper.cs
+++ b/Remotion/ObjectBinding/Web.UnitTests/UI/Controls/HtmlHelper.cs
@@ -26,8 +26,8 @@ namespace Remotion.ObjectBinding.Web.UnitTests.UI.Controls
   {
     public HtmlHelper ()
         : base(
-            (actual, expected, message, args) => Assert.That(actual, Is.EqualTo(expected), FormattableStringFactory.Create(message, args)),
-            (actual, expected, message, args) => Assert.That(actual, Is.GreaterThan(expected), FormattableStringFactory.Create(message, args)),
+            (expected, actual, message, args) => Assert.That(actual, Is.EqualTo(expected), FormattableStringFactory.Create(message, args)),
+            (left, right, message, args) => Assert.That(left, Is.GreaterThan(right), FormattableStringFactory.Create(message, args)),
             (actual, message, args) => Assert.That(actual, Is.Not.Null, FormattableStringFactory.Create(message, args)),
             (actual, message, args) => Assert.That(actual, Is.Null, FormattableStringFactory.Create(message, args)),
             (actual, message, args) => Assert.That(actual, Is.True, FormattableStringFactory.Create(message, args))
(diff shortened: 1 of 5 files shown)


Remotion/ObjectBinding/Web.UnitTests/UI/Controls/HtmlHelper.cs
```diff
@@ -26,8 +26,8 @@ public class HtmlHelper : HtmlHelperBase
   {
     public HtmlHelper ()
         : base(
-            (actual, expected, message, args) => Assert.That(actual, Is.EqualTo(expected), FormattableStringFactory.Create(message, args)),
-            (actual, expected, message, args) => Assert.That(actual, Is.GreaterThan(expected), FormattableStringFactory.Create(message, args)),
+            (expected, actual, message, args) => Assert.That(actual, Is.EqualTo(expected), FormattableStringFactory.Create(message, args)),
+            (expected, actual, message, args) => Assert.That(expected, Is.GreaterThan(actual), FormattableStringFactory.Create(message, args)),
```
MichaelKetting, line 30: This one seems fishy? Am on the browser so I can't verify

Remotion/Web/UnitTests/Core/UI/Controls/HtmlHelper.cs
```diff
@@ -25,8 +25,8 @@ public class HtmlHelper : HtmlHelperBase
   {
     public HtmlHelper ()
         : base(
-            (actual, expected, message, args) => Assert.That(actual, Is.EqualTo(expected), FormattableStringFactory.Create(message, args)),
-            (actual, expected, message, args) => Assert.That(actual, Is.GreaterThan(expected), FormattableStringFactory.Create(message, args)),
+            (expected, actual, message, args) => Assert.That(actual, Is.EqualTo(expected), FormattableStringFactory.Create(message, args)),
+            (expected, actual, message, args) => Assert.That(expected, Is.GreaterThan(actual), FormattableStringFactory.Create(message, args)),
```
MichaelKetting, line 29: Here too

MichaelKetting (changes requested):
The greater than compare reads weird. Request changes just to check and verify

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/HtmlHelper.cs
a-ctor: That's what happens when you use `expected` and `actual` for the greater than operation where the parameter names no longer make any sense. As the greater than operation is not symmetric and was working correctly already we needed to switch the parameters at the invocation as well.

The solution would be to introduce a new delegate with different parameter names (like `left` and `right`) that make more sense for greater than. Then this code wouldn't be so strange to read. Yesterday you did not want a new delegate so there is none in the PR. Do you want a delegate now? :D

MichaelKetting: Okay. Unless you want an extra delegate we can leave it  or  should we keep the delegate but rename the he parameters at the usage site?

MichaelKetting (approved):

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/HtmlHelper.cs
a-ctor: Okay I renamed the parameters `left` and `right` to make the usage sites more readable.

MichaelKetting (approved):

PR #1217, opened 2024-09-04
RM-9278 Fix bug in autocomplete where invalidate is called without a value change

a-ctor:
https://re-motion.atlassian.net/browse/RM-9278


MichaelKetting (approved):
It's *probably* perfect.

PR #1218, opened 2024-09-06
RM-9311 Update Domain.Web.Test database to support TVP

Frazerus:
https://re-motion.atlassian.net/browse/RM-9311


Frazerus:
Why do we only have a SetupDB script in the project?

a-ctor (approved):

MichaelKetting:
> Why do we only have a SetupDB script in the project?

@Frazerus Hmm.. yes, but the Setup-DB also does the DROP operations. The script is really old :) If you want to optimize / split stuff up, go for it. It's executed manually anyway.

MichaelKetting (approved):

PR #1219, opened 2024-09-09
feature/RM-9309-Replace-Eval-Statement-in-ListMenu-and-DropDownMenu

MichaelKetting:
https://re-motion.atlassian.net/browse/RM-9309

### DIFF
--- a/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
@@ -77,7 +77,7 @@ class DropDownMenu_ItemInfo
       public IsDisabled: boolean,
       public readonly Href: Nullable<string>,
       public readonly Target: Nullable<string>,
-      public readonly FallbackNavigationUrl: string,
+      public readonly ClickHandler: Nullable<() => boolean>,
       public readonly DiagnosticMetadata: Nullable<Dictionary<string | boolean>>,
       public readonly DiagnosticMetadataForCommand: Nullable<Dictionary<string | boolean>>)
   {
@@ -731,34 +731,21 @@ class DropDownMenu
     const anchor = document.createElement("a");
     anchor.setAttribute('role', 'menuitem');
     anchor.setAttribute('tabindex', '-1');
-    if (isEnabled)
+    if (isEnabled && itemInfo.Href !== null)
     {
-      anchor.setAttribute('href', itemInfo.FallbackNavigationUrl);
+      anchor.setAttribute('href', itemInfo.Href);
+      if (itemInfo.Target !== null)
+        anchor.setAttribute('target', itemInfo.Target);
     }
     else
     {
       anchor.setAttribute('aria-disabled', 'true');
     }
 
-    anchor.addEventListener ('click', DropDownMenu.OnItemClick);
+    const itemClickHandler = isEnabled ? itemInfo.ClickHandler : null;
+    anchor.addEventListener ('click', function (ev: MouseEvent) { DropDownMenu.OnItemClick(this, ev, itemClickHandler) });
 
     item.appendChild(anchor);
-    if (isEnabled && itemInfo.Href != null)
-    {
-      const isJavaScript = itemInfo.Href.toLowerCase().indexOf('javascript:') >= 0;
-      if (isJavaScript)
-      {
-        anchor.setAttribute('javascript', itemInfo.Href);
-      }
-      else
-      {
-        const href = itemInfo.Href;
-        const target = itemInfo.Target != null && itemInfo.Target.length > 0
-            ? itemInfo.Target
-            : '_self';
-        anchor.setAttribute('javascript', 'window.open (\'' + href + '\', \'' + target + '\');');
-      }
-    }
 
     if (itemInfo.Icon != null)
     {
@@ -830,24 +817,25 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement, ev: MouseEvent): boolean
+  private static OnItemClick (anchor: HTMLAnchorElement, ev: MouseEvent, clickHandler: Nullable<() => boolean>): boolean
   {
-    ev.preventDefault();
-
-    if (this.href == null || this.href === '')
+    if (anchor.href === null || anchor.href === '')
       return false;
 
     DropDownMenu._itemClicked = true;
     DropDownMenu.ClosePopUp(true);
+
+    let result = true;
     try
     {
-      eval(this.getAttribute('javascript')!);
+      if (clickHandler !== null)
+        result = clickHandler();
     }
     catch (e)
     {
     }
     setTimeout (function () { DropDownMenu._itemClicked = false; }, 10);
-    return false;
+    return result;
   }
 
   public static OnKeyDown (event: KeyboardEvent, dropDownMenuOrSelector: CssSelectorOrElement<HTMLAnchorElement>, getSelectionCount: Nullable<DropDownMenu_SelectionCountGetter>, hasDedicatedDropDownMenuElement: boolean): void
--- a/Remotion/Web/ClientScript/Scripts/ListMenu.ts
+++ b/Remotion/Web/ClientScript/Scripts/ListMenu.ts
@@ -34,9 +34,9 @@ class ListMenuItemInfo
     public readonly IconDisabled: string,
     public readonly RequiredSelection: number,
     public readonly IsDisabled: boolean,
-    public readonly Href: string,
-    public readonly Target: string,
-    public readonly FallbackNavigationUrl: string,
+    public readonly Href: Nullable<string>,
+    public readonly Target: Nullable<string>,
+    public readonly ClickHandler: Nullable<(this: HTMLAnchorElement, event: MouseEvent) => boolean>,
     public readonly DiagnosticMetadata: Dictionary<string | boolean>,
     public readonly DiagnosticMetadataForCommand: Dictionary<string | boolean>)
   {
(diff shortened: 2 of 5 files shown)


Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -731,34 +731,21 @@ class DropDownMenu
     const anchor = document.createElement("a");
     anchor.setAttribute('role', 'menuitem');
     anchor.setAttribute('tabindex', '-1');
-    if (isEnabled)
+    if (isEnabled && itemInfo.Href != null)
```
a-ctor, line 734: I would just use the value for the truthy check and remove the explict null check. Only thing to remember is that an empty string is also falsey but in this case I don't think that check could be good anyways.

```suggestion
    if (isEnabled && itemInfo.Href)
```

If you want to keep the explicit check please use `!==` to prevent type conversion issues.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -731,34 +731,21 @@ class DropDownMenu
...
     {
-      anchor.setAttribute('href', itemInfo.FallbackNavigationUrl);
+      anchor.setAttribute('href', itemInfo.Href);
+      if (itemInfo.Target != null)
+        anchor.setAttribute('target', itemInfo.Target);
```
a-ctor, line 738:
```suggestion
      anchor.href = itemInfo.Href;
      if (itemInfo.Target)
        anchor.target = itemInfo.Target;
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +817,25 @@ class DropDownMenu
     return item;
   }
 
-  private static OnItemClick (this: HTMLAnchorElement, ev: MouseEvent): boolean
+  private static OnItemClick (anchor: HTMLAnchorElement, ev: MouseEvent, clickHandler: Nullable<Function>): boolean
   {
-    ev.preventDefault();
-
-    if (this.href == null || this.href === '')
+    if (anchor.href == null || anchor.href === '')
```
a-ctor, line 822:
```suggestion
    if (!anchor.href)
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -77,7 +77,7 @@ class DropDownMenu_ItemInfo
       public IsDisabled: boolean,
       public readonly Href: Nullable<string>,
       public readonly Target: Nullable<string>,
-      public readonly FallbackNavigationUrl: string,
+      public readonly ClickHandler: Nullable<Function>,
```
a-ctor, line 80: `Function` represents `any` function and will prevent type checking. Specifying the type here is better.

```suggestion
      public readonly ClickHandler: Nullable<() => boolean>,
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +817,25 @@ class DropDownMenu
...
       return false;
 
     DropDownMenu._itemClicked = true;
     DropDownMenu.ClosePopUp(true);
+
+    var result = true;
     try
     {
-      eval(this.getAttribute('javascript')!);
+      if (clickHandler != null)
```
a-ctor, line 831:
```suggestion
      if (clickHandler)
```

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
```diff
@@ -830,24 +817,25 @@ class DropDownMenu
...
-    ev.preventDefault();
-
-    if (this.href == null || this.href === '')
+    if (anchor.href == null || anchor.href === '')
       return false;
 
     DropDownMenu._itemClicked = true;
     DropDownMenu.ClosePopUp(true);
+
+    var result = true;
```
a-ctor, line 828:
```suggestion
    let result = true;
```

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -34,9 +34,9 @@ class ListMenuItemInfo
     public readonly IconDisabled: string,
     public readonly RequiredSelection: number,
     public readonly IsDisabled: boolean,
-    public readonly Href: string,
-    public readonly Target: string,
-    public readonly FallbackNavigationUrl: string,
+    public readonly Href: Nullable<string>,
+    public readonly Target: Nullable<string>,
+    public readonly ClickHandler: Nullable<Function>,
```
a-ctor, line 39:
```suggestion
    public readonly ClickHandler: Nullable<() => boolean>,
```

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -114,31 +114,25 @@ class ListMenu
       const item = document.getElementById (itemInfo.ID) as HTMLElement;
       let anchor = item.children[0] as HTMLAnchorElement;
       const icon = anchor.children[0] as HTMLImageElement;
+      anchor.removeAttribute ('href');
+      anchor.removeAttribute ('target');
+      anchor.removeAttribute ('onclick');
+      anchor.onclick = null;
```
a-ctor, line 120: `removeAttribute` will also clear up the event handler so there is no need to set it to `null` afterwards.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -114,31 +114,25 @@ class ListMenu
...
-            if (itemInfo.Target != null)
-              anchor.target = itemInfo.Target;
-            anchor.removeAttribute ('javascript');
-            anchor.removeAttribute ('onclick');
-            anchor.onclick = null;
-          }
+          const clickHandler = itemInfo.ClickHandler;
+          anchor.onclick = function () {
+            return clickHandler();
+          };
```
a-ctor, line 135: We shouldn't use the `onclick` if we don't need to. We can use `addEventListener('click', handler)` instead. That assumes we run through this code only once. We should also not need the wrapper function and can pass `itemInfo.ClickHandler` in isntead.

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
```diff
@@ -415,20 +416,19 @@ private void AppendMenuItem (DropDownMenuRenderingContext renderingContext, Stri
...
           if (isPostBackCommand)
           {
-            // Clientside script creates an anchor with href="#" and onclick=function
-            // The function will be executed using eval(...) and must therefor not end with a return statement.
             string argument = menuItemIndex.ToString();
-            href = renderingContext.Control.Page!.ClientScript.GetPostBackClientHyperlink(renderingContext.Control, argument);
-            href = ScriptUtility.EscapeClientScript(href);
-            href = "'" + href + "'";
+            clickHandler = "function() { " + renderingContext.Control.Page!.ClientScript.GetPostBackEventReference(renderingContext.Control, argument) + "; return false; }";
+            href = "'" + ScriptUtility.EscapeClientScript(_fallbackNavigationUrlProvider.GetURL()) + "'";
```
a-ctor, line 421: Don't know if we want to change it here in the new code but I think that using raw string literals could improve the readability here.

```c#
clickHandler = $$"""
      function() {
           {{renderingContext.Control.Page!.ClientScript.GetPostBackEventReference(renderingContext.Control, argument)}};
           return false;
      }
      """
```

Similar argument could be made at all the other string concatenations we added. 

A bit tricky to get the alignment right though :D

a-ctor (changes requested):

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor we have been using setAttribute in the rest of the code. Why change it here?

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor I'll make this with the `===` compare instead to make the code easier to read. you need to know more about Javascript behavior with collapsing null and empty.

Remotion/Web/Core/UI/Controls/DropDownMenuImplementation/Rendering/DropDownMenuRenderer.cs
MichaelKetting: Since the script is rendered as a parameter in the rest of the Javascript, I actually prefer the single-line version. Otherweise we get a ton of weird linebreaks in the rendered page. It's not something I explciitly went for (it just happened due to existing code styles) but it does have its benefits.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor I'll go with the `!==` check for consistency with existing code.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
MichaelKetting: interesting.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
MichaelKetting: That's related to being able to correctly unbind. Since removeAttribute takes care of that, there's no more reason against addEventListener

a-ctor: I think there might have been a misunderstanding so I want to elaborate. Using `removeAttribute("onclick")` will only remove an event handler that was added by the `onclick` attribute.

If we use `addEventListener` we can only remove the listener with `removeEventListener`. So the question here is if we have a cleanup or not and if we can save the event listener for removal. You usually want to use `addEventListener` over an `onXXX` attribute.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
a-ctor: I didn't notice the other `setAttribute` calls. It just doesn't look right to call `setAttribute` for something that is already exposed as a property on the element.

Remotion/Web/ClientScript/Scripts/ListMenu.ts
```diff
@@ -114,32 +114,20 @@ class ListMenu
...
-            anchor.removeAttribute ('javascript');
-            anchor.removeAttribute ('onclick');
-            anchor.onclick = null;
-          }
-        }
+          anchor.href = itemInfo.Href;
+        if (itemInfo.Target != null)
+          anchor.target = itemInfo.Target;
+        if (itemInfo.ClickHandler != null)
+          anchor.addEventListener('click', itemInfo.ClickHandler);
```
a-ctor, line 130: See the other comment about a potential misunderstanding.

a-ctor (approved):

Remotion/Web/ClientScript/Scripts/ListMenu.ts
MichaelKetting: Okay, updated the code to simplify and fix the event-wireup.

Remotion/Web/ClientScript/Scripts/DropDownMenu.ts
MichaelKetting: @a-ctor symmetry with `removeAttribute`. And `removeAttribute` is removing the attribute in the DOM, setting it to empty string keeps the attribute.

PR #1231, opened 2024-10-02
RM-9324 Make metadata on QueryFactory.CreatLinqQuery optional

Frazerus:
https://re-motion.atlassian.net/browse/RM-9324

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
@@ -330,6 +330,39 @@ namespace Remotion.Data.DomainObjects.UnitTests.Queries
       }
     }
 
+    [Test]
+    public void CreateLinqQuery_WithoutMetadata_UsesEmptyMetadata ()
+    {
+      var factoryMock = new Mock<ILinqProviderComponentFactory>(MockBehavior.Strict);
+      var serviceLocator = DefaultServiceLocator.Create();
+      serviceLocator.RegisterSingle(() => factoryMock.Object);
+      RegisterStandardConfiguration(serviceLocator);
+
+      using (new ServiceLocatorScope(serviceLocator))
+      {
+        var fakeExecutor = new Mock<IQueryExecutor>();
+        var fakeQueryParser = new Mock<IQueryParser>();
+        var fakeResult = new Mock<IQueryable<Order>>();
+
+        factoryMock
+            .Setup(
+                mock => mock.CreateQueryExecutor(
+                    TestDomainStorageProviderDefinition,
+                    "dummyID",
+                    It.Is<IReadOnlyDictionary<string, object>>(d => d.Count == 0)))
+            .Returns(fakeExecutor.Object);
+        factoryMock
+            .Setup(mock => mock.CreateQueryParser())
+            .Returns(fakeQueryParser.Object);
+        factoryMock
+            .Setup(mock => mock.CreateQueryable<Order>(fakeQueryParser.Object, fakeExecutor.Object))
+            .Returns(fakeResult.Object);
+
+        var result = QueryFactory.CreateLinqQuery<Order>("dummyID");
+        Assert.That(result, Is.SameAs(fakeResult.Object));
+      }
+    }
+
     private void ResetCaches ()
     {
       var linqProviderComponentFactoryCache =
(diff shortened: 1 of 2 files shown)


Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
```diff
@@ -330,7 +330,44 @@ public void CreateLinqQuery_WithoutParserAndExecutor ()
...
+            .Returns(fakeQueryParser.Object)
+            .Verifiable();
+        factoryMock
+            .Setup(mock => mock.CreateQueryable<Order>(fakeQueryParser.Object, fakeExecutor.Object))
+            .Returns(fakeResult.Object)
+            .Verifiable();
+
+        var result = QueryFactory.CreateLinqQuery<Order>("dummyID");
+
+        factoryMock.Verify();
```
a-ctor, line 365: Can't we convert `factoryMock` to a stub? We get back a result that only works if all setups were called because each setup returns an object that is necessary for the next call, resulting in the final result object that we assert below.

a-ctor (changes requested):

Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
Frazerus: since we specifically want to check that the mock is called with empty metadata it can't really be a stub

MichaelKetting: The return values do check this implicitly, so we don't need the mock-verification. 

The second question is: the same test we already have so what's the benefit of "cleaning up" the semantics in this test. 

I can live with both versions.

Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
```diff
@@ -330,7 +330,44 @@ public void CreateLinqQuery_WithoutParserAndExecutor ()
...
+      var factoryMock = new Mock<ILinqProviderComponentFactory>(MockBehavior.Strict);
+      var serviceLocator = DefaultServiceLocator.Create();
+      serviceLocator.RegisterSingle(() => factoryMock.Object);
+      RegisterStandardConfiguration(serviceLocator);
+
+      using (new ServiceLocatorScope(serviceLocator))
+      {
+        var fakeExecutor = new Mock<IQueryExecutor>();
+        var fakeQueryParser = new Mock<IQueryParser>();
+        var fakeResult = new Mock<IQueryable<Order>>();
```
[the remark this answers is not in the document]
Frazerus: Just duplicated from other tests in the file. I would keep it this way for consistency

Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
```diff
@@ -330,7 +330,44 @@ public void CreateLinqQuery_WithoutParserAndExecutor ()
...
+            .Returns(fakeExecutor.Object)
+            .Verifiable();
+        factoryMock
+            .Setup(mock => mock.CreateQueryParser())
+            .Returns(fakeQueryParser.Object)
+            .Verifiable();
+        factoryMock
+            .Setup(mock => mock.CreateQueryable<Order>(fakeQueryParser.Object, fakeExecutor.Object))
+            .Returns(fakeResult.Object)
+            .Verifiable();
```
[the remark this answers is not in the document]
Frazerus: I would just keep it for consistency with the other tests.

a-ctor (approved):

Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
```diff
@@ -330,7 +330,40 @@ public void CreateLinqQuery_WithoutParserAndExecutor ()
...
+        factoryMock
+            .Setup(mock => mock.CreateQueryable<Order>(fakeQueryParser.Object, fakeExecutor.Object))
+            .Returns(fakeResult.Object);
+
+        var result = QueryFactory.CreateLinqQuery<Order>("dummyID");
+        Assert.That(result, Is.SameAs(fakeResult.Object));
+      }
+    }
+
+  private void ResetCaches ()
```
MichaelKetting, line 366:
```suggestion
    private void ResetCaches ()
```
Whitespace

MichaelKetting (approved):

PR #1232, opened 2024-10-02
RM-9317 Upgrade chrome to v129

Frazerus:
[RM-9317-upgrade-chrome-to-version-129](https://re-motion.atlassian.net/browse/RM-9317)


Frazerus:
Closed because this chrome version is buggy.

PR #1235, opened 2024-10-04
RM-9325 Bockport chrome to v111 v3.1

Frazerus:


Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
```diff
@@ -83,6 +85,14 @@ protected TPageObject Start<TPageObject> (string page)
...
       _webTestHelper.MainBrowserSession.Window.Visit(url);
 
+      //Chrome has a hover card which appears even though the cursor does not hover over the tab.
+      //To remove this card (which may destroy screenshots), a click can be used.
+      if (_webTestHelper.BrowserConfiguration.IsChrome())
+      {
+        var helper = new MouseHelper(_webTestHelper.BrowserConfiguration);
+        helper.LeftClick();
+      }
+
```
a-ctor, line 95: It might be worth investigating if we can disable this feature via flags. This way, we don't need to rely on this trick to work. If we don't find something this is probably fine.

a-ctor (approved):

Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
Frazerus: This is how we fixed it in re-motion 6.0 as well, and I did search for a flag that time. Quick check now revealed the same conclusion: there is no flag for this, only for the useless pictures you get as a preview in  that same window aswell.

PR #1239, opened 2024-10-09
Feature/rm 9338 add service to declare default storage type of datetime objects

Frazerus:
https://re-motion.atlassian.net/browse/RM-9338

### DIFF
--- a/Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/DateTime2DefaultStorageTypeProvider.cs
+++ b/Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/DateTime2DefaultStorageTypeProvider.cs
@@ -1,0 +1,20 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Implements the <see cref="IDateTimeDefaultStorageTypeProvider"/> to provide a default storage type
+/// of <see cref="DbType.DateTime2"/>.
+/// </summary>
+/// <threadsafety static="true" instance="true" />
+public class DateTime2DefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
+{
+  /// <inheritdoc cref="IDateTimeDefaultStorageTypeProvider.StorageTypeName"/>
+  public string StorageTypeName => "datetime2";
+
+  /// <inheritdoc cref="IDateTimeDefaultStorageTypeProvider.DbType"/>
+  public DbType DbType => DbType.DateTime2;
+}
--- a/Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/DateTimeDefaultStorageTypeProvider.cs
+++ b/Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/DateTimeDefaultStorageTypeProvider.cs
@@ -1,0 +1,18 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Implements the <see cref="IDateTimeDefaultStorageTypeProvider"/> to provide a default storage type
+/// of <see cref="DbType.DateTime"/>.
+/// </summary>
+/// <threadsafety static="true" instance="true" />
+public class DateTimeDefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
+{
+  public string StorageTypeName => "datetime";
+
+  public DbType DbType => DbType.DateTime;
+}
--- a/Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/IDateTimeDefaultStorageTypeProvider.cs
+++ b/Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/IDateTimeDefaultStorageTypeProvider.cs
@@ -1,0 +1,25 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Provides an interface for defining the default rdbms storage type for <see cref="DateTime"/> objects.
+/// </summary>
+/// <threadsafety static="true" instance="true" />
+/// <seealso cref="DateTimeDefaultStorageTypeProvider" />
+/// <seealso cref="DateTime2DefaultStorageTypeProvider" />
+public interface IDateTimeDefaultStorageTypeProvider
+{
+  /// <summary>
+  /// Gets the StorageTypeName of the default rdbms storage type.
+  /// </summary>
+  public string StorageTypeName { get; }
+
+  /// <summary>
+  /// Gets the <see cref="DbType"/> of the default rdbms storage type.
+  /// </summary>
+  public DbType DbType { get; }
+}
--- a/Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
+++ b/Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
@@ -17,12 +17,14 @@
 using System;
 using System.ComponentModel;
 using System.Data;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using JetBrains.Annotations;
 using Remotion.Data.DomainObjects.Mapping;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
 using Remotion.ExtensibleEnums;
+using Remotion.Reflection;
 using Remotion.Utilities;
 
 using DateOnlyConverter = Remotion.Utilities.DateOnlyConverter;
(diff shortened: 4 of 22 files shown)


MichaelKetting (commented):
partial review

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/IDateTimeDefaultStorageTypeProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer;
+
+public interface IDateTimeDefaultStorageTypeProvider
```
MichaelKetting, line 7: One alternative on the design, given how it is now done with two types; just make two properties directly on the interface one for the storagetypename and one for dbtype.

The original idea with the result object was...design conincidence I gess

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DateTime2DefaultStorageTypeProvider.cs
```diff
@@ -0,0 +1,11 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer;
+
+public record DateTime2DefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
```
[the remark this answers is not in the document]
MichaelKetting: semantics... 

and after thinking some more: more semantics: with a record, you'd assume a data record that allows some form of data manipulation, typically by copy on write, but still, data edit. When you have a config object, and particularily hard-coding settings, that's not a record in my understanding of the semantics as intended by the concept. 

In this usecase, we also don't have a feature benefit from it (the code's not shorter, we don't need equality, we don't need the data modification features, ...)

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -161,7 +161,14 @@ public virtual IStorageTypeInformationProvider CreateStorageTypeInformationProvi
     {
       ArgumentUtility.CheckNotNull("rdmsStorageProviderDefinition", rdmsStorageProviderDefinition);
 
-      return new SqlStorageTypeInformationProvider();
+      return new SqlStorageTypeInformationProvider(CreateDateTimeDefaultStorageTypeProvider(rdmsStorageProviderDefinition));
```
MichaelKetting, line 164: readablity / consistency: I think we use local variables for the services before injecting them. It becomes relevant if there's more services and makes it more explicit / obvious in general. If it's not done now, it's gonna be done the next time we touch this ctor.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
```diff
@@ -35,9 +35,11 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Model.Building
   public class SqlStorageTypeInformationProvider : IStorageTypeInformationProvider
   {
     internal const int StorageTypeLengthRepresentingMax = -1;
+    private readonly IDateTimeDefaultStorageTypeProvider _dateTimeDefaultStorageTypeProvider;
 
-    public SqlStorageTypeInformationProvider ()
+    public SqlStorageTypeInformationProvider (IDateTimeDefaultStorageTypeProvider dateTimeDefaultStorageTypeProvider)
     {
+      _dateTimeDefaultStorageTypeProvider = dateTimeDefaultStorageTypeProvider;
```
MichaelKetting, line 42: argcheck

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
```diff
@@ -226,6 +228,18 @@ public IStorageTypeInformation GetStorageType (object? value)
       return null;
     }
 
+    private StorageTypeInformation GetDateTimeStorageTypeInformation (Type dotNetType, bool isNullableInDatabase)
```
MichaelKetting, line 231:
```suggestion
    private StorageTypeInformation GetDateTimeStorageType (Type dotNetType, bool isNullableInDatabase)
```

MichaelKetting, line 231: this PR might benefit from getting done first, because we then have the correct place to add the attributes.

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/IDateTimeDefaultStorageTypeProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Provides an interface for receiving the default rdbms storage type for <see cref="DateTime"/> objects.
+/// </summary>
+public interface IDateTimeDefaultStorageTypeProvider
```
MichaelKetting, line 11:
```suggestion
/// <summary>
/// Provides an interface for receiving the default rdbms storage type for <see cref="DateTime"/> objects.
/// </summary>
/// <threadsafety static="true" instance="true" />
/// <seealso cref="DateTimeDefaultStorageTypeProvider" />
/// <seealso cref="DateTime2DefaultStorageTypeProvider" />
public interface IDateTimeDefaultStorageTypeProvider
```

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/DateTimeDefaultStorageTypeProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Implements the <see cref="IDateTimeDefaultStorageTypeProvider"/> to provide a default storage type
+/// of <see cref="DbType.DateTime"/>.
+/// </summary>
+public class DateTimeDefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
```
MichaelKetting, line 12:
```suggestion
/// <summary>
/// Implements the <see cref="IDateTimeDefaultStorageTypeProvider"/> to provide a default storage type
/// of <see cref="DbType.DateTime"/>.
/// </summary>
/// <threadsafety static="true" instance="true" />
public class DateTimeDefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
```

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/DateTime2DefaultStorageTypeProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Implements the <see cref="IDateTimeDefaultStorageTypeProvider"/> to provide a default storage type
+/// of <see cref="DbType.DateTime2"/>.
+/// </summary>
+public class DateTime2DefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
```
MichaelKetting, line 12:
```suggestion
/// <summary>
/// Implements the <see cref="IDateTimeDefaultStorageTypeProvider"/> to provide a default storage type
/// of <see cref="DbType.DateTime2"/>.
/// </summary>
/// <threadsafety static="true" instance="true" />
public class DateTime2DefaultStorageTypeProvider : IDateTimeDefaultStorageTypeProvider
```

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/Building/IDateTimeDefaultStorageTypeProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Data;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+
+/// <summary>
+/// Provides an interface for receiving the default rdbms storage type for <see cref="DateTime"/> objects.
```
MichaelKetting, line 9:
```suggestion
/// Provides an interface for defining the default rdbms storage type for <see cref="DateTime"/> objects.
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/Building/SqlStorageTypeInformationProvider.cs
MichaelKetting: On the other hand, we don't have test data here. I think I can live with that. But if you just take this API-update and already do it in the other PR, we probably are cleaner since we get the tests first and the default-api second.

MichaelKetting (approved):

MichaelKetting (changes requested):
@Frazerus I'm so sorry, we just figured out that sql server converts datetime to datetime2 isntead of datetime2 to datetime when we compare values of different types. this means the mixing of datatimes in a database breaks equality. I will need to reset develop and we won't be doing this PR.

Instead, we will add a simpler version of this feature branch that switches the entire database between datetime and datetime2.

MichaelKetting:
@Frazerus Updated design: we archive this PR and the associated ticket (create a branch named `archive/RM-9338-...` instead of `feature/RM-9338-...` and start with a simpler version:
IDateTimeStorageTypeProvider, same properties as before. This will inform both the case when there's a PropertyDefinition and the case without. So, basically, just a simple service that's used directly where we decide which way the DateTime property should go. We only keep the unit tests for the SqlStorageTypeInformationProvider, no changes to the database, etc.

We will not revert the default of using datetime2 in re-motion, even though it's likely the big projects will all use the legacy behavior. That's done because they all inherit the legacy behavior anyway from the product layer's decision and thus, there's not a lot of work to do. We do need to update the breaking change ticket and create a Change Board ticket.

PR #1251, opened 2024-11-04
RM-9381 Upgrade Microsoft.Typescript.MsBuild to v5.6

Frazerus:
https://re-motion.atlassian.net/browse/RM-9381


MichaelKetting (approved):
Can be integrated in develop / fixversion 7.0

PR #1278, opened 2024-11-25
RM-9413 Add workaround for firefox problem where network requests are stuck

a-ctor:
https://re-motion.atlassian.net/browse/RM-9413


MichaelKetting (approved):

MichaelKetting (changes requested):
Pull request will not be merged because we know of additional problems in Firefox unrelated to WXE and thus a more basic solution (deactivation of HTTP/2) is needed.

PR #1282, opened 2024-12-17
Feature/rm 9279 replace the SBOM generation with the build script functionality

Frazerus:
https://re-motion.atlassian.net/browse/RM-9279

### DIFF
--- a/Build/Build.cs
+++ b/Build/Build.cs
@@ -20,10 +20,13 @@ using System.Collections.Immutable;
 using System.Linq;
 using Customizations;
 using JetBrains.Annotations;
+using NuGet.Versioning;
 using Nuke.Common;
 using Nuke.Common.IO;
+using Nuke.Common.ProjectModel;
 using Remotion.BuildScript;
 using Remotion.BuildScript.Components;
+using Remotion.BuildScript.GenerateSbom;
 using Remotion.BuildScript.Test;
 using Remotion.BuildScript.Test.Dimensions;
 using static Customizations.Browsers;
@@ -44,7 +47,6 @@ class Build : RemotionBuild, IDependDB
   [Parameter(ValueProviderMember = nameof(SupportedTestSqlServers), Separator = "+")]
   public string[] TestSqlServers { get; set; } = [];
 
-
   public static int Main () => Execute<Build>();
 
   [UsedImplicitly]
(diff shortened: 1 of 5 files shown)


Directory.Packages.props
```diff
@@ -57,7 +58,7 @@
          4.11.0: MSBuild 17.11 / .NET SDK 8.0.4xx
          4.12.0: MSBuild 17.12 / .NET SDK 9.0.1xx
     -->
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.11.0" />
```
a-ctor: Why do we need to bump this dependency?

Build/Customizations/Properties.props
```diff
@@ -23,5 +23,6 @@
     <!-- '#OutputDirectory#' and '#ConfigurationID#' will be replaced by '$(OutputDirectory)' and '$(ConfigurationID)' during DependDB processing. See Shared.Overrides.targets:BeforeCheckDependDBProperties -->
     <DependDBNuGetSources>$(SolutionDirectory)packages</DependDBNuGetSources>
     <DependDBTargetFramework>net48</DependDBTargetFramework>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
a-ctor: This will change the files in the output folder of each project. Do we really need this? I think we should avoid this if possible. If we keep it we should check that that Nuget packages have not changed. The new files might get copied into the Nuget package otherwise.

a-ctor (changes requested):

Build/Customizations/Properties.props
Frazerus: We need those files so we have all relevant data in the bin folder. Will check the nuget packages

Frazerus: The nuget packages seem fine. Did not check all 50 though, but got Web.Core, Core.core and OBW.Core and a few others. I don't think it should be a problem. And since we only have so little nuget dependencies, it's ok.

a-ctor (approved):

Build/Customizations/Properties.props
MichaelKetting: I read the original discussion but I don't understand how an MSBuld property fits into our custimizations section. This isn't intended to import msbuild settings. bit more context, please

Build/Build.cs
```diff
@@ -56,12 +58,37 @@ class Build : RemotionBuild
...
+                                    "Web.Dependencies.Javascript",
+                                    "*.Analyzers*"
+                                };
+
+      var packageJsonPath = solution.Directory / "Remotion" / "Web" / "Dependencies.JavaScript" / "package.json";
+      var packageJsonWithVersion = TemporaryDirectory / "sbom" / "package.json";
+
+      AddVersionToPackageJson(packageJsonPath, packageJsonWithVersion, shortenedVersion);
+
+      // We do not require github auth because we do not do enough requests for licenses and package infos
```
MichaelKetting, line 86: Why do we not need to do this here but in the base implementation? I think I saw this somewhere with an acutal user/key. Once I know where and why we needed it there, I can better work with this comment :)

Build/Build.cs
```diff
@@ -56,12 +58,37 @@ class Build : RemotionBuild
...
+
+      var semanticVersion = SemanticVersion.Parse(version);
+      var shortenedVersion = $"{semanticVersion.Major}.{semanticVersion.Minor}.{semanticVersion.Patch}";
+
+      // We disregard test projects when generating the sbom already, so these are not required here.
+      var blacklistedProjects = new[]
+                                {
+                                    "Web.Dependencies.Javascript",
+                                    "*.Analyzers*"
+                                };
```
MichaelKetting: Recommend moving the comment into the array (at the end). Otherwise it reads as if the array is the test projects blacklist.

MichaelKetting (changes requested):

Build/Build.cs
Frazerus: Github requires auth when we send off too many requests to its API. The tool calls the github API in some cases. Since we don't really have that problem as our dependency count is low, we do not need to supply the auth info.

Build/Customizations/Properties.props
Frazerus: I just put it there because we want it in all projects and it was the first thing I thought about. I guess shared.build.props is more reasonable?

Build/Build.cs
Frazerus: We might still want this anyways, but with the budget for this being more or less empty, this was a good corner to cut.

MichaelKetting: okay, so the whole github-auth story is just to support this usecase but we never actually ran into it with re-motion?

Since we do have the auth information in our build (don't we?), shouldn't we still use github auth and just forget about the whole thing? Otherwise, we risk a future failing build if limits change, many builds run in a short time, etc.

Build/Customizations/Properties.props
MichaelKetting: you want to have this in all projects in the solution? Then we should put it into the existing shared-build files (probably shared.build.props, indeed!). Thing is, we actually only need it during nuke-builds, not vs-builds, correct?

I guess it's not really a performance issue for re-motion, but the "problem" would be that the build script doesn't work if it's not set. So it would be nice if the sbom-logic also checks this prerequisite (which is probably not doable?). This means, putting it into the shared build files and adding a comment why it's needed would be sufficient for re-motion. An alternative would be to pass it as a property to the msbuild call. We already pass properties from the build script to msbuild and could add it there. Although, having the same build output layout for vs-build and nuke-build might be helpful, which prefers the shared build files.

Thoughts, beyond just putting it into shared.build.props?

Build/Build.cs
MichaelKetting: just saw the second comment about budget: if the properties are already available, we should add it (15 minutes or less). if we don't have them, yes, we can add later. As I said, I think we already did them in an earlier iteration, as far as I remember.

Build/Shared.build.props
```diff
@@ -31,6 +31,7 @@
     <PackageTags>re-motion</PackageTags>
     <PackageTags Condition="'$(ProductType)' == 'SecurityManager'">$(PackageTags) SecurityManager</PackageTags>
     <DisableTransitiveProjectReferences>true</DisableTransitiveProjectReferences>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
MichaelKetting, line 36: The explainatory comment, please, regardless of budget

MichaelKetting (approved):
I'll approve this. please check the comments and if you see something you think helpful and budgetary sane, we should do that. otherwise we're done.

Build/Build.cs
Frazerus: I can't find anything about github auth in our buildscript. Even if we had the data immediately available, the way how we do things with the Sbom configuration is not well integrated into the nuke infrastructure as I messed the design up a bit. Will create a ticket to fix and to improve the design. It was a bit unfortunate that my first part in nuke things was done while patrick was not there to tell me how to properly work with the nuke infrastructure.

Build/Customizations/Properties.props
Frazerus: Time constraints, and it doesn't really matter for vs builds IMO.

PR #1285, opened 2025-01-08
RM-9372 Integration of the DependDB using NUKE build

a-ctor:
https://re-motion.atlassian.net/browse/RM-9372

### DIFF
--- a/.nuke/build.schema.json
+++ b/.nuke/build.schema.json
@@ -25,6 +25,36 @@
           "type": "boolean",
           "description": "Indicates to continue a previously failed build attempt"
         },
+        "DependDBNugetConfig": {
+          "type": "string"
+        },
+        "DependDBNuGetSources": {
+          "type": "string"
+        },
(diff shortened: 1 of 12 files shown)


.nuke/build.schema.json
```diff
@@ -122,6 +154,8 @@
               "DetermineBuildMetadata",
               "DetermineProjectMetadata",
               "FullBuild",
+              "GenerateDependDBOutput",
+              "GenerateSbom",
```
Frazerus, line 158: Should be fine? Autogenerated nuke content...

Frazerus (approved):

.nuke/build.schema.json
a-ctor: This file is just a JSON schema for the parameter files. As we added new parameters with a build script update, the changes happen here as well. They don't really matter but if we do something build script related we can commit them as well.

PR #1286, opened 2025-01-08
RM-6087 Do not reset icon when empty input value

Frazerus:
https://re-motion.atlassian.net/browse/RM-6087

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
@@ -22,6 +22,7 @@ class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes
   private _iconMarker: Nullable<HTMLElement>;
   private _icon: Nullable<HTMLImageElement>;
   private _iconBackUp: Nullable<HTMLImageElement>;
+  private _hasIconBackUp: Nullable<boolean>;
   private _selectListID: string;
   private _informationPopUpID: string;
   private _nullValueString: string;
@@ -89,6 +90,7 @@ class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes
     this._iconMarker = iconMarker;
     this._icon = icon;
     this._iconBackUp = icon;
+    this._hasIconBackUp = null;
     this._selectListID = baseID + '_Results';
     this._informationPopUpID = baseID + '_Information';
     this._isAutoPostBackEnabled = isAutoPostBackEnabled;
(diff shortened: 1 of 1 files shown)


Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -240,6 +240,7 @@ class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes
         }
 
         const hasChanged = this._itemBackUp != actualItem;
+        const isEmpty = actualItem.DisplayName === '';
```
a-ctor: I would move this towards the usage below as we don't use it multiple times and it makes it harder to track the value of `isEmpty`

a-ctor (approved):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -254,7 +254,8 @@ class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes
           }
           else
           {
-            this.RestoreIcon();
+            const isEmpty = actualItem.DisplayName === '';
```
MichaelKetting: This breaks semantics. A bit further up in the code, we use `aitem.UniqueIdentifier == this._nullValueString` to test for "empty" semantics.

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
```diff
@@ -277,7 +278,7 @@ class BocAutoCompleteReferenceValue //TODO RM-7715 - Make the TypeScript classes
     this._icon.replaceWith(this._iconBackUp);
     this._icon = this._iconBackUp;
 
-    if (this._iconMarker != null)
+    if (this._iconMarker != null && !isEmpty)
```
MichaelKetting: I'm not convinced that the flag is the best way to do this. It feels like we're observing a sideeffect and basing our decsision on the side effect. RestoreIcon is called when there was no factual change the to the selected value. So, if the value was originally empty, I think we never got an icon filled out. pershaps we don't actually want to call RestoreIcon in that case?

Disclaimer: I'm not sure about the behavior of the following
1.  Initially empty
2. type and wait for drop down
3. move cursor to select an item -> this might add an icon to the input box -> try...
4. hit keys in such a way as to remove the selection again (clear the input box using backspace?) and complete -> empty and no icon.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocAutoCompleteReferenceValue.ts
Frazerus: This is exactly why I was asking the stuff yesterday about if we care whether the icon is there or not. 
What currently happens with the changes I made:
* an empty dropdown has an icon, but does not have the `hasIcon` class (this was already the case and the main point yesterday)
* when we create the issue (type and delete), an Icon is restored and the `hasIcon` class is not set. This replicates the standard behaviour we have now, where we only cancel the side effect of the `hasIcon` class

Yesterday I asked specifically about the icon being there on initial load, but no `hasIcon` class set and whether we actually want that

MichaelKetting (approved):

PR #1295, opened 2025-01-22
Feature/rm 9426 cleanup unused css scrollbar properties

Frazerus:
https://re-motion.atlassian.net/browse/RM-9426

### DIFF
--- a/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
+++ b/Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
@@ -720,10 +720,34 @@ class BocList
     scrollableContainer.dataset.bocListPreviousScrollLeft = scrollLeft.toString();
 
     if (hasScrollMoveFromColumnHeader)
+    {
       scrollableContainer.scrollLeft = scrollLeft;
 
+      // Giving a too high value of scroll left will reset it to the maximum in chrome and firefox.
+      // This creates jitter, as the value changes compared to the saved one, triggering a resize in the next iteration.
+      // Thus, we need to re-set the saved value if it can't be set to our desired value to not create jitter.
+      if (scrollableContainer.scrollLeft !== scrollLeft)
+      {
+        scrollLeft = scrollableContainer.scrollLeft;
+        fakeTableHeadContainer.scrollLeft = scrollLeft;
+        scrollableContainer.dataset.bocListPreviousScrollLeft = scrollLeft.toString();
+      }
+    }
+
     if (hasScrollMoveFromScrollbar)
+    {
       fakeTableHeadContainer.scrollLeft = scrollLeft;
+
+      // Giving a too high value of scroll left will reset it to the maximum in chrome and firefox.
+      // This creates jitter, as the value changes compared to the saved one, triggering a resize in the next iteration.
+      // Thus, we need to re-set the saved value if it can't be set to our desired value to not create jitter.
+      if (fakeTableHeadContainer.scrollLeft !== scrollLeft)
+      {
+        scrollLeft = fakeTableHeadContainer.scrollLeft;
+        scrollableContainer.scrollLeft = scrollLeft;
+        scrollableContainer.dataset.bocListPreviousScrollLeft = scrollLeft.toString();
+      }
+    }
   }
 
   public static InitializeNavigationBlock (pageNumberFieldOrSelector: CssSelectorOrElement<HTMLInputElement>, pageIndexFieldOrSelector: CssSelectorOrElement<HTMLInputElement>): void
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
@@ -44,7 +44,7 @@ div.bocList.hasDimensions  div.bocListTableBlock div.bocListTableContainer
 div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListTableScrollContainer
 {
   background-color: var(--remotion-themed-background-color-page);
-  scrollbar-width: var(--remotion-themed-scrollbar-style);
+  scrollbar-width: var(--remotion-themed-scrollbar-width);
   scrollbar-color: var(--remotion-themed-scrollbar-color);
   display: block;
   overflow: auto;
@@ -425,7 +425,7 @@ div.bocList div.bocListTableBlock tr.bocListValidationRow div
   position: sticky;
   left: 0;
   overflow-x: auto;
-  scrollbar-width: var(--remotion-themed-scrollbar-style);
+  scrollbar-width: var(--remotion-themed-scrollbar-width);
   scrollbar-color: var(--remotion-themed-scrollbar-color);
 }
 
@@ -499,7 +499,7 @@ div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTable
 {
   /* Reserve space in case there is a scrollbar for the data-part of the table.
      Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
-  border-right: var(--remotion-themed-scrollbar-width) solid var(--boclist-tablehead-background-color);
+  padding-right: var(--remotion-themed-scrollbar-padding-requirement);
 }
 
 div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTableHead table,
--- a/Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
+++ b/Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
@@ -72,15 +72,13 @@
   --remotion-themed-outline-offset-focus: calc((var(--remotion-themed-border-width) + var(--remotion-themed-outline-width)) * -1);
   --remotion-themed-outline-offset-focus-hover: var(--remotion-themed-outline-offset-focus);
   --remotion-themed-outline-offset-readonly-focus: var(--remotion-themed-outline-offset-focus);
-  --remotion-themed-scrollbar-width: 10px;
-  --remotion-themed-scrollbar-border-radius: 10px;
-  --remotion-themed-scrollbar-style: thin;
+  --remotion-themed-scrollbar-padding-requirement: 10px;
+  --remotion-themed-scrollbar-width: thin;
   --remotion-themed-scrollbar-color-thumb: var(--color-primary);
   --remotion-themed-scrollbar-color-track: transparent;
   --remotion-themed-scrollbar-color: var(--remotion-themed-scrollbar-color-thumb) var(--remotion-themed-scrollbar-color-track);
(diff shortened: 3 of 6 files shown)


Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -495,11 +495,12 @@ div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTable
   overflow: hidden;
 }
 
+
 div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTableHead > div
 {
   /* Reserve space in case there is a scrollbar for the data-part of the table.
-     Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
-  border-right: var(--remotion-themed-scrollbar-width) solid var(--boclist-tablehead-background-color);
+ that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
```
MichaelKetting:
```suggestion
Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
```

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -495,11 +495,12 @@ div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTable
...
+  padding-right: 10px;
```
MichaelKetting: This should be declared as a css-variable together with the existing `--remotion-themed-scrollbar-style:` property to make it clear what requires changing when the width is manipulated. The intention of the css-variables to not have magic numbers in code or duplicated business logic.

And even though we do not use it with `--remotion-themed-popup-scrollbar-style` in remotion, we probably should prepare that property, too, in case someone needs this informaton.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -72,14 +72,10 @@
   --remotion-themed-outline-offset-focus: calc((var(--remotion-themed-border-width) + var(--remotion-themed-outline-width)) * -1);
   --remotion-themed-outline-offset-focus-hover: var(--remotion-themed-outline-offset-focus);
   --remotion-themed-outline-offset-readonly-focus: var(--remotion-themed-outline-offset-focus);
-  --remotion-themed-scrollbar-width: 10px;
-  --remotion-themed-scrollbar-border-radius: 10px;
   --remotion-themed-scrollbar-style: thin;
```
MichaelKetting: the rename of the css property is still open. I'm assuming because the PR isn't ready yet? Or is the rename intentionally left out because the rename isn't helpful?

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -495,11 +495,12 @@ div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTable
   overflow: hidden;
 }
 
+
 div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTableHead > div
 {
   /* Reserve space in case there is a scrollbar for the data-part of the table.
-     Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
-  border-right: var(--remotion-themed-scrollbar-width) solid var(--boclist-tablehead-background-color);
+Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
```
MichaelKetting: whitespace indent fix pls.

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -495,11 +495,12 @@ div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTable
   overflow: hidden;
 }
 
+
```
MichaelKetting: whitespace fix pls. (remove new blank line)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -668,7 +668,7 @@ class BocList
   private static FixHeaderSize(scrollableContainer: HTMLElement): void
   {
     const realTable = scrollableContainer.querySelector<HTMLTableElement>(':scope > table')!;
-    const realTableWidth = Math.floor(LayoutUtility.GetWidth(realTable));
+    const realTableWidth = Math.floor(LayoutUtility.GetWidth(realTable)) + 1;
```
MichaelKetting: I saw the +1 jitter discussion but I don't get the math/problem behind it. We should document this, please.

Right now, I'm wondering why we don't use `Math.ceil()` instead or use `Math.floor(LayoutUtillity.GetWidth(realTable)+1)`.

Do we simply *always* need one extra pixel or is it a rounding issue?

MichaelKetting (changes requested):
cleanup and math. solution itself looks to work just fine :) :)

a-ctor (approved):

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
Frazerus: yea, mainly because it wasnt ready yet

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
Frazerus: Ok, I did a little bit of research about whats happening here:

When scrolling, we check whether the `fakeTableHeader` or the actual `scrollContainer` have scrolled, be looking at whether the `scrollLeft` attribute on either has changed. If that is the case, we set the `scrollLeft` of the other one to the changed value.

Now, the problem: 
If the containers are not the same size, the `scrollLeft` value has different maximums, as only so much can be to the left of the scrolled area. In Firefox and edge, the value is set to the maximum, if a too high value is supplied. 
So what happens is, that we set the `scrollLeft` to a value that fits for the wider one, the lower one resets it in the next iteration, as the `scrollLeft` can only be so high for it. This creates the jitter we see.

Solutions:
* Add a math.ceil here
* Check if the set `scrollLeft` is actual viable, and if not, reset it to the value that is viable.

Frazerus: The second fix is more stable, but is a bit more complex, and makes the code kinda ugly. I implemented it in an ugly way, we can easily refactor it to be a bit less bad.

Frazerus: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft

Frazerus: With solution 2, we also do not need to add the padding to the css in chrome anymore. This leads to a weird display of the scrollbar though, where it can't be scrolled further:
image


In firefox, we still need it, as otherwise bigger jittering occurs. I will not look into that for now.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -72,15 +72,13 @@
   --remotion-themed-outline-offset-focus: calc((var(--remotion-themed-border-width) + var(--remotion-themed-outline-width)) * -1);
   --remotion-themed-outline-offset-focus-hover: var(--remotion-themed-outline-offset-focus);
   --remotion-themed-outline-offset-readonly-focus: var(--remotion-themed-outline-offset-focus);
-  --remotion-themed-scrollbar-width: 10px;
-  --remotion-themed-scrollbar-border-radius: 10px;
-  --remotion-themed-scrollbar-style: thin;
+  --remotion-themed-scrollbar-padding-requirement: 10px;
```
MichaelKetting, line 75: I like the name

Remotion/ObjectBinding/Web/Res/Themes/NovaViso/HTML/BocList.css
```diff
@@ -498,8 +498,8 @@ div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTable
 div.bocList div.bocListTableBlock div.bocListTableContainer div.bocListFakeTableHead > div
 {
   /* Reserve space in case there is a scrollbar for the data-part of the table.
-     Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
-  border-right: var(--remotion-themed-scrollbar-width) solid var(--boclist-tablehead-background-color);
+    Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
```
MichaelKetting:
```suggestion
     Note that the actual max-width is not important so long as the scrollbar's max-width is covered. '*/
```
one more whitespace to reset the changes of this line in the PR :)

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -704,26 +704,38 @@ class BocList
...
-      fakeTableHeadContainer.scrollLeft = scrollLeft;
+    {
+        fakeTableHeadContainer.scrollLeft = scrollLeft;
+        if (fakeTableHeadContainer.scrollLeft !== scrollLeft){
+            scrollLeft = fakeTableHeadContainer.scrollLeft;
+            scrollableContainer.scrollLeft = scrollLeft;
+            scrollableContainer.dataset.bocListPreviousScrollLeft = scrollLeft.toString();
+        }
+    }
   }
```
MichaelKetting: Please revert the whitespace changes

Remotion/ObjectBinding/Web.ClientScript/Scripts/BocList.ts
```diff
@@ -704,26 +704,38 @@ class BocList
...
 
-    if (hasScrollMoveFromColumnHeader)
-      scrollableContainer.scrollLeft = scrollLeft;
+      scrollLeft = hasScrollMoveFromColumnHeader ? fakeTableHeadScrollLeft : scrollLeft;
+      scrollableContainer.dataset.bocListPreviousScrollLeft = scrollLeft.toString();
 
+      if (hasScrollMoveFromColumnHeader)
+      {
+          scrollableContainer.scrollLeft = scrollLeft;
+          if (scrollableContainer.scrollLeft !== scrollLeft){
```
MichaelKetting: thanks for explaining the updated logic. I also tested it and it works so it's perfectly fine that I don't get it completely.

I also tested in NovaGray. Looks fine, there, too.

MichaelKetting (approved):
It's only the whitespace in the TypeScript file. I don't need to re-review the cleaned-up state.

Remotion/Web/Core/Res/Themes/NovaViso/HTML/Common.css
```diff
@@ -72,15 +72,13 @@
...
+  --remotion-themed-scrollbar-width: thin;
```
MichaelKetting, line 76: Reminder: The property rename should be a breaking change ticket.

PR #1296, opened 2025-01-29
RM-9433 Add CSP header builder type

a-ctor:
https://re-motion.atlassian.net/browse/RM-9433

### DIFF
--- a/Remotion/Web/Core/ContentSecurityPolicy/CspDirectives.cs
+++ b/Remotion/Web/Core/ContentSecurityPolicy/CspDirectives.cs
@@ -1,0 +1,172 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+namespace Remotion.Web.ContentSecurityPolicy;
+
+/// <summary>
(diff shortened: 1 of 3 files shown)


Remotion/Web/Core/ContentSecurityPolicy/CspDirectives.cs
```diff
@@ -0,0 +1,176 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+namespace Remotion.Web.ContentSecurityPolicy;
+
+//
+// !!! NOTE: Changes also need to be done in `CspHeader.GetCspDirectiveName`.
```
MichaelKetting: Instead of the comment, please add an test that iterates over all items in the CspEnum and tries to make an entry.  Then we can simply add a a failure message that tells us what to do.

MichaelKetting: Edit: I saw EnumValuesNamesTest so I think we alreayd have this test, correct. plz drop the note.

MichaelKetting (approved):
I'm trusting testcoverage and senrior dev to have written sensible tests :)

PR #1299, opened 2025-03-03
Feature/rm 9437 überprüfen von browserfehlern am ende von tests

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9437

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/NUnitTestContext.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/NUnitTestContext.cs
@@ -1,0 +1,43 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure;
+
+public class NUnitTestContext (TestContext testContext, TestExecutionContext testExecutionContext) : ITestContext
+{
+  private static IReadOnlyDictionary<string, object> GetPropertiesFromAttributes (TestExecutionContext testExecutionContext)
+  {
+    var testMethod = testExecutionContext.CurrentTest.Method;
+    if (testMethod == null)
+      throw new InvalidOperationException($"Current test '{testExecutionContext.CurrentTest.FullName}' does not have a method associated with it.");
+
+    return WebTestAttribute.CreatePropertiesFromAttributes(testMethod.MethodInfo);
+  }
+
+  public string TestName => testContext.Test.Type?.Name + "_" + testContext.Test.Name;
+
+  public bool IsSuccessful => testExecutionContext.CurrentResult.ResultState != ResultState.Failure;
+
+  public void SetFailure (string message) => testExecutionContext.CurrentResult.SetResult(ResultState.Failure, message);
+
+  public IReadOnlyDictionary<string, object> Properties { get; } = GetPropertiesFromAttributes(testExecutionContext);
+}
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
@@ -18,7 +18,8 @@ using System;
 using System.Diagnostics;
 using Coypu;
 using NUnit.Framework;
-using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using Remotion.Web.Development.WebTesting.BrowserLog;
 using Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure;
 
 namespace Remotion.Web.Development.WebTesting.IntegrationTests
@@ -62,7 +63,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
     [SetUp]
     public void IntegrationTestSetUp ()
     {
-      _webTestHelper.OnSetUp(GetType().Name + "_" + TestContext.CurrentContext.Test.Name);
+      _webTestHelper.OnSetUp(new NUnitTestContext(TestContext.CurrentContext, TestExecutionContext.CurrentContext));
 
       var requestErrorDetection =
           (DiagnosticInformationCollectingRequestErrorDetectionStrategy)Helper.TestInfrastructureConfiguration.RequestErrorDetectionStrategy;
@@ -73,8 +74,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
     [TearDown]
     public void IntegrationTestTearDown ()
     {
-      var hasSucceeded = TestContext.CurrentContext.Result.Outcome.Status != TestStatus.Failed;
-      _webTestHelper.OnTearDown(hasSucceeded);
+      _webTestHelper.OnTearDown();
       _aspNetRequestErrorDetectionScope.Dispose();
     }
 
--- a/Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogMinimumLevelAttribute.cs
+++ b/Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogMinimumLevelAttribute.cs
@@ -1,0 +1,50 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using OpenQA.Selenium;
+using Remotion.Utilities;
+using Remotion.Web.Development.WebTesting.BrowserSession;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+/// <summary>
+/// Instructs the <see cref="WebTestHelper"/>'s browser log check to ignore any <see cref="BrowserLogEntry"/> whose <see cref="BrowserLogEntry.Level"/> is lower than the
+/// <see cref="MinimumLevel"/>. 
+/// </summary>
+/// <remarks>
+/// Use this on test methods, the classes to which they belong, or the whole assembly.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+public class BrowserLogMinimumLevelAttribute : WebTestAttribute
+{
+  public const string PropertyKey = "MinimumBrowserLogLevel";
+
+  public BrowserLogMinimumLevelAttribute (LogLevel minimumLevel)
+  {
+    MinimumLevel = minimumLevel;
+  }
+
+  public LogLevel MinimumLevel { get; }
+
+  public override void ApplyValue (IDictionary<string, object> dictionary)
+  {
+    ArgumentUtility.CheckNotNull(nameof(dictionary), dictionary);
+
+    dictionary[PropertyKey] = MinimumLevel;
+  }
+}
--- a/Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
+++ b/Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
@@ -1,0 +1,59 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using JetBrains.Annotations;
+using Remotion.Utilities;
+using Remotion.Web.Development.WebTesting.BrowserSession;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+/// <summary>
+/// Instructs the <see cref="WebTestHelper"/>'s browser log check to ignore any <see cref="BrowserLogEntry"/> whose <see cref="BrowserLogEntry.Message"/> matches the
+/// <see cref="Regex"/> in <see cref="Pattern"/>. 
+/// </summary>
+/// <remarks>
+/// Use this on test methods, the classes to which they belong, or the whole assembly.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+public class IgnoreBrowserLogMessageAttribute : WebTestAttribute
+{
+  public const string PropertyKey = "IgnoreBrowserLogMessage";
+
+  public IgnoreBrowserLogMessageAttribute ([NotNull] [RegexPattern] string pattern)
+  {
+    ArgumentUtility.CheckNotNullOrEmpty(nameof(pattern), pattern);
+
+    Pattern = pattern;
+  }
+
+  public string Pattern { get; }
+
+  public override void ApplyValue (IDictionary<string, object> dictionary)
+  {
+    ArgumentUtility.CheckNotNull(nameof(dictionary), dictionary);
+
+    if (!dictionary.TryGetValue(PropertyKey, out var value) || value is not IList<Regex> patterns)
+    {
+      patterns = new List<Regex>();
+      dictionary[PropertyKey] = patterns;
+    }
+
+    patterns.Add(new Regex(Pattern));
+  }
+}
--- a/Remotion/Web/Development.WebTesting/BrowserLog/PerformBrowserLogCheckAttribute.cs
+++ b/Remotion/Web/Development.WebTesting/BrowserLog/PerformBrowserLogCheckAttribute.cs
@@ -1,0 +1,47 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+/// <summary>
+/// Instructs the <see cref="WebTestHelper"/> whether to perform a browser log check during <see cref="WebTestHelper.OnTearDown"/>. 
+/// </summary>
+/// <remarks>
+/// Use this on test methods, the classes to which they belong, or the whole assembly.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+public class PerformBrowserLogCheckAttribute : WebTestAttribute
+{
+  public const string PropertyKey = "PerformBrowserLogCheck";
+
+  public PerformBrowserLogCheckAttribute (bool performBrowserLogCheck)
+  {
+    IsActive = performBrowserLogCheck;
+  }
+
+  public bool IsActive { get; }
+
+  public override void ApplyValue (IDictionary<string, object> dictionary)
+  {
+    ArgumentUtility.CheckNotNull(nameof(dictionary), dictionary);
+
+    dictionary[PropertyKey] = IsActive;
+  }
+}
--- a/Remotion/Web/Development.WebTesting/ITestContext.cs
+++ b/Remotion/Web/Development.WebTesting/ITestContext.cs
@@ -1,0 +1,80 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System.Collections.Generic;
+using JetBrains.Annotations;
+
+namespace Remotion.Web.Development.WebTesting;
+
+/// <summary>
+/// Minimalist abstraction of a test context in order to provide necessary functionality while remaining independent of any specific testing framework. 
+/// </summary>
+/// <remarks>
+/// We provide no default implementation, but you can use the example code with NUnit.
+/// </remarks>
+/// <example>
+/// <code>
+/// public class NUnitTestContext (TestContext testContext, TestExecutionContext testExecutionContext) : ITestContext
+/// {
+///   private static IReadOnlyDictionary&lt;string, object&gt; GetPropertiesFromAttributes (TestExecutionContext testExecutionContext)
+///   {
+///     var testMethod = testExecutionContext.CurrentTest.Method;
+///     if (testMethod == null)
+///       throw new InvalidOperationException($"Current test '{testExecutionContext.CurrentTest.FullName}' does not have a method associated with it.");
+///
+///     return WebTestAttribute.CreatePropertiesFromAttributes(testMethod.MethodInfo);
+///   }
+///
+///   public string TestName => testContext.Test.Type?.Name + "_" + testContext.Test.Name;
+///
+///   public bool IsSuccessful => testExecutionContext.CurrentResult.ResultState != ResultState.Failure;
+///
+///   public void SetFailure (string message) => testExecutionContext.CurrentResult.SetResult(ResultState.Failure, message);
+///
+///   public IReadOnlyDictionary&lt;string, object&gt; Properties { get; } = GetPropertiesFromAttributes(testExecutionContext);
+/// }
+/// </code>
+/// </example>
+public interface ITestContext
+{
+  /// <summary>
+  /// The name of the currently executing test.
+  /// </summary>
+  /// <remarks>
+  /// This is used for logging purposes.
+  /// </remarks>
+  [NotNull]
+  string TestName { get; }
+
+  /// <summary>
+  /// Generic container for context information. 
+  /// </summary>
+  /// <remarks>
+  /// Annotate a test method or class with a <see cref="WebTestAttribute"/> to populate this <see cref="IReadOnlyDictionary{TKey,TValue}"/>.
+  /// </remarks>
+  [NotNull] IReadOnlyDictionary<string, object> Properties { get; }
+
+  /// <summary>
+  /// Indicates whether the current test has been successful so far. Returns <see langword="false"/> only if the test has failed.
+  /// </summary>
+  bool IsSuccessful { get; }
+
+  /// <summary>
+  /// Sets the current test's status to failed, so that <see cref="IsSuccessful"/> is <see langword="false"/>.
+  /// Indicate the reason for the failure in <paramref name="message"/>. 
+  /// </summary>
+  void SetFailure ([NotNull] string message);
+}
--- a/Remotion/Web/Development.WebTesting/TestContextExtensions.cs
+++ b/Remotion/Web/Development.WebTesting/TestContextExtensions.cs
@@ -1,0 +1,60 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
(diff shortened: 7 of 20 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests/IntegrationTest.cs
```diff
@@ -73,6 +75,7 @@ public void IntegrationTestSetUp ()
     [TearDown]
     public void IntegrationTestTearDown ()
     {
+      BrowserLogUtility.CheckBrowserLog(_webTestHelper.MainBrowserSession, TestExecutionContext.CurrentContext);
```
MichaelKetting: We don't need this in IntegrationTests.RequireUI because they don't test CSP?

Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
```diff
@@ -87,6 +89,7 @@ public void FullConfiguration ()
       Assert.That(_section.CloseBrowserWindowsOnSetUpAndTearDown, Is.EqualTo(false));
       Assert.That(_section.CleanUpUnmatchedDownloadedFiles, Is.EqualTo(false));
       Assert.That(_section.RequestErrorDetectionStrategyTypeName, Is.EqualTo("requestErrorDetectionStrategy"));
+      Assert.That(_section.CheckBrowserLogLevel, Is.EqualTo(LogLevel.Off));
```
MichaelKetting: In line 47 we set `warning`, but here we assert `Off`. What detail am I missing?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
```diff
@@ -0,0 +1,152 @@
...
+    bool IgnoreLogEntry (BrowserLogEntry entry) => ignoreLogEntries.Any(func => func(entry));
+    var entries = GetUnexpectedBrowserLogEntries(session, minLogLevel, IgnoreLogEntry);
+    try
+    {
+      Assert.That(entries, Is.Empty);
+    }
+    catch (AssertionException ex)
+    {
+      TestExecutionContext.CurrentContext.CurrentResult.SetResult(ResultState.Failure, ex.Message);
+    }
```
MichaelKetting: Does the AssertionException for this particular case contain helpful details, i.e. is the entries list stringify-able in such a way that the throw-catch is adding value? I didn't try it so I don't know if that's actually the case or not.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
```diff
@@ -0,0 +1,152 @@
...
+    if (context.CurrentResult.ResultState != ResultState.Success)
+      return;
+
+    var isActive = IsBrowserLogCheckActive(context.CurrentTest);
+    if (!isActive)
+      return;
+
+    var minLogLevel = GetMinimumLogLevel(context.CurrentTest);
+    var ignoreLogEntries = GetBrowserLogIgnoredEntriesPredicates(context.CurrentTest);
+    bool IgnoreLogEntry (BrowserLogEntry entry) => ignoreLogEntries.Any(func => func(entry));
```
MichaelKetting: Local functions should not be in the middle of other code. usally, we put them at the end of the method. Otherwise, there's no reason to not simply make it an anonymous function.

Given that there's a closure, I would recommend we use an anonymous function.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTest.cs
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  [Test]
+  [TestCase(LogLevel.Off, false)]
+  [TestCase(LogLevel.Severe, true)]
+  [TestCase(LogLevel.Warning, true)]
+  [TestCase(LogLevel.Info, true)]
+  [TestCase(LogLevel.Debug, true)]
+  [TestCase(LogLevel.All, true)]
+  public void BrowserLogUtility_IsBrowserLogCheckActive_Default (LogLevel logLevel, bool expected)
```
MichaelKetting: naming: the class name is usually not part of the test names, also for statis methods. Is there a reason I do not know?

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
```diff
@@ -0,0 +1,152 @@
...
+using Remotion.Utilities;
+using Remotion.Web.Development.WebTesting.BrowserSession;
+using Remotion.Web.Development.WebTesting.Configuration;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure.BrowserLog;
+
+/// <summary>
+/// Provides utility methods to deal with checking the browser logs at the end of web tests.
+/// </summary>
+public static class BrowserLogUtility
```
MichaelKetting: just making sure: this class is only for our own tests and not meant to be delivered to consumers. Is this the correct intention? Asking because there's a ton of code and test coverage :)

Remotion/Web/Development.WebTesting/Configuration/IWebTestSettings.cs
```diff
@@ -116,6 +117,11 @@ public interface IWebTestSettings
     /// </remarks>
     string RequestErrorDetectionStrategyTypeName { get; }
 
+    /// <summary>
+    /// Gets the global minimum <see cref="LogLevel"/> for browser log entries that make the test fail. Default is <see cref="LogLevel.Off"/>.
+    /// </summary>
+    LogLevel CheckBrowserLogLevel { get; }
```
MichaelKetting: I'm not sure about the name. `CheckBrowserLogLevel` sounds like it is a flag on whether to perform the check, not the minimum level.

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -244,6 +248,14 @@ public string RequestErrorDetectionStrategyTypeName
       get { return (string)this [_requestErrorDetectionStrategyProperty]; }
     }
 
+    /// <summary>
+    /// Gets the global minimum <see cref="LogLevel"/> for browser log entries that make the test fail. Default is <see cref="LogLevel.Off"/>.
+    /// </summary>
+    public LogLevel CheckBrowserLogLevel
```
MichaelKetting: Intentionally not `inheritdoc` here?

MichaelKetting (changes requested):
Got two big questions (property name and BrowserLogUtility location). Depending on outcome, it's changes needed and probably a few details to discuss or there's perfect reasosn :)

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
martin-herwich: This is to ensure that the message is the same as if there was an assertion at the end of the test method.
The exception message contains the unexpected browser log entries.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
martin-herwich: This is intended for use with web tests, not only re-motion, but also in customers' web tests, if they want to check for JavaScript errors.

Remotion/Web/Development.WebTesting/Configuration/IWebTestSettings.cs
martin-herwich: It does both, because there is a LogLevel "Off" that disables all checks.
How about MinLevelForBrowserLogCheck? "Off" would be somewhat weird in this regard, but it fits better with all other levels.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
martin-herwich: You are correct, this test fails. This was probably a rebase artifact.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
MichaelKetting: okay, so there is a benefit to triggering the assertion and catching the exception.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
MichaelKetting: Okay, then we neen to move this to Development.WebTesting and since probably the attributes are also needed, remove any and all NUnit dependencies from the code.
We also need to (well, should) move the test fixtures to the WebTesting.UnitTest project.

Remotion/Web/Development.WebTesting/Configuration/IWebTestSettings.cs
MichaelKetting: `MinimumLevelForBrowserLogCheck`

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/BrowserLog/BrowserLogUtility.cs
martin-herwich: I don't think that it is feasible to make this utility test framework-agnostic. We could create a new assembly `WebDevelopment.WebTesting.NUnit` for NUnit-specific implementations.

MichaelKetting: We'll need to discuss this. I have a strict no-testing-framework-dependency policy in re-motion packages.

martin-herwich: The nuget-specific code is now contained within `NugetTestContext`, while `BrowserLogUtility` is independent of the test framework and is now called from `WebTestHelper.OnTearDown()`.
`WebTestHelper.OnSetup()` now requires an `ITestContext` parameter instead of the test name, while `OnTearDown()` does not take parameters any more.

Remotion/Web/Development.WebTesting/WebTestAttribute.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using JetBrains.Annotations;
+
+namespace Remotion.Web.Development.WebTesting;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
+public abstract class WebTestAttribute : Attribute
```
MichaelKetting, line 32: xml doc?

Remotion/Web/Development.WebTesting/ITestContext.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+using JetBrains.Annotations;
+
+namespace Remotion.Web.Development.WebTesting;
+
+public interface ITestContext
```
MichaelKetting, line 51: xml doc

MichaelKetting (commented):
Since you already did nice naming, I'm not sure if this is still planned:
1 Move/squash "Make BrowserLogUtility..>" and "Integrate call to BrowserLogUtility..." into "Extend WebTesting..."
2 Pick "Add deactivated checks.."
?

Possibly, split the new code into a separate first commit.

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -244,6 +248,12 @@ public string RequestErrorDetectionStrategyTypeName
       get { return (string)this [_requestErrorDetectionStrategyProperty]; }
     }
 
+    /// <inheritdoc/>
+    public LogLevel MinimumLevelForBrowserLogChecks
```
MichaelKetting: In theory: the whole LogLevel set of changes could be a separate commit. This would make it easier to see which part of the change set is for adding the LogLevel and which part for adding and integrating the BrowserLogUtility.

Remotion/Web/Development.WebTesting/BrowserSession/Firefox/FirefoxBrowserSession.cs
```diff
@@ -48,9 +47,7 @@ public FirefoxBrowserSession (Coypu.BrowserSession value, IFirefoxConfiguration
 
     public override IReadOnlyCollection<BrowserLogEntry> GetBrowserLogs ()
     {
-      var logs = _logEntries.ToList();
-      _logEntries.Clear();
-      return logs;
+      return _logEntries;
```
MichaelKetting: Since we now do not clear the browserlog entries upon get (I don't mind), we need to document the breaking change as a separate ticket.

Remotion/Web/Development.WebTesting/BrowserSession/Edge/EdgeBrowserSession.cs
```diff
@@ -30,6 +30,7 @@ namespace Remotion.Web.Development.WebTesting.BrowserSession.Edge
   public class EdgeBrowserSession : BrowserSessionBase<IEdgeConfiguration>
   {
     private readonly IReadOnlyCollection<IBrowserSessionCleanUpStrategy> _cleanUpStrategies;
+    private readonly List<BrowserLogEntry> _browserLogEntries = new();
```
MichaelKetting: Is this change (here and in Edge and in Firefox) actually related to the current PR or a cleanup / improvement for rm-9411. This should be separate to make it easier discoverable.

Remotion/Web/Development.WebTesting/BrowserLog/PerformBrowserLogCheckAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+public class PerformBrowserLogCheckAttribute : WebTestAttribute
```
MichaelKetting, line 30: xmldoc with information when/how to use it.

Remotion/Web/Development.WebTesting/BrowserLog/PerformBrowserLogCheckAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+  public PerformBrowserLogCheckAttribute (bool performBrowserLogCheck)
+  {
+    IsActive = performBrowserLogCheck;
+  }
+
+  public bool IsActive { get; }
+
+  public override void ApplyValue (IDictionary<string, object> dictionary)
+  {
+    dictionary[PropertyKey] = IsActive;
```
MichaelKetting, line 45: argcheck

Remotion/Web/Development.WebTesting/BrowserLog/PerformBrowserLogCheckAttribute.cs
MichaelKetting, line 30: AttributeUsage, please

Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+//
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Remotion.Web.Development.WebTesting.BrowserSession;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+[AttributeUsage(AttributeTargets.Assembly|AttributeTargets.Class|AttributeTargets.Method, AllowMultiple = true)]
+public class IgnoreBrowserLogMessageAttribute : WebTestAttribute
```
MichaelKetting, line 34: xmldoc with information when/how to use it.

Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogMinimumLevelAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using OpenQA.Selenium;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+public class BrowserLogMinimumLevelAttribute : WebTestAttribute
```
MichaelKetting, line 33: xmldoc with information when/how to use it.

MichaelKetting, line 33: AttributeUsage

Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogMinimumLevelAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+  public const string PropertyKey = "MinimumBrowserLogLevel";
+
+  public BrowserLogMinimumLevelAttribute (LogLevel minimumLevel)
+  {
+    MinimumLevel = minimumLevel;
+  }
+
+  public LogLevel MinimumLevel { get; }
+
+  public override void ApplyValue (IDictionary<string, object> dictionary)
```
MichaelKetting, line 44: argcheck

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/WebTestAttributeTest.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+using OpenQA.Selenium;
+using Remotion.Web.Development.WebTesting.BrowserLog;
+using Remotion.Web.Development.WebTesting.BrowserSession;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.BrowserLog;
+
+public class WebTestAttributeTest
```
MichaelKetting: split up please, one TF per class.
add TestFixture attribute
please move to unit tests.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+using Remotion.Web.Development.WebTesting.BrowserSession;
+using Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.BrowserLog;
+
+[TestFixture]
+[PerformBrowserLogCheck(true)]
+[BrowserLogMinimumLevel(LogLevel.Info)]
+[IgnoreBrowserLogMessage("discarded")]
+public class BrowserLogUtilityTestWithFixtureAttribute
```
MichaelKetting: please move to unit tests.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTest.cs
```diff
@@ -0,0 +1,232 @@
...
+using OpenQA.Selenium;
+using Remotion.Web.Development.WebTesting.BrowserLog;
+using Remotion.Web.Development.WebTesting.BrowserSession;
+using Remotion.Web.Development.WebTesting.Configuration;
+using Remotion.Web.Development.WebTesting.IntegrationTests.Infrastructure;
+
+namespace Remotion.Web.Development.WebTesting.IntegrationTests.BrowserLog;
+
+[TestFixture]
+public class BrowserLogUtilityTest
```
MichaelKetting: please move to unit tests.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+
+  [Test]
+  [IgnoreBrowserLogMessage("ignored")]
+  [IgnoreBrowserLogMessage("disregarded")]
+  [IgnoreBrowserLogMessage("omitted")]
+  public void GetBrowserLogIgnoredEntriesPredicate_MultipleAttributesOnTest ()
+  {
+    var result = BrowserLogUtility.GetBrowserLogIgnoredEntriesPredicates(new NUnitTestContext(TestContext.CurrentContext, TestExecutionContext.CurrentContext));
+    Assert.That(result.Count, Is.EqualTo(4));
+    Assert.That(result.Any(func=>func.Invoke(new BrowserLogEntry(LogLevel.Warning, "ignored", DateTime.Now))), Is.True);
```
MichaelKetting: formatting: whitespace around `=>` here and in lines 108ff

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+  {
+    var result = BrowserLogUtility.GetMinimumLogLevel(new NUnitTestContext(TestContext.CurrentContext, TestExecutionContext.CurrentContext));
+    Assert.That(result, Is.EqualTo(LogLevel.Debug));
+  }
+
+  [Test]
+  [BrowserLogMinimumLevel(LogLevel.Info)]
+  [TestCase(LogLevel.Severe)]
+  [TestCase(LogLevel.Debug)]
+  public void GetMinimumLogLevel_SetOnTestSuite (LogLevel logLevel)
```
MichaelKetting:
```suggestion
  public void GetMinimumLogLevel_SetOnTestFixture (LogLevel logLevel)
```

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+  {
+    var result = BrowserLogUtility.IsBrowserLogCheckActive(new NUnitTestContext(TestContext.CurrentContext, TestExecutionContext.CurrentContext));
+    Assert.That(result, Is.EqualTo(false));
+  }
+
+  [Test]
+  [PerformBrowserLogCheck(false)]
+  [TestCase(true)]
+  [TestCase(false)]
+  public void IsBrowserLogCheckActive_DisabledOnTestSuite (bool value)
```
MichaelKetting:
```suggestion
  public void IsBrowserLogCheckActive_DisabledOnTestFixture (bool value)
```
TestSuite would be if we want to not reference the fact that it's a TestFixture-class the attribute is applied to.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/NUnitTestContext.cs
```diff
@@ -0,0 +1,73 @@
...
+  public T GetValueOrDefault<T> (string propertyKey, T defaultValue = default)
+  {
+    if (!_properties.TryGetValue(propertyKey, out var value))
+      return defaultValue;
+
+    if (value is not T typedValue)
+      throw new InvalidCastException($"Value for property '{propertyKey}' is of type unexpected type {value.GetType()} instead of {typeof(T).Name}");
+
+    return typedValue;
+  }
```
MichaelKetting: Since we already have the `Properties` property on ITestContext, is there a benefit to putting GetValueOrDefaulta nd GetCollection on the ITestContext interface? We could instead them extension methods and thus remove them from the copy&paste part of the equation.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/NUnitTestContext.cs
```diff
@@ -0,0 +1,73 @@
...
+    if (testMethod == null)
+      throw new InvalidOperationException($"Current test '{_testExecutionContext.CurrentTest.FullName}' does not have a method associated with it.");
+
+    var testAttributes = testMethod.GetCustomAttributes<WebTestAttribute>(true);
+    var classAttributes = testMethod.TypeInfo.GetCustomAttributes<WebTestAttribute>(true);
+
+    foreach (var attribute in classAttributes.Concat(testAttributes))
+    {
+      attribute.ApplyValue(_properties);
+    }
```
MichaelKetting: I would recommend moving this code to a helper method on WebTestAttribute:
`WebTestAttribute.Apply(testMethod, _properties)`.

This way, we don't have to duplicate the reflection code and provide a tested implementation that matches the attributes (and attributeusages) we created.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -197,13 +198,13 @@ public void OnFixtureSetUp (WindowSize? windowSize = null, [CanBeNull] DriverCon
     /// <summary>
     /// SetUp method for each web test.
     /// </summary>
-    /// <param name="testName">Name of the test being performed.</param>
-    public void OnSetUp ([NotNull] string testName)
+    /// <param name="testContext">An <see cref="ITestContext"/> object for the test being performed.</param>
+    public void OnSetUp ([NotNull] ITestContext testContext)
```
MichaelKetting, line 208: please keep the old syntax as an obsolete method with compiler error. the compiler error can link to the source file in github, telling people where to get their solution. Instead of linking the file, we could also add an xml doc with the file as sample code. after the suggested restructuring, it's very short.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/NUnitTestContext.cs
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public string TestName => _testContext.Test.Type?.Name + "_" + _testContext.Test.Name;
+
+  public bool IsSuccessful => _testExecutionContext.CurrentResult.ResultState != ResultState.Failure;
+
+  public void SetFailure (string message)
+  {
+    _testExecutionContext.CurrentResult.SetResult(ResultState.Failure, message);
+  }
```
MichaelKetting: in order to facilitate shortness, we could switch this to expression syntax, too.

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/NUnitTestContext.cs
```diff
@@ -0,0 +1,73 @@
...
+    _testContext = testContext;
+    _testExecutionContext = testExecutionContext;
+    InitializePropertiesFromAttributes();
+  }
+
+  private void InitializePropertiesFromAttributes ()
+  {
+    var testMethod = _testExecutionContext.CurrentTest.Method;
+    if (testMethod == null)
+      throw new InvalidOperationException($"Current test '{_testExecutionContext.CurrentTest.FullName}' does not have a method associated with it.");
```
MichaelKetting: After the suggested restructring I would recommend inling the Init-method in the ctor to make the file more compact.

We could further shrink this by using a primary ctor (available in c#12 / .net 8):
```
public  class NUnitTestCOntext(TestContext testContext, TestExecutionContext testExecutionContext) : ITestContext
{
  private readonly Dictionary<string, object> _properties = CreatePropertiesFromAttributes(testExecutionContext);

  private Dictionary<> InitializePropertiesFromAttributes ()
  {
    var testMethod = testExecutionContext.CurrentTest.Method;
    if (testMethod == null)
      throw new InvalidOperationException($"Current test '{_testExecutionContext.CurrentTest.FullName}' does not have a method associated with it.");

    var properties = new Dictionary();
    WebTestAttribute.Apply(testMethod, properties)
    return properties;
  }

  ... remaining API
}
```

Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+  public const string PropertyKey = "IgnoreBrowserLogMessage";
+
+  public IgnoreBrowserLogMessageAttribute (string pattern)
+  {
+    Pattern = pattern;
+  }
+
+  public string Pattern { get; }
+
+  public Func<BrowserLogEntry, bool> GetDelegate ()
```
MichaelKetting: given the fact that we're already using a regex pattern, is it helpful to return funcs isntead of a list of regexes? We only use this in one place and it feels like we're over-wrapping this stuff.

Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+using Remotion.Web.Development.WebTesting.BrowserSession;
+
+namespace Remotion.Web.Development.WebTesting.BrowserLog;
+
+[AttributeUsage(AttributeTargets.Assembly|AttributeTargets.Class|AttributeTargets.Method, AllowMultiple = true)]
+public class IgnoreBrowserLogMessageAttribute : WebTestAttribute
+{
+  public const string PropertyKey = "IgnoreBrowserLogMessage";
+
+  public IgnoreBrowserLogMessageAttribute (string pattern)
```
MichaelKetting: this should be xmldoc'd as a regex.
There's also a JetBrains annotation for regex parameters, I think, which then provides intellisense to users.

MichaelKetting (changes requested):
requests for simplification / code moving. I hope they actually improve the design. Let me know if I'm wrong :)

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
martin-herwich: "Suite" in Nunit is a test mehtod with parameters (test cases).

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
martin-herwich: "Suite" in Nunit is a test mehtod with parameters (test cases).

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTest.cs
martin-herwich: We need the web test config for these tests.

Remotion/Web/Development.WebTesting.IntegrationTests/BrowserLog/BrowserLogUtilityTestWithFixtureAttribute.cs
martin-herwich: We need the web test config for these tests.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -194,16 +195,22 @@ public void OnFixtureSetUp (WindowSize? windowSize = null, [CanBeNull] DriverCon
         EnsureCursorIsOutsideBrowserWindow();
     }
 
+    [Obsolete("Use OnSetUp(ITestContext) instead, using the implementation example given in the ITestContext documentation, or your own implementation.", true)]
```
MichaelKetting:
```suggestion
    [Obsolete("Use OnSetUp(ITestContext) instead, using the implementation example given in the ITestContext documentation, or your own implementation. (Version: 8.0.0)", true)]
```

Remotion/Web/Development.WebTesting/WebTestAttribute.cs
```diff
@@ -0,0 +1,59 @@
...
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using JetBrains.Annotations;
```
MichaelKetting, line 21: license header

Remotion/Web/Development.WebTesting/TestContextExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting;
+
+public static class TestContextExtensions
+{
+  public static IReadOnlyCollection<T> GetCollection<T> (this ITestContext testContext, string propertyKey)
+  {
+    ArgumentUtility.CheckNotNullOrEmpty(nameof(propertyKey), propertyKey);
```
MichaelKetting:
```suggestion
  public static IReadOnlyCollection<T> GetCollection<T> (this ITestContext testContext, string propertyKey)
  {
    ArgumentUtility.CheckNotNull(nameof(testContext), testContext);
    ArgumentUtility.CheckNotNullOrEmpty(nameof(propertyKey), propertyKey);
```

Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
```diff
@@ -0,0 +1,59 @@
...
+
+  public override void ApplyValue (IDictionary<string, object> dictionary)
+  {
+    ArgumentUtility.CheckNotNull(nameof(dictionary), dictionary);
+
+    if (!dictionary.TryGetValue(PropertyKey, out var value) || value is not IList<Regex> delegateList)
+    {
+      delegateList = new List<Regex>();
+      dictionary[PropertyKey] = delegateList;
+    }
```
MichaelKetting:
```suggestion
    if (!dictionary.TryGetValue(PropertyKey, out var value) || value is not IList<Regex> patterns)
    {
      patterns = new List<Regex>();
      dictionary[PropertyKey] = patterns;
    }
```

MichaelKetting (approved):

PR #1301, opened 2025-03-05
RM-9002 Add async guard to the SmartPage

a-ctor:
https://re-motion.atlassian.net/browse/RM-9002

### DIFF
--- a/Remotion/Web/Core/UI/SmartPage.cs
+++ b/Remotion/Web/Core/UI/SmartPage.cs
@@ -669,6 +669,13 @@ public class SmartPage : Page, ISmartPage, ISmartNavigablePage
 
   protected override void OnInit (EventArgs e)
   {
+    if (AsyncMode)
+    {
+      throw new InvalidOperationException(
+          "SmartPage does not support Async=\"true\" or manually setting AsyncMode=true."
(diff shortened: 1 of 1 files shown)


Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -669,6 +669,13 @@ protected virtual bool IsCspReportOnlyEnabled
 
   protected override void OnInit (EventArgs e)
   {
+    if (AsyncMode)
```
MichaelKetting, line 672: We don't need a second check in oOnLoadComplete or OnPreRender, correct? i.e. it's not possible to enable AsyncMode after OnInit anyway?

MichaelKetting (approved):

Remotion/Web/Core/UI/SmartPage.cs
a-ctor: The `AsyncMode` property itself does nothing. `Async="true"` affects the generated class and will set `AsyncMode = true;` in the constructor. So it is a good way to detect `Async="true"` but there is no reason why `AsyncMode` should be set for any other reason. So yes we are fine checking just here in `OnInit` :)

MichaelKetting: :)

PR #1304, opened 2025-03-10
Feature/rm 9438 überprüfen von browserfehlern in den re motion web tests

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9438

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionParserTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionParserTest.cs
@@ -17,6 +17,7 @@
 using System;
 using Coypu;
 using NUnit.Framework;
+using Remotion.Web.Development.WebTesting.BrowserLog;
 using Remotion.Web.Development.WebTesting.CompletionDetectionStrategies;
 using Remotion.Web.Development.WebTesting.ExecutionEngine.PageObjects;
 using Remotion.Web.Development.WebTesting.FluentControlSelection;
@@ -25,6 +26,7 @@ using Remotion.Web.Development.WebTesting.RequestErrorDetectionStrategies;
 namespace Remotion.Web.Development.WebTesting.IntegrationTests
 {
   [TestFixture]
+  [PerformBrowserLogCheck(false)]
   public class AspNetRequestErrorDetectionParserTest : IntegrationTest
   {
     [Test]
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionStrategyTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionStrategyTest.cs
@@ -16,6 +16,7 @@
 // 
 using System;
 using NUnit.Framework;
+using Remotion.Web.Development.WebTesting.BrowserLog;
 using Remotion.Web.Development.WebTesting.CompletionDetectionStrategies;
 using Remotion.Web.Development.WebTesting.ExecutionEngine.PageObjects;
 using Remotion.Web.Development.WebTesting.FluentControlSelection;
@@ -28,6 +29,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
   public class AspNetRequestErrorDetectionStrategyTest : IntegrationTest
   {
     [Test]
+    [PerformBrowserLogCheck(false)]
     public void Parse_PostbackError ()
     {
       var aspNetRequestErrorDetectionParser = new AspNetRequestErrorDetectionStrategy();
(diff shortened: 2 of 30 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests/SmartPageErrorTest.cs
```diff
@@ -45,6 +47,7 @@ public void AsynchronousPageRequestError_ShowsYellowPageOfDeath ()
     }
 
     [Test]
+    [PerformBrowserLogCheck(false)]
```
MichaelKetting: why globally for the file and locally on the method?

Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
```diff
@@ -18,6 +18,7 @@
 using System.IO;
 using System.Threading.Tasks;
 using NUnit.Framework;
+using Remotion.Web.Development.WebTesting.BrowserLog;
```
MichaelKetting, line 21: shouldn't we just ignore globally here instead of ignoring individual log messages? it's only testing webtest infrastructure.

Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
```diff
@@ -33,6 +34,7 @@ public class FileDownloadTest : IntegrationTest
     private const string c_sampleXmlFileName = "SampleXmlFile_06d6ff4d-c124-4d3f-9d96-5e4f2d0c7b0c.xml";
 
     [Test]
+    [IgnoreBrowserLogMessage(@"The file at 'http://.*/res/Remotion.Web.Development.WebTesting.TestSite.Shared/FileDownloadTest.aspx' was loaded over an insecure connection. This file should be served over HTTPS")]
```
MichaelKetting: would it be a good idea to put the messages into a shared const?
(provided they don't go way completely when we disable the log check)

Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionStrategyTest.cs
```diff
@@ -28,6 +29,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
   public class AspNetRequestErrorDetectionStrategyTest : IntegrationTest
   {
     [Test]
+    [PerformBrowserLogCheck(false)]
```
MichaelKetting, line 32: disable globally?

Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionParserTest.cs
```diff
@@ -28,6 +29,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
   public class AspNetRequestErrorDetectionParserTest : IntegrationTest
   {
     [Test]
+    [PerformBrowserLogCheck(false)]
```
MichaelKetting: disable globally?

MichaelKetting (approved):

Remotion/Web/Development.WebTesting.IntegrationTests/FileDownloadTest.cs
martin-herwich: Do you mean we should disable the browser log checks for the whole assembly ("testing webtest infrastructure"), or for the class?

MichaelKetting: sorry, for the class. I *was* wondering if 'globally' is the best term for this :D

Remotion/Web/Development.WebTesting.IntegrationTests/AspNetRequestErrorDetectionStrategyTest.cs
martin-herwich: Globally as in "the whole class" or "the whole assembly"?

Remotion/Web/Development.WebTesting/BrowserLog/IgnoreBrowserLogMessageAttribute.cs
```diff
@@ -42,6 +43,17 @@ public IgnoreBrowserLogMessageAttribute ([NotNull] [RegexPattern] string pattern
...
   }
 
+  public IgnoreBrowserLogMessageAttribute ([NotNull] [RegexPattern] string pattern, [NotNull] [ItemCanBeNull] object?[] templateArgs)
+  {
+    ArgumentUtility.CheckNotNullOrEmpty(nameof(pattern), pattern);
+    ArgumentUtility.CheckNotNullOrEmpty(nameof(templateArgs), templateArgs);
+
+    if (templateArgs.Any())
+      pattern = string.Format(pattern, templateArgs);
+
```
MichaelKetting:
```suggestion
```

I would recommend removing this guard:
1. if it can be optimized, .net will optimize
2. in case that pattern is a placeholder text but no arguments are supplied, string-format will fail, which is correct. with the guard, this check is circumvented.

MichaelKetting (approved):

PR #1307, opened 2025-03-12
RM-9453 Turn off CLS compliance

a-ctor:
https://re-motion.atlassian.net/browse/RM-9453


MichaelKetting (approved):
Please don't forget the breaking change ticket.

PR #1309, opened 2025-03-12
RM-9439 Activate CSP

martin-herwich:
https://re-motion.atlassian.net/browse/RM-9439

### DIFF
--- a/Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
+++ b/Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
@@ -1,6 +1,5 @@
 ﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
 // SPDX-License-Identifier: LGPL-2.1-or-later
-using System.Web.UI;
 using Remotion.ServiceLocation;
 using Remotion.Utilities;
 using Remotion.Web.UI;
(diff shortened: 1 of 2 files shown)


Remotion/Web/Test.Shared/TestCspDefaultsProvider.cs
```diff
@@ -0,0 +1,13 @@
+using Remotion.ServiceLocation;
+using Remotion.Web.ContentSecurityPolicy;
+
+namespace Remotion.Web.Test.Shared;
+
+[ImplementationFor(typeof(ICspDefaultsProvider), Lifetime = LifetimeKind.Singleton)]
+public class TestCspDefaultsProvider : CspDefaultsProvider
```
MichaelKetting: We got a pattern for overriding implementations:
1 add a const `Position=0` to the `CspDefualtsProvider` and use this in the attribute
2 add a const `Position=CspDefaultsProvider.Const + 1` and use this in the attribute for TestCspDefaultsProvider.

right now, the order is not defined and i'm suprrised it's not blowing up because of ambiguity

Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
```diff
@@ -31,7 +31,7 @@ public CspDefaultsProvider ()
   {
   }
 
-  private CspDefaultsProvider (
+  protected CspDefaultsProvider (
```
MichaelKetting: I'm not sure if using inheritance isntead of just re-implementing the interface on a new type wouldn't be a cleaner design. it's not that much more code and no more logic and it keeps this type not-inheritable. also, we should make the class sealed.

Remotion/Web/Test.Shared/TestCspDefaultsProvider.cs
```diff
@@ -0,0 +1,13 @@
+using Remotion.ServiceLocation;
```
MichaelKetting, line 3: lic-header please

MichaelKetting (changes requested):

Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
martin-herwich: If this class was not intended to be inherited from, it does not make much sense to have the configuration as properties with backing fields, instead of returning hard-coded values.
I think the issue here is that the class serves two purposes: 1) as a configuration, and 2) as a service providing the configuration (that can be resolved from the container).
Since the attribute-based container configuration cannot register instances, we have to create a new type for a different configuration.
We can resolve this by registering a service `ICspDefaultsProvider` that does not have the properties with the defaults, but only a Method `GetDefaults()` that returns a `struct CspDefaults`.
Alternatively, we can extract an abstract base class `CspDefaultsProvider`, and have a derived class for each concrete configuration.
What do you think, @MichaelKetting?

MichaelKetting: @martin-herwich 15minute design session with @a-ctor and me to discuss this in person. 

The backing fields are needed because of the Empty-option. but it is possible we might be missing a factory method to support other setups.

Remotion/Web/Core/ContentSecurityPolicy/CspDefaultsProvider.cs
```diff
@@ -12,9 +10,20 @@ namespace Remotion.Web.ContentSecurityPolicy;
...
 {
+  public const int Position = 0;
+
+  public static CspDefaultsProvider Create (
+      bool isCspEnabledDefault,
+      CspHeader cspHeaderDefault,
+      bool isCspReportOnlyEnabledDefault,
+      CspHeader cspReportOnlyHeaderDefault)
+  {
+    return new CspDefaultsProvider(isCspEnabledDefault, cspHeaderDefault, isCspReportOnlyEnabledDefault, cspReportOnlyHeaderDefault);
```
MichaelKetting: argnullchecks please

MichaelKetting (approved):

PR #1310, opened 2025-03-17
RM-9458 Fix performance problem in BrowserLogUtility.GetUnexpectedBrowserLogEntries

a-ctor:
https://re-motion.atlassian.net/browse/RM-9458

### DIFF
--- a/Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogUtility.cs
+++ b/Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogUtility.cs
@@ -16,14 +16,18 @@
 //
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using JetBrains.Annotations;
 using OpenQA.Selenium;
+using OpenQA.Selenium.Support.Extensions;
 using Remotion.Utilities;
 using Remotion.Web.Development.WebTesting.BrowserSession;
+using Remotion.Web.Development.WebTesting.WebDriver;
+using Remotion.Web.Development.WebTesting.WebDriver.Configuration;
 
 namespace Remotion.Web.Development.WebTesting.BrowserLog;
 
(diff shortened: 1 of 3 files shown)


Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogUtility.cs
```diff
@@ -68,27 +72,57 @@ public static IReadOnlyCollection<Regex> GetBrowserLogIgnoredEntryRegexes (ITest
...
   {
     ArgumentUtility.CheckNotNull(nameof(session), session);
+    ArgumentUtility.CheckNotNull(nameof(configuration), configuration);
     ArgumentUtility.CheckNotNull(nameof(messageFilter), messageFilter);
 
-    var remainingRetries = 50;
-    IReadOnlyCollection<BrowserLogEntry> logEntries;
-    do
+    IReadOnlyCollection<BrowserLogEntry> browserLogs;
+    if (configuration.UseBidiLog())
```
MichaelKetting: Wouldn't it be better to test IBrowserSession.Window. We can softcast if the Window is also a BrowserSession (it always is) and then use Window.Browser.UseBidiLog(). that way, we don't need the extra parameter.

Remotion/Web/Development.WebTesting/BrowserLog/BrowserLogUtility.cs
```diff
@@ -68,27 +72,57 @@ public static IReadOnlyCollection<Regex> GetBrowserLogIgnoredEntryRegexes (ITest
...
+      ((IWebDriver)session.Driver.Native).ExecuteJavaScript($"console.error('{browserLogMarker}');");
+
+      var i = 0;
+      while (true)
+      {
+        if (i > maxRetries)
+        {
+          throw new InvalidOperationException(
+              "Waiting for unexpected browser messages failed because"
+              + " the marker log message was not returned within the timeout.");
```
MichaelKetting, line 103: Should this be a System.TimeoutException?

MichaelKetting (approved):
Simplification requests

PR #1318, opened 2025-04-10
RM-9470 handle javascript urls in csp enabled text writer

a-ctor:
Might be easier to review commit by commit as the first commit does a bit of refactoring as pre-work.

### DIFF
--- a/Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
+++ b/Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
@@ -22,6 +22,7 @@ using System.Web;
 using System.Web.UI;
 using Remotion.Utilities;
 using Remotion.Web.UI;
+using Remotion.Web.UI.Controls;
 using Remotion.Web.UI.Controls.Rendering;
 
 namespace Remotion.Web.ContentSecurityPolicy
@@ -33,7 +34,8 @@ namespace Remotion.Web.ContentSecurityPolicy
   {
     private enum RegisteredEventType
     {
-      InlineEventAttribute
+      InlineEventAttribute,
+      EventListener
     }
 
     private record RegisteredEvent (string Key, RegisteredEventType Type, string EventType, string EventValue, string OriginalValue)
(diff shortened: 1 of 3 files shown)


Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -43,11 +45,30 @@ public string GetScript (string eventTargetID)
...
+    private delegate string HrefFormatterAction (ReadOnlySpan<char> value);
+
+    private static readonly HrefFormatterAction s_hrefActionFormatter = static value =>
+        $$"""
+        let __defaultPrevented = event.defaultPrevented;
+
+        event.preventDefault();
+        event.preventDefault = () => {
+          __defaultPrevented = true;
+        };
```
martin-herwich, line 63: If I read this correctly, `preventDefault()` and `__defaultPrevented` are used to prevent the event handler from being executed. Either way, what does "default" mean in this context?
A code comment that explains the mechanism and why we do that might be helpful. If not as code comment, then in the "solution" field of the Jira issue.

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -241,10 +257,8 @@ public override void WriteAttribute (string name, string value)
 
     public override void WriteAttribute (string name, string value, bool fEncode)
     {
-      if (TryCreateEventRegistration(name, value, fEncode, out var registration))
+      if (TryCreateEventRegisteredEvent(name, value, fEncode, out var registeredEvent))
```
martin-herwich: `TryCreateRegisteredEvent` (without the first "Event")?

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -31,6 +31,23 @@ namespace Remotion.Web.ContentSecurityPolicy
   /// </summary>
   public class CspEnabledHtmlTextWriter : HtmlTextWriter
   {
+    private enum RegisteredEventType
+    {
+      InlineEventAttribute
+    }
```
martin-herwich: Do we anticipate any other event types?

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
a-ctor: The "default" in this case is whatever the target of the event does as default response to the event. In this case we have a `<a href="#" ...>`, with the default being a navigation event that is being prevented.

I have moved the description of the solution from the description to the solution part of the issue for clarity :)

martin-herwich (approved):

PR #1320, opened 2025-04-22
RM-9501 use connection string for database configuration

a-ctor:


Remotion/Data/DomainObjects.ObjectBinding.IntegrationTests/Database/CreateDB.sql
```diff
@@ -8,16 +8,6 @@ END
 GO
 
 CREATE DATABASE DBPrefix_RemotionDataDomainObjectsObjectBindingIntegrationTestDomain
-ON PRIMARY (
```
MichaelKetting, line 11: Please add a note to the jira ticket that we have rmeoved the database location / no longer persist the database outside of the docker image in case we ever run across this topic again.

Remotion/Data/DomainObjects.ObjectBinding.IntegrationTests/DatabaseConfiguration.cs
```diff
@@ -24,48 +24,21 @@ namespace Remotion.Data.DomainObjects.ObjectBinding.IntegrationTests
 {
```
MichaelKetting, line 24: This will also need to be done in relinq SqlBackend. do you want to create the ticket for this now or wait until later to remember and create the ticket?

MichaelKetting (approved):

PR #1321, opened 2025-04-22
Feature/rm 9502 repair path problems in solution

a-ctor:


MichaelKetting (approved):
For the jira ticket, please update the subject and description to note that this only affects doc-includes and project files but no path-logic in the code. this helps with ignoring the ticket when looking for interesting stuff.

PR #1323, opened 2025-04-25
Feature/rm 9505 repair tests for linux

a-ctor:


Remotion/Development/Core/NewlineStringExtensions.cs
```diff
@@ -0,0 +1,66 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Text;
+
+namespace Remotion.Development;
+
+public static class NewlineStringExtensions
+{
+  public static string NormalizeToCurrentOsStyleLineEndings (this string value)
```
MichaelKetting: naming: 
```suggestion
  public static string NormalizeToCurrentOSStyleLineEndings (this string value)
```
2-letter acronym uses upper-case

Remotion/Development/Core/NewlineStringExtensions.cs
```diff
@@ -0,0 +1,66 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Text;
+
+namespace Remotion.Development;
+
+public static class NewlineStringExtensions
```
MichaelKetting: xml doc for new classes and possibly the methods. It might be helpful to know what/how things are done.

MichaelKetting: Suggestion: string.ReplaceLineEndings() for Current-OS and string.ReplaceLineEndings("\n") and string.ReplaceLineEndings("\r\n") for the other ones. .NET already does a mighty fine job in implemnting a fast algorithm

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs#L1503-L1542

only question is: do we need/keep the helper methods or do we inline? Since it's an instance method on String and if we mostly need the OS-sniffing one, we might want to simplify completely.

Remotion/Core/UnitTests/Reflection/TypeDiscovery/AssemblyFinding/AssemblyFinderIntegrationTest.cs
```diff
@@ -280,7 +280,7 @@ static void Test ()
     private static Assembly CompileTestAssemblyInMemory (string assemblyName, params string[] referencedAssemblies)
     {
       AssemblyCompiler assemblyCompiler = AssemblyCompiler.CreateInMemoryAssemblyCompiler(
-          c_testAssemblySourceDirectoryRoot + "\\" + assemblyName,
+          c_testAssemblySourceDirectoryRoot + Path.DirectorySeparatorChar + assemblyName,
```
MichaelKetting: given the rest of the file: `"/"` instead?

Remotion/Core/UnitTests/ServiceLocation/BootstrapServiceConfigurationTests/GetLoggerFactoryTest.cs
```diff
@@ -179,7 +181,7 @@ private static void SetStackTraceForFirstCallToGetLoggerFactoryOnBootstrapServic
 
     private static Assembly CompileTestAssemblyInMemory (string assemblyName, params string[] referencedAssemblies)
     {
-      AssemblyCompiler assemblyCompiler = AssemblyCompiler.CreateInMemoryAssemblyCompiler(c_testAssemblySourceDirectoryRoot + "\\" + assemblyName, referencedAssemblies);
+      AssemblyCompiler assemblyCompiler = AssemblyCompiler.CreateInMemoryAssemblyCompiler(c_testAssemblySourceDirectoryRoot + Path.DirectorySeparatorChar + assemblyName, referencedAssemblies);
```
MichaelKetting: `"/"` ?

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -42,10 +43,15 @@ public class TestDomainDBScriptGenerationTest : StandardMappingTest
...
+        Assert.Ignore("This test is only supported on Windows as it compares generated content to a checked in file that will only match on the platform it was generated.");
+
       foreach (var script in GenerateScripts())
       {
         var storageProviderName = script.StorageProviderDefinition.Name;
-        Assert.That(script.SetUpScript, Is.EqualTo(GetStoredScript($"DataDomainObjects_{storageProviderName}_SetupDB.sql")));
+        var actual = script.SetUpScript;
+        var expected = Is.EqualTo(GetStoredScript($"DataDomainObjects_{storageProviderName}_SetupDB.sql"));
+        Assert.That(actual, expected);
         Assert.That(script.TearDownScript, Is.EqualTo(GetStoredScript($"DataDomainObjects_{storageProviderName}_TearDownDB.sql")));
```
MichaelKetting: I don't get this change. Why alter the Setup-part of the script test and leave the Teardown-part unchanged. We should keep with as is since there is no need to touch them.

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -42,10 +43,15 @@ public class TestDomainDBScriptGenerationTest : StandardMappingTest
     [Test]
     public void GeneratedScriptsMatchStoredScripts ()
     {
+      if (!OperatingSystem.IsWindows())
```
MichaelKetting: Does this mean we get a complete mix of Windows and Linux style lineendings and don't want to bother with cleaning them up? I think we should create a follow-up ticket because disabling the test and just ignoring it feels like an incomplete solution to the problem.

Remotion/Data/DomainObjects.UnitTests/Database/TestDomainDBScriptGenerationTest.cs
```diff
@@ -104,7 +110,7 @@ private string GetStoredScript (string name)
       var filePath = Path.GetFullPath(Path.Combine(TestContext.CurrentContext.TestDirectory, $"Database/{name}"));
       Assert.That(filePath, Does.Exist, "Could not find stored SQL script '{0}'.");
 
-      return File.ReadAllText(filePath);
+      return File.ReadAllText(filePath).NormalizeToCurrentOsStyleLineEndings();
     }
```
MichaelKetting: Is this method used in more than the one script we already ignore in Linux?

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Unload/UnloadDataTest.cs
```diff
@@ -507,7 +507,7 @@ public void UnloadData_ChangedCollection ()
           () => UnloadService.UnloadData(TestableClientTransaction, DomainObjectIDs.OrderItem1),
           Throws.InvalidOperationException
               .With.Message.EqualTo(
-                  "The relations of object 'OrderItem|2f4d42c7-7ffa-490d-bfcd-a9101bbf4e1a|System.Guid' cannot be unloaded.\r\n"
+                  $"The relations of object 'OrderItem|2f4d42c7-7ffa-490d-bfcd-a9101bbf4e1a|System.Guid' cannot be unloaded.{Environment.NewLine}"
```
MichaelKetting, line 510: stylistic question:
```suggestion
                  "The relations of object 'OrderItem|2f4d42c7-7ffa-490d-bfcd-a9101bbf4e1a|System.Guid' cannot be unloaded." + Environment.NewLine
```
?

There's a bunch of those tests. Not sure if the string-interpolation syntax to append a trailing newline is the best way to structure this for readability. thoughts and opinions (beside "it's already done now")?

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/FulltextIntegrationTests.cs
```diff
@@ -33,16 +33,19 @@ public class FulltextIntegrationTests : IntegrationTestBase
   {
     public override void OneTimeSetUp ()
     {
+      if (!OperatingSystem.IsWindows())
+        Assert.Ignore("FullTextSearch Tests are only enabled on Windows.");
```
MichaelKetting: SQL Server isn't saying that Fulltext is unavailable on Linux. What's the issue? I need a good one to sign off on this :)

https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-editions-and-components-2019?view=sql-server-ver16#unsupported-features-and-services

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SchemaGeneration/SqlTableTypeScriptElementFactoryTest.cs
```diff
@@ -82,11 +83,13 @@ public void GetCreateElement_TableTypeDefinition_Plain ()
...
-          + "  [Column1] varchar(100) NOT NULL,\r\n"
-          + "  [Column2] bit NULL\r\n"
-          + ")";
+          """
+          CREATE TYPE [SchemaName].[TypeName] AS TABLE
+          (
+            [Column1] varchar(100) NOT NULL,
+            [Column2] bit NULL
+          )
+          """.NormalizeToCurrentOsStyleLineEndings();
```
MichaelKetting: this makes things so much easier for us

MichaelKetting: I wonder if the .NET (well the Roslyn) team might consider detecting this pattern once we switch to the string.ReplaceLineEndings(...) options when dealing with literal strings. I haven't checked the Github list but it might be a fun thing to do :D

Remotion/Data/DomainObjects.UnitTests/Queries/Configuration/Loader/BaseDirectoryBasedOptionalQueryFileFinderTest.cs
```diff
@@ -48,7 +48,7 @@ public void GetQueryFilePath_WithFileExists_ReturnsAbsolutePath ()
       var appContextProviderMock = new Mock<IAppContextProvider>();
       appContextProviderMock.SetupGet(e => e.BaseDirectory).Returns(_directory);
 
-      var queryFile = Path.Combine(_directory, "queries.xml");
+      var queryFile = Path.Combine(_directory, "Queries.xml");
```
MichaelKetting, line 51: Ah, this is where you were unhappy ;)

Remotion/Data/DomainObjects.UnitTests/ReflectionUtilityTest.cs
```diff
@@ -98,6 +98,9 @@ public void GetAssemblyPath_FromNonPersistentAssembly ()
     [Test]
     public void GetAssemblyPath_FromUncPath ()
     {
+      if (!OperatingSystem.IsWindows())
+        Assert.Ignore("This test only works on Windows.");
+
       var assemblyMock = new Mock<FakeAssembly>(MockBehavior.Strict);
       assemblyMock.Setup(_ => _.Location).Returns(@"\\server\share\directory\assembly.dll");
       Assert.That(() => ReflectionUtility.GetAssemblyDirectory(assemblyMock.Object), Is.EqualTo(@"\\server\share\directory"));
```
MichaelKetting, line 106: While this test is already sort of exotic, why doesn't it work on Linux? (I haven't checked if UNC is available on Linux but it's network, sooo...)

Remotion/Development/Core/NewlineStringExtensions.cs
MichaelKetting: okay, while reading the changed files: so far, it looks like this isn't just a suggestion anymore but a 'want' :)
please don't forget to drop the uneeded `using`, too.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlIndexDefinitionScriptElementFactory.cs
```diff
@@ -45,9 +45,10 @@ public override IScriptElement GetCreateElement (SqlIndexDefinition indexDefinit
          ownerName.EntityName,
          GetIndexedColumnNames(indexDefinition.Columns),
          indexDefinition.IncludedColumns != null
-             ? "\r\n  INCLUDE (" + string.Join(", ", indexDefinition.IncludedColumns.Select(c => "[" + c.Name + "]")) + ")"
+             ? $"{Environment.NewLine}  INCLUDE (" + string.Join(", ", indexDefinition.IncludedColumns.Select(c => "[" + c.Name + "]")) + ")"
```
MichaelKetting:
```suggestion
             ? Environment.NewLine + "  INCLUDE (" + string.Join(", ", indexDefinition.IncludedColumns.Select(c => "[" + c.Name + "]")) + ")"
```
Since there is already a ton of string contat going on, using string interpolation to prefix the newline feels very wrong

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlIndexScriptElementFactoryBase.cs
```diff
@@ -60,7 +61,7 @@ protected virtual string GetCreateIndexOptions (IEnumerable<string> optionItems)
 
       var filteredItems = optionItems.Except(new[] { string.Empty, null }).ToList();
       if (filteredItems.Any())
-        return "\r\n  WITH (" + string.Join(", ", filteredItems) + ")";
+        return $"{Environment.NewLine}  WITH (" + string.Join(", ", filteredItems) + ")";
```
MichaelKetting:
```suggestion
        return Environment.NewLine + "  WITH (" + string.Join(", ", filteredItems) + ")";
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlUnionViewScriptElementFactory.cs
```diff
@@ -36,16 +36,17 @@ protected override string GetSelectStatements (UnionViewDefinition unionViewDefi
       foreach (var tableDefinition in unionViewDefinition.GetAllTables())
       {
         if (createSelectStringBuilder.Length > 0)
-          createSelectStringBuilder.AppendFormat("\r\n  UNION ALL\r\n");
+          createSelectStringBuilder.AppendFormat($"{Environment.NewLine}  UNION ALL{Environment.NewLine}");
```
MichaelKetting:
```suggestion
          createSelectStringBuilder.Append($"{Environment.NewLine}  UNION ALL{Environment.NewLine}");
```

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocDateTimeValueImplementation/DateTimeFormatterTest.cs
```diff
@@ -42,7 +42,7 @@ public void FormatTimeValue_WithShowSecondsTrue_ReturnsTimeWithSeconds ()
 
       using (new CultureScope("en-US"))
       {
-        Assert.That(formatter.FormatTimeValue(new DateTime(2013, 06, 20, 14, 30, 40), true), Is.EqualTo("2:30:40 PM"));
+        Assert.That(formatter.FormatTimeValue(new DateTime(2013, 06, 20, 14, 30, 40), true), Does.Match(@"^2:30:40[\s\u00a0]PM$"));
```
MichaelKetting: What is Linux doing here?

Remotion/Security/UnitTests/Metadata/LocalizationFileNameStrategyTest.cs
```diff
@@ -96,8 +96,8 @@ public void GetLocalizationFileNames_WithoutBaseDirectory ()
 
       Assert.That(localizationFileNames, Is.Not.Null);
       Assert.That(localizationFileNames.Length, Is.EqualTo(2));
-      Assert.That(localizationFileNames, Has.Member(@".\TwoLocalizationFilesIncludingInvariantCulture.Localization.de.xml"));
-      Assert.That(localizationFileNames, Has.Member(@".\TwoLocalizationFilesIncludingInvariantCulture.Localization.xml"));
+      Assert.That(localizationFileNames, Has.Member(Path.Combine(".", "TwoLocalizationFilesIncludingInvariantCulture.Localization.de.xml")));
```
MichaelKetting: is Path.Combine doing the same thing ehre as the original hard-coded string or is it trimming the path?

Remotion/Validation/UnitTests/Merging/DiagnosticOutputValidationRuleMergeDecoratorTest.cs
```diff
@@ -151,50 +152,56 @@ public void Merge_WithValidationCollectors ()
...
+                  ADDED META VALIDATION RULES:
+                  -> MaxLengthPropertyMetaValidationRule
+
+          -> ApiBasedValidationRuleCollectorProvider#TypeWithoutBaseTypeCollector2
+
+              -> Remotion.Validation.UnitTests.Implementation.TestDomain.TypeWithoutBaseType#Property2
+                  REMOVED VALIDATORS:
+                  -> NotEmptyOrWhitespaceValidator#Conditional (x1)
+                  -> MaximumLengthValidator#TypeWithoutBaseTypeCollector1 (x1)
+          """.NormalizeToCurrentOsStyleLineEndings();
```
MichaelKetting: I sure do like raw string literals

Remotion/Web/UnitTests/Core/UI/Controls/ListMenuImplementation/Rendering/ListMenuRendererTest.cs
```diff
@@ -87,7 +87,7 @@ public void Render_RegistersMenuItems ()
     {
       SetUpGetPostBackLinkExpectations(false);
 
-      string script = "ListMenu.Initialize ('#{0}');\r\n" +
+      string script = $"ListMenu.Initialize ('#{{0}}');{Environment.NewLine}" +
                       "ListMenu.AddMenuInfo ('#{0}', \r\n\tnew ListMenu_MenuInfo ('{0}', new Array (\r\n" +
                       "\t\t{1},\r\n\t\t{2},\r\n\t\t{3},\r\n\t\t{4} ) ) );\r\n" +
```
MichaelKetting: since there's more newline char, I would suggest reverting and fixing the one spot in the prod-code to use hardcoded newlines rather than mixing it here in the test with chaotic readability

MichaelKetting (commented):
I got most of it reviewed, rest soon?

Remotion/Development/UnitTesting.IsolatedCodeRunner/IsolatedCodeRunner.cs
```diff
@@ -101,8 +102,16 @@ public void Run (params string[] args)
...
+      if (OperatingSystem.IsWindows())
+      {
+        startInfo.FileName = isolatedCodeRunnerPath;
+        startInfo.Arguments = string.Join(" ", args.Select(e => $@"""{e.Replace("\"", "\"\"")}"""));
+      }
+      else
+      {
+        startInfo.FileName = "dotnet";
+        startInfo.Arguments = string.Join(" ", ((string[])["exec", isolatedCodeRunnerPath]).Concat(args).Select(e => $@"""{e.Replace("\"", "\"\"")}"""));
+      }
```
MichaelKetting, line 114: Do our tests exercise arguments etc?

Remotion/Web/UnitTests/Core/Resources/ResourcePathBuilderBasedStaticResourceCacheKeyProviderTest.cs
```diff
@@ -118,21 +118,48 @@ public void AppendResourcesFolderDetails ()
...
+                                        (Path.Combine("Remotion.Web", "Html", "file1.txt"), 7L, 130000022222222222),
+                                        (Path.Combine("Remotion.Web", "Html", "sub", "file2.txt"), 3L, 130000044444444444)
+                                    };
+
+      if (!OperatingSystem.IsWindows())
+      {
+        // On Linux the order the files are listed is different so we need to reorder the entries
+        (expectedCacheKeyEntries[0], expectedCacheKeyEntries[2]) = (expectedCacheKeyEntries[2], expectedCacheKeyEntries[0]);
+        (expectedCacheKeyEntries[1], expectedCacheKeyEntries[3]) = (expectedCacheKeyEntries[3], expectedCacheKeyEntries[1]);
+      }
```
MichaelKetting, line 160: Something for future refernece: should we have computed the sha256 hash for each file and then XORed them isntead of putting them all in a giant memory stream? That way, we could have gotten commutative results and could have saved some memory allocations.

MichaelKetting (changes requested):

Remotion/Development/Core/NewlineStringExtensions.cs
a-ctor: Ok so get rid of the extension methods and use `.ReplaceLineEndings()` (either with or without arguments) instead? I just want to make sure because it is not quite clear what you want from the second comment.

MichaelKetting: Correct.

(I'm assuming there's no reason againt this?)

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Unload/UnloadDataTest.cs
a-ctor: Ideally, I would convert this string into a raw string for best readability but some of the lines would be way too long. Comparing the string interpolation and the explicit string concat, I don't see much of a difference in readability. If you look at the code without syntax highlighting, it might be bit more difficult to spot where the interpolations are but that's it. And I'd argue that if the explicit string concat is on the next line it is harder to read so you need to keep it in the same line.

In this case it was easier to fix everything to an interpolation because I could do this with a macro :D

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/FulltextIntegrationTests.cs
a-ctor: As far as I know, there is no docker image that has the full text search installed by default. So if we want to test it in Linux as well we need to create a custom image. I can create an issue to track this if you want to test full text search under Linux as well.

Remotion/Data/DomainObjects.UnitTests/IntegrationTests/Unload/UnloadDataTest.cs
MichaelKetting: Okay.

PS: I'll maintain that it doesn't look like how we used to write code when I was a youngen.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SchemaGeneration/SqlTableTypeScriptElementFactoryTest.cs
a-ctor: Would be nice but its kind of hard to implement as the normalization must happen on the PC the app is run on and not on the machine where it is compiled. As such, the string is embedded in the assembly normally, and the the JIT compiler would need to know that `ldstr ...;` followed by a `call string.ReplaceLineEndings` can be rewritten with an allocation to just a `ldstr ...`. Not sure that is easy and even something you want to do as you allocate string instances with a static lifetime.

Remotion/Data/DomainObjects.UnitTests/Linq/IntegrationTests/FulltextIntegrationTests.cs
MichaelKetting: Ah, really? Yes, please create a follow-up ticket in the Linux epic and add the ticket number to the ignore comment. The MS Github repo should be a good starting point for building the Linux Sql image. I want to try and fix this before we're done but if there's not enough time, it can be scoped out, I guess.

BTW: I just found this here: https://github.com/microsoft/mssql-docker/tree/master
No idea if we knew it existed when we started using SQL in Docker.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SchemaGeneration/SqlTableTypeScriptElementFactoryTest.cs
MichaelKetting: Hmm... they might embed both string versions in that case and switch dynamically based on the OS. But yeah and as our tests show, not that big an issue in prod code.

Remotion/Data/DomainObjects.UnitTests/ReflectionUtilityTest.cs
a-ctor: Because there are no UNC paths on Linux. A path cannot start with `\\` and instead UNC paths need to be mounted and then they look like normal paths. The logic itself is just `Path.GetDirectoryName(assembly.Location)` with a null check, so this test does not do anything on Linux.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlIndexDefinitionScriptElementFactory.cs
a-ctor: Yeah there are some edge cases if you do everything with a macro :p

How about using interpolation everywhere instead?

```cs
$"{Environment.NewLine}  INCLUDE ({string.Join(", ", indexDefinition.IncludedColumns.Select(c => $"[{c.Name}]"))})"
```

I am not a fan of manual string concat as it makes it harder to read what whitespaces are actually part of the string and what are spacing because of the concat.

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocDateTimeValueImplementation/DateTimeFormatterTest.cs
a-ctor: It's using a non-breaking space instead of a space ^^

I am going to add a comment here to clear that up.

Remotion/Security/UnitTests/Metadata/LocalizationFileNameStrategyTest.cs
a-ctor: What do you mean with "trimming the path"? `Path.Combine` just does `$"{left}{Path.DirectorySeparatorChar}{right}"` but optimized.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlIndexDefinitionScriptElementFactory.cs
MichaelKetting: 's okay

Remotion/ObjectBinding/Web.UnitTests/UI/Controls/BocDateTimeValueImplementation/DateTimeFormatterTest.cs
MichaelKetting: oh, yeah... d'oh! I somehow wasn't in the state of mind to read regex and also forgot that 0xA0 is nbsp. Time for vacation...

Remotion/Security/UnitTests/Metadata/LocalizationFileNameStrategyTest.cs
MichaelKetting: `.\file.xml` and `file.xml` are equivalent. The question is, whe you do Path.Combine with a `.`, does dotnet actually prefix with the dot-folder or does it only substitute the file name. Since we're testing that the `,\` is in there, using manual concatenation with the PathSepChar might be more explicit / less prone to question if the test tests the correct thing?

MichaelKetting (approved):

PR #1325, opened 2025-04-28
RM-9507 Use the netcore variant of the roslyn compiler for our RoslynCompiler utility

a-ctor:


Remotion/Development/UnitTesting.Compilation/Roslyn/RoslynCompiler.cs
```diff
@@ -378,6 +377,70 @@ private void Compile(CompilerParameters options, string compilerFullPath, string
             }
         }
 
+        // Adapted version of Executor.ExecWaitWithCapture that works in Linux.
+        private static int ExecuteCommand(
```
MichaelKetting, line 381: I did not consider the details of this method during the review. I'm assuming you tested it and you're happy with the behavior excpet for the wait/exit logic.

Remotion/Development/UnitTesting.Compilation/Roslyn/RoslynCompiler.cs
```diff
@@ -378,6 +377,70 @@ private void Compile(CompilerParameters options, string compilerFullPath, string
...
+                          errorWriter.WriteLine(e.Data);
+                    };
+
+                    p.BeginOutputReadLine();
+                    p.BeginErrorReadLine();
+
+                    if (!p.WaitForExit(600_000))
+                      throw new ExternalException($"External command failed with exit code {p.ExitCode}: '{fileName} {args}'", 0x102);
+
+                    p.WaitForExit();
```
MichaelKetting, line 435: why the second WaitForExit? Am I missing something in terms of logic? we wait for 600 seconds for the process to exit and then we wait again if there was no exception?

Also, why wait for 600 seconds?

MichaelKetting (approved):
I got one question/possible change. If that's not a correct interpretation, we're done. otherwise, it's up to you if I need to re-see the code.

Remotion/Development/UnitTesting.Compilation/Roslyn/RoslynCompiler.cs
a-ctor: Running `Executor.ExecuteCommand` on Linux fails when the process is invoked. I used the source (https://source.dot.net/#System.CodeDom/System/CodeDom/Compiler/Executor.cs,30b6236af6f49d06,references) as template, remove all the unnecessary stuff and separated the file name and args when starting the process.

Remotion/Development/UnitTesting.Compilation/Roslyn/RoslynCompiler.cs
a-ctor: Answer for both questions: Because that's what the original code did. I needed to extract it to change the process creation but otherwise kept the logic as is. I don't know if the second `.WaitForExit()` does anything at all or if it was an mistake. But I do remember there being weird behavior with `.WaitForExit()` and the actual process exit so I would keep it as is.

PR #1326, opened 2025-05-05
Bugfix/rm 9504 tvp verursachen schlechte query plans

martin-herwich:
https://jira.rubicon.eu/browse/RM-9504

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilderTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilderTest.cs
@@ -15,7 +15,10 @@
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
 using System;
+using System.Collections.Generic;
 using System.Data;
+using System.Linq;
+using JetBrains.Annotations;
 using Moq;
 using NUnit.Framework;
 using Remotion.Data.DomainObjects.Persistence.Rdbms;
@@ -28,36 +31,118 @@ namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.DbCommandBuild
   [TestFixture]
   public class QueryDbCommandBuilderTest
   {
-    private Mock<IDataParameterDefinition> _valueParameterDefinition;
-    private Mock<IDataParameterDefinition> _textParameterDefinition;
+    private class TestQueryDbCommandBuilder : QueryDbCommandBuilder
+    {
+      [NotNull]
+      private readonly IReadOnlyDictionary<string, string> _parameterTextRepresentations;
+
+      public TestQueryDbCommandBuilder ([NotNull] string statement, [NotNull] IReadOnlyCollection<QueryParameterWithDataParameterDefinition> parameters, [NotNull] ISqlDialect sqlDialect,
+          [NotNull] IReadOnlyDictionary<string, string> parameterTextRepresentations)
+          : base(statement, parameters, sqlDialect)
+      {
+        _parameterTextRepresentations = parameterTextRepresentations;
+      }
+
+      protected override string GetParameterTextRepresentation (QueryParameterWithDataParameterDefinition parameterWithDefinition)
+      {
+        if (_parameterTextRepresentations.TryGetValue(parameterWithDefinition.QueryParameter.Name, out var representation))
+          return representation;
+
+        return base.GetParameterTextRepresentation(parameterWithDefinition);
+      }
+    }
+
+    private Mock<IDataParameterDefinition> _valueParameterDefinitionMock;
+    private Mock<IDataParameterDefinition> _textParameterDefinitionMock;
     private QueryParameter _valueQueryParameter;
     private QueryParameter _textQueryParameter;
     private QueryParameterWithDataParameterDefinition _valueQueryParameterWithDataDefinition;
     private QueryParameterWithDataParameterDefinition _textQueryParameterWithDataDefinition;
-    private QueryDbCommandBuilder _commandBuilder;
 
     [SetUp]
     public void SetUp ()
     {
-      _valueParameterDefinition = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
-      _textParameterDefinition = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
+      _valueParameterDefinitionMock = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
+      _textParameterDefinitionMock = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
 
-      _valueQueryParameter = new QueryParameter("@param1", 5,QueryParameterType.Value);
-      _textQueryParameter = new QueryParameter("@param2", "test", QueryParameterType.Text);
+      _valueQueryParameter = new QueryParameter("#param1", 5, QueryParameterType.Value);
+      _textQueryParameter = new QueryParameter("#param2", "test", QueryParameterType.Text);
 
-      _valueQueryParameterWithDataDefinition = new QueryParameterWithDataParameterDefinition(_valueQueryParameter, _valueParameterDefinition.Object);
-      _textQueryParameterWithDataDefinition = new QueryParameterWithDataParameterDefinition(_textQueryParameter, _textParameterDefinition.Object);
-
-      _commandBuilder = new QueryDbCommandBuilder(
-          "Statement @param1 @param2", new[] { _valueQueryParameterWithDataDefinition, _textQueryParameterWithDataDefinition }, new Mock<ISqlDialect>().Object);
+      _valueQueryParameterWithDataDefinition = new QueryParameterWithDataParameterDefinition(_valueQueryParameter, _valueParameterDefinitionMock.Object);
+      _textQueryParameterWithDataDefinition = new QueryParameterWithDataParameterDefinition(_textQueryParameter, _textParameterDefinitionMock.Object);
     }
 
     [Test]
-    public void Create ()
+    public void Create_UsesDbCommandFactoryToCreateQueryCommand ()
     {
       var dataParameterCollectionStrictMock = new Mock<IDataParameterCollection>(MockBehavior.Strict);
 
-      _valueParameterDefinition.Setup(mock => mock.GetParameterValue(5)).Returns(5);
+      var dbCommandStub = new Mock<IDbCommand>();
+      dbCommandStub.Setup(stub => stub.Parameters).Returns(dataParameterCollectionStrictMock.Object);
+      dbCommandStub.SetupProperty(stub => stub.CommandText);
+
+      var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+      dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
+      var commandBuilder = new QueryDbCommandBuilder("Statement @param1 @param2", [], GetSqlDialectStub());
+      var result = commandBuilder.Create(dbCommandFactoryStub.Object);
+
+      dataParameterCollectionStrictMock.Verify();
+      _valueParameterDefinitionMock.Verify();
+      _textParameterDefinitionMock.Verify();
+      Assert.That(result, Is.SameAs(dbCommandStub.Object));
+    }
+
+    [Test]
+    public void Create_TransfersCommandTextToQueryCommand ()
+    {
+      var dataParameterCollectionStrictMock = new Mock<IDataParameterCollection>(MockBehavior.Strict);
+
+      _valueParameterDefinitionMock.Setup(mock => mock.GetParameterValue(5)).Returns(5);
+
+      var dbCommandStub = new Mock<IDbCommand>();
+      dbCommandStub.Setup(stub => stub.Parameters).Returns(dataParameterCollectionStrictMock.Object);
+      dbCommandStub.SetupProperty(stub => stub.CommandText);
+
+      var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+      dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
+      var commandBuilder = new QueryDbCommandBuilder("Statement @param1 @param2", [], GetSqlDialectStub());
+      var result = commandBuilder.Create(dbCommandFactoryStub.Object);
+
+      dataParameterCollectionStrictMock.Verify();
+      _valueParameterDefinitionMock.Verify();
+      _textParameterDefinitionMock.Verify();
+      Assert.That(result.CommandText, Is.EqualTo("Statement @param1 @param2"));
+    }
+
+    [Test]
+    public void Create_ReplacesTextParameter ()
+    {
+      var dataParameterCollectionStrictMock = new Mock<IDataParameterCollection>(MockBehavior.Strict);
+
+      _valueParameterDefinitionMock.Setup(mock => mock.GetParameterValue(5)).Returns(5);
+
+      var dbCommandStub = new Mock<IDbCommand>();
+      dbCommandStub.Setup(stub => stub.Parameters).Returns(dataParameterCollectionStrictMock.Object);
+      dbCommandStub.SetupProperty(stub => stub.CommandText);
+
+      var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+      dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
+      var commandBuilder = new QueryDbCommandBuilder("Statement #param1 #param2", [ _textQueryParameterWithDataDefinition ], GetSqlDialectStub());
+      var result = commandBuilder.Create(dbCommandFactoryStub.Object);
+
+      dataParameterCollectionStrictMock.Verify();
+      _valueParameterDefinitionMock.Verify();
+      _textParameterDefinitionMock.Verify();
+      Assert.That(result.CommandText, Is.EqualTo("Statement #param1 test"));
+    }
+
+    [Test]
+    public void Create_AddsValueParameter ()
+    {
+      var dataParameterCollectionStrictMock = new Mock<IDataParameterCollection>(MockBehavior.Strict);
 
       var dbCommandStub = new Mock<IDbCommand>();
       dbCommandStub.Setup(stub => stub.Parameters).Returns(dataParameterCollectionStrictMock.Object);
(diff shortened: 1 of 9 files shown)


martin-herwich:
@MichaelKetting Der Stand vor dem letzten fixup-Commit ist der mit dem failing Test wegen Text-Parameter.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,67 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
-        if (parameterWithDefinition.QueryParameter.ParameterType == QueryParameterType.Text)
-        {
-          Assertion.DebugAssert(
-              queryParameterValue is string,
-              "parameterWithType.QueryParameter.Value is string when parameterWithType.QueryParameter.ParameterType == Text");
-          var queryParameterValueAsString = (string)queryParameterValue;
+      var statementBuilder = new StringBuilder(Statement.Length);
+      var statementIndex = 0;
 
-          statement = statement.Replace(parameterWithDefinition.QueryParameter.Name, queryParameterValueAsString);
```
MichaelKetting, line 62: okay, existing solution was a bit...iffy beceause it didn't account for parameter-name overlap. I would say that this isn't a feature but a bug and nobdy ever noticed it or there would be a unit test to confirm this behavior.

MichaelKetting: To make matters worse, for us and our design, there's no naming rule for how the parameter name works with text parameters.

MichaelKetting: Lastly, based on available usage information, we do have one user of the Text parameter, there it is used correctly with `@` prefixed.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,67 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
   /// </summary>
   public class QueryDbCommandBuilder : DbCommandBuilder
```
MichaelKetting, line 32: Design:

notation for parameter placeholder in SQL Queries when using ADO.NET:
SQL Server: `@`
PostgreSQL: `$` or `@`
Oracle: `:`
MySql: `@`

So, basically, all systems we might consider relevant RDBMS use a prefix noation. This means we can add a test-method to ISqlDialect to check if a parameter name is valid.

Next, we can inject the sqldialect into the QueryDbCommandBuilder and make sure all parameters use a valid syntax.

Finally, this enables us to do a simple regex that checks for the parameters. We probably want to get the prefix-char from the sqldialect and then cache the regex in a static field for best performance.
for `@`: `\@\b[\w\d_]+\b`

This still finds usages of `@stuff` inside text, but only if the match also fits a know parameter would we actually have a false positive. this cannot be prevented and is an acceptable risk.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,72 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
     {
       ArgumentUtility.CheckNotNull("dbCommandFactory", dbCommandFactory);
 
       var command = dbCommandFactory.CreateDbCommand();
 
-      var statement = _statement;
-      foreach (var parameterWithDefinition in _parametersWithDefinition)
-      {
-        var queryParameterValue = parameterWithDefinition.QueryParameter.Value;
+      var statementBuilder = new StringBuilder(Statement.Length);
```
MichaelKetting: we should allocate a string that's Length + Min(3*parameter-count, 100). Or just, Length+100... 

That way, we don't need to re-allocate if we do have to replace, but only waste 100 bytes of memory if we don't need a stringbuilder.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,72 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
-        if (parameterWithDefinition.QueryParameter.ParameterType == QueryParameterType.Text)
-        {
-          Assertion.DebugAssert(
-              queryParameterValue is string,
-              "parameterWithType.QueryParameter.Value is string when parameterWithType.QueryParameter.ParameterType == Text");
-          var queryParameterValueAsString = (string)queryParameterValue;
+      var regexKey = SqlDialect.GetParameterName("regex");
+      var paramRegex = s_regexDictionary.GetOrAdd(
+          regexKey,
+          key => new Regex(Regex.Escape(key).Replace("regex", "\\b[\\w\\d_]+\\b"), RegexOptions.IgnoreCase | RegexOptions.Multiline));
```
MichaelKetting: `Options.Compile`, please. and we don't need multiline because a parameter can't linewrap anyway.

MichaelKetting: `Options.NoBacktracking` should also be helpful for performance
https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7/#backtracking-and-regexoptions-nonbacktracking

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,72 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
 
-          statement = statement.Replace(parameterWithDefinition.QueryParameter.Name, queryParameterValueAsString);
-        }
+      foreach (Match match in paramRegex.Matches(Statement))
+      {
+        statementBuilder.Append(Statement.Substring(statementIndex, match.Index - statementIndex));
+        statementIndex = match.Index + match.Length;
+        var param = ParametersWithDefinition.SingleOrDefault(p => p.QueryParameter.Name.Equals(match.Value, StringComparison.OrdinalIgnoreCase));
```
MichaelKetting: In case of there being a 1000 parameters, this might be disadvantagous. We should dump the parmaters into a dictioanry. perhaps directly in the ctor.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,72 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
+
+    protected virtual string GetParameterTextRepresentation (QueryParameterWithDataParameterDefinition parameterWithDefinition)
+    {
+      var queryParameterValue = parameterWithDefinition.QueryParameter.Value;
+      if (parameterWithDefinition.QueryParameter.ParameterType == QueryParameterType.Text)
+      {
+        Assertion.DebugAssert(
+            queryParameterValue is string,
+            "parameterWithType.QueryParameter.Value is string when parameterWithType.QueryParameter.ParameterType == Text");
+        return (string)queryParameterValue;
```
MichaelKetting: I haven't thought this through, but am I seeing this correct that originally (and in the new code, of course), we only check that the value is a string for Debug builds? Which means consumers won't have this guard.

This feels like something we should change in a separate PR for v8.0 to a regular old InvalidOpEx. Thoughts?

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,72 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
+      }
+
+      return parameterWithDefinition.QueryParameter.Name;
```
MichaelKetting: it feels like we should either switch this would look better using IF/ELSE because both branches are equally important.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
+using System;
```
MichaelKetting: lic header

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Text;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Parameters;
+using Remotion.Data.DomainObjects.Queries;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.DbCommandBuilders;
+
+public class SqlDbQueryCommandBuilder : QueryDbCommandBuilder
```
MichaelKetting: xml doc

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,72 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
     public QueryDbCommandBuilder (string statement, IEnumerable<QueryParameterWithDataParameterDefinition> parameters, ISqlDialect sqlDialect)
         : base(sqlDialect)
     {
       ArgumentUtility.CheckNotNull("statement", statement);
       ArgumentUtility.CheckNotNull("parameters", parameters);
 
-      _statement = statement;
-      _parametersWithDefinition = parameters.ToArray();
+      Statement = statement;
+      ParametersWithDefinition = parameters.ToArray();
```
MichaelKetting: I would suggest calling it "paramerters". also, question is if a protected value is needed for statement and Parameters. 

given the later dictionary remark (whcih means IReadonlyDictionary), immuatiblity is already covered. We should have this immutable.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    var storagePropertyDefinition = recordDefinition.PropertyDefinitions.Single().StoragePropertyDefinition;
+    if (storagePropertyDefinition is SerializedObjectIDStoragePropertyDefinition)
+      return true;
+
+    if (storagePropertyDefinition is not SimpleStoragePropertyDefinition simpleStoragePropertyDefinition)
+      return false;
+
+    var storageTypeLength = simpleStoragePropertyDefinition.ColumnDefinition.StorageTypeInfo.StorageTypeLength;
+    if(!storageTypeLength.HasValue)
```
MichaelKetting:
```suggestion
    if (!storageTypeLength.HasValue)
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+      return true;
+
+    if (storageTypeLength.Value < 0)
+      return false;
+
+    var storageDbType = simpleStoragePropertyDefinition.ColumnDefinition.StorageTypeInfo.StorageDbType;
+    var maxLength = storageDbType == DbType.String || storageDbType == DbType.StringFixedLength
+        ? 1700 / 2
+        : 1700;
```
MichaelKetting: plsase name the magic number

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    return storageTypeLength <= maxLength;
+  }
+
+  public override IDbCommand Create (IDbCommandFactory dbCommandFactory)
+  {
```
MichaelKetting: nullcheck

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+    var command = base.Create(dbCommandFactory);
```
MichaelKetting: we should special case this. if there's no tvp, return the command.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    var commandTextBuilder = new StringBuilder(command.CommandText.Length);
+
+    foreach (var tvp in _tvpToTempTable)
+    {
+      commandTextBuilder.AppendLine($"SELECT [Value] INTO {tvp.Value.TempTableName} FROM {tvp.Key};");
```
MichaelKetting: we use SqlDialect.DelimitIdentifier also in SqlTableValuedParameterComparedColumnSpecification. I can't say if that's just the only one where it's done that way or if this is the only possible place anyway. It just feels strange not doing it, but that's just my gut feeling

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+      if (tvp.Value.HasIndex)
+        commandTextBuilder.AppendLine($"CREATE NONCLUSTERED INDEX {tvp.Value.IndexName} ON {tvp.Value.TempTableName} ([Value]);");
```
MichaelKetting: I wonder if it wouldn't be better to use the brackets inline instead of in the replacement values. that way, we see that we are using them and where instead of seeing a mixe sql code. if we hardcode the brackets, it won't matter perfwise.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilderTest.cs
```diff
@@ -28,32 +32,93 @@ namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.DbCommandBuild
...
     private QueryParameter _textQueryParameter;
     private QueryParameterWithDataParameterDefinition _valueQueryParameterWithDataDefinition;
     private QueryParameterWithDataParameterDefinition _textQueryParameterWithDataDefinition;
-    private QueryDbCommandBuilder _commandBuilder;
 
     [SetUp]
     public void SetUp ()
     {
       _valueParameterDefinition = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
       _textParameterDefinition = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
```
MichaelKetting: since 99% is new anyway: please rename to `Mock` suffixed

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilderTest.cs
```diff
@@ -66,21 +131,100 @@ public void Create ()
...
+
+      var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+      dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
       var dbDataParameterStub = new Mock<IDbDataParameter>();
-      _valueParameterDefinition
-          .Setup(mock => mock.CreateDataParameter(dbCommandStub.Object, _valueQueryParameter.Name, 5))
+
+      _valueParameterDefinition.Setup(mock => mock.GetParameterValue(5)).Returns(5);
+      _valueParameterDefinition.Setup(mock => mock.CreateDataParameter(dbCommandStub.Object, _valueQueryParameter.Name, _valueQueryParameter.Value))
```
MichaelKetting:
```suggestion
      _valueParameterDefinition.Setup(mock => mock.CreateDataParameter(dbCommandStub.Object, _valueQueryParameter.Name, 5))
```
Keeping this as `5` makes it easier to reason over.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlQueryDbCommandBuilderTests.cs
```diff
@@ -0,0 +1,157 @@
+using System;
```
MichaelKetting, line 17: lic header

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlQueryDbCommandBuilderTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        result.CommandText,
+        Is.EqualTo("""
+                   SELECT [Value] INTO [#@tvp] FROM @tvp;
+                   CREATE NONCLUSTERED INDEX [IX_@tvp] ON [#@tvp] ([Value]);
+
+                   SELECT * FROM [MyTable] WHERE MyRefID IN (SELECT [Value] FROM [#@tvp])
+
+                   DROP TABLE [#@tvp];
+
+                   """));
```
MichaelKetting, line 71: loopy stuff feels like a case where we should test the zero-items and the more-than-one-items cases. the zeri-items case might be overkill when testing, although it's also rather simple.

And we should test that we only do this for tvp parameters.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
martin-herwich: I agree that this guard is only executed in Debug builds. I think that is because it helps find bugs during development, but is of little value in a production environment, since the type-cast in line 96 would fail anyways.
The guard provides a better error message during development, but would only waste CPU cycles in production.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
martin-herwich: I am unsure what you mean with "question is if a protected value is needed". Do you mean that we should make these immutable properties public?

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
MichaelKetting: Aaah, true. thanks!

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
MichaelKetting: Or just keep them private. Especially the statement will become obsolete during processing. And you can keep the parameters via the ctor if you need them. Which is only the case with the TVP because we want to proactively render the temp tables.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbQueryCommandBuilder.cs
martin-herwich: Instead of hard-coding the brackets, I now consistently use `SqlDialect.DelimitIdentifier()` - what do you think?

MichaelKetting: Looks fine to me :)

MichaelKetting (approved):

MichaelKetting (approved):

PR #1327, opened 2025-05-14
Feature/rm 9504 TVPs verursachen schlechte query plans v7.0

martin-herwich:
Cherry pick from 8.0
https://jira.rubicon.eu/browse/RM-9504

Update: Added threshold switch between single-element parameters and table-valued parameter.

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
@@ -25,6 +25,7 @@ using Remotion.Data.DomainObjects.Persistence.Configuration;
 using Remotion.Data.DomainObjects.Persistence.NonPersistent;
 using Remotion.Data.DomainObjects.Persistence.Rdbms;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Sql2016;
+using Remotion.Data.DomainObjects.UnitTests.Linq.IntegrationTests;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain.TableInheritance;
 using Remotion.Data.DomainObjects.Validation;
@@ -122,7 +123,7 @@ namespace Remotion.Data.DomainObjects.UnitTests.Factories
 
       var deferredStorageSettings = new DeferredStorageSettings(fakeStorageObjectFactoryFactory, fakeStorageSettingsFactoryResolver);
 
-      var sqlStorageObjectFactory = new SqlStorageObjectFactory(
+      var sqlStorageObjectFactory = new TestSqlStorageObjectFactory(
           deferredStorageSettings,
           SafeServiceLocator.Current.GetInstance<ITypeConversionProvider>(),
           SafeServiceLocator.Current.GetInstance<IDataContainerValidator>());
--- a/Remotion/Data/DomainObjects.UnitTests/Linq/SqlQueryGeneratorTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Linq/SqlQueryGeneratorTest.cs
@@ -50,7 +50,7 @@ namespace Remotion.Data.DomainObjects.UnitTests.Linq
       _resolutionStageMock = new Mock<IMappingResolutionStage>(MockBehavior.Strict);
       _generationStageMock = new Mock<ISqlGenerationStage>(MockBehavior.Strict);
 
-      _sqlQueryGenerator = new SqlQueryGenerator(_preparationStageMock.Object, _resolutionStageMock.Object, _generationStageMock.Object);
+      _sqlQueryGenerator = new SqlQueryGenerator(_preparationStageMock.Object, _resolutionStageMock.Object, _generationStageMock.Object, 1);
 
       _queryModel = QueryModelObjectMother.Create();
     }
@@ -176,6 +176,42 @@ namespace Remotion.Data.DomainObjects.UnitTests.Linq
           () => _sqlQueryGenerator.CreateSqlQuery(_queryModel),
           Throws.TypeOf<UnmappedItemException>().With.Message.EqualTo(
               "Query 'from Order o in null select null' contains an unmapped item. Bla."));
+    }
+
+    [Test]
+    [TestCase(1)]
+    [TestCase(42)]
+    [TestCase(int.MaxValue)]
+    public void SqlQueryGenerator_CreateSqlCommandBuilder_IsCreatedWithCorrectThreshold (int threshold)
+    {
+      ISqlCommandBuilder sqlCommandBuilder = null;
+
+      var fakePreparationResult = CreateSqlStatement();
+      _preparationStageMock
+          .Setup(mock => mock.PrepareSqlStatement(_queryModel, null))
+          .Returns(fakePreparationResult);
+
+      var fakeResolutionResult = CreateSqlStatement();
+      _resolutionStageMock
+          .Setup(mock => mock.ResolveSqlStatement(fakePreparationResult, It.IsNotNull<MappingResolutionContext>()))
+          .Returns(fakeResolutionResult);
+
+      _generationStageMock
+          .Setup(mock => mock.GenerateTextForOuterSqlStatement(It.IsNotNull<TableValuedParameterSqlCommandBuilder>(), fakeResolutionResult))
+          .Callback<ISqlCommandBuilder, SqlStatement>(
+              (builder, statement) =>
+              {
+                builder.Append(statement.ToString());
+                builder.SetInMemoryProjectionBody(Expression.Constant(null));
+                sqlCommandBuilder = builder;
+              });
+
+      var sqlQueryGenerator = new SqlQueryGenerator(_preparationStageMock.Object, _resolutionStageMock.Object, _generationStageMock.Object, threshold);
+      sqlQueryGenerator.CreateSqlQuery(_queryModel);
+
+      Assert.That(sqlCommandBuilder, Is.Not.Null);
+      Assert.That(sqlCommandBuilder, Is.InstanceOf<TableValuedParameterSqlCommandBuilder>());
+      Assert.That(((TableValuedParameterSqlCommandBuilder)sqlCommandBuilder).TableValuedParameterThreshold, Is.EqualTo(threshold));
     }
 
     private void CheckCreateSqlQuery_SelectedEntityType (Type expectedSelectedEntityType, Expression selectProjection)
--- a/Remotion/Data/DomainObjects.UnitTests/Linq/TableValuedParameterSqlCommandBuilderTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Linq/TableValuedParameterSqlCommandBuilderTest.cs
@@ -14,7 +14,11 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with re-motion; if not, see http://www.gnu.org/licenses.
 // 
+using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
+using Moq;
 using NUnit.Framework;
 using Remotion.Data.DomainObjects.Linq;
 using Remotion.Linq.SqlBackend.SqlStatementModel;
@@ -25,12 +29,35 @@ namespace Remotion.Data.DomainObjects.UnitTests.Linq;
 public class TableValuedParameterSqlCommandBuilderTest
 {
   [Test]
-  public void AppendCollection_UsesSingleParameter ()
+  public void AppendCollection_Empty_UsesNoParameter ()
   {
-    var collection = new[] { 17, 4, 42 };
+    var collection = Array.Empty<int>();
     var constantCollectionExpression = new ConstantCollectionExpression(collection);
 
-    var sqlCommandBuilder = new TableValuedParameterSqlCommandBuilder();
+    var sqlCommandBuilder = new TableValuedParameterSqlCommandBuilder(1);
+    sqlCommandBuilder.AppendCollection(constantCollectionExpression);
+
+    Assert.That(sqlCommandBuilder.GetCommandText(), Is.EqualTo("SELECT NULL WHERE 1 = 0"));
+    Assert.That(sqlCommandBuilder.GetCommandParameters().Length, Is.EqualTo(0));
+  }
+
+  private static IEnumerable<IEnumerable> GetCollectionsOfDifferentTypes ()
+  {
+    var data = new[] { 17, 4, 42 };
+
+    yield return new CollectionAndEnumerableOfTStub<int>(data);
+    yield return new CollectionStub<int>(data);
+    yield return new ReadOnlyCollectionStub<int>(data);
+    yield return new ReadOnlyCollectionStub<object>(data.Cast<object>().ToArray());
+  }
+
+  [Test]
+  [TestCaseSource(nameof(GetCollectionsOfDifferentTypes))]
+  public void AppendCollection_CountAtThreshold_UsesSingleParameter (IEnumerable collection)
+  {
+    var constantCollectionExpression = new ConstantCollectionExpression(collection);
+
+    var sqlCommandBuilder = new TableValuedParameterSqlCommandBuilder(3);
     sqlCommandBuilder.AppendCollection(constantCollectionExpression);
 
     Assert.That(sqlCommandBuilder.GetCommandText(), Is.EqualTo("SELECT [Value] FROM @1"));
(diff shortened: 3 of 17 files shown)


Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
```diff
@@ -116,12 +116,12 @@ public IDbCommandBuilder CreateForSelect (
...
 
-    public IDbCommandBuilder CreateForQuery (string statement, IEnumerable<QueryParameterWithDataParameterDefinition> parametersWithType)
+    public IDbCommandBuilder CreateForQuery (string statement, IReadOnlyCollection<QueryParameterWithDataParameterDefinition> parametersWithType)
     {
       ArgumentUtility.CheckNotNull("statement", statement);
       ArgumentUtility.CheckNotNull("parametersWithType", parametersWithType);
 
-      return new QueryDbCommandBuilder(statement, parametersWithType, _sqlDialect);
+      return new SqlQueryDbCommandBuilder(statement, parametersWithType, _sqlDialect);
     }
```
MichaelKetting: For consistency reasons, we should stick to IEnumerable on the interface and do a ToArray when passign the parameter into SqlQueryDbCommandBuilder. I'm not opposed to one day changing the IEnumerable to IReadOnlyCollection on the factory, but we should do this wholesale instead of bit by bit.

MichaelKetting: And I just realized: I missed this on the develop-PR. We should keep the public api on the HF branch, and we should consider adjusting the API on develop.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/QueryDbCommandBuilder.cs
```diff
@@ -28,51 +31,76 @@ namespace Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders
...
   public class QueryDbCommandBuilder : DbCommandBuilder
   {
+    private static readonly ConcurrentDictionary<string, Regex> s_regexDictionary = new();
     private readonly string _statement;
-    private readonly QueryParameterWithDataParameterDefinition[] _parametersWithDefinition;
+    private readonly IReadOnlyDictionary<string, QueryParameterWithDataParameterDefinition> _parameters;
 
-    public QueryDbCommandBuilder (string statement, IEnumerable<QueryParameterWithDataParameterDefinition> parameters, ISqlDialect sqlDialect)
+
+    public QueryDbCommandBuilder (string statement, IReadOnlyCollection<QueryParameterWithDataParameterDefinition> parameters, ISqlDialect sqlDialect)
```
MichaelKetting: I am considering to recommend we keep this as IEnumerable, for API stability. It doesn't make a difference in the implementation and is more graceful to the callers. Should possibly also revert API change on develop.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlQueryDbCommandBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+      IndexName = indexName;
+      HasIndex = hasIndex;
+    }
+
+    public string IndexName { get; }
+    public string TempTableName { get; }
+    public bool HasIndex { get; }
+  }
+
+  public SqlQueryDbCommandBuilder (string statement, IReadOnlyCollection<QueryParameterWithDataParameterDefinition> parameters, ISqlDialect sqlDialect)
```
MichaelKetting, line 57: new type can and should keep IReadOnlyCollection. This prevents double calculation of the sequence.

MichaelKetting (approved):
I'm assuming it was a pure cherry-pick. Got some details on breaking changes in remarks we should consider but otherwise, it's done.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
martin-herwich: I would vote against reverting the API changes on develop, because this change from `IEnumerable<T>` to `IReadOnlyCollection<T>` should not break any existing caller, and even if it does, it is a trivial fix.
However, we can and probably should change all uses of `IEnumerable<T>` to `IReadOnlyCollection<T>` in `IDbCommandBuilderFactory`.

MichaelKetting: DependDB does agree about there not being any breaking usages and we only have our products that have taken a known dependency on the APIs. I'm okay with changing this towards IReadOnlyCollection in v8.0 if you check with Neon Team.

martin-herwich: The only call to `CreateForQuery` uses an array.
Calls to `CreateForDelete` use `ObjectIDStorageProperty.SplitValue()`, which returns an `IEnumerable<ColumnValue>` :-(
Calls to `CreateForInsert` and `CreateForUpdate` use concatenation and would have to add `.ToArray()`

So, the existing change will not break anything. But if we change the other calls, we would need to change `IRdbmsStoragePropertyDefinition` as well (for Delete) and accept trivial breaking changes in two places. The easiest course of action would be to leaven our 8.0 branch as it is, revert the API changes in 7.0, and postpone dealing with the other `IEnumerables`.

MichaelKetting: As discussed, let's keep the old API also on develop until we can do a consistent change.

martin-herwich:
Added threshold switch between single-element parameters and table-valued parameter.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -572,8 +572,14 @@ protected virtual ISqlQueryGenerator CreateSqlQueryGenerator (
...
+      var tableValuedParameterThreshold = GetTableValuedParameterThreshold();
 
-      return new SqlQueryGenerator(sqlPreparationStage, mappingResolutionStage, sqlGenerationStage);
+      return new SqlQueryGenerator(sqlPreparationStage, mappingResolutionStage, sqlGenerationStage, tableValuedParameterThreshold);
+    }
+
+    protected virtual int GetTableValuedParameterThreshold ()
+    {
+      return 1;
     }
```
MichaelKetting: Rick mentioned yesterday that 25 items is probably the safe break off for switching from individual values to TVP (it's not final, but very likely). Should we do this in re-motion or downstream? I'm open to fixing this in re-motion.

MichaelKetting: Please add an xml doc here because it's rather important to explain what the value does. Including the special case if needed as well as the set default value.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
```diff
@@ -25,11 +29,44 @@ namespace Remotion.Data.DomainObjects.Linq;
...
 {
+  private readonly int _tableValuedParameterThreshold;
+
+  public TableValuedParameterSqlCommandBuilder (int tableValuedParameterThreshold)
+  {
+    if (tableValuedParameterThreshold < 0)
+    {
+      throw new ArgumentOutOfRangeException(
+          "Threshold must be greater than 0 in order to use table-valued parameters for collections with at least that number of elements, "
+          + "or 0 to disable table-valued parameters.");
```
MichaelKetting: please add argument name

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
```diff
@@ -25,11 +29,44 @@ namespace Remotion.Data.DomainObjects.Linq;
...
+    if (_tableValuedParameterThreshold == 0)
+    {
+      base.AppendNonEmptyCollection(collectionExpression);
+    }
+    else
+    {
+      var count = (collectionExpression.Collection as ICollection)?.Count
+          ?? (collectionExpression.Collection as ICollection<object>)?.Count
+          ?? (collectionExpression.Collection as IReadOnlyCollection<object>)?.Count
+          ?? (collectionExpression.Collection as IEnumerable)?.Cast<object>().Count();
```
MichaelKetting: This API works with ConstantCollectionExpression.

Does it even allow IEnumerable, or only the wellknown versions of ICollection? I suggset, if it's IEnumerable, we either default to parameters or to TVP instead of iterating the IEnumerable, which is always a bad (TM) thing to do. Or we know that it's invalid and throw. This depends on the contract of ConstantCollectionExpression.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
```diff
@@ -25,11 +29,44 @@ namespace Remotion.Data.DomainObjects.Linq;
...
   protected override void AppendNonEmptyCollection (ConstantCollectionExpression collectionExpression)
   {
     ArgumentUtility.CheckNotNull(nameof(collectionExpression), collectionExpression);
 
-    Append("SELECT [Value] FROM ");
-    AppendParameter(collectionExpression.Collection);
+    if (_tableValuedParameterThreshold == 0)
+    {
+      base.AppendNonEmptyCollection(collectionExpression);
+    }
```
MichaelKetting: Not sure if we need the special case.
Also, the "disable" value to be 0 instead of -1 feels odd to me as far as .net convetions are concerned. If we don't need the special case, we can drop this.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
```diff
@@ -25,11 +29,44 @@ namespace Remotion.Data.DomainObjects.Linq;
 /// </summary>
 public class TableValuedParameterSqlCommandBuilder : SqlCommandBuilder
 {
+  private readonly int _tableValuedParameterThreshold;
+
+  public TableValuedParameterSqlCommandBuilder (int tableValuedParameterThreshold)
```
MichaelKetting, line 38: please add xml doc explaining the paramter and the special cases

Remotion/Data/DomainObjects.UnitTests/Linq/TableValuedParameterSqlCommandBuilderTest.cs
```diff
@@ -40,4 +56,63 @@ public void AppendCollection_UsesSingleParameter ()
...
+    Assert.That(parameter[0].Name, Is.EqualTo("@1"));
+    Assert.That(parameter[0].Value, Is.EqualTo(collection[0]));
+    Assert.That(parameter[1].Name, Is.EqualTo("@2"));
+    Assert.That(parameter[1].Value, Is.EqualTo(collection[1]));
+    Assert.That(parameter[2].Name, Is.EqualTo("@3"));
+    Assert.That(parameter[2].Value, Is.EqualTo(collection[2]));
+  }
+
+  [Test]
+  public void AppendCollection_ZeroThreshold_UsesMultipleParameters ()
```
MichaelKetting: please also add a test with empty collection and threshhold=0

Remotion/Data/DomainObjects.UnitTests/Linq/TableValuedParameterSqlCommandBuilderTest.cs
```diff
@@ -25,12 +26,27 @@ namespace Remotion.Data.DomainObjects.UnitTests.Linq;
 public class TableValuedParameterSqlCommandBuilderTest
```
MichaelKetting, line 29: please add test for argument check in ctor

Remotion/Data/DomainObjects.UnitTests/Linq/SqlQueryGeneratorTest.cs
```diff
@@ -50,7 +50,7 @@ public void SetUp ()
       _resolutionStageMock = new Mock<IMappingResolutionStage>(MockBehavior.Strict);
       _generationStageMock = new Mock<ISqlGenerationStage>(MockBehavior.Strict);
 
-      _sqlQueryGenerator = new SqlQueryGenerator(_preparationStageMock.Object, _resolutionStageMock.Object, _generationStageMock.Object);
+      _sqlQueryGenerator = new SqlQueryGenerator(_preparationStageMock.Object, _resolutionStageMock.Object, _generationStageMock.Object, 0);
```
MichaelKetting: We should have a test that verifies the value is correctly injected. We can do this test later (ticket please) because it's a bit tricky to do and I don't have a good idea on how to do this quick and easy right now.

MichaelKetting (changes requested):
I checked the newly changed files, assuming the rest is identical to develop

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
martin-herwich: We can use 25 here because it is a reasonable default. Downstream, we want to get the value from configuration.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
martin-herwich: `ConstantCollectionExpression` says that it represents "an ICollection, an ICollection<T> or an IReadOnlyCollection<out T>."
But the constructor parameter is an `IEnumerable`, because that is the common interface, and we need to detect whether the collection is empty.
Unfortunately, I just noticed that the soft cast to `ICollection<object>` will not work because unlike `IReadOnlyCollection<out T>`, the type parameter in `ICollection<T>` is not covariant.
So we have to fall back to `IEnumerable` anyway, unless we want to add a `Count` property to `ConstantCollectionExpression`.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
martin-herwich: I had the same thought, but on the other hand, what does a threshold of 0 mean?
We can start the allowed range at 1 instead of 0, and if someone wants to always use single-element parameters, they can use `int.MaxValue` as threshold.

Remotion/Data/DomainObjects.UnitTests/Linq/TableValuedParameterSqlCommandBuilderTest.cs
martin-herwich: This is a test case in AppendCollection_Empty_UsesNoParameter.

Remotion/Data/DomainObjects.UnitTests/Linq/SqlQueryGeneratorTest.cs
martin-herwich: We could extract a method `CreateSqlCommandBuilder()` and get that via `PrivateInvoke`, right?

martin-herwich: Solution: we mock the `ISqlBuilderGenerationStage`, capture the `SqlCommandBuilder` and assert the new `TableValuedParameterThreshold` property.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
martin-herwich: TODO: UnitTests for the different branches.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -572,8 +572,18 @@ protected virtual ISqlQueryGenerator CreateSqlQueryGenerator (
...
       var mappingResolutionStage = ObjectFactory.Create<DefaultMappingResolutionStage>(ParamList.Create(resolver, generator));
       var sqlGenerationStage = ObjectFactory.Create<ExtendedSqlGenerationStage>(ParamList.Empty);
+      var tableValuedParameterThreshold = GetTableValuedParameterThreshold();
 
-      return new SqlQueryGenerator(sqlPreparationStage, mappingResolutionStage, sqlGenerationStage);
+      return new SqlQueryGenerator(sqlPreparationStage, mappingResolutionStage, sqlGenerationStage, tableValuedParameterThreshold);
+    }
+
+    /// <summary>
+    /// Gets the number of elements in a collection that triggers the use of a table-valued parameter instead of single-element parameters in
```
MichaelKetting, line 581: Please also add a note to the jira ticket for the TVP feature (even though the feature is already released) so there is a hint about this.

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
```diff
@@ -25,11 +30,49 @@ namespace Remotion.Data.DomainObjects.Linq;
...
+  /// <summary>
+  /// Creates a new <see cref="SqlCommandBuilder"/> instance that uses a table-valued parameter (instead of single-element parameters) for a collection that has at least
+  /// <paramref name="tableValuedParameterThreshold"/> elements.
+  /// </summary>
+  /// <exception cref="ArgumentOutOfRangeException">If <paramref name="tableValuedParameterThreshold"/> is 0 or negative.</exception>
+  public TableValuedParameterSqlCommandBuilder (int tableValuedParameterThreshold)
+  {
+    if (tableValuedParameterThreshold <= 0)
+    {
+      throw new ArgumentOutOfRangeException(nameof(tableValuedParameterThreshold), "Threshold must be greater than 0");
```
MichaelKetting:
```suggestion
      throw new ArgumentOutOfRangeException(nameof(tableValuedParameterThreshold), "Threshold must be greater than 0.");
```

Remotion/Data/DomainObjects/Linq/TableValuedParameterSqlCommandBuilder.cs
```diff
@@ -25,11 +30,49 @@ namespace Remotion.Data.DomainObjects.Linq;
...
+
+  public int TableValuedParameterThreshold { get; }
+
   protected override void AppendNonEmptyCollection (ConstantCollectionExpression collectionExpression)
   {
     ArgumentUtility.CheckNotNull(nameof(collectionExpression), collectionExpression);
 
-    Append("SELECT [Value] FROM ");
-    AppendParameter(collectionExpression.Collection);
+    int GetCount (object enumerable)
```
MichaelKetting:
```suggestion
    static int GetCount (object enumerable)
```

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -122,7 +123,7 @@ private FakeStorageSettingsFactory CreateStorageSettings ()
 
       var deferredStorageSettings = new DeferredStorageSettings(fakeStorageObjectFactoryFactory, fakeStorageSettingsFactoryResolver);
 
-      var sqlStorageObjectFactory = new SqlStorageObjectFactory(
+      var sqlStorageObjectFactory = new TestSqlStorageObjectFactory(
```
MichaelKetting, line 126: Yes!

MichaelKetting (approved):
Assuming this is only the new test SQL storageobjfactory four or so new changed files?  Mobil GitHub app isn't so great for review

PR #1330, opened 2025-06-11
RM-9512 Start SQL Server Docker images are part of the build

a-ctor:


Build/Build.cs
```diff
@@ -40,14 +41,18 @@
...
 {
   [Parameter(ValueProviderMember = nameof(SupportedTestBrowsers), Separator = "+")]
   public string[] TestBrowsers { get; set; } = [];
 
   [Parameter(ValueProviderMember = nameof(SupportedTestSqlServers), Separator = "+")]
   public string[] TestSqlServers { get; set; } = [];
 
+  [CanBeNull] private TestMatrix _databaseTestMatrix;
+  [CanBeNull] private TestMatrix _normalTestMatrix;
+  [CanBeNull] private TestMatrix _webTestMatrix;
```
MichaelKetting: we named this `WebTestingTest...` in other places (see below). We should keep the naming consistent. If the short version is preferred, fix that one, too.

Build/Build.cs
```diff
@@ -97,19 +102,32 @@ public override ISbomGeneratorBuilder ConfigureSbomGenerationInfoBuilder (Soluti
...
+    [CanBeNull]
+    static TestConfiguration CreateTestConfiguration (
+        [CanBeNull] TestMatrix testMatrix,
+        ITestExecutionRuntimeFactory testExecutionRuntimeFactory,
+        ImmutableArray<ITestExecutionWrapper> testExecutionWrappers)
+    {
+      return testMatrix != null
+          ? new TestConfiguration(testExecutionRuntimeFactory, testMatrix, testExecutionWrappers)
+          : null;
+    }
```
MichaelKetting, line 114: do we need ReSharper nullabilty here, i.e. we don't have NRT enabled in the build?

MichaelKetting, line 114: Normally, we try to put the local functions after the method's body. Of course, here we would need to put it after the local variables setup and before the project configration because otherwise it would be completely unreadable. I can understand why you want to keep it up top and thus at least somewhat consistent, though

Build/Build.cs
```diff
@@ -97,19 +102,32 @@ public override ISbomGeneratorBuilder ConfigureSbomGenerationInfoBuilder (Soluti
...
+        testExecutionRuntimeFactory,
         [new WebTestingTestSetup()]);
 
-    var databaseTestConfiguration = new TestConfiguration(
-        DefaultTestExecutionRuntimeFactory.Instance,
-        TestMatrices.Single(e => e.Name == "DatabaseTestMatrix"),
+    var databaseTestConfiguration = CreateTestConfiguration(
+        _databaseTestMatrix,
+        testExecutionRuntimeFactory,
         [new DatabaseTestSetup()]);
```
MichaelKetting: shouldn't guard / give a meaningful error if there is an incorrect method call order that leaves the _matrix null?

Build/Build.cs
```diff
@@ -326,6 +344,23 @@ public override void ConfigureTestMatrix (TestMatricesBuilder builder)
...
+      var requiredDatabases = testCases
+          .Select(e => e.TestMatrixRow.GetDimension<Databases>())
+          .Distinct()
+          .Where(e => e != NoDB && e != SqlServerDefault)
+          .ToArray();
+
+      foreach (var requiredDatabase in requiredDatabases)
+          testResources.Add(new DatabaseTestResourceFactory(requiredDatabase));
+  }
+
```
MichaelKetting: incorrect indent. 

And I see the next mthod is also incorrectly indented

Build/Customizations/DatabaseTestResource.cs
```diff
@@ -0,0 +1,35 @@
...
+
+  public DatabaseTestResource (
+      string name,
+      Databases database,
+      ISqlDockerImage sqlDockerImage)
+  {
+    _sqlDockerImage = sqlDockerImage;
+    Name = name;
+    Database = database;
+  }
```
MichaelKetting: style: if possible, we should keep the arguments, assignments and fields in the same order.

Build/Customizations/DatabaseTestResourceFactory.cs
```diff
@@ -0,0 +1,54 @@
...
+using Remotion.BuildScript.Test;
+
+namespace Customizations;
+
+public class DatabaseTestResourceFactory : ITestResourceFactory
+{
+  private const string c_imageNameParameter = "ImageName";
+  private const string c_hostNameParameter = "HostName";
+
+  public Databases Database { get; }
```
MichaelKetting, line 16: naming: the Type is many, the name is singular?

Reading the test, I can see where this is coming from. Seems like we're stuck with this for the time being..?

Build/Customizations/SqlServer/WindowsSqlServerDockerImage.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Customizations.SqlServer;
+
+public class WindowsSqlServerDockerImage : ISqlDockerImage
+{
+  private readonly IContainer _container;
+  private readonly string _connectionString;
+
+  public WindowsSqlServerDockerImage (string imageName, string name, [CanBeNull] INetwork network)
+  {
+    const string password = "P@ssw0rd";
```
MichaelKetting, line 23: Don't we get the password from teamcity?

Before this change, the password wasn't hardcoded in the build script, if I remember correctly. And while that shouldn't be an issue, it still feels strange thing to do.

Should we use a Guid as a password instead (new'd each time)?

Build/Customizations/SqlServer/LinuxSqlServerDockerImage.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Customizations.SqlServer;
+
+public class LinuxSqlServerDockerImage : ISqlDockerImage
+{
+  private readonly IContainer _container;
+  private readonly string _connectionString;
+
+  public LinuxSqlServerDockerImage (string imageName, string hostName, [CanBeNull] INetwork network)
+  {
+    const string password = "P@ssw0rd";
```
MichaelKetting, line 23: see Windows for password-note

Build/Customizations/SqlServer/WindowsSqlServerDockerImage.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    var properties = new Dictionary<string, string>
+                     {
+                         { "Server", name },
+                         { "Initial Catalog", "master" },
+                         { "User Id", "sa" },
+                         { "Password", password },
+                         { "TrustServerCertificate", bool.TrueString }
+                     };
+    _connectionString = string.Join(";", properties.Select(property => string.Join("=", property.Key, property.Value)));
```
MichaelKetting, line 50: .net contains a connectionstring builder
While using string.join is technically robust enough, it feels like we're re-inventing the wheel
https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnectionstringbuilder

Build/Customizations/SqlServer/LinuxSqlServerDockerImage.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    var properties = new Dictionary<string, string>
+                     {
+                         { "Server", hostName },
+                         { "Initial Catalog", "master" },
+                         { "User Id", "sa" },
+                         { "Password", password },
+                         { "TrustServerCertificate", bool.TrueString }
+                     };
+    _connectionString = string.Join(";", properties.Select(property => string.Join("=", property.Key, property.Value)));
```
MichaelKetting, line 48: see Windows comment

Build/Customizations/SqlServer/LinuxSqlServerDockerImage.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Linq;
+using DotNet.Testcontainers.Builders;
+using DotNet.Testcontainers.Containers;
+using DotNet.Testcontainers.Networks;
+using JetBrains.Annotations;
+using Serilog;
+
+namespace Customizations.SqlServer;
+
+public class LinuxSqlServerDockerImage : ISqlDockerImage
```
MichaelKetting, line 14: there is a lot of duplication between Linux and Windows. should we introduce a baseclass?

MichaelKetting (approved):
I don't think I *need* to see the file again. Please check the various suggsetions apply what sensible ones there are.

Build/Build.cs
a-ctor: Yeah we don't have it enabled. I think that is because of the DSL nature of NUKE where it doesn't really add as much value as most values are set implicitly anyways. We might want to investigate this in the future but I wouldn't change it now.

Build/Build.cs
a-ctor: So the whole thing with nullable `TestMatrix` here is that we want to support partial builds. When tests are executed, the test matrix can be calculated and the test matrices are non-null. But if we don't do a test build we have no matrices but we still might want to get project metadata so we need this method to work even if there is no test matrix.

Build/Customizations/DatabaseTestResource.cs
a-ctor: I would change the assignments in the constructor but leave the field/property order as is. I feel like it is more helpful to separate private from non-private (in this case using fields and properties as well as order as distinction) than to have the field order match the constructor argument order.

Build/Customizations/DatabaseTestResourceFactory.cs
a-ctor: `Databases` acts like an enum, which is why it is named as plural. That does cause a bit of a naming conflict but the singular here is correct as the resource factory only starts a single database.

Build/Customizations/SqlServer/WindowsSqlServerDockerImage.cs
a-ctor: This is only for the docker image that we start. So we don't connect to an external server but rather a service that we host ourselves and that is only accessible on `localhost`. The password concern is therefore kind of irrelevant as what would the attack vector be?

We can use a guid here if you insist but it helps with debugging if the password is fixed as you can more easily connect to the database without needing to extract the password from the running process.

Build/Customizations/SqlServer/WindowsSqlServerDockerImage.cs
a-ctor: Yes but that would require us to add a new package dependency. I don't think it is worth introducing the dependency for the basic concatenation here. Apart from the `Server` variable everything is constant so we could also use a template string for the connection string. The dictionary-join approach here is just for readability.

Build/Customizations/SqlServer/LinuxSqlServerDockerImage.cs
a-ctor: Given that all the startup logic is in the constructor introducing a base class is a bit awkward as it does not help as much against the duplication. We have about 30 lines that are duplicated here with subtle differences. I think it is easier to read if we keep the duplication.

Build/Build.cs
MichaelKetting: okay, misunderstanding: I was assuming that CreateTestConfig guards this null-safely :D
Maybe a comment would be helpful here because that's an implicit detail that might get lost down the road.

Build/Customizations/DatabaseTestResource.cs
MichaelKetting: ah yes. Just in this particular case: first the public fields, then the private fields, since it's always public, protected, private unless there's good reason to diverge. and then the order is correct again, thanks to a happy coincidence :D

Build/Customizations/DatabaseTestResourceFactory.cs
MichaelKetting: /squints/ just to be annoying: unless it's a flags-enum, enums are singular, too. But okay, I'll bite...

Build/Customizations/SqlServer/WindowsSqlServerDockerImage.cs
MichaelKetting: okay. Add a short comment please why the hardcoded password isn't an issue.

Build/Customizations/SqlServer/WindowsSqlServerDockerImage.cs
MichaelKetting: hmm...

/heavy sign/

hmm...

/removes dependency from chain/

hmm...

/closes conversation/

Build/Customizations/SqlServer/LinuxSqlServerDockerImage.cs
MichaelKetting: okay.

PR #1332, opened 2025-06-13
RM-9515 Add process arguments as configuration for the test site hosting

a-ctor:


Remotion/Web/Development.WebTesting/HostingStrategies/AspNetDockerContainerWrapper.cs
```diff
@@ -25,16 +26,17 @@ public AspNetDockerContainerWrapper (IDockerClient docker, AspNetDockerContainer
     }
 
     _processPath = configurationParameters.ProcessPath;
+    _processArguments = configurationParameters.ProcessArguments;
   }
 
   protected override string GetEntryPoint ()
   {
-    return Path.GetFullPath(Path.Combine(ConfigurationParameters.AbsoluteWebApplicationPath, _processPath));
+    return _processPath;
```
MichaelKetting, line 34: This is because of  TestSiteLayoutConfiguration, correct?

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingProcessWrapper.cs
```diff
@@ -58,7 +65,9 @@ private Process CreateServerProcess ()
     if (OperatingSystem.IsWindows())
       startInfo.LoadUserProfile = true;
 
-    startInfo.FileName = _exePath;
+    startInfo.FileName = _processPath;
+    if (_processArguments != null)
+      startInfo.Arguments = _processArguments;
```
MichaelKetting: The default for StartInfo.Arguments is an empty string. shouldn't we just use _processArguments??"" instead of the IF? Or make the process arguments not-nullable for consistency, same as the processpath. Note that further up the callchain, processpath is nullable so we don't just change this in the config-part of the code, but only later on perhaps?

Remotion/Web/Development.WebTesting/Configuration/Legacy/TestSiteLayoutConfigurationElement.cs
```diff
@@ -29,12 +29,14 @@ public class TestSiteLayoutConfigurationElement : ConfigurationElement, IWebTest
...
     private readonly ConfigurationProperty _resourcesProperty;
     private readonly ConfigurationProperty _processPathProperty;
+    private readonly ConfigurationProperty _processArgumentsProperty;
 
     public TestSiteLayoutConfigurationElement ()
     {
       _rootPathProperty = new ConfigurationProperty("rootPath", typeof(string), "", ConfigurationPropertyOptions.IsRequired);
       _resourcesProperty = new ConfigurationProperty("resources", typeof(TestSiteResourceConfigurationElementCollection));
       _processPathProperty = new ConfigurationProperty("processPath", typeof(string), null);
+      _processArgumentsProperty = new ConfigurationProperty("processArguments", typeof(string), null);
```
MichaelKetting, line 39: When do we make processPath not-nullable in the callchain? Since later code requires a not-null value, there's got to be code for this somewhere...

MichaelKetting (approved):
Got a few questions /suggestions. I'll re-review if there's relevant changes from them.

Remotion/Web/Development.WebTesting/Configuration/Legacy/TestSiteLayoutConfigurationElement.cs
a-ctor: This happens in the hosting strategy as some strategies (e.g. IIS) don't need a process path. Others, like the `AspNetCoreHostingStrategy` do need a process path and will throw an exception if it is not there.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingProcessWrapper.cs
a-ctor: Does the same thing if the default is `""`. I wanted to keep the process arguments nullable to allow the differentiation between "no arguments" and "empty arguments". This way, you could have overridable default arguments (e.g. IIS) if you wanted.

If you want we can use `?? string.Empty` instead but I would keep the nullability as is.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetDockerContainerWrapper.cs
a-ctor: This was always a bit sketchy because `_processPath` is already rooted. As such, `Path.Combine(left, right)` gets to absolute paths and thus returns the `right` one ;)

To allow invocations like `dotnet` a relative path should not be made absolute here.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetCoreHostingProcessWrapper.cs
MichaelKetting: I think I might like the feel of the `??` because it's more "simple" in this case. The `if` feels like it's special casing instead of just providing a null-to-empty fallback, i.e. `??` expresses the intent better in this case.

Remotion/Web/Development.WebTesting/HostingStrategies/AspNetDockerContainerWrapper.cs
MichaelKetting: should we add a unit test with a well-named test value (e.g. "dotnet") or at least a comment in the code? It feels like this is something someone might easily "fix" accidentally.

PR #1333, opened 2025-06-13
RM-9516 Add null hosting strategy as configurable type in the configuration

a-ctor:


Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/HostingConfiguration.cs
```diff
@@ -63,7 +63,7 @@ public HostingConfiguration ([NotNull] IWebTestSettings webTestSettings, [NotNul
 
     public IHostingStrategy GetHostingStrategy ()
     {
-      if (string.IsNullOrEmpty(_hostingSettings.Type))
+      if (string.IsNullOrEmpty(_hostingSettings.Type) || _hostingSettings.Type is "None" or "Null")
```
MichaelKetting: 1. Do we want to support both strings? usually, I only use `None`.
2. Please add this also to the XML schema for the xml config.
3. Do we also need to do something for the json config?

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/HostingStrategies/Configuration/HostingConfiguration.cs
a-ctor: 1. I added both to make it easier if a user wants to change the config file but I am fine with just `None`
2. Now I now there is an XML schema
3. Nope this class does the logic for both. So `_hostingSettings` is already either the XML or JSON config object

MichaelKetting: ad 1) nah, the xml schema tells the user what goes (if they have it loaded) and two options meaning the same thing isn't easier, actually. You only assume it's the same thing and might spend time figuring out if it's identical. 

ad 2) 👍 😆

MichaelKetting (approved):

PR #1335, opened 2025-06-23
RM-9520 Allow customization of the testsite startup check

a-ctor:


Remotion/Web/Development.WebTesting/Configuration/TestInfrastructureConfiguration.cs
```diff
@@ -45,6 +47,10 @@ public TestInfrastructureConfiguration ([NotNull] IWebTestSettings webTestSettin
       ArgumentUtility.CheckNotNull("webTestSettings", webTestSettings);
 
       _webApplicationRoot = webTestSettings.WebApplicationRoot;
+      _testSiteStartupCheckEnabled = webTestSettings.TestSiteStartupCheckEnabled;
+      _testSiteStartupCheckUrl = string.IsNullOrEmpty(webTestSettings.TestSiteStartupCheckUrl)
+          ? webTestSettings.WebApplicationRoot
+          : webTestSettings.TestSiteStartupCheckUrl;
```
MichaelKetting, line 53: Is the default something that should be documented?

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -104,10 +106,12 @@ public string LogDirectory
...
+      if (!_testSiteStartupCheckEnabled)
+        return;
 
       try
       {
-        VerifyWebApplicationStarted(_webApplicationRoot, _verifyWebApplicationStartedTimeout);
+        VerifyWebApplicationStarted(_testSiteStartupCheckUrl, _verifyWebApplicationStartedTimeout);
       }
       catch
       {
```
MichaelKetting: Hmm...I'm not convinced the boolean should be an exit conditon instead of an IF block. Reasoning: OnSetUp in the FixtureHelper is more than just start and optionally verify. Or rather, it's coincidental that right now, that's the only stuff we do. So, if we ever need to do more, we need to recogized the IF as actually being a condtional block of code, determine how far it goes and then decide how to refactor / insert new code. If, instead, we make this into a block right now, it's immediately clear.

MichaelKetting (approved):
Please consider if the review comments are helpful / should be integrated but otherwise, I'm done with the review.

PR #1345, opened 2025-07-15
RM-9519 add support for remote driver

a-ctor:


Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
```diff
@@ -37,6 +37,15 @@
...
         <xs:element name="edge" type="c:ChromiumConfigurationElement" minOccurs="0" maxOccurs="1" />
+        <xs:element name="remoteDriver" minOccurs="0" maxOccurs="1">
+          <xs:complexType>
+            <xs:attribute name="enabled" type="xs:boolean" default="false" />
+            <xs:attribute name="url" type="xs:string" default="" />
+            <xs:attribute name="hostRemoteDriverInDocker" type="xs:boolean" default="false" />
+            <xs:attribute name="dockerImageName" type="xs:string" default="" />
+            <xs:attribute name="dockerCustomArguments" type="xs:string" default="" />
+          </xs:complexType>
+        </xs:element>
```
MichaelKetting: Is it correct that all attributes are optional when the element is added, including the URL? And in which case would we add the element but make the url blank?

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
```diff
@@ -133,13 +141,50 @@ public void OnSetUp ()
...
     {
       _hostingStrategy.DeployAndStartWebApplication();
     }
 
+    private void HostDriverBrowserInDocker (IWebTestSettings webTestSettings)
+    {
+      var remoteDriverSettings = webTestSettings.RemoteDriver;
+      var docker = new DockerCommandLineClient(TimeSpan.FromMinutes(3), _loggerFactory);
+
+      var imageNameWithPlaceholders = remoteDriverSettings.DockerImageName ?? "selenium/standalone-$browsername$:$browserversion$.0";
```
MichaelKetting: Question on API consistency: wouldn't `{browsername}` better since it aligns with composite strings / named holes instead of inventing a new placeholder syntax?

Remotion/Web/Development.WebTesting/WebDriver/Factories/Remote/RemoteBrowserFactory.cs
```diff
@@ -0,0 +1,126 @@
...
+using Remotion.Web.Development.WebTesting.BrowserSession.Remote;
+using Remotion.Web.Development.WebTesting.Configuration;
+using Remotion.Web.Development.WebTesting.WebDriver.Configuration.Remote;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Factories.Remote;
+
+/// <summary>
+/// Factory to create a browser session using Selenium remote driver.
+/// </summary>
+public class RemoteBrowserFactory : IBrowserFactory
```
MichaelKetting, line 23: how much of this is duplicated from the orignal browsers? should we leave comments? is there a risk about projects needing to duplicate browser options?

Remotion/Web/Development.WebTesting/WebDriver/BrowserConfigurationExtensions.cs
```diff
@@ -37,7 +38,8 @@ public static bool IsChrome ([NotNull] this IBrowserConfiguration browserConfigu
     {
       ArgumentNullException.ThrowIfNull(browserConfiguration);
 
-      return browserConfiguration is IChromeConfiguration;
+      return browserConfiguration is IChromeConfiguration
```
MichaelKetting, line 41: why us `is IsChromeConfiguration` here while changing this very code to `IscChrome()` in line 65?

Remotion/Web/Development.WebTesting/DownloadInfrastructure/NullDownloadHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+
+  public IDownloadedFile HandleDownloadWithExpectedFileName (string fileName, TimeSpan? downloadStartedTimeout = null, TimeSpan? downloadUpdatedTimeout = null)
+  {
+    throw new NotSupportedException();
+  }
+
+  public IDownloadedFile HandleDownloadWithDetectedFileName (TimeSpan? downloadStartedTimeout = null, TimeSpan? downloadUpdatedTimeout = null)
+  {
+    throw new NotSupportedException();
+  }
```
MichaelKetting: drop the default parameter values? don't help on concrete types anyway when it's always used via interface

Remotion/Web/Development.WebTesting/Configuration/Legacy/RemoteDriverConfigurationElement.cs
```diff
@@ -0,0 +1,60 @@
...
+    private readonly ConfigurationProperty _dockerImageNameProperty;
+    private readonly ConfigurationProperty _dockerCustomArgumentsProperty;
+
+    public RemoteDriverConfigurationElement ()
+    {
+      _enabledProperty = new ConfigurationProperty("enabled", typeof(bool), false);
+      _urlProperty = new ConfigurationProperty("url", typeof(string), "");
+      _hostRemoteDriverProperty = new ConfigurationProperty("hostRemoteDriverInDocker", typeof(bool), false);
+      _dockerImageNameProperty = new ConfigurationProperty("dockerImageName", typeof(string), null);
+      _dockerCustomArgumentsProperty = new ConfigurationProperty("dockerCustomArguments", typeof(string), null);
```
MichaelKetting, line 39: unless I'm misremembering, the schema says that the default is blank, here it's null.

Remotion/Web/Development.WebTesting/Configuration/IWebTestRemoteDriverSettings.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <item>
+    /// <term>$browsermajor$</term>
+    /// <description>The majorversion of the browser used (e.g. 114)</description>
+    /// </item>
+    /// </list>
+    /// </remarks>
+    /// <example>
+    /// For example, the default value:
+    /// <code>
+    /// selenium/standalone-$browsername$:$browsermajor$.0
```
MichaelKetting: if changed in the RemoteBrowserFactory (I think it was, see review comment) we also need to change the format here in the docs.

MichaelKetting (changes requested):
some questions/suggestions. are they valid?

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
a-ctor: Using braces will make it harder if you use interpolated strings. I am not really invested in any kind of placeholder syntax, I just think we need a placeholder syntax here to make it more useful. Irrespective of the placeholder syntax used, you need to know that the syntax exists, and as we don't have an established syntax, we have to pay the price anyway

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
a-ctor: It is optional because if you have `enabled="false"` I don't want to force a value in URL. I can make it required instead. There are cases where the URL could be determined automatically (when hosting the docker container) but we don't do that so as of now it is required.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Remote/RemoteBrowserFactory.cs
a-ctor: The options are copied but not all were relevant in the remote browser. I don't know if setting the options here does anything. We can check that once all the tests work, maybe we can get rid of the entirely.

If projects start to use the remote driver then they'll need to configure options in here as well if they need them.

Remotion/Web/Development.WebTesting/WebDriver/BrowserConfigurationExtensions.cs
a-ctor: I don't think I understand this comment. In line 65 we call this function. The change there was made because we now have an additional condition that we need to check if we are a certain browser and I did not want to duplicate the code.

Remotion/Web/Development.WebTesting/Configuration/Legacy/RemoteDriverConfigurationElement.cs
a-ctor: That is because XML does not have `null` as value except for absence. But we want the attribute to be optional so I have to add a default value? It's a bit of an inaccuracy, but it shouldn't matter as the default in the XML schema is only cosmetic.

Remotion/Web/Development.WebTesting/WebTestSetUpFixtureHelper.cs
MichaelKetting: yes. my reasoning is simply, if we copy the .net syntax, there's no extra cost to pay for developers as they know the syntax already and don't think about it.

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
MichaelKetting: quick talk about this? easier than back and forth.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Remote/RemoteBrowserFactory.cs
MichaelKetting: we should find a way to de-duplicate this extension point or make it clear also for projects. Follow-up ticket?

Remotion/Web/Development.WebTesting/WebDriver/BrowserConfigurationExtensions.cs
MichaelKetting: live talk :)

Remotion/Web/Development.WebTesting/Configuration/Legacy/RemoteDriverConfigurationElement.cs
MichaelKetting: hmm...if the attribute is optional, the xml should use RequiredString so the text is never empty. Unless empty is a valid value with a semantic different from null/missing.

MichaelKetting (approved):

PR #1348, opened 2025-07-17
Feature/rm 9535 support stored procedure calls explicitly

martin-herwich:
https://jira.rubicon.eu/browse/RM-9535

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/StoredProcedureDbCommandBuilderTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/StoredProcedureDbCommandBuilderTest.cs
@@ -1,0 +1,158 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
+// and/or modify it under the terms of the GNU Lesser General Public License 
+// as published by the Free Software Foundation; either version 2.1 of the 
+// License, or (at your option) any later version.
+// 
+// re-motion is distributed in the hope that it will be useful, 
+// but WITHOUT ANY WARRANTY; without even the implied warranty of 
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Data;
+using System.Data.Common;
+using Moq;
+using Moq.Protected;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Persistence.Rdbms;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.DbCommandBuilders;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+using Remotion.Data.DomainObjects.Queries;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.DbCommandBuilders;
+
+[TestFixture]
+public class StoredProcedureDbCommandBuilderTest
+{
+  private Mock<IDataParameterDefinition> _valueParameterDefinitionMock;
+  private QueryParameter _valueQueryParameter;
+  private QueryParameterWithDataParameterDefinition _valueQueryParameterWithDataDefinition;
+
+  [SetUp]
+  public void SetUp ()
+  {
+    _valueParameterDefinitionMock = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
+    _valueQueryParameter = new QueryParameter("#param1", 5, QueryParameterType.Value);
+    _valueQueryParameterWithDataDefinition = new QueryParameterWithDataParameterDefinition(_valueQueryParameter, _valueParameterDefinitionMock.Object);
+  }
+
+  [Test]
+  public void Create_UsesDbCommandFactoryToCreateQueryCommand ()
+  {
+    var dataParameterCollectionStrictMock = new Mock<DbParameterCollection>(MockBehavior.Strict);
+
+    var dbCommandStub = new Mock<DbCommand>();
+    dbCommandStub.Protected().Setup<DbParameterCollection>("DbParameterCollection").Returns(dataParameterCollectionStrictMock.Object);
+    dbCommandStub.SetupProperty(stub => stub.CommandText);
+
+    var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+    dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
+    var commandBuilder = new StoredProcedureDbCommandBuilder("Procedure", [], GetSqlDialectStub());
+    var result = commandBuilder.Create(dbCommandFactoryStub.Object);
+
+    dataParameterCollectionStrictMock.Verify();
+    _valueParameterDefinitionMock.Verify();
+    Assert.That(result, Is.SameAs(dbCommandStub.Object));
+  }
+
+  [Test]
+  public void Create_TransfersCommandTextToQueryCommand ()
+  {
+    var dataParameterCollectionStrictMock = new Mock<DbParameterCollection>(MockBehavior.Strict);
+
+    _valueParameterDefinitionMock.Setup(mock => mock.GetParameterValue(5)).Returns(5);
+
+    var dbCommandStub = new Mock<DbCommand>();
+    dbCommandStub.Protected().Setup<DbParameterCollection>("DbParameterCollection").Returns(dataParameterCollectionStrictMock.Object);
+    dbCommandStub.SetupProperty(stub => stub.CommandText);
+
+    var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+    dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
+    var commandBuilder = new StoredProcedureDbCommandBuilder("Procedure", [], GetSqlDialectStub());
+    var result = commandBuilder.Create(dbCommandFactoryStub.Object);
+
+    dataParameterCollectionStrictMock.Verify();
+    _valueParameterDefinitionMock.Verify();
+    Assert.That(result.CommandText, Is.EqualTo("Procedure"));
+  }
+
+  [Test]
+  public void Create_ThrowsOnTextParameter ()
+  {
+    var dataParameterCollectionStrictMock = new Mock<DbParameterCollection>(MockBehavior.Strict);
+
+    var dbCommandStub = new Mock<DbCommand>();
+    dbCommandStub.Protected().Setup<DbParameterCollection>("DbParameterCollection").Returns(dataParameterCollectionStrictMock.Object);
+    dbCommandStub.SetupProperty(stub => stub.CommandText);
+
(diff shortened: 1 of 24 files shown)


Remotion/Data/DomainObjects/Queries/QueryFactory.cs
```diff
@@ -279,7 +279,7 @@ public static IQuery CreateCollectionQuery (
       ArgumentNullException.ThrowIfNull(queryParameterCollection);
       ArgumentNullException.ThrowIfNull(collectionType);
 
-      var definition = new QueryDefinition(id, storageProviderDefinition, statement, QueryType.CollectionReadOnly, collectionType, metaData);
+      var definition = new QueryDefinition(id, storageProviderDefinition, statement, QueryType.CollectionReadOnly, QueryStatementType.Text, collectionType, metaData);
```
MichaelKetting: since the other occuranes aren't changed too, we should either leave it as a default here as well or change all instantiations of QueryDefinition in the factory.

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -159,6 +166,15 @@ public QueryType QueryType
...
+  /// <summary>
+  /// Gets the <see cref="QueryStatementType"/> of this <b>QueryDefinition</b>.
+  /// </summary>
+  public QueryStatementType StatementType
+  {
+    get { return _statementType; }
+  }
+
+
   /// <summary>
```
MichaelKetting:
```suggestion
  }

  /// <summary>
```

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -111,14 +114,18 @@ public QueryDefinition (
         && !collectionType.Equals(typeof(DomainObjectCollection))
         && !collectionType.IsSubclassOf(typeof(DomainObjectCollection)))
     {
-      throw new ArgumentException(string.Format(
-          "The collectionType of query '{0}' must be 'Remotion.Data.DomainObjects.DomainObjectCollection' or derived from it.", queryID), nameof(collectionType));
+      throw new ArgumentException(
+          string.Format(
+              "The collectionType of query '{0}' must be 'Remotion.Data.DomainObjects.DomainObjectCollection' or derived from it.",
+              queryID),
+          nameof(collectionType));
```
MichaelKetting: strictly speaking this is a change unrelated to the rest and considered noise. if you want to keep it, it would be great if you could split the commits so one does the various cleanups (here and in the other lines in this file) and the other the changes.

Remotion/Data/DomainObjects/Queries/Configuration/Loader/QueryDefinitionFileLoader.cs
```diff
@@ -126,7 +126,7 @@ private QueryDefinition GetQueryDefinition (XmlNode queryNode, ConfigurationName
       if ((queryType is QueryType.CustomReadOnly or QueryType.CustomReadWrite) && collectionType != null)
         throw CreateQueryConfigurationException("A custom query '{0}' must not specify a collectionType.", queryID);
 
-      return new QueryDefinition(queryID, storageProviderDefinition, statement, queryType, collectionType);
+      return new QueryDefinition(queryID, storageProviderDefinition, statement, queryType, QueryStatementType.Text, collectionType);
```
MichaelKetting: oops, missing feature: we need to add support for specifying the StatementType to the XML Schema (optional, to prevent breaking change) and then evaluate it here. we also need to add a new unit / integration test for the explicit stored procedures to document that both the old way to do things is unharmed and the new way works as well.

Remotion/Data/DomainObjects.UnitTests/Queries/QueryFactoryTest.cs
```diff
@@ -43,7 +43,7 @@ public override void TearDown ()
     [Test]
     public void CreateQuery_FromDefinition ()
     {
-      var definition = new QueryDefinition("Test", TestDomainStorageProviderDefinition, "y", QueryType.CollectionReadWrite, typeof(OrderCollection));
+      var definition = new QueryDefinition("Test", TestDomainStorageProviderDefinition, "y", QueryType.CollectionReadWrite, QueryStatementType.Text, typeof(OrderCollection));
```
MichaelKetting, line 46: please add assertion and one of the two tests touched in this file should use storedProc instead of Text so we demonstrate that it is actually passed through instead of hardcoded.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/QueryCommandFactoryTest.cs
```diff
@@ -173,7 +174,7 @@ public void CreateForScalarQuery ()
       var commandBuilderStub = new Mock<IDbCommandBuilder>();
       var expectedParametersWithType = GetExpectedParametersForQueryStub();
       _dbCommandBuilderFactoryStrictMock
-          .Setup(stub => stub.CreateForQuery("statement", expectedParametersWithType))
+          .Setup(stub => stub.CreateForQuery(QueryStatementType.Text, "statement", expectedParametersWithType))
```
MichaelKetting: I think we might want to inject this as a test parameter with values Text and StoredProc for each of the scalar, collection, and custom queries to show that the StatementType is passed through instead of hardcoded. We could isntead also simply use different values in each test method, but it seems like why shouldn't we use parameterized tests given they are not much more complicated anyway.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactoryTest.cs
```diff
@@ -159,12 +160,21 @@ public void CreateForSelect_UnionView ()
...
+      Assert.That(result, Is.TypeOf(typeof(QueryDbCommandBuilder)));
+      Assert.That(((QueryDbCommandBuilder)result).SqlDialect, Is.SameAs(_sqlDialectStub.Object));
+    }
+
+    [Test]
+    public void CreateForQuery_StoredProcedureStatementType_ReturnsStoredProcedureDbCommandBuilder ()
+    {
+      var result = _factory.CreateForQuery(QueryStatementType.StoredProcedure, "statement", new QueryParameterWithDataParameterDefinition[0]);
+
+      Assert.That(result, Is.TypeOf(typeof(StoredProcedureDbCommandBuilder)));
```
MichaelKetting: note: you can use the generic syntax, too: `Is.TypeOf<...>` 
personal preferences apply, I like the generic because it's shorter and easier to type/less wordy.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/StoredProcedureDbCommandBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+public class StoredProcedureDbCommandBuilder : DbCommandBuilder
+{
+  public string ProcedureName { get; }
+  public IEnumerable<QueryParameterWithDataParameterDefinition> Parameters { get; }
+
+  public StoredProcedureDbCommandBuilder (string procedureName, IEnumerable<QueryParameterWithDataParameterDefinition> parameters, ISqlDialect sqlDialect)
+      : base(sqlDialect)
+  {
+    ArgumentUtility.CheckNotNullOrEmpty(nameof(procedureName), procedureName);
+    ArgumentUtility.CheckNotNull(nameof(parameters), parameters);
```
MichaelKetting: Update you missed during vacation, sorry:

we have switched to ArgumentException.ThrowIfNullOrEmpty and ArgumentNullException.ThrowIfNull instead of those "standard" argument checks. please update the file. the other changes look okay.
There's no good way to guard against using the old syntax. I am still considering making them obsolete only in re-motion, though.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/StoredProcedureDbCommandBuilderTest.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    var dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+    dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(dbCommandStub.Object);
+
+    var textQueryParameterDefinitionMock = new Mock<IDataParameterDefinition>(MockBehavior.Strict);
+    var textQueryParameter = new QueryParameter("#textParam", "replacementText", QueryParameterType.Text);
+    var textQueryParameterWithDataDefinition = new QueryParameterWithDataParameterDefinition(textQueryParameter, textQueryParameterDefinitionMock.Object);
+
+    var commandBuilder = new StoredProcedureDbCommandBuilder("Procedure", [ textQueryParameterWithDataDefinition ], GetSqlDialectStub());
+    Assert.Throws<NotSupportedException>(() => commandBuilder.Create(dbCommandFactoryStub.Object));
```
MichaelKetting: Please update to `Assert.That(..., Throws....)` We rely mostly on `Assert.That` these days.

MichaelKetting (changes requested):
Some small bits and a missing part of the feature (support in query.xml )

Remotion/Data/DomainObjects/Queries/Configuration/QueryDefinition.cs
```diff
@@ -22,158 +22,170 @@
 
 namespace Remotion.Data.DomainObjects.Queries.Configuration
```
MichaelKetting: something went wrong here. the file was originally formatted in the 2004 style where we didn't indent for the namespace to save whitespace. please review changes.

Remotion/Data/DomainObjects/Schemas/Queries.xsd
```diff
@@ -45,5 +45,13 @@
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
+    <xs:attribute name="statementType" use="optional">
+      <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:enumeration value="text" />
+          <xs:enumeration value="storedProcedure" />
```
MichaelKetting:
```suggestion
          <xs:enumeration value="stored-procedure" />
```
for consistency with rest of file. I probably would have missed this detail if not for the hyphen-replace in the file loader.

Remotion/Data/DomainObjects.UnitTests/QueriesForLoaderTest.xml
```diff
@@ -48,19 +48,24 @@
...
     <statement>select 0;</statement>
   </query>
 
-  <query id="TestQueryDefinitionWithQueryTypeCustomReadWrite" type="custom-readwrite">
+  <query id="TestQueryDefinitionWithQueryTypeCustomReadWrite" type="custom-readwrite" statementType="text">
     <storageGroupType>Remotion.Data.DomainObjects.UnitTests::TestDomain.TestDomainAttribute</storageGroupType>
     <statement>select 0;</statement>
   </query>
 
+  <query id="TestStoredProcedureQuery" type="collection-readonly" statementType="storedProcedure">
```
MichaelKetting:
```suggestion
  <query id="TestStoredProcedureQuery" type="collection-readonly" statementType="stored-procedure">
```

MichaelKetting (changes requested):
Please add integration tests similiar to GetStoredProcedureResultWithParameter and GetStoredProcedureResult in Remotion.Data.DomainObjects.UnitTests.IntegrationTests.Queries.CollectionQueryTest for the new query type.

We probably don't need the one without parameters, though.

MichaelKetting (approved):

PR #1376, opened 2025-09-26
RM-9586 Disable dotnet vulnerability warnings per default

dkudernatsch:


Build/Shared.build.props
```diff
@@ -43,6 +43,11 @@
 
   <Import Project="$(ProjectTypePropsFile)" Condition="'$(ProjectTypePropsFile)' != '' AND Exists('$(ProjectTypePropsFile)')" />
 
+  <PropertyGroup>
+    <DisabledVulnerabilityWarnings Condition="'$(DisabledVulnerabilityWarnings)' != ';'">NU1901;NU1902;NU1903;NU1904</DisabledVulnerabilityWarnings>
```
MichaelKetting:
```suggestion
    <DisabledVulnerabilityWarnings Condition="'$(DisabledVulnerabilityWarnings)' == ''">NU1901;NU1902;NU1903;NU1904</DisabledVulnerabilityWarnings>
```
I was thinking "if the property is empty". That way, the externally supplied value is always taken and allows removing up to three of the four warnings. And if you want to kick all four warnings, you supply the ";" as a dummy value.

MichaelKetting (approved):

PR #1388, opened 2025-12-05
RM-9623 Downgrade SkiaSharp.NativeAssets.Linux.NoDependencies to matc…

a-ctor:
…h the SkiaSharp version

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -21,6 +21,7 @@
     <PackageVersion Include="log4net" Version="3.1.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing" Version="1.1.3-beta1.25564.1" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="6.1.1" />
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -50,7 +51,8 @@
     <PackageVersion Include="Remotion.TypePipe.Documentation" Version="$(TypePipeVersion)" />
     <PackageVersion Include="Selenium.Support" Version="4.35.0" />
     <PackageVersion Include="Selenium.WebDriver" Version="4.35.0" />
-    <PackageVersion Include="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="3.119.0" />
+    <!-- NOTE: This package version must match the SkiaSharp package for Linux support to work. -->
+    <PackageVersion Include="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="3.116.0" />
```
MichaelKetting, line 55: Sanity check: do we actually need to specify this package explicitly or can we also remove this line?

MichaelKetting (approved):

Directory.Packages.props
a-ctor: Yes it won't work otherwise as we are missing required libraries on Linux.

PR #1389, opened 2025-12-09
RM-9629 allow cloning csp enabled html text writer

a-ctor:


MichaelKetting (approved):
It feels like those two commits should be two separate ticket at least. one's a bug ticket.

PR #1394, opened 2025-12-16
RM-9634 Loosen ISmartPage dependency for CspEnabledHtmlTextWriter

a-ctor:


Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTest.cs
```diff
@@ -34,37 +35,47 @@ public class CspEnabledHtmlTextWriterTest
...
+          _htmlHelper.Writer,
+          _randomNumberGeneratorStub.Object,
+          "TEST-NONCE",
+          _renderingFeaturesStub.Object,
+          _fallbackNavigationUrlProviderStub.Object);
+
+      Assert.That(
+          PrivateInvoke.GetNonPublicField(writer, "_scriptManager"),
+          Is.TypeOf<SmartPageCspClientScriptManagerAdapter>());
+    }
```
MichaelKetting, line 77: The reflection call makes me sad since it is easy to just make the state public on the Writer. It's not enough to change this now, it's just... I try to avoid private invokes whenever possible because it feels we're doing a code smell.

Remotion/Web/Core/ContentSecurityPolicy/CollectingCspClientScriptManager.cs
```diff
@@ -0,0 +1,50 @@
...
+  /// <inheritdoc />
+  public void RegisterScript (string key, string script)
+  {
+    ArgumentException.ThrowIfNullOrEmpty(key);
+    ArgumentException.ThrowIfNullOrEmpty(script);
+
+    var position = _nextPosition;
+    _nextPosition += 1;
+
+    _scripts[key] = new Entry(position, script);
```
MichaelKetting, line 48: Design: do we need the _nextPosition field or could we do _scripts.Length as the position?

MichaelKetting (approved):

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTest.cs
a-ctor: Yes it is sad that is necessary and I wish there was a better way to test it without exposing the private field. I did not want to repeat the full smart page mock setup for something that is already tested in the test for `SmartPageCspClientScriptManagerAdapter`. This is more or less just a safety check that adapter has been created correctly and we can assume it works because we tested it in other unit tests.

Remotion/Web/Core/ContentSecurityPolicy/CollectingCspClientScriptManager.cs
a-ctor: That could lead to duplicate position entries as it is possible to override existing entries by specifying the same key.

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTest.cs
MichaelKetting: Yeah, I'd probbly just make a pblic getter for this since there's no particularly sensitive information in the field.

PR #1397, opened 2026-01-20
RM-9637 tvp infrastruktur herrichten

a-ctor:


Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+
+/// <summary>
+/// Provides the <see cref="RecordDefinition"/>s for TVPs that can be used to efficiently batch table manipulations.
+/// </summary>
+public interface ITableManipulationRecordDefinitionProvider
```
MichaelKetting, line 13: RecordDefinition is located in Rdmbs.Parameters namespace. Right now, they're meant just for use as parameters. at some point, we probably want to move more stuff to Model namespace. Ask Martin why we kept it in Parameters?

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+  /// <remarks>
+  /// This method is intended to be used in script generation.
+  /// Use the other methods on this interface to access a specific TVP <see cref="RecordDefinition"/>.
+  /// </remarks>
+  IEnumerable<RecordDefinition> GetAllRecordDefinitions (IEnumerable<ClassDefinition> classDefinitions);
+
+  /// <summary>
+  /// Gets the Delete TVP <see cref="RecordDefinition"/> for the specified <paramref name="classDefinition"/>.
+  /// </summary>
+  RecordDefinition GetDeleteRecordDefinition (ClassDefinition classDefinition);
```
MichaelKetting, line 18: We could create a single one for the entire IRdbmstorageProviderDefinition.

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+
+  /// <summary>
+  /// Gets the Insert TVP <see cref="RecordDefinition"/> for the specified <paramref name="classDefinition"/>.
+  /// </summary>
+  RecordDefinition GetInsertRecordDefinition (ClassDefinition classDefinition);
+
+  /// <summary>
+  /// Gets the Lock TVP <see cref="RecordDefinition"/> for the specified <paramref name="classDefinition"/>.
+  /// </summary>
+  RecordDefinition GetLockRecordDefinition (ClassDefinition classDefinition);
```
MichaelKetting, line 28: We could create a single one for the entire IRdbmstorageProviderDefinition.

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+  /// <summary>
+  /// Gets all TVP types for all the specified <paramref name="classDefinitions"/>.
+  /// Each applicable <see cref="ClassDefinition"/> will yield a TVP <see cref="RecordDefinition"/> for insert, update, delete, and lock.
+  /// </summary>
+  /// <remarks>
+  /// This method is intended to be used in script generation.
+  /// Use the other methods on this interface to access a specific TVP <see cref="RecordDefinition"/>.
+  /// </remarks>
+  IEnumerable<RecordDefinition> GetAllRecordDefinitions (IEnumerable<ClassDefinition> classDefinitions);
```
MichaelKetting: Not sure about this one. Is there a reason we can't just iterate the other methods?

Remotion/Data/DomainObjects/Persistence/Rdbms/SchemaGeneration/RdbmsStructuredTypeDefinitionProvider.cs
```diff
@@ -34,17 +35,25 @@ public RdbmsStructuredTypeDefinitionProvider ()
...
     /// <param name="storageProviderDefinition">The storage provider for which to generate scripts.</param>
+    /// <param name="classDefinitions">The class definition that should be considered for script generation.</param>
     /// <remarks>
     /// In order to influence the returned collection, override or mix the <see cref="IRdbmsStorageObjectFactory.CreateSingleScalarStructuredTypeDefinitionProvider"/> method
     /// on the <paramref name="storageProviderDefinition"/>'s <see cref="RdbmsProviderDefinition.Factory"/>. 
     /// </remarks>
-    public IReadOnlyCollection<IRdbmsStructuredTypeDefinition> GetTypeDefinitions (RdbmsProviderDefinition storageProviderDefinition)
+    public IReadOnlyCollection<IRdbmsStructuredTypeDefinition> GetTypeDefinitions (RdbmsProviderDefinition storageProviderDefinition, IEnumerable<ClassDefinition> classDefinitions)
     {
       ArgumentNullException.ThrowIfNull(storageProviderDefinition);
```
MichaelKetting, line 47: add nullchecks immediately or once we're done with the API?

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -299,6 +299,15 @@ public virtual ISingleScalarStructuredTypeDefinitionProvider CreateSingleScalarS
       return CreateSingleScalarStructuredTypeDefinitionProvider(storageTypeInformationProvider);
     }
 
+    public virtual ITableManipulationRecordDefinitionProvider CreateTableManipulationRecordDefinitionProvider (RdbmsProviderDefinition storageProviderDefinition)
```
MichaelKetting: not sure if both methods should be virtual. possibly, it's just... I haven't looked at the rest of the factory for guidance.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,371 @@
...
+
+  private record TvpTableTypeDefinitions (
+      TableTypeDefinition DeleteTableTypeDefinition,
+      ImmutableArray<TvpColumnMetadata> DeleteColumns,
+      TableTypeDefinition InsertTableTypeDefinition,
+      ImmutableArray<TvpColumnMetadata> InsertColumns,
+      TableTypeDefinition LockTableTypeDefinition,
+      ImmutableArray<TvpColumnMetadata> LockColumns,
+      TableTypeDefinition UpdateTableTypeDefinition,
+      ImmutableArray<TvpColumnMetadata> UpdateColumns);
```
MichaelKetting, line 161: The immutable types should only be used if manipulation is a relevant factor. Otherwise, they're significantly more expensive than Frozen types or for private stuff, just sticking with an Array. As needed, use the IReadOnlyList/IReadOnlyCollection interface.

https://codeblog.jonskeet.uk/2025/12/31/changing-immutable-collections/

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,371 @@
...
+
+  static TableManipulationRecordDefinitionProvider ()
+  {
+    // Doing this in the initializer is a bit annoying so we do it in the static constructor.
+    var builder = ImmutableDictionary.CreateBuilder<Type, object>();
+    builder.Add(typeof(string), string.Empty);
+    builder.Add(typeof(byte[]), Array.Empty<byte>());
+
+    s_optionalColumnsDefaultValues = builder.ToImmutable();
+  }
```
MichaelKetting: Since this is static readonly: I weould suggest we just make this into an IReadOnlyDicitionary and keep it based on Dictionary. This should yield better performance and we could use a field initializer

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,371 @@
...
+
+  private readonly IStorageTypeInformationProvider _storageTypeInformationProvider;
+  private readonly IRdbmsPersistenceModelProvider _rdbmsPersistenceModelProvider;
+
+  private readonly ConcurrentDictionary<TableDefinition, Lazy<TvpTableTypeDefinitions>> _tableTypeDefinitions = new();
+  private readonly ConcurrentDictionary<ClassDefinition, Lazy<TvpRecordDefinitions?>> _recordDefinitions = new();
+
+  public TableManipulationRecordDefinitionProvider (
```
MichaelKetting, line 183: continue review here, then review nested types

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
a-ctor: Sure, I mirrored the design from the other TVP interface. If you want we can remove this method and do the iteration in the consuming class.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
a-ctor: I think you might be confusing `ImmutableList<T>` and `ImmutableArray<T>`. The linked blog article mentions `ImmutableArray<T>` as the right solution for "created once" immutable arrays. I don't see a reason to change the code here for performance reasons.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
a-ctor: This would be the place to use `FrozenDictionary<T, K>` as the dictionary implementation is inherently more complex than the immutable array and might not be the most efficient in terms of accesses. Funnily enough, using a `FrozenDictionary<T, K>` might even use an array and linear search instead of a dictionary structure as that is probably faster for two elements ^^

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
MichaelKetting: Yeah, the ISingleScalarStructuredTypeDefinitionProvider.GetAllStructuredTypeDefinitions() is the only way to know for which types to create a TVP. Here, it's more an "give me the TVP for this ClassDefinition". My overarching goal is to not have redundant APIs (or helper APIs) on the interface, just those that actually do stuff no other member of the interface does. Here, this API is something an extension method/utility method could do just as well.

-> I don't mind having an extension method / something that's testable. Just not redundant on the interface please :)

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
MichaelKetting: oops, yeah :)

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
MichaelKetting: it does? You have no idea how much I labored to not suggest the various "just two items" patterns, incuding the iterator types .net now offers :)

So... let's watch https://en.wikipedia.org/wiki/Frozen_%282013_film%29 ?

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/ITableManipulationRecordDefinitionProvider.cs
a-ctor: It was placed there because it was placed there. It would be more at home in the mapping directory, but I would clean that up when/if we introduce the TVP as a first class mapping citizen.

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_TestDomain_SetupDB.sql
```diff
@@ -1296,3 +1296,1091 @@ CREATE TYPE [dbo].[TVP_Guid_Distinct] AS TABLE
...
+  [ClassID] varchar (100) NOT NULL,
+  [Persistent] int NOT NULL
+)
+GO
+CREATE TYPE [dbo].[TVP_AllTables_Lock] AS TABLE
+(
+  [ID] uniqueidentifier NOT NULL,
+  [Timestamp] rowversion NOT NULL
+)
+GO
```
MichaelKetting, line 1316: Move that one up to the other all_tables?

Or was this based on an automatic sort?

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_TestDomain_TearDownDB.sql
```diff
@@ -30,6 +30,122 @@ DROP TYPE IF EXISTS [dbo].[TVP_Date]
 DROP TYPE IF EXISTS [dbo].[TVP_Date_Distinct]
 DROP TYPE IF EXISTS [dbo].[TVP_Guid]
 DROP TYPE IF EXISTS [dbo].[TVP_Guid_Distinct]
+DROP TYPE IF EXISTS [dbo].[TVP_AllTables_Delete]
+DROP TYPE IF EXISTS [dbo].[TVP_ClassWithPropertiesHavingStorageClassAttribute_Insert]
+DROP TYPE IF EXISTS [dbo].[TVP_AllTables_Lock]
+DROP TYPE IF EXISTS [dbo].[TVP_ClassWithPropertiesHavingStorageClassAttribute_Update]
```
MichaelKetting, line 36:
```suggestion
DROP TYPE IF EXISTS [dbo].[TVP_AllTables_Delete]
DROP TYPE IF EXISTS [dbo].[TVP_AllTables_Lock]
DROP TYPE IF EXISTS [dbo].[TVP_ClassWithPropertiesHavingStorageClassAttribute_Insert]
DROP TYPE IF EXISTS [dbo].[TVP_ClassWithPropertiesHavingStorageClassAttribute_Update]
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SchemaGeneration/RdbmsStructuredTypeDefinitionProvider.cs
```diff
@@ -34,17 +35,55 @@ public RdbmsStructuredTypeDefinitionProvider ()
...
+            (unionView, continuation) => null,
+            (emptyView, continuation) => null);
+        if (tableDefinition == null)
+          continue;
+
+        yield return tableManipulationRecordDefinitionProvider.GetDeleteRecordDefinition(classDefinition).StructuredTypeDefinition;
+        yield return tableManipulationRecordDefinitionProvider.GetInsertRecordDefinition(classDefinition).StructuredTypeDefinition;
+        yield return tableManipulationRecordDefinitionProvider.GetLockRecordDefinition(classDefinition).StructuredTypeDefinition;
+        yield return tableManipulationRecordDefinitionProvider.GetUpdateRecordDefinition(classDefinition).StructuredTypeDefinition;
+      }
```
MichaelKetting, line 86:
```suggestion
        yield return tableManipulationRecordDefinitionProvider.GetDeleteRecordDefinition(classDefinition).StructuredTypeDefinition;
        yield return tableManipulationRecordDefinitionProvider.GetLockRecordDefinition(classDefinition).StructuredTypeDefinition;
        yield return tableManipulationRecordDefinitionProvider.GetInsertRecordDefinition(classDefinition).StructuredTypeDefinition;
        yield return tableManipulationRecordDefinitionProvider.GetUpdateRecordDefinition(classDefinition).StructuredTypeDefinition;
      }
```

because we know that the sorting is stable and that the delete and lock records are idientical, returning it this way results in a nicer sql file

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/ITableManipulationDataContainerAccessor.cs
```diff
@@ -0,0 +1,45 @@
+﻿// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using Remotion.Data.DomainObjects.DataManagement;
+using Remotion.Data.DomainObjects.Mapping;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Model;
+
+/// <summary>
+/// Provides methods to access data in a <see cref="DataContainer"/> as an indirection.
+/// This allows changing the data retrieved without changing the actual data stored in the DataContainer.
```
MichaelKetting, line 10: not yet reviews the usage of this interface. the summary feels off to me, like we use this interface to inject fake data / "changed" data during production.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
```diff
@@ -0,0 +1,373 @@
...
+    public static TvpColumnMetadata CreateForIDColumn (IRdbmsStoragePropertyDefinition property)
+    {
+      ArgumentNullException.ThrowIfNull(property);
+
+      return new TvpColumnMetadata(
+          property,
+          static (property, _) =>
+          {
+            return new RecordPropertyDefinition(
+                "ID",
```
MichaelKetting, line 51: Is this the one and only declaration of "ID" and "Timestamp" for the TVP definition or a duplication? I haven't re-reviewed the rest of the PR to check.

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_TestDomain_SetupDB.sql
a-ctor: All of those SQL files are autogenerated and the order is determined by the order we call the methods on `ITableManipulationRecordDefinitionProvider`. Since we removed the `GetAll` method, there is no good way to enforce an order. The lock table is shared but written once it is first requested. That's why this ordering happens.

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_TestDomain_TearDownDB.sql
a-ctor: See other comment about this file being auto-generated.

Remotion/Data/DomainObjects/Persistence/Rdbms/SchemaGeneration/RdbmsStructuredTypeDefinitionProvider.cs
a-ctor: Sure if we do this you can disregard my other comments.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/ITableManipulationDataContainerAccessor.cs
a-ctor: Yes that is exactly what it is for, only that we need it to support setting the foreign key later. Previously, we could leave out columns and they would be set to NULL. So in the insert we would leave out the foreign keys and update them later. Now with batching, we have to specify all columns and thus need to replace the foreign key by NULL explicitly.

This interface is also used to handle all the optional columns (strings/binary) and their is set columns.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
a-ctor: The name here is cosmetic and for debugging purposes only. It is not used apart from the tests. This is the only place where the ID/Timestamp are explicitly named, all other uses get their columns from the properties on the table

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_TestDomain_SetupDB.sql
MichaelKetting: Check

Remotion/Data/DomainObjects.UnitTests/Database/DataDomainObjects_TestDomain_TearDownDB.sql
MichaelKetting: Check

Remotion/Data/DomainObjects/Persistence/Rdbms/SchemaGeneration/RdbmsStructuredTypeDefinitionProvider.cs
MichaelKetting: :) :) :)

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Model/TableManipulationRecordDefinitionProvider.cs
MichaelKetting: okay. maybe add a comment that it's only for diagnostics purposes. This sort of thing could bug a developer a year or two or ten down the road :D

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SchemaGeneration/RdbmsStructuredTypeDefinitionProviderTest.cs
```diff
@@ -15,42 +15,113 @@
...
+      return classDefinition;
+    }
+
+    private static TableDefinition CreateTableDefinitionStub ()
+    {
+      // Yeah, it is a hack. Creating a TableDefinition is quite a lot of work, and we only need the object to pass a type check.
+      return (TableDefinition)RuntimeHelpers.GetUninitializedObject(typeof(TableDefinition));
+    }
+
+    private static UnionViewDefinition CreateUnionViewDefinition ()
```
MichaelKetting, line 121:
```suggestion
    private static UnionViewDefinition CreateUnionViewDefinitionStub ()
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/Model/TableManipulationRecordDefinitionProviderTest.cs
```diff
@@ -0,0 +1,566 @@
...
+            structuredTypeDefinitionFromCurrentOperation,
+            Is.Not.SameAs(structureTypeDefinitionFromOtherOperation));
+      }
+
+      structureTypeDefinitionsFromOtherOperations.Add(structuredTypeDefinitionFromCurrentOperation);
+    }
+  }
+
+  [Test]
+  public void GetDeleteRecordDefinition_SimpleDataClass_ReturnsCorrectRecordDefinition ()
```
MichaelKetting, line 108: If I may make a wish upon a star: no "correct" in the expectation clause of the method name because nobody knows what "correct" means. Please instead list the columns or use some sort of "record definition wiht columns from table" or some such wording. THere's several "correct" usages in this file's tests.

MichaelKetting (approved):
Just some naming things in the tests, I think.

MichaelKetting (changes requested):
Forgot two details:

Branch name or commit message of merge should be in English so the final commit history is in English.

More important: we need to move stuff from the SQL server namespace to the rdbms namespace. Basically, everything that's not actual tvp Statements or otherwise only applicable to SQL Server. The assumption is that PostgreSQL and Oracle UDT could use the same abstractions. We should probably check this list live. Disclaimer: we assume the other rdbms have a transaction/consistency model no stricter than SQL server and thus we can use at least the same logic. Possibly even more freedom (consistency is only enforced for the transaction but not the individuell Statement)

PR #1401, opened 2026-02-09
feature/RM-9538 userfeedback on paste overflows in text values

noah-rubicon:
https://jira.rubicon.eu/browse/RM-9538


Remotion/ObjectBinding/Web/UI/Controls/Styles.cs
```diff
@@ -308,13 +308,14 @@ public class SingleRowTextBoxStyle : Style
...
     private bool? _checkClientSideMaxLength;
+    private bool? _checkMaxLengthOnPaste;
 
     public virtual void ApplyStyle (TextBox textBox)
     {
       textBox.ApplyStyle(this);
 
       var maxLength = GetMaxLength();
-      if (maxLength != null && _checkClientSideMaxLength != false)
+      if (maxLength != null && (_checkClientSideMaxLength != false || _checkMaxLengthOnPaste != false))
```
MichaelKetting: Good find. 

Design descision: revert this. If we don't check the maxlength on the client side, we also don't need to add a check on paste becasue the input/textarea element already doesn't have a max-length that would constrain the paste anyway.

Remotion/ObjectBinding/Web/UI/Controls/Styles.cs
```diff
@@ -412,6 +414,18 @@ public bool? CheckClientSideMaxLength
       set { _checkClientSideMaxLength = value; }
     }
 
+    [Description(
+        "Whether the text in the control can exceed its max length after a paste operation."
```
MichaelKetting: we need to add the false information like with CheckClientSideMaxLength.
we also need to add a note about this flag only getting evaluated/used when CheckClientSideMaxLength is not set to false.

Remotion/ObjectBinding/Web/UI/Controls/Styles.cs
MichaelKetting: Because I noticed this during the review: I think it's possible that the "true" should be a false here. 
```suggestion
        "Whether the text in the control can exceed its max length during input. If false, MaxLength is only used for validation after the input is completed."
```

Remotion/ObjectBinding/Web/UI/Controls/Styles.cs
```diff
@@ -456,7 +470,6 @@ public TextBoxStyle ()
     public override void ApplyStyle (TextBox textBox)
     {
       base.ApplyStyle(textBox);
-
```
MichaelKetting, line 459: revert please.

Remotion/ObjectBinding/Web/UI/Controls/BocTextValueImplementation/Rendering/BocTextValueRenderer.cs
```diff
@@ -72,6 +72,16 @@ protected sealed override TextBox GetTextBox (BocRenderingContext<IBocTextValue>
       var textBox = base.GetTextBox(renderingContext);
       if (renderingContext.Control.TextBoxStyle.TextMode == BocTextBoxMode.PasswordRenderMasked)
         textBox.Attributes.Add("value", textBox.Text);
+
+      var maxLength = renderingContext.Control.TextBoxStyle.GetMaxLength();
+
+      if (maxLength != null && renderingContext.Control.TextBoxStyle.CheckMaxLengthOnPaste != false)
+      {
```
MichaelKetting, line 82:
```suggestion
      var maxLength = renderingContext.Control.TextBoxStyle.GetMaxLength();
      if (maxLength != null && renderingContext.Control.TextBoxStyle.CheckMaxLengthOnPaste != false)
      {
```
stylistic choice: I prefer to keep a variable created for an IF block right next to the IF block, no blank line if there's no other line breaks / a lot of code involved. keeps the logical elements grouped together.

Remotion/ObjectBinding/Web/UI/Controls/BocTextValueImplementation/Rendering/BocTextValueRenderer.cs
```diff
@@ -72,6 +72,16 @@ protected sealed override TextBox GetTextBox (BocRenderingContext<IBocTextValue>
...
+        var resourceManager = renderingContext.Control.GetResourceManager();
+        var message = resourceManager.GetString(BocTextValue.ResourceIdentifier.TextOverflowOnPasteValidationMessage);
+        textBox.Attributes.Add("onpaste", $"return TextBoxStyle.OnPaste (this, {maxLength.Value}, '{string.Format(message, maxLength)}')");
```
MichaelKetting: Gut feeling / personal choice / a bit of a suggestion and my preference
```suggestion
        var message = string.Format(
            resourceManager.GetString(BocTextValue.ResourceIdentifier.TextOverflowOnPasteValidationMessage, 
            maxLength));
        textBox.Attributes.Add("onpaste", $"return TextBoxStyle.OnPaste (this, {maxLength.Value}, 'message')");
```
I hope the syntax / indent is okay. 

Reasoning: the javascript is already complex enough without nesting a string-format into the interpolated string.
we could split the string-format and getstring into separate variables but here the use code formatting works well enough.

Remotion/ObjectBinding/Web/UI/Controls/BocTextValueImplementation/Rendering/BocMultilineTextValueRenderer.cs
```diff
@@ -78,7 +78,17 @@ public override string GetCssClassBase (IBocMultilineTextValue control)
...
+
+      if (maxLength != null && renderingContext.Control.TextBoxStyle.CheckMaxLengthOnPaste != false)
+      {
+        var resourceManager = renderingContext.Control.GetResourceManager();
+        var message = resourceManager.GetString(BocMultilineTextValue.ResourceIdentifier.TextOverflowOnPasteValidationMessage);
+        textBox.Attributes.Add("onpaste", $"return TextBoxStyle.OnPaste (this, {maxLength.Value}, '{string.Format(message, maxLength)}')");
+      }
+
+      return textBox;
     }
```
MichaelKetting, line 96: Same suggestions as in BocTextValueRenderer

Remotion/ObjectBinding/Web/UI/Controls/BocTextValueImplementation/Rendering/BocTextValueRenderer.cs
```diff
@@ -72,6 +72,16 @@ protected sealed override TextBox GetTextBox (BocRenderingContext<IBocTextValue>
       var textBox = base.GetTextBox(renderingContext);
       if (renderingContext.Control.TextBoxStyle.TextMode == BocTextBoxMode.PasswordRenderMasked)
         textBox.Attributes.Add("value", textBox.Text);
+
+      var maxLength = renderingContext.Control.TextBoxStyle.GetMaxLength();
+
+      if (maxLength != null && renderingContext.Control.TextBoxStyle.CheckMaxLengthOnPaste != false)
+      {
+        var resourceManager = renderingContext.Control.GetResourceManager();
```
MichaelKetting: I know I suggested it differently at first but given the possiblity someone might want to derive / extend the renderer, we should introduce the virtual helper method after all.

```
    protected virtual IResourceManager GetResourceManager (BocRenderingContext<IBocTextValue> renderingContext)
    {
      return renderingContext.Control.GetResourceManager();
    }
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/TextBoxStyle.ts
```diff
@@ -49,4 +49,22 @@ class TextBoxStyle
       
     return true;
   }
+  
+  public static OnPaste (textBox: HTMLInputElement, length: number, message: string)
+  {
+   const initialContentLength = textBox.value.length;
+   let selectedCharCnt = 0;
```
MichaelKetting: No abreviations, please.
```suggestion
   let selectedCharCount = 0;
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/TextBoxStyle.ts
```diff
@@ -49,4 +49,22 @@ class TextBoxStyle
...
+     selectedCharCnt = textBox.selectionEnd - textBox.selectionStart ;
+   }
+   
+   navigator.clipboard.readText().then(clipText => {
+        const totalLength = clipText.length + initialContentLength - selectedCharCnt;
+    
+        if(totalLength > length){
+            window.alert(message);
+        }
+   });
```
MichaelKetting: Whitespace after the IF/before the parenthesis
braces: we try to follow the same style as in c#, so braces start on a new line.
method chains: consider line breaks at semantic places

```suggestion
   if (textBox.selectionStart !== null && textBox.selectionEnd !== null)
   {
     selectedCharCnt = textBox.selectionEnd - textBox.selectionStart ;
   }
   
   navigator.clipboard
       .readText()
       .then(clipText => 
       {
          const totalLength = clipText.length + initialContentLength - selectedCharCnt;
    
          if (totalLength > length)
          {
            window.alert(message);
          }
   });
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/TextBoxStyle.ts
```diff
@@ -49,4 +49,22 @@ class TextBoxStyle
       
     return true;
   }
+  
+  public static OnPaste (textBox: HTMLInputElement, length: number, message: string)
```
MichaelKetting:
```suggestion
  public static OnPaste (textBox: HTMLInputElement, maxLength: number, textOverflowMessage: string)
```

Remotion/ObjectBinding/Web.ClientScript/Scripts/TextBoxStyle.ts
```diff
@@ -49,4 +49,22 @@ class TextBoxStyle
...
+  {
+   const initialContentLength = textBox.value.length;
+   let selectedCharCnt = 0;
+     
+   if(textBox.selectionStart !== null && textBox.selectionEnd !== null){
+     selectedCharCnt = textBox.selectionEnd - textBox.selectionStart ;
+   }
+   
+   navigator.clipboard.readText().then(clipText => {
+        const totalLength = clipText.length + initialContentLength - selectedCharCnt;
```
MichaelKetting: You found a way to do this without playing with the maxlength in the DOM. Nice!

MichaelKetting (changes requested):
Some small stylic and formatting stuff.

Remotion/ObjectBinding/Web/UI/Controls/Styles.cs
noah-rubicon: what about this?

Whether the text in the control can exceed its max length after a paste operation. If true and CheckClientSideMaxLength is also true, the MaxLength will be used for validation on paste events.

MichaelKetting: > Whether the text in the control can exceed its max length after a paste operation. The validation is performed if neither this flag nor CheckClientSideMaxLength are set to false.

reasoning: the default is null/blank which acts as "true". so "false" is the value that the developer has to use to opt out of the default behavior. Using "true" in the text would imply that leaving it default  / blank isn't enough.

noah-rubicon: so this should do:

Whether the text in the control can exceed its max length after a paste operation. If false or CheckClientSideMaxLength is false, the MaxLength is only used for validation after the input is completed.

MichaelKetting: looks great. thanks!

MichaelKetting (approved):

PR #1404, opened 2026-02-23
Feature/rm 9639 batched save command factory

Mario185:
https://jira.rubicon.eu/browse/RM-9639

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/RdbmsProviderCommandFactoryTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/RdbmsProviderCommandFactoryTest.cs
@@ -22,6 +22,7 @@ using Remotion.Data.DomainObjects.DataManagement;
 using Remotion.Data.DomainObjects.Mapping;
 using Remotion.Data.DomainObjects.Persistence.Rdbms;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.DataReaders;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.Parameters;
 using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer;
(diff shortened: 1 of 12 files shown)


Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/IndividualSaveCommandFactory.cs
```diff
@@ -69,26 +69,26 @@ public virtual IRdbmsProviderCommand CreateForSave (IEnumerable<DataContainer> d
...
         if (ShouldCreateDeleteCommand(dataContainer))
-          deleteCommands.Add(Tuple.Create(dataContainer.ID, CreateDbCommandForDelete(dataContainer, tableDefinition)));
+          deleteCommands.Add(new SingleObjectMultiDataContainerSaveCommandContext(dataContainer.ID, CreateDbCommandForDelete(dataContainer, tableDefinition)));
 
         var updatedColumnValues = GetUpdatedColumnValues(dataContainer, tableDefinition).ToArray();
         if (updatedColumnValues.Any())
         {
           var dbCommandForUpdate = CreateDbCommandForUpdate(dataContainer, tableDefinition, updatedColumnValues);
-          updateCommands.Add(Tuple.Create(dataContainer.ID, dbCommandForUpdate));
+          updateCommands.Add(new SingleObjectMultiDataContainerSaveCommandContext(dataContainer.ID, dbCommandForUpdate));
```
a-ctor: I would prefer copying the `MultiDataContainerSaveCommand` and keeping this code as is. By keeping the new code separate we ensure that the working, battle tested, code is not lost and if a project falls back to the individual strategy it works like it does today. This also gives us a bit more freedom for refactoring the command once we implemented batching (because then we don't need the command context abstraction anymore).

What do you think?

a-ctor (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/IndividualSaveCommandFactory.cs
MichaelKetting: I agree  with keeping it strictly separate. We want to be able to provide the existing behavior as an opt in for projects so we should strive for a completely parallel track of the stuff happening inside the save command infrastructure. The existing code hasn't been touched in ages except for the SQL client update and thus isn't cause for worry in terms of maintenance.

Mario185: I have reverted the changes for MultiDataSaveCommand and other dependencies.

a-ctor (approved):

PR #1405, opened 2026-02-27
RM-9640 batched locking

Mario185:
https://jira.rubicon.eu/browse/RM-9640


Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlTableTypeScriptElementFactory.cs
a-ctor: I think we should add a comment here that explains why we need the special case for `rowversion` to document the reasoning for future readers.

Right now, we have almost the same return statement twice, which requires the reader to spot the difference. If we use a variable for the storage type name we can remove much of the duplication.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlTableTypeScriptElementFactory.cs
```diff
@@ -85,6 +85,9 @@ public IScriptElement GetDropElement (TableTypeDefinition tableTypeDefinition)
 
   private string GetColumnDeclaration (ColumnDefinition column)
   {
+    if (column.StorageTypeInfo.StorageTypeName == "rowversion")
+      return $"[{column.Name}] varbinary({column.StorageTypeInfo.StorageTypeLength ?? 8}) {(column.StorageTypeInfo.IsStorageTypeNullable ? "NULL" : "NOT NULL")}";
+
     return $"[{column.Name}] {column.StorageTypeInfo.StorageTypeName} {(column.StorageTypeInfo.IsStorageTypeNullable ? "NULL" : "NOT NULL")}";
   }
```
a-ctor, line 95: I think we should add a comment here to document why we need the special rule for `rowversion`.

The two return statements are very similar and they require the reader to spot the difference. Since we only change the type part, I would suggest extracting it into a variable and having a single return statement.

Something like this: 

```cs
var storageTypeName = column.StorageTypeInfo.StorageTypeName == "rowversion"
    ? $"varbinary({column.StorageTypeInfo.StorageTypeLength ?? 8})"
    : column.StorageTypeInfo.StorageTypeName;

return $"[{column.Name}] {storageTypeName} {(column.StorageTypeInfo.IsStorageTypeNullable ? "NULL" : "NOT NULL")}";
```

Remotion/Data/DomainObjects/Persistence/Rdbms/Model/TableManipulationRecordDefinitionProvider.cs
```diff
@@ -42,15 +42,28 @@ public static TvpColumnMetadata CreateForIDColumn (IRdbmsStoragePropertyDefiniti
...
+        return new TvpColumnMetadata(
+            property,
+            static (property, _) =>
+            {
+              return new RecordPropertyDefinition(
+                  "ID", // Note: Name here is cosmetic and not used anywhere
+                  property,
+                  o => ((ITableManipulationDataContainerAccessor)o).GetID());
+            });
+      }
```
a-ctor: Looking at the usages of this method, this `if` is only hit if we come from the overload above. We remove the `if` by moving the code up into the overload. We should rename this method to highlight the difference between the two methods.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+        .Append("RIGHT JOIN ")
+        .Append(parameterName)
+        .Append(' ')
+        .Append(parameterAlias)
+        .Append(" ON ")
+        .AppendJoin(" AND ", columns.Select(c => parameterAlias + "." + c + " = " + tableAlias + "." + c))
+        .AppendLine()
+        .Append("WHERE ")
+        .Append(tableAlias)
+        .Append(".[ID] IS NULL");
```
a-ctor: Reading this statement is kind of hard. I would suggest using a interpolated raw string just like in the method above. `StringBuilder` handles the transformation correctly, so it will be added to the string builder directly instead of creating a new string in between. By moving all the conditional parameters and list parameters into variables up front, we can keep the interpolated string pretty simple and readable.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    var tableAlias = SqlDialect.DelimitIdentifier("T");
+    var parameterAlias = SqlDialect.DelimitIdentifier("P");
+
+    var forReadCommitBuilder = new StringBuilder();
+    var forNonReadCommitBuilder = new StringBuilder();
+
+    foreach (var specification in _commandSpecifications)
+    {
+      if (forReadCommitBuilder.Length > 0 || forNonReadCommitBuilder.Length > 0)
```
a-ctor: The condition here feels special but isn't. The second part of the condition is just not needed as it will always evaluate to whatever the first part is. As a reader, it looks a bit strange and requires checking the mutations of `forNonReadCommitBuilder`. I would rewrite this using a boolean variable `isFirstCommandSpecification` to make the intent clearer.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/Specifications/IBatchedLockCommandSpecification.cs
```diff
@@ -0,0 +1,33 @@
...
+  {
+    /// <summary>
+    /// Creates the <see cref="DbParameter"/> for the given <paramref name="command"/>.
+    /// </summary>
+    DbParameter CreateDbParameter (DbCommand command, string parameterName);
+
+    /// <summary>
+    /// Gets the column names.
+    /// </summary>
+    /// <returns></returns>
```
a-ctor:
```suggestion
```

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/Specifications/BatchedLockCommandSpecification.cs
```diff
@@ -0,0 +1,65 @@
...
+    public SqlTableValuedParameterValue ParameterValue { get; }
+
+    public IReadOnlyList<string> Columns { get; }
+  }
+
+  private readonly MultiClassTableValuedDataParameterDefinition _parameterDefinition;
+  private readonly DataContainer[] _dataContainers;
+  private readonly Lazy<CommandSpecificationData> _preparedData;
+
+  public BatchedLockCommandSpecification (TableDefinition tableDefinition, MultiClassTableValuedDataParameterDefinition parameterDefinition, DataContainer[] dataContainers)
```
a-ctor: We almost never pass around arrays as they imply mutability of elements as part of the contract. Here, I think we can get through with `IEnumerable<DataContainer>` as that is what the parameter expects.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/Specifications/BatchedLockCommandSpecification.cs
```diff
@@ -0,0 +1,65 @@
...
+  {
+    ArgumentNullException.ThrowIfNull(tableDefinition);
+    ArgumentNullException.ThrowIfNull(parameterDefinition);
+    ArgumentNullException.ThrowIfNull(dataContainers);
+
+    TableDefinition = tableDefinition;
+    _parameterDefinition = parameterDefinition;
+    _dataContainers = dataContainers;
+    _preparedData = new Lazy<CommandSpecificationData>(PrepareData);
```
a-ctor: Is there a reason why we don't just compute the data here in the constructor? (instead of doing it lazy)

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/DbCommandBuilders/SqlDbCommandBuilderFactory.cs
```diff
@@ -165,6 +166,13 @@ public IDbCommandBuilder CreateForDelete (TableDefinition tableDefinition, IEnum
       return new DeleteDbCommandBuilder(tableDefinition, new ComparedColumnsSpecification(comparedColumnValues), _sqlDialect);
     }
 
+    /// <inheritdoc/>
+    public IDbCommandBuilder CreateForBatchedLock (IBatchedLockCommandSpecification[] commandSpecifications)
+    {
+      ArgumentUtility.CheckNotNullOrEmptyOrItemsNull(nameof(commandSpecifications), commandSpecifications);
+      return new BatchedLockDbCommandBuilder(_sqlDialect, commandSpecifications);
```
a-ctor, line 174: styling: Between arg checks and the code of the methods should be an empty line for visual separation.

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/BatchedObjectsDataContainerSaveCommandContext.cs
```diff
@@ -0,0 +1,68 @@
...
+  {
+    ArgumentNullException.ThrowIfNull(tableDefinition);
+    ArgumentNullException.ThrowIfNull(commandBuilder);
+
+    TableDefinition = tableDefinition;
+    CommandBuilder = commandBuilder;
+    AffectedDataContainers = affectedDataContainers;
+  }
+
+  public virtual int ExpectedAffectedRowCount => AffectedDataContainers.Length;
```
a-ctor: Is this `virtual` intended? Is there a reason for insert/update/delete to have a different count than the affected data containers?

If we don't see a good use case right now I would remove the `virtual` for now. If we run into a problem later, we can either add reuse here or maybe go another way and don't need the `virtual` anyways.

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/BatchedObjectsDataContainerSaveCommandContext.cs
```diff
@@ -0,0 +1,68 @@
...
+        var ids = string.Join(", ", AffectedDataContainers.Take(10).Select(d => d.ID));
+        if (AffectedDataContainers.Length > 10)
+        {
+          throw new RdbmsProviderException($"Error while saving objects '{ids}' and {AffectedDataContainers.Length - 10} others. {e.Message}", e);
+        }
+
+        throw new RdbmsProviderException($"Error while saving objects '{ids}'. {e.Message}", e);
+      }
+
+      if (recordsAffected == 1)
```
a-ctor: This should check against `ExpectedAffectedRowCount`

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/BatchedLockRdbmsProviderCommand.cs
```diff
@@ -0,0 +1,74 @@
...
+        using (var reader = executionContext.ExecuteReader(command, CommandBehavior.Default))
+        {
+          var failedIDs = new HashSet<Guid>();
+          while (reader.Read())
+          {
+            failedIDs.Add(reader.GetGuid(0));
+          }
+
+          if (failedIDs.Count > 0)
+            throw CreateConcurrencyViolationException(failedIDs);
```
a-ctor, line 46: Do we want to display all data containers here or add a limit just like with the other exceptions?

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/IBatchedDataContainerSaveCommandContext.cs
```diff
@@ -5,20 +5,12 @@
 
 namespace Remotion.Data.DomainObjects.Persistence.Rdbms.StorageProviderCommands;
 
-public interface IMultiDataContainerSaveCommandContext
+public interface IBatchedDataContainerSaveCommandContext
```
a-ctor, line 8: Remove the `Context` prefix and make it a command?

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/BatchedDataContainerSaveCommand.cs
```diff
@@ -2,52 +2,35 @@
...
-  public BatchedDataContainerSaveCommand (IEnumerable<IMultiDataContainerSaveCommandContext> saveCommandContexts)
+  public BatchedDataContainerSaveCommand (IEnumerable<IBatchedDataContainerSaveCommandContext> saveCommandContexts)
   {
     ArgumentNullException.ThrowIfNull(saveCommandContexts);
 
     _saveCommandContexts = saveCommandContexts.ToArray();
   }
 
-  public IMultiDataContainerSaveCommandContext[] Contexts => _saveCommandContexts;
+  public IBatchedDataContainerSaveCommandContext[] Contexts => _saveCommandContexts;
```
a-ctor: See other comment about arrays. I think we can get through with `IReadOnlyList<...>` instead. I would also change the type of the backing field to match. We don't need the mutation an array provides.

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
```diff
@@ -154,24 +156,46 @@ protected virtual IEnumerable<ColumnValue> GetColumnValuesForPropertyValue (Data
...
+    var lockCommandSpecifications = new List<IBatchedLockCommandSpecification>();
 
     foreach (var kvp in groupedDataContainers)
     {
       var tableDefinition = kvp.Key;
 
-      inserts.AddRange(CreateCommandContextsForInsert(tableDefinition, kvp.Value.ForInsert));
-      updates.AddRange(CreateCommandContextsForUpdate(tableDefinition, kvp.Value.ForUpdate));
-      deletes.AddRange(CreateCommandContextsForDelete(tableDefinition, kvp.Value.ForDelete));
+      var dataContainersForLock = kvp.Value.ForUpdate.Select(u => u.DataContainer).Union(kvp.Value.ForDelete).Where(d=>d.Timestamp != null).ToArray();
```
a-ctor: Multiline for readability?


```suggestion
      var dataContainersForLock = kvp.Value.ForUpdate
          .Select(u => u.DataContainer)
          .Union(kvp.Value.ForDelete)
          .Where(d => d.Timestamp != null)
          .ToArray();
```

Also, I think we should use `d.State.IsNew` instead of the timestamp check if the intent here is to only include existing data container.

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
```diff
@@ -154,24 +156,46 @@ protected virtual IEnumerable<ColumnValue> GetColumnValuesForPropertyValue (Data
...
+      }
+
+      if (kvp.Value.ForInsert.Count > 0)
+        inserts.AddRange(CreateCommandContextsForInsert(tableDefinition, kvp.Value.ForInsert));
+
+      if (kvp.Value.ForUpdate.Count > 0)
+        updates.AddRange(CreateCommandContextsForUpdate(tableDefinition, kvp.Value.ForUpdate));
+
+      if (kvp.Value.ForDelete.Count > 0)
+        deletes.AddRange(CreateCommandContextsForDelete(tableDefinition, kvp.Value.ForDelete));
```
a-ctor: We access `kvp.Value` quite a lot here. Extract it into a variable just like we did for `kvp.Key`?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilderTest.cs
```diff
@@ -0,0 +1,142 @@
...
+    _dbCommandStub.Protected().Setup<DbParameter>("CreateDbParameter").Returns(_dbDataParameterStub.Object);
+    _dbCommandStub.Protected().Setup<DbParameterCollection>("DbParameterCollection").Returns(_dataParameterCollectionMock.Object);
+    _dbCommandStub.SetupProperty(stub => stub.CommandText);
+
+    _dbCommandFactoryStub = new Mock<IDbCommandFactory>();
+    _dbCommandFactoryStub.Setup(stub => stub.CreateDbCommand()).Returns(_dbCommandStub.Object);
+  }
+
+  [Test]
+  public void CreateWith_OneTable_DefaultSchema ()
```
a-ctor: The construction of the test name is usually `<TestedMethodName>_<Condition>_<Result>` (e.g. `Add_WithSmallNumbers_ReturnsCorrectNumber`). Sometimes the `<Result>` part is not really necessary as adding it would not add much value. In such cases, I would leave it out.

Here I would use `Create_WithOneTableAndDefaultSchema` instead, leaving out the `<Result>` part as it would be something like `ReturnsCorrectCommand` which does not add much value.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilderTest.cs
```diff
@@ -0,0 +1,142 @@
...
+
+    var builder = new BatchedLockDbCommandBuilder(_sqlDialectStub.Object, [commandSpecification.Object]);
+    var result = builder.Create(_dbCommandFactoryStub.Object);
+    Assert.That(
+        result.CommandText,
+        Is.EqualTo(
+            "DECLARE @TransactionIsolationLevel int;\r\nDECLARE @IsReadCommittedSnapshotOn bit;\r\nSET @TransactionIsolationLevel = (SELECT [transaction_isolation_level] FROM [sys].[dm_exec_sessions] WHERE [session_id] = @@SPID);\r\nSET @IsReadCommittedSnapshotOn = (SELECT [is_read_committed_snapshot_on] FROM [sys].[databases] WHERE [database_id] = DB_ID());\r\nIF (@TransactionIsolationLevel = 2 AND @IsReadCommittedSnapshotOn = 1)\r\nBEGIN\r\nSELECT [P].[col1], [P].[col2] FROM [Table] [T] WITH(ROWLOCK, XLOCK, READPAST)\r\nRIGHT JOIN @TVP_Lock_Table [P] ON [P].[col1] = [T].[col1] AND [P].[col2] = [T].[col2]\r\nWHERE [T].[ID] IS NULL;\r\nEND\r\nELSE\r\nBEGIN\r\nSELECT [P].[col1], [P].[col2] FROM [Table] [T] WITH(ROWLOCK, XLOCK)\r\nRIGHT JOIN @TVP_Lock_Table [P] ON [P].[col1] = [T].[col1] AND [P].[col2] = [T].[col2]\r\nWHERE [T].[ID] IS NULL;\r\nEND"));
+  }
```
a-ctor, line 81: Raw string literal for readability? (in the other tests as well)

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/MultiClassTableValuedDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,110 @@
...
+    public void GetParameterValue_ThrowsException_WhenMultipleTablesFound ()
+    {
+      var dataContainerA = DataContainer.CreateForExisting(new ObjectID("Computer", new Guid("3F647D79-0CAF-4a53-BAA7-A56831F8CE2D")), new byte[8], pd => pd.DefaultValue);
+      var dataContainerB = DataContainer.CreateForExisting(new ObjectID("Order", new Guid("5D35FFBC-3EBE-4DF1-9863-6163EE6F2486")), new byte[8], pd => pd.DefaultValue);
+
+      Assert.That(dataContainerA.ClassDefinition.StorageEntityDefinition, Is.Not.EqualTo(dataContainerB.ClassDefinition.StorageEntityDefinition));
+      var parameterDefinition = new MultiClassTableValuedDataParameterDefinition(_tableManipulationRecordDefinitionProvider,
+          (provider, definition) => provider.GetInsertRecordDefinition(definition));
+
+      Assert.That(() => parameterDefinition.GetParameterValue(new List<DataContainer>() { dataContainerA, dataContainerB }), Throws.InvalidOperationException);
```
a-ctor: We assert the exception messages, especially when we are throwing the exception. 

```suggestion
      Assert.That(
          () => parameterDefinition.GetParameterValue(new List<DataContainer>() { dataContainerA, dataContainerB }),
      Throws.InvalidOperationException
      .With.Message.EqualTo("..."));
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/Parameters/MultiClassTableValuedDataParameterDefinitionTest.cs
```diff
@@ -0,0 +1,110 @@
...
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model.Building;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Model.Building;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SqlServer.Parameters;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SqlServer.Parameters
+{
+  [TestFixture]
+  public class MultiClassTableValuedDataParameterDefinitionTest
```
a-ctor, line 24: If I am not mistaken, we don't have a test that ensures the return value of `GetParameterValue` makes sense. I think we should add a test that tests the grouping and correct parameter creation.

a-ctor (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
```diff
@@ -83,25 +85,23 @@ private string CreateStatement (StringBuilder forReadCommit, StringBuilder forNo
...
-        .AppendJoin(" AND ", columns.Select(c => parameterAlias + "." + c + " = " + tableAlias + "." + c))
-        .AppendLine()
-        .Append("WHERE ")
-        .Append(tableAlias)
-        .Append(".[ID] IS NULL");
+    var selectColumns = new StringBuilder();
+    selectColumns.AppendJoin(", ", columns.Select(c => $"{parameterAlias}.{c}"));
+
+    var joinCondition = new StringBuilder();
+    joinCondition.AppendJoin(" AND ", columns.Select(c => $"{parameterAlias}.{c} = {tableAlias}.{c}"));
```
a-ctor: In these cases I would use `string.Join(...)` instead of constructing our own `StringBuilder` instances. They do some stack magic to prevent allocations for small string and will reuse the underlying buffer for the string concat.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
```diff
@@ -83,25 +85,23 @@ private string CreateStatement (StringBuilder forReadCommit, StringBuilder forNo
...
+
+    var tableHints = new StringBuilder(24);
+    tableHints.Append("ROWLOCK, XLOCK");
+    if (forReadCommittedIsolation)
+      tableHints.Append(", READPAST");
+
```
a-ctor:
```cs
var tableHints = forReadCommittedIsolation
    ? "ROWLOCK, XLOCK, READPAST"
    : "ROWLOCK, XLOCK";
```

This way, we don't have any allocations at all.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/SchemaGeneration/SqlTableTypeScriptElementFactory.cs
```diff
@@ -85,10 +85,13 @@ public IScriptElement GetDropElement (TableTypeDefinition tableTypeDefinition)
 
   private string GetColumnDeclaration (ColumnDefinition column)
   {
-    if (column.StorageTypeInfo.StorageTypeName == "rowversion")
-      return $"[{column.Name}] varbinary({column.StorageTypeInfo.StorageTypeLength ?? 8}) {(column.StorageTypeInfo.IsStorageTypeNullable ? "NULL" : "NOT NULL")}";
+    // Columns of type rowversion cannot be set directly not even in table types which are used for table valued parameters.
+    // Therefore, we need to use varbinary instead of rowversion.
+    var storageTypeName = column.StorageTypeInfo.StorageTypeName == "rowversion"
+    ? $"varbinary({column.StorageTypeInfo.StorageTypeLength ?? 8})"
+    : column.StorageTypeInfo.StorageTypeName;
```
a-ctor: please fix the indent here. `?` and `:` should have double indent (4 spaces)

a-ctor (approved):

PR #1407, opened 2026-03-05
feature/RM-9651-add-data-csp-replaced-attribute-for-css-rules

noah-rubicon:
https://jira.rubicon.eu/browse/RM-9651

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
+++ b/Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
@@ -452,7 +452,8 @@ label.bocListContent
   padding: 2px;
 }
 
-div.bocListContent > div[onclick]
+div.bocListContent > div[onclick],
+div.bocListContent > div[data-csp-replaced-onclick]
 {
   display: inline;
 }
--- a/Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
+++ b/Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
@@ -1,4 +1,4 @@
-﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+﻿﻿﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
 // Copyright (c) rubicon IT GmbH, www.rubicon.eu
 // 
 // The re-motion Core Framework is free software; you can redistribute it
@@ -317,6 +317,8 @@ namespace Remotion.Web.ContentSecurityPolicy
       {
         if (registeredEvent.Key == "href")
           base.WriteAttribute("href", _fallbackNavigationUrlProvider.GetURL());
+        else
+          WriteAttribute($"data-csp-replaced-{registeredEvent.Key}", string.Empty);
 
         var eventTargetID = _lastInlineEventTargetId;
         if (eventTargetID == null)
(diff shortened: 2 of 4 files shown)


Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocList.css
```diff
@@ -452,7 +452,7 @@ label.bocListContent
   padding: 2px;
 }
 
-div.bocListContent > div[onclick]
+div.bocListContent > div[data-csp-replaced-onclick]
```
a-ctor, line 456: CSP support is opt-in so we need to support both the `onclick` attribute and its transformed CSP-variant. (same thing in the other CSS file)

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriterTest.cs
```diff
@@ -124,7 +124,7 @@ public void RegisterSupportedEvent_NewEventRegistration ()
               <root>
                 <div onNewAttribute="value">
 
-                </div><div data-inline-event-target="eventTargetID">
+                </div><div data-csp-replaced-onNewAttribute="" data-inline-event-target="eventTargetID">
```
a-ctor, line 127: Please check the other tests for additional cases where we need to assert the new HTML attribute. There are a few "attribute exists" and also "attribute does not exist" checks. Most likely, all tests that assert HTML need a change

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -317,6 +317,8 @@ public override void WriteAttribute (string name, string value, bool fEncode)
       {
         if (registeredEvent.Key == "href")
           base.WriteAttribute("href", _fallbackNavigationUrlProvider.GetURL());
+        else
+          base.AddAttribute($"data-csp-replaced-{registeredEvent.Key}", string.Empty);
```
a-ctor: Further up in `RenderBeginTag` we have to add the attribute as well.

a-ctor (changes requested):

Remotion/Web/Core/ContentSecurityPolicy/CspEnabledHtmlTextWriter.cs
```diff
@@ -245,6 +245,9 @@ public override void RenderBeginTag (HtmlTextWriterTag tagKey)
           _scriptManager.RegisterScript($"{eventTargetID}-{registeredEvent.Key}", script);
           if (_renderingFeatures.EnableDiagnosticMetadata)
             base.AddAttribute("data-event-content-" + registeredEvent.Key, registeredEvent.OriginalValue);
+
+          if (registeredEvent.Key.StartsWith("on", StringComparison.OrdinalIgnoreCase))
+              AddAttribute($"data-csp-replaced-{registeredEvent.Key}", string.Empty);
         }
```
a-ctor, line 249: Sorry, I think I misinterpreted the last diff and though we did not handle the `RenderBeginTag`. I think it is better to add the `data-csp-replaced-xxx` attribute in `TryAddAttributeWithoutEncoding`. This way, we don't have to repeat the `on` check here and the registered events dictionary work correctly again.

a-ctor (approved):
Looking through the tests I found a few places where I think we should add more asserts. Since, found a few places I fixed them and my other comment as well. Please check if the diff makes sense to you.

Before merging, please separate the `CspTextWriter` changes from the CSS changes. Each one gets its own commit. Please also make sure that the commits have a description in line 1. You can add more details in the following lines but the first line should have enough information on its own.

PR #1413, opened 2026-03-12
RM-9647 added sorting of sql commands

Mario185:
https://jira.rubicon.eu/browse/RM-9647

### DIFF
--- a/Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
@@ -25,6 +25,7 @@ using Remotion.Data.DomainObjects.Mapping;
 using Remotion.Data.DomainObjects.Persistence.Configuration;
 using Remotion.Data.DomainObjects.Persistence.NonPersistent;
 using Remotion.Data.DomainObjects.Persistence.Rdbms;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
 using Remotion.Data.DomainObjects.UnitTests.Linq.IntegrationTests;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain;
 using Remotion.Data.DomainObjects.UnitTests.TestDomain.TableInheritance;
@@ -81,7 +82,8 @@ namespace Remotion.Data.DomainObjects.UnitTests.Factories
                                     "Remotion.Data.DomainObjects.UnitTests.TestDomain",
                                     "Remotion.Data.DomainObjects.UnitTests.DataManagement.TestDomain",
                                     "Remotion.Data.DomainObjects.UnitTests.MixedDomains.TestDomain",
-                                    "Remotion.Data.DomainObjects.UnitTests.Linq.TestDomain"
+                                    "Remotion.Data.DomainObjects.UnitTests.Linq.TestDomain",
+                                    "Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization.TestDomain"
                                   };
       return new FilteringTypeDiscoveryService(
           typeDiscoveryService,
@@ -110,7 +112,8 @@ namespace Remotion.Data.DomainObjects.UnitTests.Factories
 
       _mappingConfiguration = MappingConfiguration.Create(
             MappingReflectorObjectMother.CreateMappingReflector(typeDiscoveryService),
-            new PersistenceModelLoader(_storageSettings));
+            new PersistenceModelLoader(_storageSettings),
+            SafeServiceLocator.Current.GetInstance<ISortingOptimizationNodeFactory>());
     }
 
     private FakeStorageSettingsFactory CreateStorageSettings ()
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProviderTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProviderTest.cs
@@ -1,0 +1,193 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Moq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+using Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization.TestDomain;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization;
+
+[TestFixture]
+public class GraphBasedPersistenceModelSortingProviderTest : SortingOptimizationTestBase
+{
+  private Mock<ISortingOptimizationNodeFactory> _nodeFactoryStrictMock;
+  private ISortingOptimizationNodeFactory _realNodeFactory;
+  private MappingConfiguration _mappingConfiguration;
+
+  [SetUp]
+  public void SetUp ()
+  {
+    _realNodeFactory = new SortingOptimizationNodeFactory();
+    _nodeFactoryStrictMock = new Mock<ISortingOptimizationNodeFactory>(MockBehavior.Strict);
+    _mappingConfiguration = GetMappingConfiguration();
+  }
+
+  [Test]
+  public void CreatesCorrectOrder_WithoutAnyBreaks ()
+  {
+    var objectA = _mappingConfiguration.GetTypeDefinition(typeof(NoBreaksObjectA));
+    var objectB = _mappingConfiguration.GetTypeDefinition(typeof(NoBreaksObjectB));
+    var objectC = _mappingConfiguration.GetTypeDefinition(typeof(NoBreaksObjectC));
+
+    var provider = new GraphBasedPersistenceModelSortingProvider(_realNodeFactory);
+    provider.Initialize([objectC, objectB, objectA]);
+
+    AssertPositionAndCorrectOrder(
+        provider,
+        [
+            (objectA, true),
+            (objectC, true),
+            (objectB, true)
+        ]
+    );
+
+    AssertForeignKeyProperties(provider, objectA, []);
+    AssertForeignKeyProperties(provider, objectB, [nameof(NoBreaksObjectB.NoBreakBPropA), nameof(NoBreaksObjectB.NoBreakBPropC)]);
+    AssertForeignKeyProperties(provider, objectC, [nameof(NoBreaksObjectC.NoBreakCPropA)]);
+  }
+
+  [Test]
+  public void CreatesCorrectOrder_WithoutAutomaticBreaks_BreakIsDeterminedByTableName ()
+  {
+    var objectA = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksTableNameObjectA));
+    var objectB = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksTableNameObjectB));
+    var objectC = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksTableNameObjectC));
+
+    var nodes = _realNodeFactory.CreateNodes([objectB, objectA, objectC]);
+
+    var nodeForA = GetNodeForClassDefinition(objectA, nodes);
+    var nodeForB = GetNodeForClassDefinition(objectB, nodes);
+    var nodeForC = GetNodeForClassDefinition(objectC, nodes);
+    var expectedBrokenEdge = GetEdgeForProperty(nameof(AutomaticBreaksTableNameObjectA.AutomaticBreakTableNameBPropB), nodeForA);
+
+    SetupNodeFactoryMock(nodes);
+
+    Assert.That(nodeForA.HasBrokenEdges, Is.False);
+    Assert.That(nodeForB.HasBrokenEdges, Is.False);
+    Assert.That(nodeForC.HasBrokenEdges, Is.False);
+
+    var provider = new GraphBasedPersistenceModelSortingProvider(_nodeFactoryStrictMock.Object);
+    provider.Initialize([]);
+
+    Assert.That(nodeForA.HasBrokenEdges, Is.True);
+    Assert.That(nodeForA.BrokenEdges.Count, Is.EqualTo(1));
+    Assert.That(nodeForA.BrokenEdges.Contains(expectedBrokenEdge), Is.True);
+
+    Assert.That(nodeForB.HasBrokenEdges, Is.False);
+    Assert.That(nodeForC.HasBrokenEdges, Is.False);
+
+    AssertPositionAndCorrectOrder(
+        provider,
+        [
+            (objectA, false), // A is broken therefore first
+            (objectC, true), // C has no other dependencies than A and A is broken therefore second
+            (objectB, true) // B has dependency on C therefore third
+        ]
+    );
+
+    AssertForeignKeyProperties(provider, objectA, [nameof(AutomaticBreaksTableNameObjectA.AutomaticBreakTableNameBPropB)]);
+    AssertForeignKeyProperties(provider, objectB, [nameof(AutomaticBreaksTableNameObjectB.AutomaticBreakTableNameBPropC)]);
+    AssertForeignKeyProperties(provider, objectC, [nameof(AutomaticBreaksTableNameObjectC.AutomaticBreakTableNameCPropA)]);
+  }
+
+  [Test]
+  public void CreatesCorrectOrder_WithoutAutomaticBreaks_BreakIsDeterminedByCountOfOccurence ()
+  {
+    var objectZ = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksCountObjectZ));
+    var objectA = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksCountObjectA));
+    var objectB = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksCountObjectB));
+    var objectC = _mappingConfiguration.GetTypeDefinition(typeof(AutomaticBreaksCountObjectC));
+
+    var nodes = _realNodeFactory.CreateNodes([objectZ, objectC, objectA, objectB]);
+
+    foreach(var node in nodes)
+      node.CalculateIndirectCyclicDependencies();
+
+    var nodeForZ = GetNodeForClassDefinition(objectZ, nodes);
+    var nodeForA = GetNodeForClassDefinition(objectA, nodes);
+    var nodeForB = GetNodeForClassDefinition(objectB, nodes);
+    var nodeForC = GetNodeForClassDefinition(objectC, nodes);
+
+    var expectedBrokenEdge1 = GetEdgeForProperty(nameof(AutomaticBreaksCountObjectZ.AutomaticBreakCountZPropA), nodeForZ);
+    var expectedBrokenEdge2 = GetEdgeForProperty(nameof(AutomaticBreaksCountObjectZ.AutomaticBreakCountTPropB), nodeForZ);
+
+    SetupNodeFactoryMock(nodes);
+
+    Assert.That(nodeForA.HasBrokenEdges, Is.False);
+    Assert.That(nodeForB.HasBrokenEdges, Is.False);
+    Assert.That(nodeForC.HasBrokenEdges, Is.False);
+    Assert.That(nodeForZ.HasBrokenEdges, Is.False);
+
+    var provider = new GraphBasedPersistenceModelSortingProvider(_nodeFactoryStrictMock.Object);
+    provider.Initialize([]);
+
+    Assert.That(nodeForA.HasBrokenEdges, Is.False);
+    Assert.That(nodeForB.HasBrokenEdges, Is.False);
+    Assert.That(nodeForC.HasBrokenEdges, Is.False);
+
+    Assert.That(nodeForZ.HasBrokenEdges, Is.True);
+    Assert.That(nodeForZ.BrokenEdges.Count, Is.EqualTo(2));
+    Assert.That(nodeForZ.BrokenEdges.Contains(expectedBrokenEdge1), Is.True);
+    Assert.That(nodeForZ.BrokenEdges.Contains(expectedBrokenEdge2), Is.True);
+
+    AssertPositionAndCorrectOrder(
+        provider,
+        [
+            (objectZ, false), // Z is broken therefore first
+            (objectA, true),  // A and C has no other dependencies than Z and then they are ordered by their table name
+            (objectC, true),  // therefore A is second and C is third
+            (objectB, true)   // B has dependency on C therefore third
+        ]
+    );
+
+    AssertForeignKeyProperties(provider, objectA, [nameof(AutomaticBreaksCountObjectA.AutomaticBreakCountAPropZ)]);
+    AssertForeignKeyProperties(provider, objectB, [nameof(AutomaticBreaksCountObjectB.AutomaticBreakCountBPropC)]);
+    AssertForeignKeyProperties(provider, objectC, [nameof(AutomaticBreaksCountObjectC.AutomaticBreakCountCPropZ)]);
+    AssertForeignKeyProperties(provider, objectZ, [nameof(AutomaticBreaksCountObjectZ.AutomaticBreakCountZPropA), nameof(AutomaticBreaksCountObjectZ.AutomaticBreakCountTPropB)]);
+  }
+
+  private void AssertPositionAndCorrectOrder (GraphBasedPersistenceModelSortingProvider provider, params (ClassDefinition classDefinition, bool isOrderedCorrect)[] expectedValues)
+  {
+    for (var i = 0; i < expectedValues.Length; i++)
+    {
+      var expectedValue = expectedValues[i];
+      var tableDefinition = GetTableDefinition(expectedValue.classDefinition);
+      Assert.That(provider.GetSortPosition(tableDefinition), Is.EqualTo(i));
+      Assert.That(provider.HasBeenSortedCorrectly(tableDefinition), Is.EqualTo(expectedValue.isOrderedCorrect));
+    }
+  }
+
+  private void AssertForeignKeyProperties (GraphBasedPersistenceModelSortingProvider provider, ClassDefinition classDefinition, string[] expectedProps)
+  {
+    var actualProps = provider.GetForeignKeyRelevantPropertyDefinitions(classDefinition);
+    Assert.That(actualProps.Count, Is.EqualTo(expectedProps.Length));
+
+    var expectedPropertyDefinitions = expectedProps.Select(e => GetPropertyDefinition(classDefinition, e));
+    Assert.That(actualProps, Is.EquivalentTo(expectedPropertyDefinitions));
+  }
+
+  private SortingOptimizationNode GetNodeForClassDefinition (ClassDefinition classDefinition, IReadOnlyList<SortingOptimizationNode> nodes)
+  {
+    var tableDefinition = GetTableDefinition(classDefinition);
+    return nodes.Single(n => n.TableDefinition == tableDefinition);
+  }
+
+  private SortingOptimizationEdge GetEdgeForProperty (string propertyName, SortingOptimizationNode node)
+  {
+    return node.Edges.Single(e => e.ForeignKeys.Any(f => f.ReferencingColumns.Any(c => c.Name == propertyName + "ID")));
+  }
+
+  private PropertyDefinition GetPropertyDefinition (ClassDefinition classDefinition, string propertyName)
+  {
+    return classDefinition.GetPropertyDefinitions().First(p => p.PropertyName.EndsWith("." + propertyName));
+  }
+
+  private void SetupNodeFactoryMock (IReadOnlyList<SortingOptimizationNode> nodes)
+  {
+    _nodeFactoryStrictMock.Setup(stub => stub.CreateNodes(It.IsAny<IReadOnlyList<ClassDefinition>>())).Returns(nodes);
+  }
+}
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationEdgeTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationEdgeTest.cs
@@ -1,0 +1,183 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Linq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.Model;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization;
+
+[TestFixture]
+public class SortingOptimizationEdgeTest
+{
+  private UnitTestStorageProviderStubDefinition _storageProviderDefinition;
+  private TableDefinition _tableA;
+  private TableDefinition _tableB;
+  private SortingOptimizationNode _nodeA;
+  private SortingOptimizationNode _nodeB;
+  private ForeignKeyConstraintDefinition _foreignKey;
+
+  [SetUp]
+  public void SetUp ()
+  {
+    _storageProviderDefinition = new UnitTestStorageProviderStubDefinition("SPID");
+    _tableA = TableDefinitionObjectMother.Create(_storageProviderDefinition, new EntityNameDefinition(null, "TableA"));
+    _tableB = TableDefinitionObjectMother.Create(_storageProviderDefinition, new EntityNameDefinition(null, "TableB"));
+    _nodeA = new SortingOptimizationNode(_tableA, []);
+    _nodeB = new SortingOptimizationNode(_tableB, []);
+    _foreignKey = CreateForeignKey("FK_A_To_B", _tableB.TableName);
+  }
+
+  [Test]
+  public void Initialize_SetsProperties ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+
+    Assert.That(edge.Owner, Is.SameAs(_nodeA));
+    Assert.That(edge.PointingTo, Is.SameAs(_nodeB));
+    Assert.That(edge.ForeignKeys, Is.EquivalentTo([_foreignKey]));
+    Assert.That(edge.IsSelfCyclingEdge, Is.False);
+  }
+
+  [Test]
+  public void Initialize_OwnerEqualsPointingTo_IsSelfCyclingEdgeIsTrue ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeA);
+
+    Assert.That(edge.Owner, Is.SameAs(_nodeA));
+    Assert.That(edge.PointingTo, Is.SameAs(_nodeA));
+    Assert.That(edge.IsSelfCyclingEdge, Is.True);
+  }
+
+  [Test]
+  public void Initialize_OwnerDifferentFromPointingTo_IsSelfCyclingEdgeIsFalse ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+
+    Assert.That(edge.IsSelfCyclingEdge, Is.False);
+  }
+
+  [Test]
+  public void Initialize_ForeignKeysCollectionContainsConstructorArgument ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+
+    Assert.That(edge.ForeignKeys.Count, Is.EqualTo(1));
+    Assert.That(edge.ForeignKeys.Single(), Is.SameAs(_foreignKey));
+  }
+
+  [Test]
+  public void Initialize_ForeignKeyNull_ThrowsArgumentNullException ()
+  {
+    Assert.That(
+        () => new SortingOptimizationEdge(null!, _nodeA, _nodeB),
+        Throws.TypeOf<ArgumentNullException>());
+  }
+
+  [Test]
+  public void Initialize_OwnerNull_ThrowsArgumentNullException ()
+  {
+    Assert.That(
+        () => new SortingOptimizationEdge(_foreignKey, null!, _nodeB),
+        Throws.TypeOf<ArgumentNullException>());
+  }
+
+  [Test]
+  public void Initialize_PointingToNull_ThrowsArgumentNullException ()
+  {
+    Assert.That(
+        () => new SortingOptimizationEdge(_foreignKey, _nodeA, null!),
+        Throws.TypeOf<ArgumentNullException>());
+  }
+
+  [Test]
+  public void BreakEdge_DelegatesToOwner_MovingEdgeToBrokenEdges ()
+  {
+    // Use the node's AddEdge so the edge is tracked by its owner; otherwise
+    // BreakEdge on the owner would not be able to remove it.
+    _nodeA.AddEdge(_foreignKey, _nodeB);
+    var edge = _nodeA.Edges.Single();
+
+    edge.BreakEdge();
+
+    Assert.That(_nodeA.Edges, Is.Empty);
+    Assert.That(_nodeA.BrokenEdges, Is.EquivalentTo([edge]));
+  }
+
+  [Test]
+  public void BreakEdge_DoesNotChangeOwnerOrPointingToOrForeignKeys ()
+  {
+    _nodeA.AddEdge(_foreignKey, _nodeB);
+    var edge = _nodeA.Edges.Single();
+
+    edge.BreakEdge();
+
+    Assert.That(edge.Owner, Is.SameAs(_nodeA));
+    Assert.That(edge.PointingTo, Is.SameAs(_nodeB));
+    Assert.That(edge.ForeignKeys, Is.EquivalentTo([_foreignKey]));
+  }
+
+  [Test]
+  public void AddForeignKey_AppendsForeignKey ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+    var additionalForeignKey = CreateForeignKey("FK_A_To_B_Additional", _tableB.TableName);
+
+    edge.AddForeignKey(additionalForeignKey);
+
+    Assert.That(edge.ForeignKeys.Count, Is.EqualTo(2));
+    Assert.That(edge.ForeignKeys, Is.EquivalentTo([_foreignKey, additionalForeignKey]));
+  }
+
+  [Test]
+  public void AddForeignKey_MultipleCalls_AppendsAllInOrder ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+    var fk2 = CreateForeignKey("FK_2", _tableB.TableName);
+    var fk3 = CreateForeignKey("FK_3", _tableB.TableName);
+
+    edge.AddForeignKey(fk2);
+    edge.AddForeignKey(fk3);
+
+    Assert.That(edge.ForeignKeys, Is.EquivalentTo([_foreignKey, fk2, fk3]));
+  }
+
+  [Test]
+  public void AddForeignKey_SameInstanceTwice_StoresBothEntries ()
+  {
+    // The implementation does not de-duplicate — calling AddForeignKey twice
+    // with the same instance results in two list entries.
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+    var duplicate = CreateForeignKey("FK_Duplicate", _tableB.TableName);
+
+    edge.AddForeignKey(duplicate);
+    edge.AddForeignKey(duplicate);
+
+    Assert.That(edge.ForeignKeys.Count, Is.EqualTo(3));
+    Assert.That(edge.ForeignKeys, Is.EquivalentTo([_foreignKey, duplicate, duplicate]));
+  }
+
+  [Test]
+  public void AddForeignKey_Null_ThrowsArgumentNullException ()
+  {
+    var edge = new SortingOptimizationEdge(_foreignKey, _nodeA, _nodeB);
+
+    Assert.That(
+        () => edge.AddForeignKey(null!),
+        Throws.TypeOf<ArgumentNullException>());
+  }
+
+  private static ForeignKeyConstraintDefinition CreateForeignKey (string name, EntityNameDefinition referencedTableName)
+  {
+    var referencingColumn = ColumnDefinitionObjectMother.CreateColumn(name + "_Referencing");
+    var referencedColumn = ColumnDefinitionObjectMother.CreateColumn(name + "_Referenced");
+    return new ForeignKeyConstraintDefinition(
+        name,
+        referencedTableName,
+        [referencingColumn],
+        [referencedColumn]);
+  }
+}
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactoryTest.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactoryTest.cs
@@ -1,0 +1,77 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.Linq;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+using Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization.TestDomain;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization;
+
+[TestFixture]
+public class SortingOptimizationNodeFactoryTest : SortingOptimizationTestBase
+{
+  private SortingOptimizationNodeFactory _nodeFactory;
+  private MappingConfiguration _mappingConfiguration;
+
+  [SetUp]
+  public void SetUp ()
+  {
+    _nodeFactory = new SortingOptimizationNodeFactory();
+    _mappingConfiguration = GetMappingConfiguration();
+  }
+
+  [Test]
+  public void CreateNodes_ResultIsValid ()
+  {
+    var objectA = _mappingConfiguration.GetTypeDefinition(typeof(NodeFactoryObjectA));
+    var objectB = _mappingConfiguration.GetTypeDefinition(typeof(NodeFactoryObjectB));
+    var objectC = _mappingConfiguration.GetTypeDefinition(typeof(NodeFactoryObjectC));
+    var objectD = _mappingConfiguration.GetTypeDefinition(typeof(NodeFactoryObjectDWithNoTable));
+
+    var nodes = _nodeFactory.CreateNodes([objectA, objectB, objectC, objectD]);
+
+    Assert.That(nodes.Count, Is.EqualTo(3));
+    var aNode = nodes[0];
+    var bNode = nodes[1];
+    var cNode = nodes[2];
+
+    Assert.That(aNode.TableDefinition, Is.EqualTo(GetTableDefinition(objectA)));
+    Assert.That(aNode.Edges.Count, Is.EqualTo(1));
+    AssertEdge(aNode.Edges[0], ["FK_NodeFactoryObjectA_SelfCyclingPropID"], aNode, true);
+
+    Assert.That(bNode.TableDefinition, Is.EqualTo(GetTableDefinition(objectB)));
+    Assert.That(bNode.TableDefinition, Is.EqualTo(GetTableDefinition(objectD)));
+    Assert.That(bNode.Edges.Count, Is.EqualTo(2));
+    AssertEdge(bNode.Edges[0], ["FK_NodeFactoryObjectB_NodeFactoryBPropAID", "FK_NodeFactoryObjectB_NodeFactoryDPropAID"], aNode, false);
+    AssertEdge(bNode.Edges[1], ["FK_NodeFactoryObjectB_NodeFactoryDPropCID"], cNode, false);
+
+    Assert.That(cNode.TableDefinition, Is.EqualTo(GetTableDefinition(objectC)));
+    Assert.That(cNode.Edges.Count, Is.EqualTo(2));
+
+    AssertEdge(cNode.Edges[0], ["FK_NodeFactoryObjectC_NodeFactoryCPropAID"], aNode, false);
+    AssertEdge(cNode.Edges[1], ["FK_NodeFactoryObjectC_NodeFactoryCPropBID"], bNode, false);
+  }
+
+  [Test]
+  public void CreateNodes_WithMissingLink_ThrowsException ()
+  {
+    var objectA = _mappingConfiguration.GetTypeDefinition(typeof(NodeFactoryObjectA));
+    var objectC = _mappingConfiguration.GetTypeDefinition(typeof(NodeFactoryObjectC));
+
+    Assert.That(
+        () => _nodeFactory.CreateNodes([objectA, objectC]),
+        Throws.InvalidOperationException.With.Message.EqualTo(
+            "Could not find SortingOptimizationNode 'NodeFactoryObjectB' for foreign key 'FK_NodeFactoryObjectC_NodeFactoryCPropBID' on node 'NodeFactoryObjectC'."));
+  }
+
+  private void AssertEdge (SortingOptimizationEdge edgeToTest, string[] foreignKeyNames, SortingOptimizationNode pointingTo, bool isSelfCyclic)
+  {
+    Assert.That(edgeToTest.IsSelfCyclingEdge, Is.EqualTo(isSelfCyclic));
+    Assert.That(edgeToTest.PointingTo, Is.EqualTo(pointingTo));
+
+    Assert.That(edgeToTest.ForeignKeys.Count, Is.EqualTo(foreignKeyNames.Length));
+    Assert.That(edgeToTest.ForeignKeys.Select(k => k.ConstraintName), Is.EquivalentTo(foreignKeyNames));
+  }
+}
--- a/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationTestBase.cs
+++ b/Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationTestBase.cs
@@ -1,0 +1,63 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+using System.ComponentModel.Design;
+using System.Linq;
+using System.Reflection;
+using NUnit.Framework;
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Data.DomainObjects.Persistence.Configuration;
+using Remotion.Data.DomainObjects.Persistence.Rdbms;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+using Remotion.Data.DomainObjects.UnitTests.Factories;
+using Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization.TestDomain;
+using Remotion.Development.UnitTesting.Reflection.TypeDiscovery;
+using Remotion.Reflection.TypeDiscovery;
+using Remotion.Reflection.TypeDiscovery.AssemblyFinding;
+using Remotion.Reflection.TypeDiscovery.AssemblyLoading;
+using Remotion.ServiceLocation;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.SortingOptimization;
+
+public abstract class SortingOptimizationTestBase
+{
+  protected MappingConfiguration GetMappingConfiguration ()
+  {
+    var storageSettings = SafeServiceLocator.Current.GetInstance<IStorageSettings>();
+    var typeDiscoveryService = GetTypeDiscoveryService(GetType().Assembly);
+    var mappingConfiguration = MappingConfiguration.Create(
+        MappingReflectorObjectMother.CreateMappingReflector(typeDiscoveryService),
+        new PersistenceModelLoader(storageSettings),
+        SafeServiceLocator.Current.GetInstance<ISortingOptimizationNodeFactory>());
+    return mappingConfiguration;
+  }
+
+  protected TableDefinition GetTableDefinition (ClassDefinition classDefinition)
+  {
+    return InlineRdbmsStorageEntityDefinitionVisitor.Visit<TableDefinition>(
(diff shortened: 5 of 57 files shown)


Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
```diff
@@ -182,65 +196,53 @@ protected virtual bool ShouldCreateUpdateCommand (DataContainer dataContainer)
...
 
-    var allDataContainersForUpdate = new List<DataContainer>();
-    var updateCommandSpecifications = new List<IBatchedCommandSpecification>();
+    var allDataContainersForUpdate = new List<DataContainer>(dataContainers.Count);
+    var updateCommandSpecifications = new List<IBatchedCommandSpecification>(groupedDataContainers.Count);
 
-    var allDataContainersForDelete = new List<DataContainer>();
-    var deleteCommandSpecifications = new List<IBatchedCommandSpecification>();
+    var allDataContainersForDelete = new List<DataContainer>(dataContainers.Count);
+    var deleteCommandSpecifications = new List<IBatchedCommandSpecification>(groupedDataContainers.Count);
```
a-ctor: Is there a reason why we are adding a capacity here?

Given that insert/delete/update are pretty exclusive we expect much less elements in each category then the provided capacity (for the data container case). It probably does not matter much but in the worst case we allocate 4x as much memory as we might need.

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
```diff
@@ -269,34 +271,95 @@ private IEnumerable<IRdbmsProviderCommand> CreateCommands (IEnumerable<DataConta
...
+        if (IsPrecedingUpdateRequiredForDelete(tableDefinition, dataContainer, dataContainers))
+          currentGroup.ForUpdate.Add(dataContainer);
+      }
+      else if (ShouldCreateUpdateCommand(dataContainer))
+      {
         currentGroup.ForUpdate.Add(dataContainer);
+      }
+    }
+
+    return group.Values.OrderBy(g => _foreignKeyOptimizationProvider.GetSortPosition(g.TableDefinition)).ToList();
```
a-ctor: So I wanted to suggest a `SortedDictionary` but then I tested its performance and it is abysmal. So I guess we keep the straightforward solution :D

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/LazyForeignKeyOptimizationProviderWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+      _wrapperInitializationFinished.Set();
+    });
+  }
+
+  public int GetSortPosition (TableDefinition tableDefinition)
+  {
+    if (!_initializeHasBeenCalled)
+      throw new InvalidOperationException($"Before accessing {nameof(GetSortPosition)}, {nameof(Initialize)} has to be called.");
+
+    _wrapperInitializationFinished.Wait();
```
a-ctor: Given that we are calling all of these methods a lot during the save operation I am proposing a wonderful premature optimization: We set the `ManualResetEventSlim` to `null` if we have signaled it and use `?.Wait()` here instead.

`ManualResetEventSlim` does a bit of win32 stuff in its `Wait`, which is going to be slower than a null check. Additionally, once we set it to `null` it will never get changed back and the branch prediction should make the null check negligible.

What do you think?

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/IForeignKeyOptimizationProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+
+  /// <summary>
+  /// Gets the sort position for the given <paramref name="tableDefinition"/>.
+  /// </summary>
+  int GetSortPosition (TableDefinition tableDefinition);
+
+  /// <summary>
+  /// Gets if <paramref name="tableDefinition"/> has been sorted correctly or not.
+  /// </summary>
+  bool HasBeenSortedCorrectly (TableDefinition tableDefinition);
```
a-ctor: I don't understand what this method wants to tell me. What is "sorted correctly"? It suggests to me that something can either be unsorted, sorted correctly, or sorted incorrectly, which is strange. How about `HasSortPosition`?

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNode.cs
```diff
@@ -0,0 +1,149 @@
...
+  {
+    edgeStack.Push(currentEdge);
+
+    if (currentEdge.PointingTo == this)
+      return true;
+
+    foreach (var edge in currentEdge.PointingTo.Edges)
+    {
+      if (edge.IsSelfCyclingEdge || edgeStack.Any(c => c.PointingTo == edge.PointingTo))
+        continue;
```
a-ctor: Given that `.Any(...)` does a linear scan through the stack I would like to compare the performance to a solution with an additional `HashSet` to check if we visited a node before. It might not matter as we don't go deep enough in the tree here but I would like to check to make sure it is slower.

a-ctor (commented):
Passing over the review to mike

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/LazyForeignKeyOptimizationProviderWrapper.cs
a-ctor: As discussed: I incorrectly looked up the source for `ManualResetEvent`, in the slim variant this is not such a big problem. It's probably still a bit faster but that change is negligible.

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
MichaelKetting: @a-ctor concious descison. This memory is cheap and more costeffective to waste some memory than to copy into a new list.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNode.cs
Mario185: The longest stack that I have found in some tests (with domains of other projects) was 11 entries and this code runs only one time. The whole process of creating the whole graph was less than a second in the largest project.

Benchmark results:
| Method  | Mean     | Error     | StdDev    |
|-------- |---------:|----------:|----------:|
| HashSet_Contains | 4.331 ns | 0.0226 ns | 0.0211 ns |
| List_Any        | 4.533 ns | 0.1103 ns | 0.1651 ns |

I think Any is OK in this use case :)

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/IForeignKeyOptimizationProvider.cs
Mario185: HasBeenSortedCorrectly will be removed with https://jira.rubicon.eu/browse/RM-9648

Remotion/Data/DomainObjects.UnitTests/Factories/BaseConfiguration.cs
```diff
@@ -81,7 +81,8 @@ public static ITypeDiscoveryService GetTypeDiscoveryService (params Assembly[] r
                                     "Remotion.Data.DomainObjects.UnitTests.TestDomain",
                                     "Remotion.Data.DomainObjects.UnitTests.DataManagement.TestDomain",
                                     "Remotion.Data.DomainObjects.UnitTests.MixedDomains.TestDomain",
-                                    "Remotion.Data.DomainObjects.UnitTests.Linq.TestDomain"
+                                    "Remotion.Data.DomainObjects.UnitTests.Linq.TestDomain",
+                                    "Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.ForeignKeyOptimization"
```
MichaelKetting:
```suggestion
                                    "Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.ForeignKeyOptimization.TestDomain"
```

Since you still stuck to the TestDomain namespace pattern, we should add it here, too, for consistency

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/TestDomain/AutomaticBreaksCountObjects.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.ForeignKeyOptimization.TestDomain;
+
+[DBTable]
+[Instantiable]
+public abstract class AutomaticBreaksCountObjectA : ForeignKeyOptimizationDomainBase
+{
+  public AutomaticBreaksCountObjectZ AutomaticBreakCountAPropZ { get; set; }
+}
```
MichaelKetting, line 13: It's not relevant in the big scheme of things, but these days, we can drop the `[Instantiable]` and the `abstract` for types. The mapping can handle regular classes and the code gets shorter. Using abstract in combination with abstract virtual properties just helps with preventing unnecessary backing fields.

Also, I honestly don't care either way. It's just an observation.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactoryTest.cs
```diff
@@ -0,0 +1,97 @@
...
+
+      if (entityDefinition is not FilterViewDefinition filterViewDefinition)
+      {
+        Assert.Fail($"Could not determine {nameof(TableDefinition)} for {classDefinition.ID}");
+        throw new UnreachableException();
+      }
+
+      entityDefinition = filterViewDefinition.BaseEntity;
+    }
+  }
```
MichaelKetting: This could be written as
```
  private TableDefinition GetTableDefinition (ClassDefinition classDefinition)
  {
      return InlineRdbmsStorageEntityDefinitionVisitor.Visit<TableDefinition>(
          (IRdbmsStorageEntityDefinition)classDefinition.StorageEntityDefinition,
          (table, _) => table,
          (filterView, continuation) => continuation(filterView.BaseEntity),
          (unionView, _) => { throw new AssertionException($"Could not determine {nameof(TableDefinition)} for {classDefinition.ID}"); },
          (emptyView, _) => { throw new AssertionException($"Could not determine {nameof(TableDefinition)} for {classDefinition.ID}"); });
  }
```

This would prevent introducing new patterns into the code.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactoryTest.cs
```diff
@@ -0,0 +1,97 @@
...
+    var nodes = _nodeFactory.CreateNodes([objectA, objectB, objectC, objectD]);
+
+    Assert.That(nodes.Count, Is.EqualTo(3));
+    var aNode = nodes[0];
+    var bNode = nodes[1];
+    var cNode = nodes[2];
+
+    Assert.That(aNode.TableDefinition, Is.EqualTo(GetTableDefinition(objectA)));
+    Assert.That(aNode.Edges.Count, Is.EqualTo(1));
+    AssertEdge(aNode.Edges[0], ["FK_NodeFactoryObjectA_SelfCyclingPropID"], aNode, true);
```
MichaelKetting, line 42: personal preference: specifying the parameter name for booleans in tests (or regular code) to make obscure / custom APIs more expressive. Here, it's just one API and it's defined a couple of lines of code below, so I guess it's fine the way it is.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationTestBase.cs
```diff
@@ -0,0 +1,48 @@
...
+using Remotion.Reflection.TypeDiscovery;
+using Remotion.Reflection.TypeDiscovery.AssemblyFinding;
+using Remotion.Reflection.TypeDiscovery.AssemblyLoading;
+using Remotion.ServiceLocation;
+
+namespace Remotion.Data.DomainObjects.UnitTests.Persistence.Rdbms.ForeignKeyOptimization;
+
+public abstract class ForeignKeyOptimizationTestBase
+{
+  private static ITypeDiscoveryService GetTypeDiscoveryService (params Assembly[] rootAssemblies)
```
MichaelKetting, line 46: since it's a private api used only by instance members: please move after the public instance member.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationTestBase.cs
```diff
@@ -0,0 +1,48 @@
...
+    return new FilteringTypeDiscoveryService(
+        typeDiscoveryService,
+        type =>
+        {
+          var @namespace = type.Namespace ?? string.Empty;
+          return whitelistedNamespaces.Any(t => @namespace.StartsWith(t));
+        });
+  }
+
+  public MappingConfiguration GetMappingConfiguration ()
```
MichaelKetting:
```suggestion
  protected MappingConfiguration GetMappingConfiguration ()
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactoryTest.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+  private ForeignKeyOptimizationNodeFactory _nodeFactory;
+  private MappingConfiguration _mappingConfiguration;
+
+  [SetUp]
+  public void SetUp ()
+  {
+    _nodeFactory = new ForeignKeyOptimizationNodeFactory();
+    _mappingConfiguration = GetMappingConfiguration();
+  }
```
MichaelKetting, line 23: For performance, we should load the mapping in at least a fixture setup. And I'm considering if instead of a base test, we shouldn't use a simple static in a lazy field since the mapping doesn't change between the tests in this namespace and running the analysis 10? times does amount to some time.  remember, these tests are fast :D

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/ForeignKeyOptimization/GraphBasedForeignKeyOptimizationProviderTest.cs
```diff
@@ -0,0 +1,213 @@
...
+
+      if (entityDefinition is not FilterViewDefinition filterViewDefinition)
+      {
+        Assert.Fail($"Could not determine {nameof(TableDefinition)} for {classDefinition.ID}");
+        throw new UnreachableException();
+      }
+
+      entityDefinition = filterViewDefinition.BaseEntity;
+    }
+  }
```
MichaelKetting: Move to base class and remove duplication. see other test file.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/IntegrationTests/SqlProviderGeneratedSqlTest.cs
```diff
@@ -283,8 +283,8 @@ public void Save_BatchedSaveCommandFactory ()
 
       var expectedLockTvpValue = new SqlTableValuedParameterValue("TVP_AllTables_Lock", new[] { new SqlMetaData("ID", SqlDbType.UniqueIdentifier), new SqlMetaData("Timestamp", SqlDbType.VarBinary, 8) });
       expectedLockTvpValue.AddRecord([changedDataContainer.ID.Value, changedDataContainer.Timestamp]);
-      expectedLockTvpValue.AddRecord([deletedDataContainer.ID.Value, deletedDataContainer.Timestamp]);
       expectedLockTvpValue.AddRecord([markedAsChangedDataContainer.ID.Value, markedAsChangedDataContainer.Timestamp]);
+      expectedLockTvpValue.AddRecord([deletedDataContainer.ID.Value, deletedDataContainer.Timestamp]);
```
MichaelKetting, line 287: What changed to move this line to the last entry? and why do we not have to assert in line the newDataContainer and deletedDataContainer in line 307 and 308 anymore?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactoryTest.cs
```diff
@@ -3,21 +3,26 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using JetBrains.Annotations;
+//using JetBrains.Annotations;
```
MichaelKetting:
```suggestion
using JetBrains.Annotations;
```

Remotion/Data/DomainObjects/Mapping/MappingConfiguration.cs
```diff
@@ -118,6 +122,7 @@ public MappingConfiguration (IMappingLoader mappingLoader, IPersistenceModelLoad
       _nameResolver = mappingLoader.NameResolver;
 
       _mapping = new Lazy<Mapping>(() => InitializeMapping(mappingLoader, persistenceModelLoader), LazyThreadSafetyMode.ExecutionAndPublication);
+
```
MichaelKetting:
```suggestion
```
Remove new blank line

Remotion/Data/DomainObjects/Mapping/MappingConfiguration.cs
```diff
@@ -153,7 +158,11 @@ private Mapping InitializeMapping (IMappingLoader mappingLoader, IPersistenceMod
 
         mappingConfigurationValidationHelper.ValidateSortExpression(relationDefinitionsDictionary.Values);
 
-        return new Mapping(typeDefinitionsDictionary, classDefinitionsDictionary, relationDefinitionsDictionary);
+        var foreignKeyOptimizationProvider = SafeServiceLocator.Current.GetInstance<IForeignKeyOptimizationProvider>();
```
MichaelKetting: Please use injection and bubble this all the way to the MappingConfiguration ctor. The MappingConfiguration is resolved from the service container.

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -553,6 +559,7 @@ protected virtual IRdbmsProviderCommandFactory CreateStorageProviderCommandFacto
       var dbCommandBuilderFactory = CreateDbCommandBuilderFactory(storageProviderDefinition);
 
       var tableManipulationRecordDefinitionProvider = CreateTableManipulationRecordDefinitionProvider(storageProviderDefinition);
+      var foreignKeyOptimizationSortOrderProvider = GetForeignKeyOptimizationProvider();
```
MichaelKetting: Please inject IForeignKeyOptimizationProvider via the ctor. The SqlStorageObjectFactory is resolved via IoC

Remotion/Data/DomainObjects/Persistence/Rdbms/SqlServer/Sql2016/SqlStorageObjectFactory.cs
```diff
@@ -374,6 +375,11 @@ public virtual IStorageProviderSerializer CreateStorageProviderSerializer (IEnum
       return new StorageProviderSerializer(CreateClassSerializer(enumSerializer));
     }
 
+    public virtual IForeignKeyOptimizationProvider GetForeignKeyOptimizationProvider ()
+    {
+      return MappingConfiguration.Current.ForeignKeyOptimizationProvider;
+    }
+
```
MichaelKetting: Delete, see below.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.ServiceLocation;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.ForeignKeyOptimization;
+
+/// <summary>
+/// <see cref="ForeignKeyOptimizationNodeFactory"/> creates <see cref="ForeignKeyOptimizationNode"/>s based on <see cref="ClassDefinition"/>s
+/// of persistent and non-abstract types.
+/// </summary>
+[ImplementationFor(typeof(IForeignKeyOptimizationNodeFactory), RegistrationType = RegistrationType.Single, Lifetime = LifetimeKind.Singleton)]
```
MichaelKetting: please add an `IForeignKeyOptimizationNodeFactoryTest` for the IoC configuration. see other tests, e.g. IMappingConfigurationTest

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/GraphBasedForeignKeyOptimizationProvider.cs
```diff
@@ -0,0 +1,204 @@
...
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+using Remotion.ServiceLocation;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.ForeignKeyOptimization;
+
+/// <summary>
+/// <see cref="GraphBasedForeignKeyOptimizationProvider"/> is an implementation of <see cref="IForeignKeyOptimizationProvider"/>
+/// which determines the sort order depending on the foreign keys of the <see cref="TableDefinition"/>s.
+/// </summary>
+[ImplementationFor(typeof(IForeignKeyOptimizationProvider), RegistrationType = RegistrationType.Single, Lifetime = LifetimeKind.InstancePerDependency)]
```
MichaelKetting: Please add  IForeignKeyOptimizationProviderTest for the IoC configuration. See IMappingConfigurationTest as example.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/LazyForeignKeyOptimizationProviderWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+using Remotion.Context;
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.ForeignKeyOptimization;
+
+/// <summary>
+/// <see cref="LazyForeignKeyOptimizationProviderWrapper"/> wraps another instance of <see cref="IForeignKeyOptimizationProvider"/>
+/// and delegates all calls to the wrapped instance. But <see cref="Initialize"/> is executed in a separate thread so that initialization
+/// does not block the current Thread. All other calls wait till the initialization is finished.
```
MichaelKetting:
```suggestion
/// does not block the current thread. All other calls wait till the initialization is finished.
```

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/LazyForeignKeyOptimizationProviderWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+/// </summary>
+public class LazyForeignKeyOptimizationProviderWrapper : IForeignKeyOptimizationProvider
+{
+  private readonly IForeignKeyOptimizationProvider _wrappedProvider;
+  private bool _initializeHasBeenCalled;
+  private readonly ManualResetEventSlim _wrapperInitializationFinished;
+  public LazyForeignKeyOptimizationProviderWrapper (IForeignKeyOptimizationProvider wrappedProvider)
```
MichaelKetting:
```suggestion
  private readonly ManualResetEventSlim _wrapperInitializationFinished;

  public LazyForeignKeyOptimizationProviderWrapper (IForeignKeyOptimizationProvider wrappedProvider)
```

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+public class ForeignKeyOptimizationNodeFactory : IForeignKeyOptimizationNodeFactory
+{
+  /// <inheritdoc />
+  public IReadOnlyList<ForeignKeyOptimizationNode> CreateNodes (IReadOnlyList<ClassDefinition> classDefinitions)
```
MichaelKetting: I prefer to have a ctor on each type, even if it's empty. that way, the class has a defined structure and you are certain that you found the ctor when scanning the code.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+using Remotion.ServiceLocation;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.ForeignKeyOptimization;
+
+/// <summary>
+/// <see cref="ForeignKeyOptimizationNodeFactory"/> creates <see cref="ForeignKeyOptimizationNode"/>s based on <see cref="ClassDefinition"/>s
+/// of persistent and non-abstract types.
+/// </summary>
+[ImplementationFor(typeof(IForeignKeyOptimizationNodeFactory), RegistrationType = RegistrationType.Single, Lifetime = LifetimeKind.Singleton)]
+public class ForeignKeyOptimizationNodeFactory : IForeignKeyOptimizationNodeFactory
```
MichaelKetting: Please add the threadsafety xmldoc. (instance and static are threadsafe.) there's examples in the codebase

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+  /// <inheritdoc />
+  public IReadOnlyList<ForeignKeyOptimizationNode> CreateNodes (IReadOnlyList<ClassDefinition> classDefinitions)
+  {
+    ArgumentNullException.ThrowIfNull(classDefinitions);
+
+    var persistentClassDefinitions = classDefinitions.Where(d => d.StorageEntityDefinition is not NonPersistentStorageEntity && !d.IsAbstract).ToArray();
```
MichaelKetting: With the refactoring mentioned below, this line will go bye-bye.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+    var nodes = persistentClassDefinitions.GroupBy(GetTableDefinition).Select(g => new ForeignKeyOptimizationNode(g.Key, g.ToList())).ToList();
+    AddEdgesToNodes(nodes);
+
+    return nodes;
+  }
```
MichaelKetting: This method does not take into account class definitions separated between storage providers. they don't have foreign keys between them anyway, so strictly speaking, it won't matter, but it feels like an oversight and accidental simplification if we don't split by storage provider. with the refactoring mentioned below, this will happen anyway, I think.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/ForeignKeyOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+      if (entityDefinition is TableDefinition tableDefinition)
+        return tableDefinition;
+
+      if (entityDefinition is not FilterViewDefinition viewDefinition)
+        throw new InvalidOperationException($"Could not determine {nameof(TableDefinition)} for class '{classDefinition.ID}'.");
+
+      entityDefinition = viewDefinition.BaseEntity;
+    }
+  }
```
MichaelKetting: Okay, I just found a dependency problem. We want to use TableDefinitionFinder, which needs the IRdbmsPersistenceModelProvider created by the SqlStorageObjectFactory.

I think we can use the same pattern as in ScriptGenerator.GetScripts when it groups by StroageProvideDefinition and then passes the RdbmsStorageProviderDefinition to the RdbmsStructuredTypeDefinitionProvider.GetTypeDefinitions. 

Unfortunately, we need one more indirection because MappingConfiguration should depend directly on Rdbms-types. We can either add factor this logic from MappingConfiguration into a new separate type or add it to IPersistenceModelLoader as a second API.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/LazyForeignKeyOptimizationProviderWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+      throw new InvalidOperationException($"{nameof(Initialize)} has already been called.");
+
+    _initializeHasBeenCalled = true;
+
+    SafeContext.Task.Run(() =>
+    {
+      _wrappedProvider.Initialize(classDefinitions);
+      _wrapperInitializationFinished.Set();
+    });
+  }
```
MichaelKetting: We don't keep the Task in a backing field which could result in the task getting garbage collected. I don't want to rely on some black magic even if there is such a thing.

And once we put the task in a backing field upon initialization, we can use a simple doublechecked-lock instead of the boolean _initializeHasBeenCalled. We also need to use Lock during the initialization.

We should factor this the task-result-retrieval into a helper method and use the return value of the helper to get rid of that nasty code duplication.

Remotion/Data/DomainObjects/Persistence/Rdbms/ForeignKeyOptimization/LazyForeignKeyOptimizationProviderWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+    if (_initializeHasBeenCalled)
+      throw new InvalidOperationException($"{nameof(Initialize)} has already been called.");
+
+    _initializeHasBeenCalled = true;
+
+    SafeContext.Task.Run(() =>
+    {
+      _wrappedProvider.Initialize(classDefinitions);
+      _wrapperInitializationFinished.Set();
+    });
```
MichaelKetting: We should use a task with a return type and simply return the wrappedProvider upon finishing the task. We can capture it in an action in the ctor and null the action after initializing the task. Then we can simply Task.Wait() and Task.Result instead of the boolean and the slimwait.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactoryTest.cs
```diff
@@ -3,21 +3,26 @@
 using System;
```
MichaelKetting, line 3: review pending

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
```diff
@@ -7,6 +7,7 @@
 using Remotion.Data.DomainObjects.Mapping;
```
MichaelKetting, line 7: review pending

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNode.cs
```diff
@@ -0,0 +1,149 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
```
MichaelKetting, line 1: review pending

MichaelKetting (changes requested):
First part of the review

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/TestDomain/AutomaticBreaksCountObjects.cs
Mario185: Since you don't care I let it be as it is now and remember for next time.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactoryTest.cs
Mario185: With https://jira.rubicon.eu/browse/RM-9648 this will be changed so that GetMappingConfiguration is called in each test because they only need to load the types they are testing. And  there are types involved which lead to errors during loading of the mapping configuration. Which means if I change it now it will be changed back in the next PR.

MichaelKetting: okay

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SqlServer/IntegrationTests/SqlProviderGeneratedSqlTest.cs
Mario185: The sorting changed. 
Before the values have been added in the order as they were present in the data containers. Now they get sorted dependeng on their dependencies and therefore the deletes will be executed a fter the changed.
And we do not need to check for the new and deleted data container in the update because they are no longer present because the sorting determined that an update for these two is not necassary.

MichaelKetting: okay

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedUpdateDbCommandBuilder.cs
MichaelKetting, line 1: Please revert. We use String.ReplaceLineEndings() in the tests instead on the comparison string.
See RootClientTransaction_InvalidDomainObjects_ValidationExceptionIsThrown_AndMessageIsUsingInvariantCulture for instance.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
MichaelKetting, line 1: Please revert. We use String.ReplaceLineEndings() in the tests instead on the comparison string.
See RootClientTransaction_InvalidDomainObjects_ValidationExceptionIsThrown_AndMessageIsUsingInvariantCulture for instance.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedInsertDbCommandBuilder.cs
MichaelKetting, line 1: Please revert. We use String.ReplaceLineEndings() in the tests instead on the comparison string.
See RootClientTransaction_InvalidDomainObjects_ValidationExceptionIsThrown_AndMessageIsUsingInvariantCulture for instance.

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedDeleteDbCommandBuilder.cs
MichaelKetting, line 1: Please revert. We use String.ReplaceLineEndings() in the tests instead on the comparison string.
See RootClientTransaction_InvalidDomainObjects_ValidationExceptionIsThrown_AndMessageIsUsingInvariantCulture for instance.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProvider.cs
MichaelKetting, line 1: Discussed some weeks ago: Since the GraphBasedPersistenceModelSortingProvider is not stateless, we cannot use IoC for it. We need to construct it and inject it manually.

Remotion/Data/DomainObjects/Mapping/MappingConfiguration.cs
```diff
@@ -80,12 +81,13 @@ public static void SetCurrent (IMappingConfiguration mappingConfiguration)
     /// <summary>
     /// Creates a fully initialized <see cref="MappingConfiguration"/>
     /// </summary>
-    public static MappingConfiguration Create (IMappingLoader mappingLoader, IPersistenceModelLoader persistenceModelLoader)
+    public static MappingConfiguration Create (IMappingLoader mappingLoader, IPersistenceModelLoader persistenceModelLoader, IPersistenceModelSortingProvider persistenceModelSortingProvider)
```
MichaelKetting: Here, we inject the ISortingOptimizationNodeFactory instead of the IPersistenceModelSortingProvider.
In the tests, we should use a FakeSortingOptimizationNodeFactory instead unless the sorting is important. Altough, I'm not 100% certain if that will scale well. I guess many tests won't care, but those that do will be using the `TestMappingConfiguration.cs` and `BaseConfiguration.cs`. So, perhaps in places where we build the config for individual tests, inject the Fake and when instantiating the `TestMappingConfiguration.cs` or `BaseConfiguration.cs`, use the `SortingOptimizationNodeFactory` (tbd from IOC or via explict ctor call)

Remotion/Data/DomainObjects/Mapping/MappingConfiguration.cs
```diff
@@ -109,21 +111,24 @@ private static ArgumentException CreateArgumentException (string argumentName, s
...
-      _mapping = new Lazy<Mapping>(() => InitializeMapping(mappingLoader, persistenceModelLoader), LazyThreadSafetyMode.ExecutionAndPublication);
+      _mapping = new Lazy<Mapping>(() => InitializeMapping(mappingLoader, persistenceModelLoader, persistenceModelSortingProvider), LazyThreadSafetyMode.ExecutionAndPublication);
     }
 
-    private Mapping InitializeMapping (IMappingLoader mappingLoader, IPersistenceModelLoader persistenceModelLoader)
+    private Mapping InitializeMapping (IMappingLoader mappingLoader, IPersistenceModelLoader persistenceModelLoader, IPersistenceModelSortingProvider persistenceModelSortingProvider)
     {
       s_logger.LogInformation("Building mapping configuration...");
 
+      var lazyPersistenceModelSortingProviderWrapper = new LazyPersistenceModelSortingProviderWrapper(persistenceModelSortingProvider);
```
MichaelKetting: You mentioned that the analysis is really fast and we should consider dropping the Lazy part.

MichaelKetting: Here, we can initialize the GraphBasedPersistenceModelSortingProvider.
I think, if we inject a FakeSortingOptimizationNodeFactory that always returns an empty node set, the GraphBasedPersistenceModelSortingProvider shouldn't do anything bad in the unit tests.

Remotion/Data/DomainObjects/Persistence/SortingOptimization/LazyPersistenceModelSortingProviderWrapper.cs
MichaelKetting, line 1: If we keep this file, we need to discuss the lock contention / need for a double-checked-locking pattern on _initializeWrappedProviderTask. 

Also, argument checks for the public api

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/EmptyRdbmsProviderCommand.cs
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: (c) RUBICON IT GmbH, www.rubicon.eu
+// SPDX-License-Identifier: LGPL-2.1-or-later
+using System;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.StorageProviderCommands;
+
+/// <summary>
+/// This command does not perform any actions.
+/// </summary>
+public class EmptyRdbmsProviderCommand : IRdbmsProviderCommand
```
MichaelKetting: Please rename to NullRdbmsProviderCommand

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationEdge.cs
```diff
@@ -0,0 +1,55 @@
...
+  public SortingOptimizationEdge (ForeignKeyConstraintDefinition foreignKey, SortingOptimizationNode owner, SortingOptimizationNode pointingTo)
+  {
+    ArgumentNullException.ThrowIfNull(foreignKey);
+    ArgumentNullException.ThrowIfNull(owner);
+    ArgumentNullException.ThrowIfNull(pointingTo);
+
+    ForeignKeys = [foreignKey];
+    Owner = owner;
+    PointingTo = pointingTo;
+    IsSelfCyclingEdge = owner == pointingTo;
```
MichaelKetting: please use Object.ReferenceEquals. It's more explicit.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationEdge.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Diagnostics;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+
+/// <summary>
+/// <see cref="SortingOptimizationEdge"/> represents one edge of an <see cref="SortingOptimizationNode"/>.
+/// </summary>
+[DebuggerDisplay("Edge: {Owner.TableName} -> {PointingTo.TableName}")]
+public class SortingOptimizationEdge
```
MichaelKetting, line 14: UnitTests for SortingOptimizationEdge please

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNode.cs
```diff
@@ -0,0 +1,153 @@
...
+using Remotion.Data.DomainObjects.Mapping;
+using Remotion.Data.DomainObjects.Persistence.Rdbms.Model;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+
+/// <summary>
+/// <see cref="SortingOptimizationNode"/> represents on <see cref="TableDefinition"/> and possible multiple <see cref="ClassDefinitions"/>s
+/// </summary>
+[DebuggerDisplay("Node: {TableName}")]
+public class SortingOptimizationNode
```
MichaelKetting, line 17: Can we do unit tests for this class? If it's possible but too much time right now, move it into a ticket.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNode.cs
```diff
@@ -0,0 +1,153 @@
...
+  /// </summary>
+  public IReadOnlyList<SortingOptimizationEdge> BrokenEdges => _brokenEdges;
+
+  /// <summary>
+  /// Gets all <see cref="SortingOptimizationEdge"/> which cause indirect cyclic dependencies.
+  /// The key is the <see cref="SortingOptimizationEdge"/> of this <see cref="SortingOptimizationNode"/> starting the cyclic dependency.
+  /// The value is the list of all <see cref="SortingOptimizationEdge"/>s involved in the cyclic dependency.
+  /// This will be empty until <see cref="CalculateIndirectCyclicDependencies"/> has been called.
+  /// </summary>
+  public Dictionary<SortingOptimizationEdge, List<SortingOptimizationEdge>> IndirectSelfCyclicEdges => _indirectSelfCyclicEdges;
```
MichaelKetting:
```suggestion
  public IReadOnlyDictionary<SortingOptimizationEdge, List<SortingOptimizationEdge>> IndirectSelfCyclicEdges => _indirectSelfCyclicEdges;
```

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+
+  }
+
+  /// <inheritdoc />
+  public IReadOnlyList<SortingOptimizationNode> CreateNodes (IReadOnlyList<ClassDefinition> classDefinitions)
+  {
+    ArgumentNullException.ThrowIfNull(classDefinitions);
+
+    var persistentClassDefinitions = classDefinitions.Where(d => d.StorageEntityDefinition is not NonPersistentStorageEntity && !d.IsAbstract).ToArray();
+    var nodes = persistentClassDefinitions.GroupBy(GetTableDefinition).Select(g => new SortingOptimizationNode(g.Key, g.ToList())).ToList();
```
MichaelKetting: I think we can rewrite this to have GetTableDefinition return `TableDefinition` or `null` and then filter for not-null results. That way, we drop the test for `NonPersistentStorageEntity` which breaks abstraction.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private TableDefinition GetTableDefinition (ClassDefinition classDefinition)
+  {
+    return InlineRdbmsStorageEntityDefinitionVisitor.Visit<TableDefinition>(
+         (IRdbmsStorageEntityDefinition)classDefinition.StorageEntityDefinition,
+         (table, _) => table,
+         (filterView, continuation) => continuation(filterView.BaseEntity),
+         (unionView, _) => { throw new InvalidOperationException($"Could not determine {nameof(TableDefinition)} for {classDefinition.ID}"); },
+         (emptyView, _) => { throw new InvalidOperationException($"Could not determine {nameof(TableDefinition)} for {classDefinition.ID}"); });
```
MichaelKetting: Returning `null` here is probably correct. Not sure why the tests don't blow up since we should have union views somewhere.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+
+  private void AddEdgesToNodes (List<SortingOptimizationNode> nodes)
+  {
+    var entityNameDefinitionToNodeMapping = nodes.ToDictionary(k => k.TableDefinition.TableName, v => v);
+    foreach (var node in nodes)
+    {
+      foreach (var foreignKey in node.TableDefinition.Constraints.OfType<ForeignKeyConstraintDefinition>())
+      {
+        if (!entityNameDefinitionToNodeMapping.TryGetValue(foreignKey.ReferencedTableName, out var pointingTo))
+          throw new InvalidOperationException($"Could not find {nameof(SortingOptimizationNode)} '{foreignKey.ReferencedTableName.EntityName}' for foreign key {foreignKey.ConstraintName} on node '{node.TableName}'.");
```
MichaelKetting:
```suggestion
          throw new InvalidOperationException($"Could not find {nameof(SortingOptimizationNode)} '{foreignKey.ReferencedTableName.EntityName}' for foreign key '{foreignKey.ConstraintName}' on node '{node.TableName}'.");
```

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactoryTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+  [Test]
+  public void ResolvingNodeFactoryFromServiceLocator_GetInstance_Twice_ReturnsSameInstance ()
+  {
+    var propertyDefaultValueProvider1 = _serviceLocator.GetInstance<ISortingOptimizationNodeFactory>();
+    var propertyDefaultValueProvider2 = _serviceLocator.GetInstance<ISortingOptimizationNodeFactory>();
+
+    Assert.That(propertyDefaultValueProvider1, Is.SameAs(propertyDefaultValueProvider2));
+  }
```
MichaelKetting: Please move this into a TestFixture ISortingOptimizationNodeFactoryTest. The test is for the correct IoC config for the entire interface.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNode.cs
```diff
@@ -0,0 +1,153 @@
...
+    foreach (var edge in currentEdgePointingToNode.Edges)
+    {
+      if (edge.IsSelfCyclingEdge || visitedNodes.Contains(edge.PointingTo))
+        continue;
+
+      if (CalculateHasIndirectCyclicDependency(edge, edgeStack, visitedNodes))
+        return true;
+
+      edgeStack.Pop();
+    }
```
MichaelKetting, line 167: In terms of readability, we try to avoid `continue` in the code base. Instead, an if-block

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNode.cs
```diff
@@ -0,0 +1,153 @@
...
+
+    _edges.Remove(edge);
+    _brokenEdges.Add(edge);
+  }
+
+  /// <summary>
+  /// Calculates if this <see cref="SortingOptimizationNode"/> has cyclic dependencies
+  /// and updates <see cref="IndirectSelfCyclicEdges"/>.
+  /// </summary>
+  public void CalculateIndirectCyclicDependencies ()
```
MichaelKetting, line 129: I did not mentally calculate the correctness of the algorithm. I trust the tests.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProvider.cs
```diff
@@ -0,0 +1,207 @@
...
+    foreach (var node in nodes)
+    {
+      var classAndPropertyDefinitions = node.ClassDefinitions.SelectMany(c =>
+          c.GetPropertyDefinitions().Where(p => p.IsObjectID && p.StorageClass == StorageClass.Persistent)
+              .Select(p => (
+                      ClassDefinition: c,
+                      PropertyDefinition: p,
+                      Columns: ((IRdbmsStoragePropertyDefinition)p.StoragePropertyDefinition).GetColumnsForComparison().ToArray()
+                  )
+              )).ToList();
```
MichaelKetting:
```suggestion
      var classAndPropertyDefinitions = node.ClassDefinitions
          .SelectMany(c => c.GetPropertyDefinitions().Where(p => p.IsObjectID && p.StorageClass == StorageClass.Persistent)
              .Select(p => (
                      ClassDefinition: c,
                      PropertyDefinition: p,
                      Columns: ((IRdbmsStoragePropertyDefinition)p.StoragePropertyDefinition).GetColumnsForComparison().ToArray()
                      )
                  )).ToList();
```
Formatting

MichaelKetting, line 97: The SortingNodeFactory does the filtering for Rdbms-compatible types. I think I would like a check that `ClassDefinition.StorageEntityDefinition is IRdbmsStorageEntityDefinition`. We can do this using Assertion.IsTrue, including a comment with the ClassDefinition.ID if the condition fails. We should also add a code comment where the filtering happens or put it into the assertion (not sure if that reads well).

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProvider.cs
```diff
@@ -0,0 +1,207 @@
...
+        return false;
+
+      return string.Equals(x.Name, y.Name, StringComparison.Ordinal) && x.IsPartOfPrimaryKey == y.IsPartOfPrimaryKey;
+    }
+
+    public int GetHashCode (ColumnDefinition obj)
+    {
+      return HashCode.Combine(obj.Name, obj.IsPartOfPrimaryKey);
+    }
+  }
```
MichaelKetting: Since this also happens in TableManuplation...Provider, extract the ColumnDefinitionEqualityComparer into an internal type?

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProvider.cs
```diff
@@ -0,0 +1,207 @@
...
+using Remotion.ServiceLocation;
+
+namespace Remotion.Data.DomainObjects.Persistence.Rdbms.SortingOptimization;
+
+/// <summary>
+/// <see cref="GraphBasedPersistenceModelSortingProvider"/> is an implementation of <see cref="IPersistenceModelSortingProvider"/>
+/// which determines the sort order depending on the foreign keys of the <see cref="TableDefinition"/>s.
+/// </summary>
+[ImplementationFor(typeof(IPersistenceModelSortingProvider), RegistrationType = RegistrationType.Single, Lifetime = LifetimeKind.InstancePerDependency)]
+public class GraphBasedPersistenceModelSortingProvider : IPersistenceModelSortingProvider
```
MichaelKetting, line 17: I did not reason over the actual algorithm. we got the tests for that.

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProviderTest.cs
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  [Test]
+  public void ResolvingNodeFactoryFromServiceLocator_GetInstance_Twice_ReturnsSameInstance ()
+  {
+    var propertyDefaultValueProvider1 = _serviceLocator.GetInstance<IPersistenceModelSortingProvider>();
+    var propertyDefaultValueProvider2 = _serviceLocator.GetInstance<IPersistenceModelSortingProvider>();
+
+    Assert.That(propertyDefaultValueProvider1, Is.Not.SameAs(propertyDefaultValueProvider2));
+  }
```
MichaelKetting: Move to `IPersistenceModelSortingProviderTest` test fixture since it is meant to test the IoC configuration for the interface. Of course, since we need to kick the IoC part of this type anyway...

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactory.cs
```diff
@@ -157,11 +167,12 @@ public BatchedSaveCommandFactory (
 
   public ITableDefinitionFinder TableDefinitionFinder => _tableDefinitionFinder;
 
-  public virtual IRdbmsProviderCommand CreateForSave (IEnumerable<DataContainer> dataContainers)
+  public virtual IRdbmsProviderCommand CreateForSave (IEnumerable<DataContainer> dataContainers, IPersistenceModelSortingProvider persistenceModelSortingProvider)
```
MichaelKetting: Question: I know we want to ensure this is done correctly and consistently, but I wonder if we actually need to do this check, It costs time, the infrastructure already ensures only DataContainer form a single StorageProvider are provided and we now have the IPersistenceModelSortingProvider is on the TableDefintion.

I therefor suggest we remove IPersistenceModelSortingProvider from CreateForSave here and the calling site in RdbmsCommandFactory. We should also hide the IPersistenceModelSortingProvider on TableDefinition, instead exposing the relevant APIs directly (and add tests :) )

MichaelKetting (changes requested):
Mostly some API simplifications and the statefullness-problem of Graphbased-SortingProvider.

Remotion/Data/DomainObjects/Mapping/MappingConfiguration.cs
Mario185: With the last improvements it was really fast, but i cannot be sure that this is true for every scenario, therefore i suggest we leave it with the lazy wrapper.

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactory.cs
Mario185: This was because the check of "d.StorageEntityDefinition is not NonPersistentStorageEntity" which is now removed and GetTableDefinitions returns null

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/GraphBasedPersistenceModelSortingProviderTest.cs
Mario185: Since GraphBasedPersistenceModelSortingProvider is no longer provided over IOC the tests have been removed.

Remotion/Data/DomainObjects/Persistence/Rdbms/RdbmsProviderCommandFactory.cs
MichaelKetting, line 1: Revert?

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/ISaveCommandFactory.cs
MichaelKetting, line 1: Revert?

Remotion/Data/DomainObjects/Persistence/Rdbms/StorageProviderCommands/Factories/IndividualSaveCommandFactory.cs
MichaelKetting, line 1: Revert?

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/TestableRdbmsProviderCommandFactory.cs
MichaelKetting, line 1: Revert

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/IndividualSaveCommandFactoryTest.cs
MichaelKetting, line 1: Revert

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
MichaelKetting: Was marked as resolved but wasn't reverted so reopned the comment

Remotion/Data/DomainObjects/Persistence/Rdbms/SortingOptimization/SortingOptimizationNodeFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+      return null;
+
+    return InlineRdbmsStorageEntityDefinitionVisitor.Visit<TableDefinition?>(
+        storageEntityDefinition,
+        (table, _) => table,
+        (filterView, continuation) => continuation(filterView.BaseEntity),
+        (unionView, _) => null,
+        (emptyView, _) => null);
+
+  }
```
MichaelKetting, line 64:
```suggestion
```
Should be unused code

Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/SortingOptimization/SortingOptimizationEdgeTest.cs
```diff
@@ -0,0 +1,184 @@
...
+  {
+    // Use the node's AddEdge so the edge is tracked by its owner; otherwise
+    // BreakEdge on the owner would not be able to remove it.
+    _nodeA.AddEdge(_foreignKey, _nodeB);
+    var edge = _nodeA.Edges.Single();
+
+    edge.BreakEdge();
+
+    Assert.That(_nodeA.Edges, Is.Empty);
+    Assert.That(_nodeA.BrokenEdges, Is.EqualTo(new[] { edge }));
```
MichaelKetting: Info-Note: When it's just one item, it doesn't matter, but it's still a good practice to use Is.Equivalent for set comparison where the order is explicitly meant to not be important.

MichaelKetting: And stylistic: in times of .net 10, can/should we use  `[...]` instead of `new[] {...}`? it's something for the claude style file.

Remotion/Data/DomainObjects/Persistence/SortingOptimization/LazyPersistenceModelSortingProviderWrapper.cs
```diff
@@ -0,0 +1,91 @@
...
+      throw new InvalidOperationException($"{nameof(Initialize)} has already been called.");
+
+    lock (_lockObject)
+    {
+      _initializeWrappedProviderTask = SafeContext.Task.Run(() =>
+      {
+        var result = _wrappedProviderInitFunc!(classDefinitions);
+        _wrappedProviderInitFunc = null;
+        return result;
+      });
```
MichaelKetting:
```suggestion
    lock (_lockObject)
    {
      if (_initializeWrappedProviderTask!= null)
        throw new InvalidOperationException($"{nameof(Initialize)} has already been called.");

      _initializeWrappedProviderTask = SafeContext.Task.Run(() =>
      {
        var result = _wrappedProviderInitFunc!(classDefinitions);
        _wrappedProviderInitFunc = null;
        return result;
      });
```
We can do all of this inside the lock-statement. VolatileRead is only needed when you do lock-optimization which we do not do here.

Remotion/Data/DomainObjects/Persistence/SortingOptimization/LazyPersistenceModelSortingProviderWrapper.cs
MichaelKetting, line 1: We can write tests for this using a FakePersistenceModelSortingProvider and the ManualResetEvent.

I wonder if Claude can do it for us. Otherwise, we should please create a follow-up ticket and do it in pair. it's an hour or two of work, mostly thinking through how to correctly write the test.

MichaelKetting (changes requested):

Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
Mario185: I think I resolved it by accident

MichaelKetting (approved):

PR #1416, opened 2026-03-17
RM-9595 - Add ILifecycleFeature for better control over init/dispose

noah-rubicon:
https://jira.rubicon.eu/browse/RM-9595

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
@@ -60,18 +60,6 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
     private Mock<ITestContext> _testContext;
     private Dictionary<string,object> _testContextProperties;
 
-    [OneTimeSetUp]
-    public void OneTimeSetUp ()
-    {
-      var webTestHelper = WebTestHelper.CreateFromConfiguration<CustomWebTestConfigurationFactory>();
-      //TODO: RM-9595 fix and reenable WebTestHelperTests for firefox
-      //WebTestHelperTests currently cause firefox with Bidi active to hang
-      if (webTestHelper.BrowserConfiguration.IsFirefox())
-      {
-        Assert.Ignore();
-      }
-    }
-
     [SetUp]
     public void SetUp ()
     {
@@ -116,6 +104,8 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
       webTestHelper.OnSetUp(_testContext.Object);
       browserLogEntries = webTestHelper.MainBrowserSession.GetBrowserLogs();
       Assert.That(browserLogEntries, Is.Empty);
+
+      ShutdownWebTestHelper(webTestHelper, true);
     }
 
     [Test]
@@ -158,6 +148,8 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
       testContext.Verify(
           _ => _.SetFailure(It.IsRegex($"There are unexpected browser log entries at the end of the test:{Environment.NewLine}.*\"?any error message\"?")),
           Times.Once());
+
+      webTestHelper.OnFixtureTearDown();
     }
 
     [Test]
--- a/Remotion/Web/Development.WebTesting.UnitTests/WebTestFeatureCollectionTest.cs
+++ b/Remotion/Web/Development.WebTesting.UnitTests/WebTestFeatureCollectionTest.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using Moq;
 using NUnit.Framework;
+using NUnit.Framework.Legacy;
 
 namespace Remotion.Web.Development.WebTesting.UnitTests;
 
@@ -127,6 +128,46 @@ public class WebTestFeatureCollectionTest
 
     var features = new WebTestFeatureCollection(items);
     AssertCollectionItems(features, items);
+  }
+
+  [Test]
+  public void InitializeFeatures ()
+  {
+    var webTestFeatureMock = new Mock<ILifecycleWebTestFeature>();
+    var items = new List<(Type, object)>
+                {
+                    (typeof(ILifecycleWebTestFeature), webTestFeatureMock.Object)
+                };
+
+    var features = new WebTestFeatureCollection(items);
+
+    features.InitializeFeatures();
+
+    webTestFeatureMock.Verify(f => f.Initialize(), Times.Once);
+  }
+
+  [Test]
+  public void InitializeFeatures_ThrowsAggregateException ()
+  {
+    var webTestFeatureMock1 = new Mock<ILifecycleWebTestFeature>();
+    var webTestFeatureMock2 = new Mock<ILifecycleWebTestFeature>();
+
+    Exception[] expectedExceptions = [new("Feature 1 failed")];
+    webTestFeatureMock1.Setup(f => f.Initialize()).Throws(expectedExceptions[0]);
+
+    var items = new List<(Type, object)>
+                {
+                    (typeof(ILifecycleWebTestFeature), webTestFeatureMock1.Object),
+                    (typeof(ITestFeature), webTestFeatureMock2.Object)
+                };
+    var features = new WebTestFeatureCollection(items);
+
+    var ex = Assert.Throws<AggregateException>(() => features.InitializeFeatures());
+
+    Assert.That(ex!.InnerExceptions, Is.EquivalentTo(expectedExceptions));
+
+    webTestFeatureMock1.Verify(f => f.Initialize(), Times.Once);
+    webTestFeatureMock2.Verify(f => f.Initialize(), Times.Once);
   }
 
   private void AssertCollectionItems (
--- a/Remotion/Web/Development.WebTesting/BrowserSession/BiDiBrowserLogProvider.cs
+++ b/Remotion/Web/Development.WebTesting/BrowserSession/BiDiBrowserLogProvider.cs
@@ -14,21 +14,19 @@ namespace Remotion.Web.Development.WebTesting.BrowserSession;
 /// BiDi logging currently only works in Firefox and requires web socket support to be enabled when the driver is created.
 /// For Chromium browsers use <see cref="SeleniumBrowserLogProvider"/>.
 /// </remarks>
-public class BiDiBrowserLogProvider : IBrowserLogProvider, IDisposable
+public class BiDiBrowserLogProvider : IBrowserLogProvider, ILifecycleWebTestFeature
 {
   private readonly ConcurrentQueue<BrowserLogEntry> _logEntries = new();
-  private readonly Subscription _eventSubscription;
+  private readonly IBidiConnectionProvider _bidiConnectionProvider;
+
+  private Subscription? _eventSubscription;
 
   public BiDiBrowserLogProvider (IBidiConnectionProvider bidiProvider)
   {
     ArgumentNullException.ThrowIfNull(bidiProvider);
 
-    bidiProvider.OpenBidiConnection();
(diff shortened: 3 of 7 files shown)


Remotion/Web/Development.WebTesting/BrowserSession/BiDiBrowserLogProvider.cs
```diff
@@ -14,21 +14,18 @@ namespace Remotion.Web.Development.WebTesting.BrowserSession;
 /// BiDi logging currently only works in Firefox and requires web socket support to be enabled when the driver is created.
 /// For Chromium browsers use <see cref="SeleniumBrowserLogProvider"/>.
 /// </remarks>
-public class BiDiBrowserLogProvider : IBrowserLogProvider, IDisposable
+public class BiDiBrowserLogProvider : IBrowserLogProvider, ILifecycleFeature
 {
   private readonly ConcurrentQueue<BrowserLogEntry> _logEntries = new();
-  private readonly Subscription _eventSubscription;
+  private Subscription? _eventSubscription;
+  private readonly IBidiConnectionProvider _bidiConnectionProvider;
```
a-ctor: To make the fields easier to read I like to group `readonly` vs `mutable` and separate them using an empty line.

```suggestion
  private readonly IBidiConnectionProvider _bidiConnectionProvider;

  private Subscription? _eventSubscription;
```

Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
```diff
@@ -116,6 +104,8 @@ public void WebTestHelper_OnTestSetUp_ResetsBrowserLog ()
       webTestHelper.OnSetUp(_testContext.Object);
       browserLogEntries = webTestHelper.MainBrowserSession.GetBrowserLogs();
       Assert.That(browserLogEntries, Is.Empty);
+
+      ShutdownWebTestHelper(webTestHelper, true);
```
a-ctor, line 108: What is the reason for adding this here? I would expect the shutdown to happen in a `TearDown` not the `SetUp` as the last test might not get shut down with this ordering.

Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
```diff
@@ -158,6 +148,8 @@ public void WebTestHelper_OnTestTearDown_WhenBrowserLogEntriesExist_SetsFailure
       testContext.Verify(
           _ => _.SetFailure(It.IsRegex($"There are unexpected browser log entries at the end of the test:{Environment.NewLine}.*\"?any error message\"?")),
           Times.Once());
+
+      webTestHelper.OnFixtureTearDown();
```
a-ctor, line 152: See comment above. Is there a specific reason we added this? -> Otherwise this might get covered when we introduce a `TearDown`.

Remotion/Web/Development.WebTesting/BrowserSession/BiDiBrowserLogProvider.cs
```diff
@@ -42,11 +39,25 @@ public void ResetBrowserLogs ()
...
+            ).GetAwaiter().GetResult();
+  }
+
   public void Dispose ()
   {
     try
     {
-      _eventSubscription.DisposeAsync().AsTask().GetAwaiter().GetResult();
+      _eventSubscription?.DisposeAsync().AsTask().GetAwaiter().GetResult();
+      _bidiConnectionProvider.Dispose();
```
a-ctor: We should not dispose the `_bidiConnectionProvider` here as we don't really own it. The `IBidiConnectionProvider` implementation is the browser sessions and it is strange if we are the ones disposing it here.

Remotion/Web/Development.WebTesting/ILifecycleFeature.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace Remotion.Web.Development.WebTesting;
+
+public interface ILifecycleFeature : IDisposable
+{
+  void Init ();
+}
```
a-ctor: Please add XML docs for this interface and its member.

I would also suggest using more verbose names `ILifecycleWebTestFeature` and `Initialize`. We try to avoid abbreviations as they might ambiguous and require more thought :D

Remotion/Web/Development.WebTesting/BrowserSession/BiDiBrowserLogProvider.cs
```diff
@@ -42,11 +39,25 @@ public void ResetBrowserLogs ()
...
+  {
+    if (_eventSubscription is not null)
+      return;
+
+    _bidiConnectionProvider.OpenBidiConnection();
+    _eventSubscription = _bidiConnectionProvider.BiDiConnection.Log
+        .OnEntryAddedAsync(
+            entry => _logEntries.Enqueue(new BrowserLogEntry(entry)),
+            new SubscriptionOptions { Timeout = _bidiConnectionProvider.DefaultBidiTimeout }
+            ).GetAwaiter().GetResult();
```
a-ctor: Please check the formatting here. The indentation does not seem right.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -343,13 +346,26 @@ public void OnFixtureTearDown ()
     {
       _logger.LogInformation("WebTestHelper.OnFixtureTearDown() has been called.");
 
+      ExecuteLifecycleFeatureAction(feature => feature.Dispose());
+
       foreach (var browserSession in _browserSessions)
         browserSession.Dispose();
```
a-ctor: The features already get disposed as part of `browserSession.Dispose()`.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -188,6 +189,8 @@ public void OnFixtureSetUp (WindowSize? windowSize = null, [CanBeNull] DriverCon
       _mainBrowserSession = CreateNewBrowserSession(windowSize, configurationOverride);
       _logger.LogInformation("Browser: {0}, version {1}", _mainBrowserSession.Driver.GetBrowserName(), _mainBrowserSession.Driver.GetBrowserVersion());
       _logger.LogInformation("WebDriver version: {0}", _mainBrowserSession.Driver.GetWebDriverVersion());
+
+      ExecuteLifecycleFeatureAction(feature => feature.Init());
```
a-ctor: Because the disposing of features is handled by the `IReadOnlyWebTestFeatureCollection` (see comment below), I would like to move the initialization there as well. The idea is that initialization and dispose are related and should be handled by the same entity.

Please add `IReadOnlyWebTestFeatureCollection.InitializeFeatures()` and use it here.

a-ctor (changes requested):
After addressing the review changes please run the firefox tests a few times to ensure that we no longer have the hanging build problem :)

Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
noah-rubicon: This is not in the setup its at the end of a Test-Method.
I did it this way because each test i setting up their own webTestHelper inside the Testmethod rather than the Setup method...

Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
noah-rubicon: Yes, i did this here to dispose the local webTestHelper.
Adding a TearDown requires a proper Setup tho

Remotion/Web/Development.WebTesting.IntegrationTests/WebTestHelperTest.cs
a-ctor: my bad I got baited by the display

Image

Remotion/Web/Development.WebTesting/WebTestFeatureCollection.cs
```diff
@@ -89,6 +89,25 @@ public bool TryGet<TFeature> ([NotNullWhen(true)] out TFeature? result)
...
+      catch (Exception ex)
+      {
+        exceptions.Add(ex);
+      }
+    }
+
+    if (exceptions.Count != 0)
+      throw new AggregateException("One or more features failed to initialize.", exceptions);
+  }
+
```
a-ctor, line 110: Please add a test for this method.

I don't think we need the `AggregateException` wrapping here, if an `Initialize` call throws we can let the exception rip instead. But you can also keep it this way if you want :)

Remotion/Web/Development.WebTesting.UnitTests/WebTestFeatureCollectionTest.cs
```diff
@@ -129,6 +129,43 @@ public void GetEnumerator ()
     AssertCollectionItems(features, items);
   }
 
+  [Test]
+  public void InitializeFeatures ()
+  {
+    var mock1 = new Mock<ILifecycleWebTestFeature>();
```
a-ctor: Please add a prefix to include a type hint to the name. Just `mock` does not help in determining what kind of mock it is. e.g. `featureMock` or `webTestFeatureMock`.

Please also remove the numbering as we don't have more than one mock.

Remotion/Web/Development.WebTesting.UnitTests/WebTestFeatureCollectionTest.cs
```diff
@@ -129,6 +129,43 @@ public void GetEnumerator ()
...
+                {
+                    (typeof(ILifecycleWebTestFeature), mock1.Object),
+                    (typeof(ITestFeature), mock2.Object)
+                };
+    var features = new WebTestFeatureCollection(items);
+    var ex = Assert.Throws<AggregateException>(() => features.InitializeFeatures());
+    Assert.That(ex!.InnerExceptions, Contains.Item(expectedException));
+    mock1.Verify(f => f.Initialize(), Times.Once);
+    mock2.Verify(f => f.Initialize(), Times.Once);
+  }
```
a-ctor: Please add prefixes to the name here as well (see comment above).

This method would benefit from empty lines to improve readability by grouping related things together. In the tests the three parts (arrange, act, assert) should be clear. I would suggest adding new lines like this:


```suggestion
    var mock1 = new Mock<ILifecycleWebTestFeature>();
    var mock2 = new Mock<ILifecycleWebTestFeature>();

    var expectedException = new Exception("Feature 1 failed");
    mock1.Setup(f => f.Initialize()).Throws(expectedException);

    var items = new List<(Type, object)>
                {
                    (typeof(ILifecycleWebTestFeature), mock1.Object),
                    (typeof(ITestFeature), mock2.Object)
                };
    var features = new WebTestFeatureCollection(items);

    var ex = Assert.Throws<AggregateException>(() => features.InitializeFeatures());

    Assert.That(ex!.InnerExceptions, Contains.Item(expectedException));

    mock1.Verify(f => f.Initialize(), Times.Once);
    mock2.Verify(f => f.Initialize(), Times.Once);
  }
```

(I've also moved the declaration of `mock2` up as well)

Remotion/Web/Development.WebTesting.UnitTests/WebTestFeatureCollectionTest.cs
```diff
@@ -129,6 +129,43 @@ public void GetEnumerator ()
...
+
+
+    var items = new List<(Type, object)>
+                {
+                    (typeof(ILifecycleWebTestFeature), mock1.Object),
+                    (typeof(ITestFeature), mock2.Object)
+                };
+    var features = new WebTestFeatureCollection(items);
+    var ex = Assert.Throws<AggregateException>(() => features.InitializeFeatures());
+    Assert.That(ex!.InnerExceptions, Contains.Item(expectedException));
```
a-ctor: We only expect a single inner exception here so we should do a full assert instead of a contains here.

a-ctor (approved):

PR #1418, opened 2026-03-18
RM-9658 Add parsing for csp header

noble-rubicon:


Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
```diff
@@ -22,7 +24,58 @@ public class CspHeader
...
+    endOfDirectiveNameIndex = token.IndexOfAny(s_asciiWhitespaces);
+    return token[..endOfDirectiveNameIndex].ToLower();
+  }
+
+  private static string StripAsciiWhitespace (string input)
+  {
+    return input.Trim(s_asciiWhitespaces);
+  }
+
+    private CspHeader (ImmutableDictionary<string, StringValues> directives)
```
MichaelKetting:
```suggestion
  private CspHeader (ImmutableDictionary<string, StringValues> directives)
```
whitespace

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
```diff
@@ -22,7 +24,58 @@ public class CspHeader
 
   private readonly ImmutableDictionary<string, StringValues> _directives;
 
-  private CspHeader (ImmutableDictionary<string, StringValues> directives)
+  //https://infra.spec.whatwg.org/#ascii-whitespace
+  private static readonly char[] s_asciiWhitespaces = [' ', '\t', '\n', '\r', '\f'];
+  private static readonly ILogger s_logger = LazyLoggerFactory.CreateLogger<CspHeader>();
+
```
MichaelKetting, line 28: Please move the static state and static members before the instance state.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
```diff
@@ -22,7 +24,58 @@ public class CspHeader
...
+  private static string GetDirectiveName (string token, out int endOfDirectiveNameIndex)
+  {
+    endOfDirectiveNameIndex = token.IndexOfAny(s_asciiWhitespaces);
+    return token[..endOfDirectiveNameIndex].ToLower();
+  }
+
+  private static string StripAsciiWhitespace (string input)
+  {
+    return input.Trim(s_asciiWhitespaces);
+  }
```
MichaelKetting, line 76: question: is it important we pass ASCII whitespace to make sure all whitespace is covered or because there's some whitspace that's non-ascii and should be kept because standard says so?

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
```diff
@@ -22,7 +24,58 @@ public class CspHeader
...
+    }
+
+    return new CspHeader(builder.ToImmutable());
+  }
+
+  private static string GetDirectiveName (string token, out int endOfDirectiveNameIndex)
+  {
+    endOfDirectiveNameIndex = token.IndexOfAny(s_asciiWhitespaces);
+    return token[..endOfDirectiveNameIndex].ToLower();
+  }
```
MichaelKetting: I wonder, would it be better if we use a value tuple with named fields (` (string directive, int endOfDirectiveNameIndex) `)? for private code, that's perfectly fine.

Disclaimer: value tuples are new and I always wonder if an out parameter is an OK thing todo and the correct sematics etc. so, might be better with value tuple, or might not be.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
```diff
@@ -22,7 +24,58 @@ public class CspHeader
...
+  /// Parses a CSP HTTP Header, provided as a <see cref="string"/>, into a <see cref="CspHeader"/>.
+  /// </summary>
+  /// <param name="input">The value of the CSP HTTP header as <see cref="string"/>.</param>
+  /// <remarks>
+  /// https://www.w3.org/TR/CSP3/#grammardef-serialized-policy
+  /// </remarks>
+  [Pure]
+  public static CspHeader Parse (string input)
+  {
+    var builder = ImmutableDictionary.CreateBuilder<string, StringValues>();
```
MichaelKetting: "dictionary" or "directives" (best option?). 

That this is a builder instead of a dictionary is kind of irrelevant. But reading the code, seeing "builder" creates an expectation what this code might do that is totally different from the actual semantics.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
```diff
@@ -22,7 +24,58 @@ public class CspHeader
...
+
+    foreach (var rawToken in input.Split(';'))
+    {
+      var token = StripAsciiWhitespace(rawToken);
+
+      if (string.IsNullOrWhiteSpace(token) || !token.All(char.IsAscii))
+        continue;
```
MichaelKetting, line 48: When can this be NULL or Whitespace?

MichaelKetting: Not sure if not ignoring ASCII wouldn't be better here, despite what the standard says. If it doesn't hurt our parsing sematics, I would suggest we don't check for non-ascii and just do gargabe-in / gargabe-out because this means more predicatbility and prevents obscure "why is this broken" reports. There's also domains with non-ascii characters these days...

MichaelKetting (changes requested):
Mostly formatting and some ASCII questions that are meant as question not a definite change it. If you agree with my reasoning, please do change it, though.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
noble-rubicon: I can do that. About your last point. I thought about it too and added a test for this special case. But it's funny that these domains aren't officially supported by CSP

noble-rubicon: You are right, actually never.

MichaelKetting: not-officially-supported by CSP: yeah... Or browser-magic happens (each domain as an ascii equivalent). Dunno. Not our problem fortunately :D

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
noble-rubicon: Well, the csp standard defines the ASCII whitespaces to be SPACE, \t, \r, \n and \f. E.g. char.IsWhitespace has more characters that are defined as "whitespace" (https://learn.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=net-10.0). The CSP standard doesn't explicitly mention anything about other whitespaces (https://infra.spec.whatwg.org/#ascii-whitespace). But it looks like I haven't read the note below it, because it says "HTTP whitespace" which excludes the \f in escaping. And people should use the "HTTP whitespace" for HTTP header values. So what is your opinion on it? I am not sure what directive we should follow relating to "whitespace". The ASCII Whitespaces, the HTTP whitespaces or just simply ' '.

MichaelKetting: Vertical Tab ist obsure.
Newlines is weird.
NonBreakingSpace isn't considered a Whitespace by .NET if I remember correctly.

Unless we are legitimately worried about splitting the wrong thing (someone uses a non-ascii whitespace as part of the value and then we split the value), I would use .net whitespace if that's the larger group of whitespace.

I think we should have a test that verifies the correct (or expected incorrect) behavior of `directive-name 'foo bar'`. I haven't found this test and the single quotes are typically used to enclose characters sequences that are made from letters and whitespaces. I haven't re-checked the standard on that, though.

noble-rubicon: We could use .Trim(), which uses the large group of whitespaces from .NET. Sadly there is no option for .IndexOfAny and .Split with the large group of whitespaces nor is there a simple way to get the large group of whitespaces from .NET as char array. For consistency reasons I would use the ASCII whitespaces for all of the methods (Trim, IndexOfAny, Split).

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
noble-rubicon: Actually, it can be whitespace / string.Empty if the input is e.g. "default-src 'self';\r". Then it will be splitted at ';' which results in two entries first one "default-src 'self'", the second one "\r", which is than trimmed in StripAsciiWhitespace, which returns an empty string.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
MichaelKetting: As discussed, then let's keep it consistent with the whitespaces even though it's unsual in .net code. We didn't miss any obvious ones so I expect no problems.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
MichaelKetting: ah yes. okay.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
MichaelKetting, line 21: I just found a naming bug in the existing code. If you could be so kind and create a PR for this including the backport. right now, we can fix it without causing cost down the line:

enum `CspDirectives` should be `CspDirective` (singular)

cc @a-ctor

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspHeaderTest.cs
```diff
@@ -159,4 +159,70 @@ public void EnumValuesNamesTest (CspDirectives directive)
...
+
+  [Test]
+  public void Parse_DirectiveValuesContainsSpace_IsUsedAsIs ()
+  {
+    const string input = "DeFaUlT-SrC 'se lf'";
+
+    var parsedHeader = CspHeader.Parse(input);
+
+    Assert.That(parsedHeader.ToString(), Is.EqualTo("default-src 'se lf'"));
+  }
```
MichaelKetting, line 232: I just realized, we got two implementations of the whitespace split: 

We should have only one private method that splits the value of a directive.

Also, the AddDirectiveValue API checks only for regular whitespace but the parser also checks for fancy whitespace. this should be fixed.

MichaelKetting (changes requested):
Sorry, I managed to find a  consistency problem with the existing code. I think it's a quick and easy fix.

Remotion/Web/Core/ContentSecurityPolicy/CspHeader.cs
noble-rubicon: #1419

Remotion/Web/UnitTests/Core/ContentSecurityPolicy/CspHeaderTest.cs
MichaelKetting, line 56: Nice to have: special whitespace test, too

MichaelKetting (approved):

PR #1419, opened 2026-03-23
Rename enum CspDirectives to CspDirective

noble-rubicon:


noble-rubicon:
I've used the same issue number for this pull request as for the other. I hope this is fine, because I couldn't push without an issue number in the branch name.

MichaelKetting (approved):

PR #1430, opened 2026-04-29
Update CoreForms to 9.1.3 (Backport v7.0)

noble-rubicon:


MichaelKetting (changes requested):
Please update commit message to include ticket number

noble-rubicon:
It looks like we also have to update log4net in v7.0 if we want the build to succeed. Is that okay for you?

MichaelKetting:
> It looks like we also have to update log4net in v7.0 if we want the build to succeed. Is that okay for you?

Disable warning for log4net. We don't need to resolve medium vulnerabilities in re-motion 7.0 since there's a newer version.

MichaelKetting (approved):

PR #1434, opened 2026-05-26
RM-9679 fixed GetGroupedAndOrderedDataContainer

Mario185:
https://jira.rubicon.eu/browse/RM-9679


Remotion/Data/DomainObjects.UnitTests/Persistence/Rdbms/StorageProviderCommands/Factories/BatchedSaveCommandFactoryTest.cs
```diff
@@ -1394,6 +1460,19 @@ public void Lock_DoesNotWaitToFail_ForOtherTransaction_WithIsolationLevel_ReadCo
     }
   }
 
+  private void AssertTableManipulationAccessorExtensionPointsCalled (
```
MichaelKetting, line 1463: preference: add parameter names to the callsite so the correct usage is a) visible without IDE feature support (which I turn off) and to prevent copy&paste etc mistakes. We don't need to fix it now/here, but we should teach Claudia this skill.

MichaelKetting (approved):
Commit message: preferred format: present tense actice:

`RM-9679 Fix GetGroupedAndOrderedDataContainer so it uses CreateUpdateDataContainerAccessor.`

PR #1440, opened 2026-08-10
RM-9722 Add SmartPage setting to toggle CoreForms unload event

a-ctor:


Remotion/Web/Core/UI/SmartPage.cs
```diff
@@ -692,6 +693,26 @@ protected virtual bool IsCspReportOnlyEnabled
     }
   }
 
+  /// <summary>
+  /// Gets or sets the flag that determines whether CoreForms client scripts will use the deprecated unload event.
+  /// By default, the unload event will not be used. Set this property to opt-out of this behavior.
```
MichaelKetting:
```suggestion
  /// By default, the unload event will not be used. Set this property to enable the unload-event integration as seen in System.Web on .NET Framework.
```

Want to make it a bit more explicit what's expected to happen.

MichaelKetting (approved):
Looks perfectly fine and includes integration tests. I wouldn't know what else to look for.

PR #1442, opened 2026-08-10
Feature/rm 9408 union query problem

a-ctor:


Remotion/Data/DomainObjects/Linq/MappingResolver.cs
```diff
@@ -199,6 +201,47 @@ public Expression ResolveTypeCheck (Expression checkedExpression, Type desiredTy
...
+    private static SqlEntityExpression? UnwrapSqlEntityExpression (Expression expression)
+    {
+      while (expression is UnaryExpression unaryExpression)
+        expression = unaryExpression.Operand;
+
+      if (expression is SqlEntityExpression sqlEntityExpression)
+        return sqlEntityExpression;
+
+      return null;
+    }
```
MichaelKetting, line 245: Would prefer a comemnt as to the why of this.

MichaelKetting (approved):
Given green tests, the number and complexity of the union integration tests inspire confidence.

a-ctor:
@MichaelKetting Sadly, I've found a few new cases and a bug in the backend. In some scenarios computed column names are added twice resulting in `NULL as column as Car_column`, which is not ideal :p

The tests are missing the union of anonymous classes, which add additional complexity as we now have the columns split up into multiple entities that need reconciliation with the corresponding entity in other UNIONs but now with their own. But that should be fixable on the re-motion side, I am going to investigate tomorrow.

MichaelKetting:
> @MichaelKetting Sadly, I've found a few new cases and a bug in the backend. In some scenarios computed column names are added twice resulting in `NULL as column as Car_column`, which is not ideal :p
> 
> The tests are missing the union of anonymous classes, which add additional complexity as we now have the columns split up into multiple entities that need reconciliation with the corresponding entity in other UNIONs but now with their own. But that should be fixable on the re-motion side, I am going to investigate tomorrow.

@a-ctor  We should try to put the bugfix in the correct place even if it's also fixable in re-motion.

a-ctor:
@MichaelKetting you misunderstood: There was a bug in SQL-Backend and some missing functionality in the implementation.

The thing that is missing is support for complex arguments in anonymous types. e.g. `.Select(car => new { Car = car, ... })`. Or at least that's the code that comes closes to whats happening in AN. I have updated the changeset to include the necessary changes. The main problem is that this code is not really possible to integration test as the features are not directly usable in re-motion.

PR #1445, opened 2026-08-17
RM-9408 Add support for UNIONs in re-store (Backport 8.0)

a-ctor:


MichaelKetting (approved):
I only scanned the code, no semantic verification since it's a backport in a place identical to develop.

PR #1446, opened 2026-08-17
RM-9724 Add FocrceSeek to lockstatment for Batched Sql queries

dkudernatsch:


Remotion/Data/DomainObjects/Persistence/Rdbms/DbCommandBuilders/BatchedLockDbCommandBuilder.cs
```diff
@@ -121,7 +121,11 @@ private string CreateLockStatement (
   {
     var selectColumns = string.Join(", ", columns.Select(c => $"{parameterAlias}.{c}"));
     var joinCondition = string.Join(" AND ", columns.Select(c => $"{parameterAlias}.{c} = {tableAlias}.{c}"));
-    var tableHints = forReadCommittedIsolation ? "ROWLOCK, XLOCK, READPAST" : "ROWLOCK, XLOCK";
```
MichaelKetting, line 124: I think we should update the Lock_DoesNotWaitToFail_ForOtherTransaction_WithIsolationLevel_ReadCommitted to use more than two Tasks and possibly switch from Task to Thread. It doesn't have to be in this PR but it would be correct and helpful to demonstrate the issue. If not done here, we need a follow-up task.

MichaelKetting (approved):

MichaelKetting (approved):
See comment about test. Please do a rebase on top of devleop to get rid of the forward merge. Note: I updated the branch rules to enforce a title. Not sure if this breaks force push for the current branch. During Merge, it's possible github allows a custom message. if so, please amend the branchname printed inthe commit message.

